PR #3, opened 2021-12-15
Add CCv1 Image security design document

jialez0:
This document describes the design of container image security under the confidential container V1 architecture, including the encryption-decryption and signing of container image. The components involved include image-rs, ocicrypt-rs, attestation-agent and Key Broker Service.

cc: @fitzthum @arronwy @stevenhorsman @jiazhang0 @jiangliu


docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,345 @@
...
+```json
+{
+    "name":"sigstore_config",
+    "informations":{
+        "index":"",
+        "extra_param":[]
+    },
+}
+```
+
```
fitzthum, line 298: This interface seems fine, but it is notable that none of the examples use anything in the `information` field. It's probably good to keep things flexible, though, as we should be able to use this to provide anything to the guest in the future.

One note is that the `offline_sev_kbc` is limited by the size of the injected secret. We may run into issues when we start trying to send in a bunch of stuff. As mentioned, we should think about fallbacks if certain KBC/KBS doesn't support certain endpoints. We can make an online KBS (even with SEV(-ES)) that won't have the same limitation.

We should also think about how we plan to implement this in the short-term. Maybe we should have a PR where we add these endpoints as a stub to every existing KBC and then the individual maintainers can go through and implement how they would like.

fitzthum (commented):
This looks great. Very thorough. I have a few minor suggestions.

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,345 @@
...
+policy.json is a necessary configuration file for deploying images in the confidential container V1 design. Its location is specified in the /etc/containers/ directory of boot image rootfs. Its main functions are as follows:
+
+- Declare a whitelist of container images with registry granularity
+
+- Configure the public key to use when verifying the image signature
+
+Thus, policy.json is confidential information that must be controlled by users and cannot be read or modified by untrusted entities.
+
+If image-rs finds that there is no policy.json in the /etc/containers/ directory or If the policy.json file is empty, call the GetResource grpc service provided by the attestation-agent. After remote attestation, obtain it from the user controlled relying KBS through the encrypted channel.
```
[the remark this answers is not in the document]
jialez0: In the mode described in the document, image-rs will preferentially read the configuration file under the local specified directory. If it is found that the config file does not exist, it will get it from AA. In the scenario where a trusted rootfs is provided, the configuration file in the local specified directory will be preset in advance, so the service interface of AA will not be called to obtain it. In this way, it can be compatible with the scenarios of trusted rootfs and untrusted rootfs

Do you mean that we need an explicit configuration option to configure whether image-rs is obtained directly from the local specified directory or from the service interface of AA?

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,345 @@
...
+}
+```
+
+Before pulling the container image, match the policy in the following order according to the location/registry of the container image and execute the corresponding policies:
+
+- Match single specific scopes under each member in transport:
+  - For the image from the container registry that implements the Docker Registry HTTP API V2 protocol, match each scope under the "docker" member;
+  - For images from local storage, match each scope under the "dir" member;
+- Match the default scope under each member in the transport. In the above example, there is only one default scope under the "dir" member, and the policy is reject.
+- Match the global default scope, such as the "default" field in the first line of the above example. This policy must be configured to reject.
```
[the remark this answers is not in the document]
jialez0: The policy file provides this flexible configuration function. If the user does not want to use any container image signature verification function, it can be configured in the policy file as follows:
```
{
  "default": [
    {
      "type": "insecureAcceptAnything"
    }
  ],
}
```
If the user specifies a registry and does not need to verify the signature, just change the policy type under the corresponding registry to `insecureAcceptAnything`. 

This will cause the signature verification system to no longer take effect, and any container image without signature will be pulled and deployed. We do not recommend this method. However, if users have special needs in this regard, they can inform the rootfs producer (such as CSP) in advance, and the rootfs producer will write the above unsafe configuration to the policy file.

docs/ccv1_image_security_design.md
jialez0: As mentioned above, obtaining resources from AA will only be triggered when image-rs does not find the configuration files in the local (rootfs) directory. For `offline_sev_kbc`, these resources can be preset in rootfs when making boot image, or provided by sev KBC in any special way (for example, use the simple policy of hard coding in KBC, or, as you mentioned, build an additional online KBC, etc.).

About the implementation of this interface of AA, I will start coding next week:)

docs/ccv1_image_security_design.md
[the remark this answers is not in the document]
Jakob-Naucke: @jialez0 Do you see it as a strict necessity that `policy.json` cannot be _read_ by untrusted entities? I fully understand they should not be able to _write_ it, but do we consider a list of allowed container images plus a public key for verification to be confidential data? I wouldn't understand what the threat is there.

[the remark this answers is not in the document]
fitzthum: > Do you mean that we need an explicit configuration option to configure whether image-rs is obtained directly from the local specified directory or from the service interface of AA?

No I think it's good as it is. If someone doesn't want to use signatures they can just include the permissive config that you describe below in their initrd. This adds additional overhead for the basic user who just wants to deploy containers, but I guess it's fine.

docs/ccv1_image_security_design.md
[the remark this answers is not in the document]
fitzthum: Ok, I think this is fine. Keep in mind that disabling signature verification isn't necessarily non-secure if the containers are encrypted.

docs/ccv1_image_security_design.md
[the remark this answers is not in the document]
jialez0: @Jakob-Naucke At present, ccv1, there are two distribution methods for policy files. In the protected boot image scheme, the user directly presets them in the rootfs of the boot image. In the unprotected boot image scheme, they are dynamically obtained from kbs through AA. KBS will distribute the policy only after AA's attitude evidence is verified, Sending through the encrypted channel is not to prevent untrusted entities from reading him, but to prevent malicious tampering during transmission. Of course, only Mac can be used to protect the integrity of policy files, which depends on the security requirements of specific KBC and KBS.

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,345 @@
...
+
+- container image: pushed to the specified remote registry.
+
+- signature claim file: upload to the specified sigstore server.
+
+## Deployment of protected container image
+
+In the confidential container V1 design, security functions will be performed when pulling the container image, including checking the registry whitelist, verifying the container image signature, and decrypting the container image layer. The high-level process is as follows:
+
+1. Check whether the registry where the image is located is in policy.json's whitelist.
```
bpradipt, line 143: It would be really nice to change all occurrence of `whitelist` with `allowlist`.

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,345 @@
...
+        }
+    }
+}
+```
+
+Before pulling the container image, match the policy in the following order according to the location/registry of the container image and execute the corresponding policies:
+
+- Match single specific scopes under each member in transport:
+  - For the image from the container registry that implements the Docker Registry HTTP API V2 protocol, match each scope under the "docker" member;
+  - For images from local storage, match each scope under the "dir" member;
```
bpradipt, line 198: Is this a valid case?

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,345 @@
...
+- Declare a whitelist of container images with registry granularity
+
+- Configure the public key to use when verifying the image signature
+
+Thus, policy.json is confidential information that must be controlled by users and cannot be read or modified by untrusted entities.
+
+If image-rs finds that there is no policy.json in the /etc/containers/ directory or If the policy.json file is empty, call the GetResource grpc service provided by the attestation-agent. After remote attestation, obtain it from the user controlled relying KBS through the encrypted channel.
+
+The format of the policy.json file follows the format and syntax defined in the [container/image policy.json](https://github.com/containers/image/blob/main/docs/containers-policy.json.5.md) document. In the V1 design of the confidential containers, a safe and reasonable policy.json file should look like the following example: 
+
```
bpradipt, line 166: This is good. I'm wondering if it would make sense to provide some sample policies like flexible which allows everything (`insecureAcceptAnything`), restrictive (for CC) etc.
image-rs functionality is generally usable feature (similar to api restriction feature) and I'm envisaging a scenario for development and otherwise where having sample policies will help.

bpradipt:
@jialez0 thanks so much for putting this design document. Very helpful.  Added few minor comments and questions.

docs/ccv1_image_security_design.md
jialez0: Done :)

docs/ccv1_image_security_design.md
jialez0: OK, I have added some other examples in the document. For more and more complete examples, please refer to [container/image policy.json](https://github.com/containers/image/blob/main/docs/containers-policy.json.5.md)

docs/ccv1_image_security_design.md
bpradipt: Hi @jialez0 trying to understand the local storage scenario. Is this referring to the container image already download (and decrypted) which can be used by another container in the POD ?

jialez0: The local storage scenario I envisage is the container image stored in rootfs when making the boot image, or the container image downloaded to the local after the pod is started.

bpradipt: Thanks @jialez0 .  rootfs having the container image pre-populated is an interesting one :-)

bpradipt (approved):
/lgtm
Thanks @jialez0. According to me this looks good to start with.

docs/ccv1_image_security_design.md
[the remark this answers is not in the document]
Jakob-Naucke: > Sending through the encrypted channel is not to prevent untrusted entities from reading him, but to prevent malicious tampering during transmission.

Shouldn't it be this then?
```
Thus, policy.json is confidential information that must be controlled by users and cannot be modified by untrusted entities.
```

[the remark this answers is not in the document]
jialez0: > 

My statement here may not be accurate and should be modified as follows: `The policy file must be generated and maintained by the user. Additional security means need to be used to ensure that the policy file will not be modified by malicious entities during storage and transmission. If users have high requirements for security level, they even need to use additional means to ensure the confidentiality of policy files.`

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,399 @@
...
+  ```json
+  {
+      "symkey": "54kiln1USEaKnlYhKdz+aA==",
+      "cipheroptions": {
+          "nonce": "AdcRPTAEhXx6uwuYcOquNA=="
+      }
+  }
+  ```
+
+Then, encrypt the PrivateLayerBlockCipherOptions with the user's private key. Under the design of confidential containers V1, when decrypting the image, the decryption of PrivateLayerBlockCipherOptions needs to be carried out through the key provider protocol with the attestation-agent. Therefore, when encrypting the image, thePrivateLayerBlockCipherOptions also needs to be encrypted through the key provider protocol, Moreover, the name of the key provider program used in encryption needs to be consistent with that used in decryption. Under the design of confidential containers V1, this name is uniformly specified as "attestation-agent".
```
c3d, line 48: missing space in thePrivateLayerBlockCipherOptions.

Also consider backticks around PrivateLayerBlockCipherOptions (here and other places).

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,399 @@
...
+
+The key provider service program encrypts the privatelayerblockcipheroptions with the user's OpenPGP key. After that, it will perform Base64 encoding and then package it into a JSON structure. Under the confidential container system, it is called 'annotation packet' to record some additional information, such as the ID of the user's private key, for example:
```
c3d, line 50: Capitalization of privatelayerblockcipheroptions

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+
+- Match single specific scopes under each member in transport, for example:
+  - For the image from the container registry that implements the Docker Registry HTTP API V2 protocol, match each scope under the "docker" member;
+  - For the image from local file system, match each scope under the "dir" member;
+- Match the default scope under the member in the transport. for example, the "" scope of the transport "dir".
+- Match the global default scope, such as the "default" field in the first line of the above example. We suggest that this policy must be configured to reject for security.
+
+After the matching is successful, the policy set under the corresponding scope is executed. In the design of confidential containers V1, only three policy types are currently supported now: "reject", "insecureAcceptAnything" and "signedBy". The former two is as the name implies, and the "signedBy" policy requires that there must exist at least one signature of this image can be verified by the public key ring provided by `keyPath` or `keyData` filed. In CCv1, he format of "signedBy" policy is defined as follows:
```
jodh-intel, line 224: How is an image resolved if multiple registries are specified and they all contain the same image? Will the local image be used in preference to the registry version?

How does "generic matching" affect this? `transports` is an object rather than an array, so are the entries searched in random order and the first matching image used?

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+    └── signature-1
+```
+
+The meaning of directory name is `image-name@digest-algo=digest-value`, multiple signature files can exist in the directory, which means that the same container image can have multiple signatures and be numbered and stored.
+
+### Image distribution
+
+After the encryption and signing are completed, we obtain two contents: the container image itself and the signature claim file, which are distributed and stored separately.
+
+- container image: pushed to the specified remote registry.
```
jodh-intel, line 139: Should this document show trust boundaries as by definition if a remote registry is configured, someone must trust it, even if the local host environment itself is not trusted.

Also, since a remote entity is being used, should the system require DNS over HTTPS/TLS?

jodh-intel (commented):
Thanks @jialez0 - a few comments.

Also, I think an overview diagram would make this easier to understand.

Finally, you might want to tal at https://github.com/kata-containers/kata-containers/blob/main/docs/Documentation-Requirements.md for ideas on how Kata docs handle API calls, filenames, etc.

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,399 @@
+# CCv1 Image Security Design
+
+This document describes the design of container image security under the confidential container V1 architecture, including the encryption-decryption and signing of container image. The components involved include image-rs, ocicrypt-rs, attestation-agent and Key Broker Service.
+
+## Production of protected container image
+
+In the confidential containers V1 design, after the user completes the production of the conventional container image, it is necessary to add security protection functions for the container image, including encryption and signing. It should be noted that the encryption process must be carried out before the signing process.
```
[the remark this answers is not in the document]
c3d: @jodh-intel I believe that you are asking good questions. The "atomically" part is more a question of integrity than anything else, right? Can we check the encrypted/signed image against the original file at the end of the process to ensure that nothing change in between? Is your concern about "modify the image" regarding the unencrypted source, or something else?

Also, out of curiosity, what kind of attack do you have in mind when you write "different physical system to the one where the image is to be deployed"?

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+
+The format standard of container image encryption is compatible with https://github.com/containers/ocicrypt.
```
c3d, line 9: Add link: `[OCICrypt](https://github.com/containers/ocicrypt).`

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+
+The format standard of container image signing is compatible with https://github.com/containers/image.
+
+### Image encryption
+
+Image encryption is based on layer granularity. Users can selectively encrypt layers containing private data or choose not to encrypt any layers. The high-level process of encrypting an image layer is as follows:
```
c3d, line 15: Didn't we conclude at some point late last year that our security model might not be robust for images with no encryption at all? @fitzthum were you the one raising that point (sorry, my memory is a bit fuzzy)?

docs/ccv1_image_security_design.md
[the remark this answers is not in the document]
c3d: Terminology (here and in the rest of the document): Should we use "owner" instead of "user"? (I asked on Slack)

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+
+1. Dynamically generate a random symmetric key.
+
+2. Encrypt container image layer with symmetric key.
+
+3. Encrypt the symmetric key with the user's private key.
```
c3d, line 21: user->owner ?

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,399 @@
...
+  ```json
+  {
+      "symkey": "54kiln1USEaKnlYhKdz+aA==",
+      "cipheroptions": {
+          "nonce": "AdcRPTAEhXx6uwuYcOquNA=="
+      }
+  }
+  ```
+
+Then, encrypt the `PrivateLayerBlockCipherOptions` with the user's private key. Under the design of confidential containers V1, when decrypting the image, the decryption of `PrivateLayerBlockCipherOptions` needs to be carried out through the key provider protocol with the attestation-agent. Therefore, when encrypting the image, the `PrivateLayerBlockCipherOptions` also needs to be encrypted through the key provider protocol, Moreover, the name of the key provider program used in encryption needs to be consistent with that used in decryption. Under the design of confidential containers V1, this name is uniformly specified as "attestation-agent".
```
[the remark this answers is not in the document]
c3d: @jodh-intel 

>  Is this `ocicrypt`?

The laste sentence says "attestation-agent". I was not aware we would use the attestation agent for encryption too. Is that intentional?

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+
+The key provider service program encrypts the `PrivateLayerBlockCipherOptions` with the user's OpenPGP key. After that, it will perform Base64 encoding and then package it into a JSON structure. Under the confidential container system, it is called 'annotation packet' to record some additional information, such as the ID of the user's private key, for example:
```
c3d, line 54: This is the first reference to OpenPGP. Do we want to make any reference to OpenPGP (arguably an implementation detail)?

Is there any necessary relation of that key with the user's private key in step 3 line 21 above? If so, use the same terminology. If not, let's explain cases where the keys could be different and how it would work.

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,399 @@
...
+{
+    "key_id":"1234",
+    "wrapped_data":#base64encode(Enc(PrivateLayerBlockCipherOptions)),
+    "wrap_type":"",
+}
+```
+
+The specific format of the annotation packet can be determined by the key provider program used by the user, but it is necessary to ensure that the key provider program used by the decryptor (i.e. the attestation-agent integrated with the specified Key Broker Client) supports parsing annotation packets of the same format.
+
+After completing the above work, Base64 encode the `PublicLayerBlockCipherOptions` and annotation packet, and write them into the manifest.json of the container image as the annotation field of the layer, the former is placed in the `org.opencontainers.image.enc.pubopts` field and the latter in the `org.opencontainers.image.enc.keys.provider.attestation-agent` field. Finally, add the "+encrypted" suffix to the mediaType field of this layer. For example, the container image manifest with an encrypted layer is as follows:
```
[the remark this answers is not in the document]
c3d: In addition to the JSON output, could you give an example of how you generated it (i.e. the commands you used to actually generate that JSON, which I assume you did not create manually)?

docs/ccv1_image_security_design.md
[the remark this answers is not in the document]
jodh-intel: On that topic, CC _really_ needs a glossary folks: https://github.com/confidential-containers/documentation/issues/13.

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+                "org.opencontainers.image.enc.keys.provider.attestation-agent": "eyJwcm9..."
+            }
+        }
+    ]
+}
+```
+
+### Image signing
+
+The container image signing standard used by the confidential containers V1 is follow the signing mechanism used in [containers/image](https://github.com/containers/image). At present, the only defined cryptographic signature format is a OpenPGP signature (RFC 4880). The signing process will not modify any content of the container image, including layer, config, manifest, etc. Therefore, the signature will not change the image manifest digest used to identify the image.
```
c3d, line 92: I'm a bit confused by this paragraph. On line 7, you wrote "It should be noted that the encryption process must be carried out before the signing process". Should I deduce that the signing is performed on the _encrypted_ data? But then, that seems to conflict with "the signature will not change the image manifest digest"… I suspect that @jodh-intel has a similar question when he asks below "_Is 817a12c32a39be.. the sha256 value for the value of busybox:latest at the time this file was created?_", which I interpret as "is it the sha256 for the unencrypted image" (sorry if I misunderstood the question).

docs/ccv1_image_security_design.md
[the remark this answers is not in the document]
jodh-intel: @c3d - I'm just trying to understand the security doc in the context of the CC threat model. CC does not trust the host environment so if the newly created, but unencrypted and unsigned image is left lying around in that untrusted host environment, how can you then trust the resulting encrypted and signed image, since it may have been compromised before you actually encrypted and signed it? To counter these attack vectors, presumably you would want to as a minimum only ever create the secure images on a fully trusted host. But for the extra paranoid, you would want to create the image, then immediately encrypt and sign it (and verify that it can be decrypted and its signature is valid of course).

docs/ccv1_image_security_design.md
jodh-intel: I'm also unclear about images containing both encrypted and unencrypted layers. How can CC guarantee safety in the following sort of scenario?:

| layer | encrypted? | dependent layer |
|-|-|-|
| 1 | yes | n/a |
| 2 | no | n/a |
| 3 | yes | 2 |

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+
+```json
+{
+    "critical": {
+        "type": "atomic container signature",
+        "image": {
+            "docker-manifest-digest": "sha256:817a12c32a39be..."
+        },
+        "identity": {
+            "docker-reference": "docker.io/library/busybox:latest"
```
c3d, line 110: What is `docker-reference` used for?

Is there a similar mechanism for other repositories, e.g. quay?

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+{
+  "default": [
+    {
+      "type": "insecureAcceptAnything"
+    }
+  ],
+}
+```
+
+We must remind that this configuration is very insecure. It will pull all unsigned container images. Or if you want to unconditionally accept an image from a specific registry, you just need to change the policy of the member matching the registry under the corresponding scope to `insecureAcceptAnything`.
```
jodh-intel, line 258: Why is `insecureAcceptAnything` required? For testing? Backwards compatibility with existing registries? I think explaining this might help.

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+- Cryptographic signature of JSON payload
+
+The JSON payload contains the manifest digest and identity information of the container image. Examples are as follows:
+
+```json
+{
+    "critical": {
+        "type": "atomic container signature",
+        "image": {
+            "docker-manifest-digest": "sha256:817a12c32a39be..."
```
c3d, line 107: Why "docker"? Why not just "manifest-digest"? Or "unencrypted-manifest-digest" if that is the intent?

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,399 @@
...
+    "optional": {
+        "creator": "some software package v1.0.1-35",
+        "timestamp": 1483228800,
+    }
+}
+```
+
+The process of signing a container image is as follows:
+
+1. Create a signature JSON payload and write the manifest digest together with some otheridentity information of the container image.
```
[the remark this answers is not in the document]
c3d: When you write "write the manifest digest together with …", where do you write that? Is that just some temporary file in order to encrypt the JSON payload? Or are you describing something more subtle?

Should I interpret the above as follows:

Create a signature JSON payload file following the structure above, containing the manifest digest for the original image, some additional identity information such as the original (unencrypted) container image reference name, and optional information.

?

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+2. Use the private key to sign the JSON payload generated in the previous step, and package the cryptographic signature together with the JSON payload into a signature claim file.
```
c3d, line 123: Your signature claim files directory structure below only shows the signature file, but not the JSON payload. Is that intentional?

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+
+The signature claim file is what we usually call the container image signature file. Its directory structure and file name format are as follows:
+
+```
+.
+└── busybox@sha256=817a12c32a39bbe394944ba49de563e085f1d3c5266eb8e9723256bc4448680e
+    └── signature-1
+```
+
+The meaning of directory name is `image-name@digest-algo=digest-value`, multiple signature files can exist in the directory, which means that the same container image can have multiple signatures and be numbered and stored.
```
c3d, line 133: Not sure about "and be numbered and stored". It looks like this applies to signatures, not container images, right? Why would there be multiple signatures? If you are talking about multiple revisions of the container, then the sha256 would be different, so I guess that's not it.

Or are you considering multiple signatures with different algorithms?

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+
+### Image distribution
+
+After the encryption and signing are completed, we obtain two contents: the container image itself and the signature claim file, which are distributed and stored separately.
```
c3d, line 137: "the container image itself" -> "the encrypted container image itself" ?

c3d: Could you elaborate as to why they are distributed and stored separately (ideally in a wording that would be appropriate for the upcoming threat model document)?

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+
+3. (if needed in policy) Verify the signature of the container image according to the configuration of the policy.
+
+4. Pull the container image layer by layer. If an encrypted container image layer is encrypted, decrypt it.
+
+After the image is pulled, the subsequent deployment process is consistent with that of the normal container image.
+
+### policy.json
+
+policy.json is a necessary configuration file for deploying images in the confidential container V1 design. Its location is specified in the /etc/containers/ directory of boot image rootfs. Its main functions are as follows:
```
c3d, line 159: For _deploying_ images, or for _pulling_ them?

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+
+- "dir": used to match images located in the local file system.
+
+For the "docker" transport type, the scope can be a docker reference named in a fully expanded form, and tag or digest can be additionally specified for exact matching, for example: `docker.io/library/busybox:latest`. Generic matching is also supported:
+
+- repository namespace: for example, `docker.io/openshift`, which matches the images as `openshift/hello-openshift` and `openshift/origin`.
+
+- repository host: for example, `docker.io`, matches all images from the docker.io repository.
+
+- subdomain name matching: such as `*.temporary-project.example.com`. Note: wildcards can only appear in the top sub domain name. Wildcards like `example.*.com` are not allowed to be specified.
```
c3d, line 177: drop "to be specified"

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+
+For the "dir" transport type, the scope format is the directory path. The path must be absolute and contain no symbolic links. The default scope is "".
```
c3d, line 179: Stating that the default scope is "" contradicts "the scope format is the directory path" (since path "" is not an absolute path).

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+
+The policy file must be generated and maintained by the user. Additional security means need to be used to ensure that the policy file will not be modified by malicious entities during storage and transmission. If users have high requirements for security level, they even need to use additional means to ensure the confidentiality of policy files.
```
c3d, line 181: Makes me wonder if we should include a digest of the policy.json file as part of what we send during attestation.

c3d: Strike the above, we do when we measure the rootfs anyway.

docs/ccv1_image_security_design.md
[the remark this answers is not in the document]
c3d: The implication of the above is that a local rootfs `policy.json` takes precedence over a remote AA one. This seems wrong to me. Consider if some particular registry has been compromised or simply moved/renamed, but the rest of your encrypted images are otherwise good (possibly at a new location). Wouldn't it be preferable to add a `policy.json` to your KBS that would supersede the local one, rather than force all rootfs to be re-generated?

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+        "dir": {
+            "": [{"type": "insecureAcceptAnything"}]
+        }
+    }
+}
+```
+
+Before pulling the container image layers, match the policy in the following order according to the location/registry of the container image and execute the corresponding policies:
+
+- Match single specific scopes under each member in transport, for example:
```
c3d, line 218: in transport -> "in `transports`"

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+  - For the image from the container registry that implements the Docker Registry HTTP API V2 protocol, match each scope under the "docker" member;
```
c3d, line 219: "docker" -> `docker`
"dir" -> `dir`

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+            ],
+        },
+        "dir": {
+            "": [{"type": "insecureAcceptAnything"}]
+        }
+    }
+}
+```
+
+Before pulling the container image layers, match the policy in the following order according to the location/registry of the container image and execute the corresponding policies:
```
c3d, line 216: It's useful to have this information here, but I wonder if this policy matching order is not specified somewhere else?

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+    "transports": {
+        "docker": {
+            "docker.io/my_private_registry": [
+                {
+                    "type": "signedBy",
+                    "keyType": "GPGKeys",
+                    "keyData": "<public Key>",
+                }
+            ],
+            "registry.access.redhat.com": [
```
c3d, line 201: Just curious if that would not be an "oci" example rather than "docker"?

docs/ccv1_image_security_design.md
c3d, line 224: "former two is" -> "former two are"

c3d: "reject" -> `reject` and same for others.

docs/ccv1_image_security_design.md
c3d, line 258: "remind" -> "stress" or "point out" or "highlight" or "warn"

c3d: Note that the insecurity is actually somewhat limited for any image with at least one encrypted layer, even for a weak `policy.json`, since the decryption would fail unless the KBS handed over the keys, no?

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+
+We must remind that this configuration is very insecure. It will pull all unsigned container images. Or if you want to unconditionally accept an image from a specific registry, you just need to change the policy of the member matching the registry under the corresponding scope to `insecureAcceptAnything`.
+
+### sigstore config
+
+The configuration file that specifies the container image sigstore is specified as all `.yaml` files in /etc/containers/registries.d directory. After the policy.json allowlist check is completed, the entries in the sigstore configuration file will claim where the container image is located, and the signature claim file of the container image will be downloaded from the specified URL (a webserver or a local directory). The sample sigstore configuration file matching the sample policy.json file in the above is as follows:
+
+```yaml
+default:
+    sigstore: file:///var/lib/containers/sigstore
```
c3d, line 266: Inconsistent indent on this line

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+    }
+  ],
+}
+```
+
+We must remind that this configuration is very insecure. It will pull all unsigned container images. Or if you want to unconditionally accept an image from a specific registry, you just need to change the policy of the member matching the registry under the corresponding scope to `insecureAcceptAnything`.
+
+### sigstore config
+
+The configuration file that specifies the container image sigstore is specified as all `.yaml` files in /etc/containers/registries.d directory. After the policy.json allowlist check is completed, the entries in the sigstore configuration file will claim where the container image is located, and the signature claim file of the container image will be downloaded from the specified URL (a webserver or a local directory). The sample sigstore configuration file matching the sample policy.json file in the above is as follows:
```
c3d, line 262: Since you wrote "all `.yaml` files", I suspect the sample file below could also be split into three YAML files, right?

Is there any reference for this `yaml` format? Would it accept `oci` or `tarball` like the `policy.json` file?

c3d: Is there any reference for this `yaml` format? Would it accept `oci` or `tarball` like the `policy.json` file?

c3d: What are the accepted transports for URLs? I see `file` and https`, but would `http` be accepted?

c3d, line 262: /etc/containers/registries.d -> `/etc/containers/registries.d`

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+    sigstore: https://my-sigstore.com/example/sigstore
+  registry.access.redhat.com:
+    sigstore: https://access.redhat.com/webassets/docker/content/sigstore
+```
+
+If there is no `.yaml` file in the /etc/containers/registries.d/ directory when image-rs want to read it, as the way to obtain policy.json, image-rs should call the GetResource grpc service provided by the attestation-agent. After remote attestation, it is obtained from the user controlled relying KBS through the encrypted channel.
+
+### Image signature verification
+
+After completing the white list check and signature file download, signature verification is divided into the following steps:
```
c3d, line 278: "white list"? So refreshingly XXth century ;-)

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+
+- repository host: for example, `docker.io`, matches all images from the docker.io repository.
+
+- subdomain name matching: such as `*.temporary-project.example.com`. Note: wildcards can only appear in the top sub domain name. Wildcards like `example.*.com` are not allowed to be specified.
+
+For the "dir" transport type, the scope format is the directory path. The path must be absolute and contain no symbolic links. The default scope is "".
+
+The policy file must be generated and maintained by the user. Additional security means need to be used to ensure that the policy file will not be modified by malicious entities during storage and transmission. If users have high requirements for security level, they even need to use additional means to ensure the confidentiality of policy files.
+
+If image-rs finds that there is no policy.json in the /etc/containers/ directory or If the policy.json file is empty, call the GetResource grpc service provided by the attestation-agent. After remote attestation, obtain it from the user controlled relying KBS through the encrypted channel.
```
c3d, line 183: First use of KBS acryonym. Please spell as [Key Broker Service (KBS)](https://github.com/confidential-containers/attestation-agent/blob/main/docs/IMPLEMENTATION.md)

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+
+### Image signature verification
+
+After completing the white list check and signature file download, signature verification is divided into the following steps:
+
+1. Parse the signature claim file and separate the JSON payload and cryptographic signature.
+
+2. Read the policy.json's "signedBy" policy, get the public key used to verify the signature from the "keyPath" or "keyData" field.
+
+3. Use the public key to verify the cryptographic signature in the signature claim file, and then compare whether the manifest digest in the JSON payload is consistent with that of the container image and whether the docker reference in the JSON payload is consistent with that of the container image.
```
c3d, line 284: I would split the second half as a separate step for simplicity, and then split that in two:

4. Compare whether:
  * the manifest digest in the JSON payload is consistent with that of the container image
  * the docker reference in the JSON payload is consistent with that of the container image.

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+
+### Image decryption
+
+After verifying the container image signature, will start the actual image pulling. When pulling the image layer by layer, if encounter an encrypted image layer, image-rs calls the container image cryptography operation library provided by ocicrypt-rs to decrypt the encrypted image layer as follows:
```
c3d, line 288: "will start": missing subject (who will?)

c3d, line 288: if encounter: missing subject to "encounter" (image pull service?)

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+
+1. Read the layer's two annotations fields in the image manifest. The `org.opencontainers.image.enc.keys.provider.attestation` field is the annotation packet mentioned above, which contains `PrivateLayerBlockCipherOptions` encrypted by the user's private key. The symmetric key used to encrypt this layer is recorded in this `PrivateLayerBlockCipherOptions`.
+2. ocicrypt-rs calls the keyprovider grpc service provided by the attestation-agent and sends the annotation packet to the attestation-agent.
+3. The attestation-agent performs HW-TEE remote attestation with the relying KBS deployed by the remote user. After passing the attestation, it interacts with KBS through the encrypted channel, and decrypts the `PrivateLayerBlockCipherOptions` using the user's public key in the KBS.
```
c3d, line 292: First use of HW-TEE acronym, expand and add link like for KBS above.

c3d: "after passing the attestation" is a bit ambiguous, since "pass" can mean both "passing some test successfully" or "passing some information". I would just write "after attestation".

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+4. ocicrypt-rs obtains the decrypted `PrivateLayerBlockCipherOptions` from the attestation-agent, reads the symmetric key used to decrypt the container image, and completes the decryption of the container image layer.
+
+## Attestation Agent
+
+The attestation-agent is an indispensable core component in the confidential containers architecture. It undertakes the trust distribution function of the confidential container. In the process of signature verification and decryption of the protected image, the attestation-agent serves as the source of the user's confidential information and the decryptor of the `PrivateLayerBlockCipherOptions`. As mentioned above, it provides two gRPC services for image-rs and ocicrypt-rs to support the above two functions.
```
c3d, line 297: Note the comment I wrote above that you seem to imply that the attestation agent would also be used for encryption. I don't think this is really what you meant. But this highlights that we need to discuss the encryption tools somewhere, and give some command-line examples of how to produce the encrypted images, signatures, etc.

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+{
+    "name":"example_resource",
+    "informations":{
+        "index":"example_index",
+        "extra_param":[]
+    },
+}
+```
+
+In the design of confidential containers V1, when get policy.json through GetResource grpc endpoint, the contents of ResourceDescription are as follows:
```
c3d, line 333: "when get" -> "when getting"

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+
+```json
+{
+    "name":"policy.json",
```
c3d, line 337: I don't understand that format. Where does `policy.json` data actually resides?

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+    rpc GetResource(getResourceRequest) returns (getResourceResponse) {};
+}
+```
+
+ResourceDescription is a JSON string used to describe resources, and its format is defined as follows:
+
+```json
+{
+    "name":"example_resource",
+    "informations":{
```
c3d, line 326: When you transfer a resource, the source actually was something like a file (e.g. `policy.json`). Does the `ResourceDescription` give you a URL to that file, or the data itself embedded in the JSON?

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+
+The GetResource grpc service interface provided by the attestation agent is used to obtain various confidential resources from the relying party (KBS). In the process of protected image deployment, it is usually used to obtain policiy.json and sigstore config file. 
+
+Attestation-agent performs HW-TEE attestation to the KBS, establish an encrypted channel after negotiating the key, and download the required confidential resources through the encrypted channel. The GetResource grpc service interface protobuf of the attachment agent is defined as follows:
+
+```protobuf
+message getResourceRequest {
+    string KbcName = 1;
+    string KbsUri = 2;
+    string ResourceDescription = 3;
```
c3d, line 309: What is `ResourceDescription` used for? In the examples below, it seems mostly empty.

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+
+### GetResource service
+
+The GetResource grpc service interface provided by the attestation agent is used to obtain various confidential resources from the relying party (KBS). In the process of protected image deployment, it is usually used to obtain policiy.json and sigstore config file. 
+
+Attestation-agent performs HW-TEE attestation to the KBS, establish an encrypted channel after negotiating the key, and download the required confidential resources through the encrypted channel. The GetResource grpc service interface protobuf of the attachment agent is defined as follows:
+
+```protobuf
+message getResourceRequest {
+    string KbcName = 1;
```
c3d, line 307: What is `KbcName` used for? Is that for logging / debugging purpose, or does it play a security role?

docs/ccv1_image_security_design.md
c3d: policy.json -> `policy.json` 😄

docs/ccv1_image_security_design.md
c3d: > I don't understand that format. Where does `policy.json` data actually resides?

I think I was confused and initially thought this was part of the response and not part of the request.

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+    "annotation": #annotation-packet,
+}
+```
+
+The output JSON string format is:
+
+```json
+{
+    "keyunwrapresults": {
+        "optsdata": #PrivateLayerBlockCipherOptions,
```
c3d, line 399: Do you wrap this in Base64 or something?

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,403 @@
...
+}
+```
+
+The input JSON string format is:
+
+```json
+{
+    "op":"keyunwrap",
+    "keyunwrapparams":{
+        "dc":{
```
c3d, line 381: What does `dc` stand for?

docs/ccv1_image_security_design.md
c3d: `optsdata` is a bit of a weird name. What about `PrivateLayerBlockCipherOption` for clarity?

c3d (approved):
@jialez0 Thanks a lot for the huge work. I made tons of comments. Nothing blocking. Quite to the contrary, I think you did an outstanding job putting this together. So feel free to address them my comments in later commits.

docs/ccv1_image_security_design.md
jodh-intel: +1. Pass can also mean to skip or ignore, so it's arguably a (semi-)auto-antonym.

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,399 @@
...
+}
+```
+
+The specific format of the annotation packet can be determined by the key provider program used by the user, but it is necessary to ensure that the key provider program used by the decryptor (i.e. the attestation-agent integrated with the specified Key Broker Client) supports parsing annotation packets of the same format.
+
+After completing the above work, Base64 encode the `PublicLayerBlockCipherOptions` and annotation packet, and write them into the manifest.json of the container image as the annotation field of the layer, the former is placed in the `org.opencontainers.image.enc.pubopts` field and the latter in the `org.opencontainers.image.enc.keys.provider.attestation-agent` field. Finally, add the "+encrypted" suffix to the mediaType field of this layer. For example, the container image manifest with an encrypted layer is as follows:
+
+```json
+{
+    "schemaVersion": 2,
```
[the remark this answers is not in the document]
jialez0: This is the version declaration field in the manifest defined by the OCI image format standard. Please refer to [here](https://github.com/opencontainers/image-spec/blob/693428a734f5bab1a84bd2f990d92ef1111cd60c/manifest.md#image-manifest-property-descriptions)

docs/ccv1_image_security_design.md
jialez0: cc @jiazhang0

docs/ccv1_image_security_design.md
jialez0: When deploying an image, the signature verification action occurs before the decryption action of the image layer. Therefore, when creating an image, the signature should be performed after encryption. Therefore, in fact, the manifest and reference as the signature object are for the encrypted container image.

docs/ccv1_image_security_design.md
jialez0: The format of JSON payload is defined by the image library. Please refer to this [document](https://github.com/containers/image/blob/main/docs/containers-signature.5.md)

docs/ccv1_image_security_design.md
jialez0: As above.

docs/ccv1_image_security_design.md
jialez0: The signature claim file contains JSON payload, cryptographic signature itself and some other signature information.

docs/ccv1_image_security_design.md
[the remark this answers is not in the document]
jialez0: The key provider program used for encryption does not refer to ocicrypt rs or attestation agent, but a program customized by the owner. However, when generating `PrivateLayerBlockCipherOptions`, the name signature of the key provider program must be consistent with the name of the key provider program used during decryption, that is, the name of "attachment agent".

docs/ccv1_image_security_design.md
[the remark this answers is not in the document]
c3d: @jodh-intel wrote:
>To counter these attack vectors, presumably you would want to as a minimum only ever create the secure images on a fully trusted host

Ah, now I get your point. But you could be in the same host in a different VM you own, right? So what you really mean is that you want the image to be encrypted in a owner-trusted environment, and that this should not be done on a system that cannot be considered trusted, including the host itself.

docs/ccv1_image_security_design.md
jialez0: @c3d The encrypted version and the non encrypted version of the container image are identical to two completely different images in the perspective of the registry. Only the owner knows the relationship between them. The owner can distinguish them by naming the container image. For example, the encrypted busybox:latest can be named busybox-enc-byxxx:latest.

docs/ccv1_image_security_design.md
jialez0: This configuration is mainly used to skip the image signature verification stage. For some users, only the encryption container image may meet their security requirements. For these users, our signature verification system needs to provide them with sufficient configuration flexibility.

docs/ccv1_image_security_design.md
c3d: @jodh-intel To me, the guarantee that you have in that case is that you won't be able to execute anything in that image before you decrypted the encrypted layers, which requires data from the attestation itself. With only signed layers, at least conceptually, you could run the image without going through attestation.

docs/ccv1_image_security_design.md
c3d: @jialez0 Thanks for the clarification. Since it looks like both @jodh-intel and myself were equally confused by the wording above, let me suggest to add exactly the explanation you just gave after the text above?

docs/ccv1_image_security_design.md
c3d: @jialez0 Ah, I see, so a totally different name. OK, thanks.

docs/ccv1_image_security_design.md
fitzthum: Because we don't explicitly verify the identity of the KBS, signature verification is only guaranteed by the successful decryption of an encrypted layer. In other words, your workload is secure only if its execution is gated by the secret. Note that this is scoped at the pod level, so you can trust signed but unencrypted containers as long as something in the pod depends on depends on the secret. This should be a pretty good fit to workloads that use some containers that access secret data and some generic containers or one container with some generic layers and some secret layers. 

Also note that we do signature verification at the container level, so I don't think we can have a mix of signed and unsigned layers within one container.

jiazhang0: @c3d @jodh-intel 

Hopefully this explanation can help you to understand why image encryption is useful: https://github.com/confidential-containers/documentation/issues/18 《What image encryption means for CC？》. Briefly, image encryption can enforce to inject an explicit authorization during CC bootstrap through Attestation Evidence for Authorization, causing a Pod has to request Relying Party to verify its TCB state.

jiazhang0: > I'm also unclear about images containing both encrypted and unencrypted layers. How can CC guarantee safety in the following sort of scenario?:
> layer 	encrypted? 	signed? 	dependent layer
> 1 	yes 	yes 	n/a
> 2 	no 	no 	n/a
> 3 	yes 	yes 	2

First, image signing takes its effect on image manifest digest (in other word, the entire image) rather than an individual layer. You could refer to https://github.com/confidential-containers/documentation/issues/18 《What container image signing means for CC？》for more explanations about it.

stevenhorsman (approved):
Sorry it took me so long to get around to reviewing this, but it looks like a great initial design. All the questions and comments I noted as I read through it were answered before the end, so a huge thanks to @jialez0 and the other reviewers who have helped refine it.

docs/ccv1_image_security_design.md
jialez0: Because the sigstore may be a public space, the signatures generated by the keys of different users in the same container may be stored in the same sigstore.

docs/ccv1_image_security_design.md
jialez0: > Note that the insecurity is actually somewhat limited for any image with at least one encrypted layer, even for a weak `policy.json`, since the decryption would fail unless the KBS handed over the keys, no?

Exactly :)

docs/ccv1_image_security_design.md
jialez0: > Since you wrote "all `.yaml` files", I suspect the sample file below could also be split into three YAML files, right?

Yes, when trying to parse sigstore config file, the contents of all yaml files in this directory will be merged.

docs/ccv1_image_security_design.md
jialez0: I have added the link of OpenPGP RFC. The current signature system only supports OpenPGP, but encryption can allow various standards.

In the future, when we implement this scheme, we may consider unified management of the signature key and encryption key on the owner side, but whether to unify the two keys may not be the focus of this document.

docs/ccv1_image_security_design.md
jialez0: Yes, policy.json, as the content of boot image, will indeed be measured during the attestation and sent as evidence for attestation.

docs/ccv1_image_security_design.md
jialez0: "registry.access.redhat.com/..." is a registry which implement Docker HTTPS API v2, so is placed in "docker" transport.

docs/ccv1_image_security_design.md
jialez0: The data here is directly transmitted in the form of byte array without Base64 encoding.
The `optsdata` name is defined by the key provider protocol. I think we may should follow it.

docs/ccv1_image_security_design.md
jialez0: Yes, because the service interface of KBS is not standardized, and the communication protocol between KBC and KBS may have high customization requirements, in order to leave enough flexibility for the service implementation of KBS, I think we should separate the resource name and information from the address of KBS

docs/ccv1_image_security_design.md
c3d: @jiazhang0 Thanks for the pointer to the documentation. My question was not about the usefulness of encryption (I have no doubt about it :smile:) but about the weakness of the attestation and workload process if we have only signature.

As I wrote in my comment, the risk is running a workload without a valid attestation when you don't need any secret to decrypt the image. @fitzthum reminded me of one easy method to achieve that, which is to insert a bogus KBS in the loop.

As he also pointed out, image decryption is not the only step that may require a secret. If the workload has some encrypted volume being used only by super-standard containers (e.g. some `mysql`), then arguably signature is sufficient, since what you really want to protect is the data, and that data is not accessible without a secret.

In short, I'm trying to evaluate if supporting a non-encrypted signed image makes sense or not. At the moment, I believe there are valid use cases, as long as it does not break the rest

jiazhang0: > As I wrote in my comment, the risk is running a workload without a valid attestation when you don't need any secret to decrypt the image. @fitzthum reminded me of one easy method to achieve that, which is to insert a bogus KBS in the loop.

To accurately answer this, we need to distinguish the use cases according to boot image strategy. For protected boot image strategy, the TCB state of Pod is trusted. It is safe to verify a signed but unencrypted image (w/o sensitive data) because all artifacts used to verify the signature of image and the verification result are all trusted. For unprotected boot image strategy, the initial TCB state of Pod is unverified, so the initial image to be pulled must be an encrypted, which will enforce an remote attestation to let remote party verify the TCB state of Pod and then unseal the decryption key. Once the TCB state of Pod was verified, the following images to be pulled can be unencrypted as long as it doesn't contain sensitive data.

So the TCB state of Pod is important to conclude whether unencrypted image has weakness or not. Essentially speaking, whether the verification result of signed image is trusted is determined by the TCB state of Pod. If it is unverified or untrusted, the verification result or even whether the signed image is actually occurred is doubtful. The introduction of image encryption can introduce an opportunity to check the TCB stage of Pod.

> In short, I'm trying to evaluate if supporting a non-encrypted signed image makes sense or not. At the moment, I believe there are valid use cases, as long as it does not break the rest

Yes. For sidecar containers, it doesn't contain sensitive data, so it is not necessary to be encrypted.

magowan: We did talk in the one of calls about the possibility in the future of having another component which might have wider awareness of protected v unprotected boot strategy, TCB state of the POD, status of all container images so far with respect to signatures and encryption and where signatures/keys originated from  (i.e. are they all unencrypted so far or has one been decrypted). 
This component can combine all this knowledge to take a view if the Pod has any unaddressed risk that might impact security/confidentiality.

Bottom line is that something somewhere needs to depend upon a key that is either secure in hardware (closer to protected bootimage) or provided by KBS as part of the attestation flow (could be for container decryption but could also be key for data decryption - application won't work without the data). If all these are untrue then we have a problem, if one is true then answer could be "depends"

jiazhang0: > We did talk in the one of calls about the possibility in the future of having another component which might have wider awareness of protected v unprotected boot strategy, TCB state of the POD, status of all container images so far with respect to signatures and encryption and where signatures/keys originated from (i.e. are they all unencrypted so far or has one been decrypted). This component can combine all this knowledge to take a view if the Pod has any unaddressed risk that might impact security/confidentiality.
> 

I guess that component is Root of Chain for Measurement (RTM), consisting of the initial measurement, dynamic measurement or vTPM (w/o dynamic measurement), and event log.

jialez0:
@c3d @jodh-intel Thanks for reviewing! I have refined this doc, now you can take a look again and see if there are any other comments :)

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,509 @@
...
+
+# Production of protected container image
+
+The so-called "protected container image" refers to the container image encrypted and signed by the owner.
+
+**Note**:
+- the encryption process must be carried out before the signing process,
+in other words, when an image is encrypted and signed, 
+the signed object is actually the encrypted container image.
+- The ideally the image creation, encryption and signing need to happen "atomically" and each steps progresses immediately after the last steps completes.
```
c3d, line 25: "The ideally" -> "Ideally"

c3d: "after the last steps" -> "after the previous step" (singular "step")

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,509 @@
...
+(if not so, there is a large window of opportunity for someone to potentially modify the image)
+
+Container image encryption/decryption and signing are currently [under development](https://github.com/opencontainers/image-spec/pull/775). 
+For encryption/decryption, we will rely on the Rust implementation of OCICrypt, ocicrypt-rs,
```
c3d, line 29: `[ocicrypt-rs])(https://github.com/containers/ocicrypt-rs)`

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,509 @@
...
+(if not so, there is a large window of opportunity for someone to potentially modify the image)
+
+Container image encryption/decryption and signing are currently [under development](https://github.com/opencontainers/image-spec/pull/775). 
+For encryption/decryption, we will rely on the Rust implementation of OCICrypt, ocicrypt-rs, 
+which itself is supposed to be compatible with [OCICrypt](https://github.com/containers/ocicrypt). 
+For signing, we will rely on the standard [image library](https://github.com/containers/image) or aim to be compatible with it.
+
+An overview of encrypting and signing a container image is as follows:
+
+![image-1](images/security-1.png)
```
c3d, line 35: Really nice diagram, helps a lot

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,509 @@
...
+          "nonce": "AdcRPTAEhXx6uwuYcOquNA=="
+      }
+  }
+  ```
+ 
+#### Key Wrapping
+
+The `PrivateLayerBlockCipherOptions` needs to be encrypted through the [key provider protocol](https://github.com/containers/ocicrypt/blob/main/docs/keyprovider.md) with a "key provider program", 
+this action is usually called "Key Wrapping".
+the key provider program here also allows the owner to customize the implementation,
```
c3d, line 95: Capitalize first "The"

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,509 @@
...
+as long as it meets the implementation specs described here.
+
+The key provider program should encrypt the `PrivateLayerBlockCipherOptions` with the owner's key.
+After that, it should Base64 encode the encrypted `PrivateLayerBlockCipherOptions` and then package it into a owner customized JSON structure. 
+In the confidential container design, this JOSN structure is called "annotation packet", it record some additional information,
```
c3d, line 100: JOSN -> JSON

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,509 @@
...
+
+The specific format of the annotation packet can be determined by the customized key provider program used by the owner,
+but it is necessary to ensure that the key provider program used by the decryptor 
+(i.e. the attestation-agent integrated with the specified [Key Broker Client](https://github.com/confidential-containers/attestation-agent/blob/main/docs/IMPLEMENTATION.md#kbc)) 
+supports parsing annotation packets of the same format.
+(In CCv1, when decrypting the image, the decryption of `PrivateLayerBlockCipherOptions` will be carried out by ocicrypt-rs through the [key provider protocol with the attestation-agent](https://github.com/confidential-containers/attestation-agent/blob/main/docs/IMPLEMENTATION.md#keyprovider-protocol) (as a key provider program).)
+
+#### Update manifest
+
+The manifest of container image provides two fileds in layer annotation
```
c3d, line 119: fileds -> fields

c3d: "The manifest of container image" -> "The container image manifest"

c3d: What do you mean with "layer annotation" ? I see no annotation in https://github.com/opencontainers/image-spec/blob/main/layer.md, and no layers in https://github.com/opencontainers/image-spec/blob/main/annotations.md. When I search for it, I get https://supercontainers.github.io/containers-wg/ideas/layer-annotation/, but that does not seem to match what you are talking about. So should that just be "annotations"?

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,509 @@
...
+            "annotations": {
+                "org.opencontainers.image.enc.pubopts": "eyJjaXBoZXIiOi...",
+                "org.opencontainers.image.enc.keys.provider.attestation-agent": "eyJwcm9..."
+            }
+        }
+    ]
+}
+```
+
+## Image signing
```
c3d, line 153: All this section much improved, thanks @jialez0

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,509 @@
...
+
+## Image distribution
+
+After the encryption and signing are completed, we obtain two products: 
+the encrypted container image itself and the signature claim file.
+In the design of CC V1, the distribution schemes of the two are as follows:
+
+- container image:
+  - When using the protected boot image scheme, 
+the container image can be directly placed in the rootfs of boot image or push to a remote registry;
```
c3d, line 208: "push to" -> "pushed to"

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,509 @@
...
+else, store in the specified "sigstore" (a webserver or a local dir).
+  - If image is placed in the boot image rootfs:
+Store it in the specified "sigstore" (a customized local dir in boot image rootfs).
+
+
+# Deployment of protected container image
+
+In the confidential container V1 design, security functions will be performed when image-rs pulling the container image, 
+including checking the registry allowlist, verifying the container image signature (if needed), 
+and decrypting the container image layer.
```
c3d, line 224: layer->layers?

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,509 @@
...
+  - If image is placed in the boot image rootfs:
+Store it in the specified "sigstore" (a customized local dir in boot image rootfs).
+
+
+# Deployment of protected container image
+
+In the confidential container V1 design, security functions will be performed when image-rs pulling the container image, 
+including checking the registry allowlist, verifying the container image signature (if needed), 
+and decrypting the container image layer. 
+The steps is as follows:
```
c3d, line 225: "The steps are as follows", or "The sequence is as follows", whichever you prefer.

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,509 @@
...
+2. (if needed in policy) Download the signature claim file of the container image.
+   - If the registry support `X-Registry-Supports-Signatures` API extension, 
+     download from the registry together with image.
+   - Else, download from "sigstore" according to the sigstore config file.
+
+3. (if needed in policy) Verify the signature of the container image according to the configuration of the policy.
+
+4. Pull the container image layer by layer. If an encrypted container image layer is encrypted, decrypt it.
+
+After the image is pulled, the subsequent deployment process is consistent with that of the normal container image.
```
c3d, line 241: That sentence seems to imply that you have to decrypt all layers before applying them. I believe it would be OK to apply each layer immediately after decrypting it (and earlier layers)?

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,509 @@
...
+
+After the matching is successful, the policy set under the corresponding scope is executed, as described in 
+[policy.json Policy Requirements](https://github.com/containers/image/blob/main/docs/containers-policy.json.5.md#policy-requirements),
+the `signedBy` policy requires that there must exist at least one signature of this image 
+can be verified by the public key ring provided by `keyPath` or `keyData` filed,
+so when get this policy, signature verification will be performed.
+
+For the example of the policy file given above, the meaning of it is as follows:
+
+1. For the image from the local file system, allowed to pull and run no matter there is a valid signature or not.
```
c3d, line 307: For the image -> For images

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,509 @@
...
+there must exist at least one signature that can be verified by the keys in `keyData`.
+
+3. For the image from the `registry.access.redhat.com`, 
+there must exist at least one signature that can be verified by the keys in `/etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release`.
+
+4. Reject all other images.
+
+The above example provides a secure and robust policy configuration, 
+but the policy file still provides sufficient configuration flexibility for weak security requirements. 
+For example, if the owner need to completely disable the container image signature verification function
```
c3d, line 319: need -> needs

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,509 @@
...
+If the image is pulled from the local rootfs or the registry does not support the extension, 
+read the sigstore configuration file to get the sigstore URL of the image 
+and then get signatures from the sigstore.
+
+The sigstore configuration file is specified as all `.yaml` files in `registries.d` directory. 
+The entries in the sigstore configuration file will claim the sigstore URL of the container image
+(the sigstore can be a webserver, local directory or etc.).
+
+The spec of sigstore configuration file is defined in 
+[contianers/image registry.d](https://github.com/containers/image/blob/main/docs/containers-registries.d.5.md).
```
c3d, line 353: contianers -> containers

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,509 @@
...
+
+3. Use the public key to verify the cryptographic signature in the signature claim file.
+
+4. Compare whether:
+   - the manifest digest in the JSON payload is consistent with that of the container image
+   - the docker reference in the JSON payload is consistent with that of the container image.
+
+## Image layer decryption
+
+After verifying the container image signature, image-rs will start the actual image pulling.
```
c3d, line 383: image-rs -> `image-rs`

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,509 @@
...
+   - the docker reference in the JSON payload is consistent with that of the container image.
+
+## Image layer decryption
+
+After verifying the container image signature, image-rs will start the actual image pulling. 
+for each encrypted image layer, 
+image-rs calls the container image cryptography operation APIs provided by ocicrypt-rs 
+to decrypt the encrypted image layer. ocicrypt-rs perform the following steps:
+
+1. Read the layer's two annotations fields in the image manifest.
```
c3d, line 388: Hmm, again this idea that there is a per-layer annotation. So I guess that must exist somewhere. A link would help.

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,509 @@
...
+
+# Attestation Agent
+
+The [attestation-agent](https://github.com/confidential-containers/attestation-agent) is an indispensable core component in the confidential containers architecture. 
+It undertakes the trust distribution function of the confidential container. 
+In the process of signature verification and decryption of the protected image, 
+the attestation-agent serves as the source of the owner's confidential information,
+and a key provider program of ocicrypt-rs: it helps ocicrypt-rs to decrypt the `PrivateLayerBlockCipherOptions`. 
+
+Attestation-agent provides two gRPC services at present: "GetResource" and "KeyProvider".
```
c3d, line 407: The attestation agent, or `attestation-agent`.

docs/ccv1_image_security_design.md
jialez0: Annotation here refers to [annotations.md](https://github.com/opencontainers/image-spec/blob/main/annotations.md), annotations can be flexibly placed in many locations in the manifest, such as:

```json
{
    "schemaVersion": 2,
    "config": {
        ...
    },
    "layers": [
        {
            "mediaType": "application/vnd.oci.image.layer.v1.tar+gzip,
            "digest": "sha256:f69ae40...",
            "size": 204812,
        }
    ]
   "annotations": {
                ...
    }
}
```

Or: 

```json
{
    "schemaVersion": 2,
    "config": {
        ...
    },
    "layers": [
        {
            "mediaType": "application/vnd.oci.image.layer.v1.tar+gzip+encrypted",
            "digest": "sha256:f69ae40...",
            "size": 204812,
            "annotations": {
                ...
            }
        }
    ]
}
```

The latter is what we call pre-layer annotations.

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,509 @@
...
+- container image:
+  - When using the protected boot image scheme, 
+the container image can be directly placed in the rootfs of boot image or push to a remote registry; 
+  - When using the unprotected boot image scheme, 
+the owner must push the container image to the remote registry.
+
+- signature claim file: 
+  - If image is pushed to a remote registry: 
+If the registry support `X-Registry-Supports-Signatures` API [extension](https://github.com/containers/image/issues/384), push to the registry together with image,
+else, store in the specified "sigstore" (a webserver or a local dir).
```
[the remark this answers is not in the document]
jialez0: `sigstore` can be a remote server or a specified local directory. The owner decides which one to use. 

For example, in the protected boot image scheme of CC, the local `rootfs` of pod is trusted, so the owner can place the signatures in the `rootfs` of boot image. When verification is needed, you can directly obtain the signatures from the local directory

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,511 @@
+# CCv1 Image Security Design
+
+## Backgrounds
+
+- [CC Security Solution Explained #1](https://github.com/confidential-containers/documentation/issues/18)
+- [Confidential Containers Trust Module](https://github.com/magowan/documentation/blob/TrustModel/TrustModel.md)
```
bpradipt, line 6: nit.. Module=>Model

docs/ccv1_image_security_design.md
```diff
@@ -0,0 +1,511 @@
...
+- [Confidential Containers Thread Module](https://github.com/magowan/documentation/blob/ThreatModel/ThreatModel.md)
```
bpradipt, line 7: nit.. "Thread Module" => "Threat Model"

bpradipt:
@jialez0 @sameo @c3d I think this version should be merged since there is in-principle approval to the design and PR has landed as well. Any subsequent updates can be handled as new PR. Wdyt ?

c3d:
> @jialez0 @sameo @c3d I think this version should be merged since there is in-principle approval to the design and PR has landed as well. Any subsequent updates can be handled as new PR. Wdyt ?

Actually, I had already given my approval on that basis, stating that improvements could be done later.

sameo (approved):

sameo:
@fitzthum @jodh-intel Are you good  with that one?

PR #4, opened 2022-01-19
Implement image-rs basic features

arronwy:
This PR implements image-rs basic features including OCI image download/decryption/decompression/unpack/mount.

Image signing verification is not included.


jiangliu:
As we are early in the development stages, let's merge the PR so others may contribute to it?

data/generate_test_data.md
```diff
@@ -0,0 +1,16 @@
+## Generate keypair and container image for unit and integration tests
+
```
stevenhorsman, line 2: I think it might be worth adding some more instructions here, for example getting the user to log into their container registry then set a shell variable to their user account. e.g. `export cr_user="<your container reg account>`
then updating the skopeo commands to replace `user` with `cr_user`.
I'm not sure if this is supposed to support more that just Docker Hub either?

data/generate_test_data.md
```diff
@@ -0,0 +1,16 @@
+## Generate keypair and container image for unit and integration tests
```
stevenhorsman, line 1: I wonder if it is worth linking this file on the main README.md as part of a set up guide as it's easy to miss?

stevenhorsman (commented):
I tried to check out the code and run the tests but hit some problems and had some questions.

Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+
+[dependencies]
+anyhow = ">=1.0"
+flate2 = "1.0"
+futures-util = "0.3"
+libc = "0.2"
+nix = "0.23.0"
+oci-distribution = { git = "https://github.com/arronwy/oci-distribution", branch = "export_pull_layer" }
+oci-spec = { git = "https://github.com/containers/oci-spec-rs" }
+ocicrypt-rs = { git = "https://github.com/arronwy/ocicrypt-rs", branch = "oci_distribution" }
```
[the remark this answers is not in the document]
arronwy: Yes, there are two pending PR in oci-distribution to support manifest list and public pull layer API, after these two PRs are merged, we'll use the upstream repo.

src/decrypt.rs
```diff
@@ -0,0 +1,94 @@
...
+}
+
+fn decrypt_layer_data(
+    encrypted_layer: &[u8],
+    descriptor: &OciDescriptor,
+    crypto_config: &CryptoConfig,
+) -> Result<Vec<u8>> {
+    if let Some(decrypt_config) = &crypto_config.decrypt_config {
+        let (layer_decryptor, _dec_digest) =
+            decrypt_layer(decrypt_config, encrypted_layer, descriptor, false)?;
```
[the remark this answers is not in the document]
arronwy: Yes, I'll add the keyprovider config file too and the unit test can use them.

data/generate_test_data.md
arronwy: Yes, we can add more detail info for the setup. It should support different registries, I tried with a local registry and Docker Hub, encrypted container image both works, but the server side may change the `dest-compress-format`.

src/decrypt.rs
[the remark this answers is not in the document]
arronwy: You can check the CI results: https://github.com/arronwy/image-rs/actions/runs/1729173936

stevenhorsman:
> > I tried to check out the code and run the tests but hit some problems and had some questions.
> 
> Hi @stevenhorsman, Thanks for your try, I just added the keyprovider config to the repo and the CI passed: https://github.com/arronwy/image-rs/actions/runs/1729173936
> 
> BTW for `test_pull_image` since the final mount operation need root privilege, we need test with sudo command.

Thanks for the info - and adding the GHA CI output link. I'm seeing better behaviour locally now, but still not quite right.
If I run the `test_pull_image` on it's own it works:
```
sudo -E PATH=$PATH -s cargo test image::tests::test_pull_image
    Finished test [unoptimized + debuginfo] target(s) in 0.19s
     Running unittests (/root/go/src/github.com/confidential-containers/image-rs/target/debug/deps/image_rs-9b1dcdc702e8e38c)

running 1 test
test image::tests::test_pull_image ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 8 filtered out; finished in 16.18s
```
but if I run all the tests it fails (differently from before):
```
sudo -E PATH=$PATH -s cargo test
    Finished test [unoptimized + debuginfo] target(s) in 0.18s
     Running unittests (/root/go/src/github.com/confidential-containers/image-rs/target/debug/deps/image_rs-9b1dcdc702e8e38c)

running 9 tests
test config::tests::test_image_config ... ok
test decoder::tests::test_uncompressed_decode ... ok
test config::tests::test_image_config_from_file ... ok
test decoder::tests::test_zstd_decode ... ok
test decoder::tests::test_gzip_decode ... ok
test bundle::tests::test_bundle_create_config ... ok
test unpack::tests::test_unpack ... ok
test pull::tests::test_pull_client ... ok
test image::tests::test_pull_image ... FAILED

failures:

---- image::tests::test_pull_image stdout ----
thread 'image::tests::test_pull_image' panicked at 'assertion failed: image_client.pull_image(image, bundle_dir.path(), &None, &None).await.is_ok()', src/image.rs:222:13
```
I tried single threading the tests to see if that helped, but it still failed, so it doesn't look like it was a synchronous issue.

It is super late for you, so I'm not expecting anything now, I just wanted to share what I saw. Thanks!

jiangliu:
> > > I tried to check out the code and run the tests but hit some problems and had some questions.
> > 
> > 
> > Hi @stevenhorsman, Thanks for your try, I just added the keyprovider config to the repo and the CI passed: https://github.com/arronwy/image-rs/actions/runs/1729173936
> > BTW for `test_pull_image` since the final mount operation need root privilege, we need test with sudo command.
> 
> Thanks for the info - and adding the GHA CI output link. I'm seeing better behaviour locally now, but still not quite right. If I run the `test_pull_image` on it's own it works:
> 
> ```
> sudo -E PATH=$PATH -s cargo test image::tests::test_pull_image
>     Finished test [unoptimized + debuginfo] target(s) in 0.19s
>      Running unittests (/root/go/src/github.com/confidential-containers/image-rs/target/debug/deps/image_rs-9b1dcdc702e8e38c)
> 
> running 1 test
> test image::tests::test_pull_image ... ok
> 
> test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 8 filtered out; finished in 16.18s
> ```
> 
> but if I run all the tests it fails (differently from before):
> 
> ```
> sudo -E PATH=$PATH -s cargo test
>     Finished test [unoptimized + debuginfo] target(s) in 0.18s
>      Running unittests (/root/go/src/github.com/confidential-containers/image-rs/target/debug/deps/image_rs-9b1dcdc702e8e38c)
> 
> running 9 tests
> test config::tests::test_image_config ... ok
> test decoder::tests::test_uncompressed_decode ... ok
> test config::tests::test_image_config_from_file ... ok
> test decoder::tests::test_zstd_decode ... ok
> test decoder::tests::test_gzip_decode ... ok
> test bundle::tests::test_bundle_create_config ... ok
> test unpack::tests::test_unpack ... ok
> test pull::tests::test_pull_client ... ok
> test image::tests::test_pull_image ... FAILED
> 
> failures:
> 
> ---- image::tests::test_pull_image stdout ----
> thread 'image::tests::test_pull_image' panicked at 'assertion failed: image_client.pull_image(image, bundle_dir.path(), &None, &None).await.is_ok()', src/image.rs:222:13
> ```
> 
> I tried single threading the tests to see if that helped, but it still failed, so it doesn't look like it was a synchronous issue.
> 
> It is super late for you, so I'm not expecting anything now, I just wanted to share what I saw. Thanks!

Most code is still assuming Linux. We may need a plan to support other OS .

stevenhorsman:
> Most code is still assuming Linux. We may need a plan to support other OS .

Sorry, I wasn't clear - this result in on my Ubuntu 20.04 system, not macOS

arronwy:
> >   
> > I tried single threading the tests to see if that helped, but it still failed, so it doesn't look like it was a synchronous issue.
> > It is super late for you, so I'm not expecting anything now, I just wanted to share what I saw. Thanks!
> 
> Most code is still assuming Linux. We may need a plan to support other OS .

Yes, we can add other OS support to the TODO list and add this info to the README

arronwy:
> Yes, we can add other OS support to the TODO list and add this info to the README



> > Most code is still assuming Linux. We may need a plan to support other OS .
> 
> Sorry, I wasn't clear - this result in on my Ubuntu 20.04 system, not macOS

The root cause maybe `image-rs` store the layers data to default work dir, during pull image it will check whether image layers data already downloaded, and `remove_image` API is not implemented yet:
 ```
const DEFAULT_WORK_DIR: &str = "/var/lib/image-rs/";
```
I updated to use tempdir for the work dir, it works with multiple times cargo test now.

**Comment by jiangliu** [comment]
> > Yes, we can add other OS support to the TODO list and add this info to the README
> 
> > > Most code is still assuming Linux. We may need a plan to support other OS .
> > 
> > 
> > Sorry, I wasn't clear - this result in on my Ubuntu 20.04 system, not macOS
> 
> The root cause maybe `image-rs` store the layers data to default work dir, during pull image it will check whether image layers data already downloaded, and `remove_image` API is not implemented yet:
> 
> ```
> const DEFAULT_WORK_DIR: &str = "/var/lib/image-rs/";
> ```
> 
> I updated to use tempdir for the work dir, it works with multiple times cargo test now.

It would be better to use `tempfile` crate to create tmpdir for tests, but be careful that the downloaded image may cost too much memory when tmpfs is used.

**Comment by arronwy** [comment]
> > 
> > The root cause maybe `image-rs` store the layers data to default work dir, during pull image it will check whether image layers data already downloaded, and `remove_image` API is not implemented yet:
> > ```
> > const DEFAULT_WORK_DIR: &str = "/var/lib/image-rs/";
> > ```
> >   
> > I updated to use tempdir for the work dir, it works with multiple times cargo test now.
> 
> It would be better to use `tempfile` crate to create tmpdir for tests, but be careful that the downloaded image may cost too much memory when tmpfs is used.

Yes, we use `tempfile` crate for all file related tests now. Thanks for the reminder, we use small size container image like busybox for current test cases. The following benchmark test will use bigger container image, we all make sure it works with current CI.

**Comment by stevenhorsman** [comment]
> Yes, we use `tempfile` crate for all file related tests now. Thanks for the reminder, we use small size container image like busybox for current test cases. The following benchmark test will use bigger container image, we'll ensure it works with current CI.

Thanks for the changed. Just to confirm that I can run `sudo -E PATH=$PATH -s cargo test` multiple times now and the tests all pass! 👍

File: src/snapshots/overlay.rs [hunk 5]
```diff
@@ -0,0 +1,76 @@
...
+        })
+    }
+
+    fn unmount(&self, mount_point: &MountPoint) -> Result<()> {
+        nix::mount::umount(mount_point.mount_path.as_path())?;
+        if mount_point.work_dir.exists() {
+            fs::remove_dir_all(&mount_point.work_dir)?;
+        }
+        Ok(())
+    }
```
dubek, line 75: this function doesn't use `self` at all. I wonder if it makes sense to move it as a function on MountPoint itself.

Also I found no calls to `unmount()`, but I might missed something.

src/unpack.rs
```diff
@@ -0,0 +1,132 @@
...
+                dirs.insert(path, times);
+            } else {
+                let ret = unsafe { libc::lutimes(path.as_ptr(), times.as_ptr()) };
+                if ret != 0 {
+                    return Err(anyhow!(
+                        "change symlink file: {:?} utime error: {:?}",
+                        path,
+                        io::Error::last_os_error()
+                    ));
+                }
```
dubek, line 59: 1. This code appears also below; consider extracting it into a function (set_file_timestamps ?)
2. Why do we bother changing the atime? Wouldn't it be more "correct" to leave it as-is (= "now")? (asking because I'm not sure about this)
3. filetime crate has set_symlink_file_times. Why don't we use it?

src/snapshots/overlay.rs
arronwy: @dubek Thanks for reviewing. Yes, currently we only implement the `pull_image` API which will setup the bundle, `remove_image` is not implemented yet, that API will call `unmount()` operation.

Since `SnapShot` trait need support diffent types which have different requirements for umount and clean operations,  for `OverLay` type, it is not needed. Also if this function can be shared by multiple snapshots, we can move it as function.

src/unpack.rs
arronwy: > 1. This code appears also below; consider extracting it into a function (set_file_timestamps ?)

we have two set time operations one is utime and another is lutime for symlink and each only call once, it make sense to extract into a function if it used at multiple places.

> 2. Why do we bother changing the atime? Wouldn't it be more "correct" to leave it as-is (= "now")? (asking because I'm not sure about this)

For container image, we need keep it consistent with the state when the image is packed, this is the common way for the image service.

> 3. filetime crate has set_symlink_file_times. Why don't we use it?

Yes, we can use filetime, the reason is to simplify the dependency,  maybe we can swith to filetime crate when we start working on support different platforms.

src/pull.rs
```diff
@@ -0,0 +1,234 @@
...
+
+                client
+                    .pull_layer(reference, &layer.digest, &mut layer_data)
+                    .await?;
+
+                let mut layer_meta = LayerMeta::default();
+                let mut media_type_str: &str = layer.media_type.as_str();
+
+                let mut decryptor = Decryptor::default();
+                if decryptor.is_encrypted(&layer.media_type) {
```
fitzthum, line 100: One of the issues we ran into with `skopeo` was that if a decryption key was provided but the image was not encrypted, the image would be unpacked anyway. For CCv0 this meant that if the CSP intercepted an encrypted image and replaced it with an unencrypted image, `skopeo` would unpack the unencrypted image inside the enclave without warning the user.

Does `image-rs` do something similar? Here it looks like the format of the image determines whether the image is decrypted or not. Do we check whether or not encryption is expected from by the client?

data/ocicrypt_keyprovider.conf
```diff
@@ -0,0 +1,7 @@
+{
+        "key-providers": {
+                "sample_kbs": {
```
fitzthum, line 3: Doesn't the `sample_kbs` use `attestation-agent` as the keyprovider name?

fitzthum (commented):
General comments. Looks good overall.

src/unpack.rs
dubek: > > 1. This code appears also below; consider extracting it into a function (set_file_timestamps ?)
> 
> we have two set time operations one is utime and another is lutime for symlink and each only call once, it make sense to extract into a function if it used at multiple places.

You're right, I missed that. I think that `lutimes` should work for both cases. But the existing code is fine anyway.

> 
> > 2. Why do we bother changing the atime? Wouldn't it be more "correct" to leave it as-is (= "now")? (asking because I'm not sure about this)
> 
> For container image, we need keep it consistent with the state when the image is packed, this is the common way for the image service.

OK.

> 
> > 3. filetime crate has set_symlink_file_times. Why don't we use it?
> 
> Yes, we can use filetime, the reason is to simplify the dependency, maybe we can swith to filetime crate when we start working on support different platforms.

Yes you're right, if you only need these two functions then no need to depend on the filetime crate.

dubek (commented):
small comment:

data/private_key.pem
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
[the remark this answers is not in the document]
arronwy: Thanks, good suggestion, I'll also change the directory name `data` to `test_data`.

src/pull.rs
arronwy: Yes, current PR for `image-rs` do the same way as `skopeo`, and after we pulled the image, we will return the image and layer metadata info like whether the image is signed and layer is encrypted. 

The next step work for integrating `signature` crate which also provide the policy info, we will do these verifications to ensure the image/layer info is comply with the policy.

data/ocicrypt_keyprovider.conf
arronwy: Yes, I use the old `sample_kbs` branch from jiale, I found latest `sample_kbs` Readme all use `attestation-agent` as the keyprovider name. Thanks, I'll fix that.

src/pull.rs
fitzthum: Ok, let's just make sure we enforce the guest owner's policy at some point, even for unsigned images.

src/unpack.rs
```diff
@@ -0,0 +1,132 @@
...
+
+    fs::create_dir_all(destination)?;
+
+    let mut dirs: HashMap<CString, [timeval; 2]> = HashMap::default();
+    for file in archive.entries()? {
+        let mut file = file?;
+        file.unpack_in(destination)?;
+
+        // tar-rs crate only preserve timestamps of files,
+        // symlink file and directory are not covered.
```
jodh-intel, line 33: Is this https://github.com/alexcrichton/tar-rs/pull/217? If so, I'd consider adding a URL in the comments.

jodh-intel (commented):
Thanks @arronwy. A few comments...

src/image.rs
```diff
@@ -0,0 +1,234 @@
...
+}
+
+/// The`image-rs` client whill support OCI image
+/// pulling, image signing verfication, image layer
+/// decryption/unpack/store and management.
+pub struct ImageClient {
+    /// The config for `image-rs` client.
+    pub config: ImageConfig,
+
+    /// The meta data database for `image-rs` client.
```
sameo, line 70: Nit again: `metadata`, in one word.

src/meta_store.rs
```diff
@@ -0,0 +1,31 @@
...
+}
+
+impl MetaStore {
+    /// load `MetaStore` from a local file
+    pub fn from_file(path: &str) -> Result<MetaStore> {
+        let file = File::open(path)
+            .map_err(|e| anyhow!("failed to open metastore file {}", e.to_string()))?;
+        serde_json::from_reader::<File, MetaStore>(file)
+            .map_err(|e| anyhow!("failed to parse metastore file {}", e.to_string()))
+    }
```
sameo, line 30: How do we store/serialize the store?

src/image.rs
```diff
@@ -0,0 +1,234 @@
...
+            return Err(anyhow!("unsupport OS image {:?}", image_config.os()));
+        }
+
+        create_config(&image_config, bundle_dir)?;
+        let image_id = image_data.id.clone();
+        self.meta_store
+            .lock()
+            .await
+            .image_db
+            .insert(image_data.id.clone(), image_data);
```
sameo, line 190: Do we need to clone twice?

sameo (commented):
Most of my comments are minor. This PR is very well laid out and easy to review, thanks a lot
@arronwy.

One overall comment: Could you please add some more detailed commit messages?

src/image.rs
```diff
@@ -0,0 +1,234 @@
...
+    pub snapshots: HashMap<String, Box<dyn SnapShot>>,
+}
+
+impl Default for ImageClient {
+    // construct a default instance of `ImageClient`
+    fn default() -> ImageClient {
+        let config = ImageConfig::default();
+
+        let mut meta_store = MetaStore::default();
+        if let Ok(ms) = MetaStore::from_file(METAFILE) {
```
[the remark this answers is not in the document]
arronwy: Yes, it will be much simpler, I also found `unwrap_or_default` can do the same thing.

src/decrypt.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    /// get_plaintxt_layer descrypts encrypted_layer data and return the
+    /// plaintxt_layer data. descriptor and crypto_info are required for
+    /// layer data decryption process.
+    pub async fn get_plaintxt_layer(
+        &mut self,
+        media_type: &str,
+        descriptor: &OciDescriptor,
+        encrypted_layer: Vec<u8>,
+        crypto_info: &str,
```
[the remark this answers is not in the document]
arronwy: Yes, I'll add the description for `crypto_info` in the comments.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,50 @@
...
+          override: true
+          components: rustfmt, clippy
+
+      - name: Run cargo build
+        uses: actions-rs/cargo@v1
+        with:
+          command: build
+          args: --all-features
+
+      - name: Run cargo test with root privilege
```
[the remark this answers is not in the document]
jodh-intel: I'd add a comment to make that clear. Also, is it worth running the unit tests twice as we do in Kata, once as root and once as non-root? That would allow you to assert that the mount tests (correctly) failed when not running as root for example.

src/unpack.rs
```diff
@@ -0,0 +1,132 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use filetime;
+    use std::fs::File;
+    use std::io::prelude::*;
+    use tempfile;
+
+    #[test]
+    fn test_unpack() {
```
jodh-intel, line 88: Given the comments about the current limitations of `tar-rs`, shouldn't this also test sym-links? Also, what about other file types like char+block devices? I'm just wondering if we need tests for those here or can we rely on `tar-rs` for those tests?

src/decrypt.rs
```diff
@@ -0,0 +1,94 @@
...
+
+#[derive(Default, Clone)]
+pub struct Decryptor {
+    /// The layer original media type before encryption.
+    pub media_type: String,
+}
+
+impl Decryptor {
+    /// Check whether media_type is encrypted.
+    pub fn is_encrypted(&mut self, media_type: &str) -> bool {
```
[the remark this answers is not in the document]
arronwy: Yes, this is the right way, fixed now.

src/decrypt.rs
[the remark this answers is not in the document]
arronwy: all changed to decrypt_config as parameter name to keep consistent and added the description in comments.

src/meta_store.rs
```diff
@@ -0,0 +1,31 @@
+use anyhow::{anyhow, Result};
+use serde::Deserialize;
+use std::collections::HashMap;
+use std::fs::File;
+
+use crate::image::{ImageMeta, LayerMeta};
+
+pub const METAFILE: &str = "meta_store.json";
+
+/// `image-rs` configuration information.
```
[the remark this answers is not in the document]
arronwy: my fault, fixed now.

src/snapshots/overlay.rs
arronwy: Yes, I updated the `unmount` function and remove the `remove_dir_all` operations

src/config.rs
```diff
@@ -0,0 +1,91 @@
...
+use std::path::{Path, PathBuf};
+
+const DEFAULT_WORK_DIR: &str = "/var/lib/image-rs/";
+const DEFAULT_SNAPSHOT: &str = "overlay";
+
+/// `image-rs` configuration information.
+#[derive(Clone, Debug, Deserialize)]
+pub struct ImageConfig {
+    /// The location for `image-rs` to store data.
+    pub work_dir: PathBuf,
```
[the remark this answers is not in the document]
arronwy: `work_dir` will be created if not existed, it can be relative or sym links, we did not enforce any policy or check for this    directory.

src/unpack.rs
arronwy: The reason we did not test sym-links is current `tar-rs` crate packing feature will not preserve the file time info, we may need a sepeate tool to crate the test file, I have add it as TODO for sym-links and other file types.

src/config.rs
```diff
@@ -0,0 +1,91 @@
...
+
+    /// The default snapshot for `image-rs` to use.
+    pub default_snapshot: String,
+}
+
+impl Default for ImageConfig {
+    // Construct a default instance of `ImageConfig`
+    fn default() -> ImageConfig {
+        let work_dir = PathBuf::from(
+            std::env::var("IMAGE_WORK_DIR").unwrap_or_else(|_| DEFAULT_WORK_DIR.to_string()),
```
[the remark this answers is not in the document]
arronwy: This variable allow the user to easy set the work dir without create the config file.
Prefix with `CC_` will be much better.

src/meta_store.rs
arronwy: This will be in to todo list since current API only cover `pull_image` API.

fitzthum (approved):

sameo (approved):

sameo:
@jodh-intel @stevenhorsman If you're ok with that new version, I think we could merge it.

src/config.rs
```diff
@@ -0,0 +1,91 @@
...
+        ImageConfig {
+            work_dir,
+            default_snapshot: DEFAULT_SNAPSHOT.to_string(),
+        }
+    }
+}
+
+impl ImageConfig {
+    /// Load `ImageConfig` from a configuration file
+    pub fn from_file(path: &Path) -> Result<ImageConfig> {
```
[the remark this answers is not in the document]
arronwy: Changed to `TryFrom` now.

src/decrypt.rs
```diff
@@ -0,0 +1,112 @@
...
+    /// layer data decryption process.
+    ///
+    /// * `decrypt_config` - decryption key info in following format:\
+    ///           - \<filename> \
+    ///           - \<filename>:file=\<passwordfile> \
+    ///           - \<filename>:pass=\<password> \
+    ///           - \<filename>:fd=\<filedescriptor> \
+    ///           - \<filename>:\<password> \
+    ///           - provider:<cmd/gprc>
+    pub async fn get_plaintxt_layer(
```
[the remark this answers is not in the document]
arronwy: Yes, thanks, all `plaintxt` typo fixed now.

stevenhorsman (approved):

test_data/generate_test_data.md
```diff
@@ -0,0 +1,33 @@
...
+```
+
+### Create JWE encrypted container image
+```shell
+$ skopeo copy --encryption-key jwe:public_key_for_tests.pem docker://busybox:latest docker://user/busybox_encrypted
+```
+
+### Create KBS encrypted container image
+Follow the [README](https://github.com/confidential-containers/attestation-agent/blob/main/sample_kbs/README.md)
+of attestation agent sample kbs to encrypt the container image and push to remote registry.
```
jodh-intel, line 30: Nit: s/kbs/KBS/

jodh-intel (commented):
Thanks @arronwy. A few nits and a small bug (I think)...

Cargo.toml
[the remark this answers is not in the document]
jiangliu: Any progress about the oci-distribution upstream?

src/bundle.rs
```diff
@@ -0,0 +1,133 @@
...
+
+/// create_runtime_config will create the config.json file under the bundle_path,
+/// and return the final bundle config path.
+pub fn create_runtime_config(
+    image_config: &ImageConfiguration,
+    bundle_path: &Path,
+) -> Result<PathBuf> {
+    let mut spec = Spec::default();
+
+    spec.set_hostname(Some("image-rs".to_string()));
```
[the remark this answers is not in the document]
arronwy: Yes, thanks, will add a `const` `BUNDLE_HOSTNAME`

src/bundle.rs
```diff
@@ -0,0 +1,133 @@
...
+    }
+    if let Some(created) = image_config.created() {
+        annotations.insert(ANNOTATION_CREATED.to_string(), created.to_string());
+    }
+    let mut process = Process::default();
+
+    if let Some(config) = image_config.config() {
+        if let Some(working_dir) = config.working_dir() {
+            process.set_cwd(PathBuf::from(working_dir));
+        }
```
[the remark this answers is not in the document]
arronwy: Yes, and default  `cwd` is set in `Process::default()`
```
            // Sets cwd of process to the container root by default
            cwd: "/".into(),
```

src/config.rs
```diff
@@ -0,0 +1,98 @@
...
+
+    /// The default snapshot for `image-rs` to use.
+    pub default_snapshot: SnapshotType,
+}
+
+impl Default for ImageConfig {
+    // Construct a default instance of `ImageConfig`
+    fn default() -> ImageConfig {
+        let work_dir = PathBuf::from(
+            std::env::var("CC_IMAGE_WORK_DIR").unwrap_or_else(|_| DEFAULT_WORK_DIR.to_string()),
```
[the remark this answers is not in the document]
arronwy: Ok, make sense, will move to top-level.

src/bundle.rs
```diff
@@ -0,0 +1,133 @@
...
+
+        if !args.is_empty() {
+            process.set_args(Some(args));
+        }
+
+        if let Some(labels) = config.labels() {
+            annotations.extend(labels.clone());
+        }
+
+        // TODO: parse image config user info and extract uid from rootfs passwd file
```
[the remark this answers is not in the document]
arronwy: Sure, I'll add a github issue to track this TODO work

test_data/generate_test_data.md
arronwy: thanks, fixed :-)

jodh-intel (approved):
Thanks @arronwy.

lgtm

PR #5, opened 2022-01-20
Initial implementation of image signature verification.

jialez0:
This PR is an initial implementation of image signature verification. Now only support RED HAT image simple signing format.

Note: Currently, this PR does not support:
1. "dir" scope in policy.json.
2. Get signatures from registry X-R-S-S API extension.
3. `https://` and other sigstore url. (only `file://` sigstore url is supported, In other words, only the case where the signature is stored locally is supported).
4. Signature standards other than OpenPGP.


signature/src/policy.rs
```diff
@@ -0,0 +1,825 @@
...
+impl Default for PolicyReferenceMatch {
+    fn default() -> Self {
+        PolicyReferenceMatch::Match(PrmMatch {
+            type_name: PRM_TYPE_MATCH_REPO_DIGEST_OR_EXACT.to_string(),
+        })
+    }
+}
+
+impl PolicyReferenceMatch {
+    pub fn matchs_docker_reference(
```
stevenhorsman, line 277: Sorry as this is pedantic, but `matches_docker_reference` might be a better name here, in PrmMatch, PrmExactReference, PrmExactRepository and PrmRemapIdentity

stevenhorsman (commented):
I'm not the best rust expert, and I'd like to use the library more to be sure that the APIs work, but from what I can see building it locally and skimming through the code, this is a great start, with really good test coverage (>90%). Thanks so much!

signature/src/signatures/sigstore.rs
```diff
@@ -0,0 +1,254 @@
...
+use anyhow::{anyhow, Result};
+use oci_distribution::Reference;
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use std::ffi::OsStr;
+use std::fs;
+
+// FIXME: Now, for Demo, the dir which the sigstore config files locate
+// is hardcoded to "/etc/containers/registry.d".
+const SIGSTORE_CONFIG_DIR: &str = "/etc/containers/registry.d";
```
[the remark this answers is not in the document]
jialez0: > Going forward I guess it is best to check if `$HOME/.config/containers/registries.d` (or `/usr/local/etc/containers/registries.d` on mac) exists before falling back to `/etc/containers/registries.d`?

Yes, this will be the follow-up work :)

signature/src/image.rs
```diff
@@ -0,0 +1,203 @@
...
+
+impl Image {
+    pub fn default_with_reference(image_ref: Reference) -> Self {
+        Image {
+            reference: image_ref,
+            manifest_digest: String::default(),
+            registry_xrss_api_support: false,
+            sigstore_base_url: String::default(),
+            signatures: Vec::default(),
+        }
```
jodh-intel, line 38: This seems to be required all because oci-distribution doesn't derive `Default` on `Reference` :cry: It might be a good idea to raise a PR on that crate to allow this code to be simplified in future?

signature/src/policy.rs
```diff
@@ -0,0 +1,824 @@
...
+        match self {
+            PolicyRequirement::Simple(req) => req.simple_judge_allowed(image),
+            PolicyRequirement::Signedby(req) => req.signature_judge_allowed(image),
+        }
+    }
+}
+
+// The `insecureAcceptAnything` or `reject` policy requirement.
+#[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
+pub struct PrSimple {
```
jodh-intel, line 137: For me, It's very hard to see the letters "pr" and not think "pull request" :smile: But this PR seems to overload the  acronym further since fwics it can mean:

- "policy requirement"
- "policy reference"

I'd strongly consider renaming some of the types to make this distinction clearer and make the code less confusing:

`PolicyReqSimple`, `PolicyReqSignedBy`, `PolicyRefMatch`, etc_?

signature/src/policy.rs
```diff
@@ -0,0 +1,824 @@
...
+            .matches_docker_reference(&origin_ref_case_1, "docker.io/example/busybox:tag")
+            .is_err());
+        assert!(pr_match_case_1
+            .matches_docker_reference(&origin_ref_case_2, origin_ref_str_2)
+            .is_err());
+        /*
+        assert!(pr_match_case_1
+            .matches_docker_reference(&origin_ref_case_3, origin_ref_str_1)
+            .is_ok());
+            */
```
jodh-intel, line 544: Commented out code (here and below).

signature/src/signatures/sigstore.rs
```diff
@@ -0,0 +1,254 @@
...
+use std::fs;
+
+// FIXME: Now, for Demo, the dir which the sigstore config files locate
+// is hardcoded to "/etc/containers/registries.d".
+const REGISTRY_CONFIG_DIR: &str = "/etc/containers/registries.d";
+
+// Format the sigstore name:
+// `image-repository@digest-algo=digest-value`
+#[allow(unused_assignments)]
+pub fn format_sigstore_name(image_ref: &Reference, image_digest: String) -> String {
```
jodh-intel, line 21: What if these parameters are empty/invalid? I think this function should check the values and return a `Result<String>` so that caller can be assured they are valid.

signature/src/signatures/sigstore.rs
```diff
@@ -0,0 +1,254 @@
...
+    ))
+}
+
+pub fn get_sigs_from_lookaside(sigstore: String) -> Result<Vec<Vec<u8>>> {
+    let mut res: Vec<Vec<u8>> = Vec::new();
+
+    // FIXME: Now only support get signatures from local files.
+    // only the url begins with "file://" can be parsed.
+    if sigstore[0..4] == *"file" {
+        let sigstore_dir_path = &sigstore[7..];
```
jodh-intel, line 149: **Bug**: How do you know the string is this many bytes long? If the string is shorter, this will panic and kill the application I'd strongly suggest not parsing URI's ourselves but instead using a crate to handle it.

signature/src/policy.rs
```diff
@@ -0,0 +1,824 @@
...
+    transports: HashMap<String, PolicyTransportScopes>,
+}
+
+pub type PolicyTransportScopes = HashMap<String, PolicyRequirements>;
+
+pub type PolicyRequirements = Vec<PolicyRequirement>;
+
+impl Policy {
+    // Parse the JSON file of policy (policy.json).
+    pub fn from_file(file_path: &str) -> Result<Self> {
```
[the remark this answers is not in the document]
jialez0: This API is called by image rs. using & STR as the type of file path can make the call more direct and simple, for example:
```
let policy = Policy::from_file("/path/to/policy.json");
```
Or:
```
const POLICY_FILE_PATH: &str = "path/to/policy.json";
let policy = Policy::from_file(POLICY_FILE_PATH);
```

jialez0:
Thanks @jodh-intel @stevenhorsman for reviewing! I have resolved the previous comments and let's try to continue the review and try to merge this PR. I proposed the design of modular verification signature function in #6. After this PR is merged, maybe we can create a new PR to realize modularization.

jodh-intel (commented):
Thanks @jialez0. A few comments...

signature/src/image.rs
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    #[test]
+    fn test_get_image_id_and_ns() {
+        #[derive(Debug)]
+        struct TestData<'a> {
+            image_reference: &'a str,
+            image_namespace: Vec<&'a str>,
+        }
```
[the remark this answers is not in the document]
jialez0: These illegal reference inputs have been tested [here](https://github.com/krustlet/oci-distribution/blob/main/src/reference.rs#L341). Maybe we don't need them again here?

signature/src/signatures/verify.rs
```diff
@@ -0,0 +1,299 @@
...
+            // use default match policy: "matchExact".
+            let default_signed_identity = policy::PolicyReferenceMatch::default();
+            default_signed_identity.matches_docker_reference(image_ref, &self.docker_reference())
+        }
+    }
+
+    // Compare wether the manifest digest in the JSON payload
+    // is consistent with that of the container image.
+    pub fn validate_signed_docker_manifest_digest(&self, ref_manifest_digest: &str) -> Result<()> {
+        if self.manifest_digest() != *ref_manifest_digest {
```
jodh-intel, line 47: Is the asterisk required here?

jodh-intel (commented):
Thanks @jialez0 - a few comments...

signature/src/signatures/sigstore.rs
```diff
@@ -0,0 +1,298 @@
...
+pub fn format_sigstore_name(image_ref: &Reference, image_digest: String) -> Result<String> {
+    let digest_compoents: Vec<&str> = image_digest.split(':').collect();
+    if digest_compoents.len() != 2 {
+        return Err(anyhow!("Format sigstore name: Image digest format is wrong."))
+    }
+    let image_name = image_ref.repository().to_string();
+    Ok(format!(
+        "{}@{}={}",
+        image_name, digest_compoents[0], digest_compoents[1]
+    ))
```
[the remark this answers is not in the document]
jialez0: The format obtained here is like `busybox@sha256=69704ef328d05a9f806b6b8502915e6a0a4faa4d72018dc42343f511490daf8a`, while `Reference::whole()` obtains the complete reference of the image, so it should not be used here.

signature/src/policy.rs
```diff
@@ -0,0 +1,868 @@
...
+        if !self.key_path.is_empty() && !self.key_data.is_empty() {
+            return Err(anyhow!(
+                "Internal inconsistency: both \"keyPath\" and \"keyData\" specified."
+            ));
+        }
+
+        let mut pubkey_ring = Vec::new();
+        if !self.key_data.is_empty() {
+            // ???: base64 encode the keyring data ???
+            // FIXME: Maybe the KeyData format need to re-think.
```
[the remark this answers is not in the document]
jialez0: I think the most suitable format here is still Base64 coding, so I removed the comments here.

jialez0:
Thanks @jodh-intel for reviewing! I have resolved these comments and let's continue the reviewing :-)

signature/src/signatures/sigstore.rs
```diff
@@ -0,0 +1,298 @@
...
+        if let Some(docker_config_map) = config.docker {
+            for (ns_name, ns_config) in docker_config_map.iter() {
+                if merged_config.is_namespace_exit(ns_name) {
+                    return Err(anyhow!(format!(
+                        "Error parsing sigstore config: {} defined repeatedly.",
+                        &ns_name
+                    )));
+                }
+
+                merged_config.inser_a_docker_config_entry(ns_name, ns_config);
```
fitzthum, line 104: `insert`?

signature/src/policy.rs
```diff
@@ -0,0 +1,893 @@
...
+// PolicyRequirement is a rule which must be satisfied by the image.
+// It has three types:
+//   `insecureAcceptAnything`: skip signature verification, accept the image unconditionally.
+//   `reject`: reject the image directly.
+//   `signedBy`: there must be at least a signature of the image can be verified by the specific key.
+// The first two are represented by PolicyReqSimple, the last is represented by PolicyReqSignedBy.
+#[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
+#[serde(untagged)]
+pub enum PolicyRequirement {
+    Signedby(PolicyReqSignedBy),
```
fitzthum, line 69: nit: `Signedby` -> `SignedBy`

fitzthum (commented):
A few random comments. I will add more soon.

signature/src/signatures/sigstore.rs
```diff
@@ -0,0 +1,298 @@
...
+    if let Some(default_config) = sigstore_config.default_docker {
+        return Ok(default_config.sigstore);
+    }
+
+    Err(anyhow!(
+        "There is no config info for this image in sigstore config files."
+    ))
+}
+
+pub fn get_sigs_from_lookaside(sigstore: String) -> Result<Vec<Vec<u8>>> {
```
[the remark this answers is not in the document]
jialez0: The `lookaside` here actually refers to the `sigstore` specified in the configuration file. This name follows the function with the same name implemented by simple signing in golang version, but I think the more appropriate name here is `get_sigs_from_specific_sigstore`

signature/src/policy.rs
```diff
@@ -0,0 +1,893 @@
...
+        // FIXME: only support "GPGKeys" type now.
+        if self.key_type != KeyType::Gpg.to_string() {
+            return Err(anyhow!(
+                "UNKNOWN KEY TYPE in policy config: only support 'GPGKeys' now."
+            ));
+        }
+
+        if !self.key_path.is_empty() && !self.key_data.is_empty() {
+            return Err(anyhow!(
+                "BUG: both keyPath and keyData specified."
```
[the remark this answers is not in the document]
jialez0: The logic here is not to allow both to be specified or both to be empty, which means that the owner must explicitly specify one of them. At present, the implementation of simple signaling mechanism in golang version is so required. It may be a good idea to follow them for the time being.

signature/src/policy.rs
```diff
@@ -0,0 +1,893 @@
...
+}
+
+impl PolicyReqSignedBy {
+    fn signature_judge_allowed(&self, image: &mut image::Image) -> Result<()> {
+        if self.r#type != PolicyReqType::TypeSignedBy.to_string() {
+            return Err(anyhow!(ErrorInfo::ErrUnknownPrType.to_string()));
+        }
+
+        let sigs = image.signatures()?;
+        let mut reject_reason: Vec<anyhow::Error> = Vec::new();
```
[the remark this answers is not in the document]
jialez0: just adding debug logging to `is_signature_accept` is indeed possible, but the current method can report the reason for rejecting the signature to the top-level caller (such as the pulling module of image-rs), which allows the caller to take different actions according to the reason for rejecting.

signature/src/policy.rs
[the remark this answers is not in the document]
fitzthum: Yeah it's fine to follow the existing implementation.

fitzthum (commented):
A few more comments. Will give more soon.

signature/src/signatures/sigstore.rs
```diff
@@ -0,0 +1,296 @@
...
+use crate::image;
+use anyhow::{anyhow, Result};
+use oci_distribution::Reference;
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use std::ffi::OsStr;
+use std::fs;
+
+// FIXME: Now, for Demo, the dir which the sigstore config files locate
+// is hardcoded to "/etc/containers/registries.d".
```
jodh-intel, line 15: I don't think the comment needs to repeat the code here:

```suggestion
// is hardcoded.
```

signature/src/policy.rs
```diff
@@ -0,0 +1,893 @@
...
+    key_type: String,
+
+    // KeyPath is a pathname to a local file containing the trusted key(s).
+    // Exactly one of KeyPath and KeyData must be specified.
+    #[serde(default, rename = "keyPath")]
+    key_path: String,
+    // KeyData contains the trusted key(s), base64-encoded.
+    // Exactly one of KeyPath and KeyData must be specified.
+    #[serde(default, rename = "keyData")]
+    key_data: String,
```
jodh-intel, line 126: This makes me wonder if there is value in creating a type to encapsulate all these related components:

```rust
pub struct Key {
    r#type: String,
    path: String,
    data: String,
}

pub struct PolicyReqSignedBy {
    key: Key,
}

File: signature/src/policy.rs [hunk 20]
```diff
@@ -0,0 +1,893 @@
...
+        // FIXME: only support "GPGKeys" type now.
+        if self.key_type != KeyType::Gpg.to_string() {
+            return Err(anyhow!(
+                "UNKNOWN KEY TYPE in policy config: only support 'GPGKeys' now."
+            ));
+        }
+
+        if !self.key_path.is_empty() && !self.key_data.is_empty() {
+            return Err(anyhow!(
+                "ERROR: both keyPath and keyData specified."
```
jodh-intel, line 176: I would recommend against error messages containing a prefix like this. Ideally, the top level caller would add that.

signature/src/policy.rs
```diff
@@ -0,0 +1,893 @@
...
+            reject_reason
+        )))
+    }
+
+    #[allow(unused_assignments)]
+    fn is_signature_valid(&self, image: &image::Image, sig: Vec<u8>) -> Result<()> {
+        // FIXME: only support "GPGKeys" type now.
+        if self.key_type != KeyType::Gpg.to_string() {
+            return Err(anyhow!(
+                "UNKNOWN KEY TYPE in policy config: only support 'GPGKeys' now."
```
jodh-intel, line 170: - I would make the entire message lower case.
- Why not add the value of `KeyType::Gpg.to_string()` to the error with a format rather than hard-coding it?

signature/src/policy.rs
```diff
@@ -0,0 +1,893 @@
...
+
+        Err(anyhow!(format!(
+            "The signatures do not satisfied! Reject reason: {:?}",
+            reject_reason
+        )))
+    }
+
+    #[allow(unused_assignments)]
+    fn is_signature_valid(&self, image: &image::Image, sig: Vec<u8>) -> Result<()> {
+        // FIXME: only support "GPGKeys" type now.
```
jodh-intel, line 167: Can you add a GitHub issue URL here so we don't forget? :smile:

signature/src/policy.rs
```diff
@@ -0,0 +1,893 @@
...
+            req.is_running_image_allowed(&mut image)?;
+        }
+
+        Ok(())
+    }
+
+    // selects the appropriate requirements for the image from Policy.
+    fn requirements_for_image(&self, image: &image::Image) -> PolicyRequirements {
+        // Get transport name of the image,
+        // FIXME: now only support "docker" (image from the registry which implement Docker HTTPS APIs V2)
```
jodh-intel, line 444: Can you add a GitHub issue URL here so we don't forget?

jodh-intel (commented):
Thanks @jialez0 - a few comments...

signature/src/image/digest.rs
```diff
@@ -0,0 +1,212 @@
...
+    /// Unsupported digest algorithm
+    DigestUnsupported,
+}
+
+impl fmt::Display for ParseError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            ParseError::DigestInvalidFormat => write!(f, "invalid checksum digest format"),
+            ParseError::DigestInvalidLength => write!(f, "invalid checksum digest length"),
+            ParseError::DigestUnsupported => write!(f, "unsupported digest algorithm"),
```
wainersm, line 40: Nit: my suggestion is to rename DigestUnsupported to DigestUnsupportedAlgorithm so we know exactly what's not supported in the digest.

signature/src/policy.rs
```diff
@@ -0,0 +1,460 @@
...
+//   `insecureAcceptAnything`: skip signature verification, accept the image unconditionally.
+//   `reject`: reject the image directly.
+//   `signedBy`: there must be at least a signature of the image can be verified by the specific key.
+// The first two are represented by PolicyReqSimple, the last is represented by PolicyReqSignedBy.
+#[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
+#[serde(untagged)]
+pub enum PolicyRequirement {
+    SignedBy(PolicyReqSignedBy),
+    Simple(PolicyReqSimple),
+}
```
sameo, line 71: I find this `enum` confusing as it's covering 3 cases with only 2 entries.

What I think would make more sense is using an trait instead of an enum

```Rust
pub trait PolicyRequirement {
     is_image_allowed(&self, image: &mut image::Image);
 }
```

And then have 3 structures that implement that trait:

```Rust
pub struct PolicyReqReject{};
pub struct PolicyReqAcceptAnything{};
pub struct PolicyReqSignedBy {
 ....
}
```

You can implement the `TryFrom` trait on the `PolicyRequirement` trait:

```Rust
impl TryFrom<&str> for &dyn PolicyRequirement {
```

signature/src/policy.rs
```diff
@@ -0,0 +1,460 @@
...
+            ));
+        }
+
+        let mut pubkey_ring = Vec::new();
+        if !self.key_data.is_empty() {
+            // base64 encode the keyring data.
+            pubkey_ring = base64::decode(&self.key_data)?;
+        } else {
+            pubkey_ring = fs::read(&self.key_path)?;
+        }
```
sameo, line 186: A more idiomatic implementation:

```rust
let pubkey_ring = if !self.key_data.is_empty() {
     base64::decode(&self.key_data)?
} else {
     fs::read(&self.key_path)?
};
```

signature/src/policy.rs
```diff
@@ -0,0 +1,460 @@
...
+    // MatchRemapIdentity: "remapIdentity".
+    RemapIdentity(MatchRemapIdentity),
+    // MatchExactRepository: "exactRepository".
+    ExactRepository(MatchExactRepository),
+    // MatchExactReference: "exactReference".
+    ExactReference(MatchExactReference),
+    // MatchSimple: "matchExact", "matchRepoDigestOrExact", "matchRepository".
+    Match(MatchSimple),
+}
+
```
sameo, line 216: Here as well, the code would probably benefit from using a trait instead:

```rust
put trait PolicyReferenceMatcher {
     fn matches_docker_reference(&self, origin: &Reference, signed_reference: &str) -> Result<()>
}
```

And have a *separate* structure for the kind of matches that you currently want to support.

sameo (commented):
Thanks a lot for the PR @jialez0 !

I am not done with the full review yet, but I'm sending a few comments already.
I'll finish the review tomorrow hopefully.

signature/src/image/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    pub fn set_manifest_digest(&mut self, digest: &str) -> Result<()> {
+        self.manifest_digest = Digest::try_from(digest)?;
+    }
+}
+
+// Get repository full name:
+// `registry-name/repository-name`
+pub fn get_image_repository_full_name(image_ref: &Reference) -> String {
```
wainersm, line 117: wouldn't be easier to split by ':' then get the first element?

wainersm: Nit: I am not sure about `repository_full_name` in the name `get_image_repository_full_name`. The `Reference` encapsulates a string formatted as `registry/repository(:[tag|disgest])?`. And what you consider `repository` is a different meaning in `repository` in `Reference`. Does OCI gives a proper name of the `registry/repository` segment?

sameo (commented):
Done with the first round of review for this PR, sorry for being so late.

An additional, general comment: Please add more descriptive commit messages. Right now, all your 10 commits only have a subject, no body at all. Please briefly summarize what the commit is about, and how it's implemented.

signature/src/signatures/sigstore.rs
```diff
@@ -0,0 +1,167 @@
...
+const REGISTRY_CONFIG_DIR: &str = "/etc/containers/registries.d";
+
+// Format the sigstore name:
+// `image-repository@digest-algo=digest-value`
```
wainersm, line 19: Could you please expand "algo" to "algorithm" here (and throughout this PR?)

signature/src/signatures/sigstore.rs
```diff
@@ -0,0 +1,167 @@
...
+        let entry = entry?;
+        let path = entry.path();
+        if path.is_dir() || path.extension() != Some(yaml_extension) {
+            continue;
+        }
+        let path_str = path
+            .to_str()
+            .ok_or(anyhow!("Unknown error: path parsed failed."))?;
+        let config_yaml_string = fs::read_to_string(path_str)?;
+        let config = serde_yaml::from_str::<SigstoreConfig>(&config_yaml_string)?;
```
wainersm, line 78: Just wanted to check if the intention is to really stop the loading of the sigstore files if one of them is a malformed YAML.

signature/src/signatures/sigstore.rs
```diff
@@ -0,0 +1,167 @@
...
+        } else {
+            let mut new_docker = HashMap::new();
+            new_docker.insert(ns_name.to_string(), ns_config.clone());
+            self.docker = Some(new_docker)
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
+struct RegistryNamespace {
```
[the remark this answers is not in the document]
wainersm: IIUC this sigstore module is design to parse the [`containers-registries.d(5)`](https://github.com/containers/image/blob/main/docs/containers-registries.d.5.md) yaml files. As such, the man-page says "sigstore defines an URL of the signature storage", therefore, this `RegistryNamespace` seems a "sigstore" => URL mapping.

signature/src/signatures/sigstore.rs
wainersm: What happens if there is a `sigstore-staging` entry like below? Does it ignore the entry or fail to parse?

```
              docker:
                  registry.example.com:
                      sigstore: https://registry-sigstore.example.com
                  registry.example.com/mydepartment:
                      sigstore: https://sigstore.mydepartment.example.com
                      sigstore-staging: file:///mnt/mydepartment/sigstore-staging
                  registry.example.com/mydepartment/myproject:mybranch:
                      sigstore: http://localhost:4242/sigstore
                      sigstore-staging: file:///home/useraccount/webroot/sigstore
```

signature/src/signatures/sigstore.rs
```diff
@@ -0,0 +1,167 @@
...
+        for ns in image_ns.iter() {
+            if let Some(base_url) = docker_config_map.get(ns) {
+                return Ok(base_url.sigstore.to_string());
+            }
+        }
+    }
+
+    // Look for a default location
+    if let Some(default_config) = sigstore_config.default_docker {
+        return Ok(default_config.sigstore);
```
wainersm, line 128: Shouldn't you convert to string (i.e. `default_config.sigstore.to_string()`) here as well?

signature/src/signatures/sigstore.rs
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    Err(anyhow!(
+        "There is no config info for this image in sigstore config files."
+    ))
+}
+
+pub fn get_sigs_from_specific_sigstore(sigstore: String) -> Result<Vec<Vec<u8>>> {
```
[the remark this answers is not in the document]
wainersm: Agree. And I would check if the url is valid when parsing the registries files (i.e. in `load_sigstore_config`), so at this point you already have a sane URL object (or struct, sorry, I'm not expert on rust).

Then possible `RegistryNamespace` should be a `sigstore: URL` struct rather than `sigstore: String`.

signature/src/signatures/sigstore.rs
```diff
@@ -0,0 +1,167 @@
...
+        for entry in fs::read_dir(sigstore_dir_path)? {
+            let entry = entry?;
+            let path = entry.path();
+            if path.is_dir() {
+                continue;
+            }
+            let path_str = path
+                .to_str()
+                .ok_or(anyhow!("Unknown error: path parsed failed."))?;
+            let sig = fs::read(path_str)?;
```
wainersm, line 156: Here (and in the policy module) the signature is represented as a `Vec<u8>` but I suggest you to use a higher-level abstraction. For example:
```rust
struct Signature {
  content: Vec<u8>,
  path: URL,
  sigstore: Sigstore, 
}

struct Sigstore {
  namespace: RegistryNamespace,
  path: URL,
  type: "docker"
}
```

Because you might need to find the reference to the signature path (from where you read it) in case of, for example, throw an error when the signature is invalid.

signature/src/policy.rs
```diff
@@ -0,0 +1,460 @@
...
+    TypeAccept,
+    #[strum(to_string = "reject")]
+    TypeReject,
+    #[strum(to_string = "signedBy")]
+    TypeSignedBy,
+}
+
+#[derive(EnumString, Display, Debug, PartialEq)]
+pub enum KeyType {
+    #[strum(to_string = "GPGKeys")]
```
wainersm, line 30: Maybe it should be "GPGKey" (singular)?

signature/src/policy.rs
```diff
@@ -0,0 +1,460 @@
...
+    Gpg,
+}
+
+#[derive(EnumString, Display, Debug, PartialEq)]
+pub enum PolicyReqMatchType {
+    #[strum(serialize = "matchExact")]
+    TypeMatchExact,
+    #[strum(serialize = "matchRepoDigestOrExact")]
```
wainersm, line 38: I would expand "Repo" to "Repository" so it is kept consistent with the other match types.

signature/src/policy.rs
jialez0: The format design of signedBy policy comes from [image simple signing](https://github.com/containers/image/blob/main/docs/containers-policy.json.5.md#signedby). I think we'd better keep consistent with their policy format for the time being.

signature/src/image/digest.rs
```diff
@@ -0,0 +1,212 @@
...
+impl Digest {
+    pub fn get_algorithm(&self) -> String {
+        self.algorithm.clone()
+    }
+
+    pub fn get_value(&self) -> String {
+        self.value.clone()
+    }
+
+    pub fn is_empty(&self) -> bool {
```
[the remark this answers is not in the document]
jialez0: When you create a digest using `Digest::default()` (which often happens when create a default `image` structure), it will be empty.

signature/src/image/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+    if image_ref.registry().is_empty() {
+        image_ref.repository().to_string()
+    } else {
+        format!("{}/{}", image_ref.registry(), image_ref.repository())
+    }
+}
+
+// Returns a list of other policy configuration namespaces to search.
```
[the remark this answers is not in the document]
jialez0: Yes, because `docker.io/library/busybox:latest` or` docker.io/library` or `docker.io` may be constrained in the policy file, and for the image `docker.io/library/busybox:latest`,no matter which of the three cases, it needs to meet the policy configuration in it.

Maybe you can refer to this: https://github.com/containers/image/blob/main/docs/containers-policy.json.5.md#docker

signature/src/signatures/sigstore.rs
```diff
@@ -0,0 +1,167 @@
...
+        "{}@{}={}",
+        image_name, image_digest.get_algorithm(), image_digest.get_value()
+    )
+}
+
+// Defines sigstore locations (sigstore base url) for a single namespace.
+#[derive(Serialize, Deserialize, Default, Debug, PartialEq, Clone)]
+struct SigstoreConfig {
+    #[serde(rename = "default-docker")]
+    default_docker: Option<RegistryNamespace>,
```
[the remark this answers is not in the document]
jialez0: Maybe you can refer to this: https://github.com/containers/image/blob/main/docs/containers-registries.d.5.md#registries-scopes-and-search-order

signature/src/signatures/sigstore.rs
```diff
@@ -0,0 +1,167 @@
...
+    // The key is a namespace, using fully-expanded Docker reference format or parent namespaces.
```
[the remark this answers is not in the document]
jialez0: I have added a [link](https://github.com/containers/image/blob/main/docs/containers-registries.d.5.md) to the document explaining the details of this in the comment.

signature/src/signatures/sigstore.rs
```diff
@@ -0,0 +1,167 @@
...
+pub fn get_sigs_from_specific_sigstore(sigstore: String) -> Result<Vec<Vec<u8>>> {
+    let mut res: Vec<Vec<u8>> = Vec::new();
+    let sigstore_uri = url::Url::parse(&sigstore[..])
+        .map_err(|e| anyhow!("Failed to parse sigstore_uri: {:?}", e))?;
+
+    // FIXME: Now only support get signatures from local files.
+    // only the uri with "file" scheme can be parsed.
+    //
+    // issue: https://github.com/confidential-containers/image-rs/issues/9
+    if sigstore_uri.scheme() == "file" {
```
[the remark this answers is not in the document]
jialez0: What's the meaning? I don't seem to understand the advice here :)

signature/src/signatures/sigstore.rs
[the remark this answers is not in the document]
jialez0: @wainersm I agree to check if the url is valid when parsing the registries files, but the `RegistryNamespace` can not be a `sigstore: URL` struct, becuase `SigstoreConfig` is the origin parsed struct, it's from `serde_yaml::from_str::<SigstoreConfig>(&config_yaml)?`, the `serde_yaml` can not directly parse a string to `URL` struct.

signature/src/signatures/sigstore.rs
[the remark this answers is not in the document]
sameo: I meant something like that:

```rust
match sigstore_uri.scheme() {
   "file" => {},
   _  => {},
```

as I expect we will handle more schemes in the future.

signature/src/policy.rs
jialez0: I tried to do this, but it will cause problems in deserialization. This is because `serde_json` does not support deserializing to trait objects. `PolicyRequirement` needs to deserialize from JSON strings in the policy configuration file. Using `enum` can make `serde_json::from_str` work now.

I have tried to customize the implementation of the `serde::Deserialize` macro for our `PolicyRequirement` trait object, but it would be very complex and make our code difficult to understand.

Therefore, I think we can use enum here. Maybe serde has better support in the future, and we can consider optimizing the code here.

signature/src/policy.rs
jialez0: As above

signature/src/image/mod.rs
jialez0: This function implements the `[full_name](https://github.com/krustlet/oci-distribution/blob/main/src/reference.rs#L137)` function in the `oci_distribution::Reference` library because the `full_name` function is private. OCI standards call both `busybox` and `docker.io/library/busybox` segments "repository". The former is a relative path and the latter is a full path.

> wouldn't be easier to split by ':' then get the first element?

Because ':' may also appear in namespace and some custom tags

signature/src/signatures/verify.rs
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    // Try to use each pubkey in the pubkey ring to verify the signature.
+    // If the signature is verified,
+    // verify that the keyID of the pubkey is consistent with the keyID recorded in the signature.
+    for packet in keyring_iter {
+        if let openpgp::Packet::PublicKey(pubkey) = packet {
+            if let Some(openpgp::Packet::Signature(ref mut signature)) = sig_packet.path_ref_mut(&[0, 2]) {
+                if signature.verify(pubkey).is_ok() {
+                    validate_key_id.trusted_key_id = pubkey.fingerprint().as_bytes().to_vec();
```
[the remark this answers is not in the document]
jialez0: `pubkey.fingerprint()` returns a enum [`Fingerprint`](https://docs.sequoia-pgp.org/sequoia_openpgp/packet/key/struct.Key4.html#method.fingerprint), so we need to convert it into vec.

signature/src/policy.rs
```diff
@@ -0,0 +1,460 @@
...
+    key_path: String,
+    // KeyData contains the trusted key(s), base64-encoded.
+    // Exactly one of KeyPath and KeyData must be specified.
+    #[serde(default, rename = "keyData")]
+    key_data: String,
+
+    // SignedIdentity specifies what image identity the signature must be claiming about the image.
+    // Defaults to "match-exact" if not specified.
+    #[serde(default, rename = "signedIdentity")]
+    signed_identity: Option<PolicyReferenceMatch>,
```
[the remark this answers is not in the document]
jialez0: The format and field names of the policy file now follow [the definitions in containers/image](https://github.com/containers/image/blob/main/docs/containers-policy.json.5.md#signedby), so we'd better keep the field names consistent with them.

signature/src/signatures/sigstore.rs
jialez0: > Just wanted to check if the intention is to really stop the loading of the sigstore files if one of them is a malformed YAML.

This will only leads to the failure when the file content does not conform to the yaml format standard and the required fields are missing, which naturally leads to the failure of signature verification because signature cannot be obtained.


> What happens if there is a `sigstore-staging` entry like below? Does it ignore the entry or fail to parse?
> 
> ```
>               docker:
>                   registry.example.com:
>                       sigstore: https://registry-sigstore.example.com
>                   registry.example.com/mydepartment:
>                       sigstore: https://sigstore.mydepartment.example.com
>                       sigstore-staging: file:///mnt/mydepartment/sigstore-staging
>                   registry.example.com/mydepartment/myproject:mybranch:
>                       sigstore: http://localhost:4242/sigstore
>                       sigstore-staging: file:///home/useraccount/webroot/sigstore
> ```

`sigstore-staging` is the sigstore url only for writing. CC doesn't need to generete signatures so this filed is not needed, if there is a `sigstore-staging` entry, our code won't crash, but will ignore it.

signature/src/policy.rs
jialez0: The field names of the policy file now follow [the definitions in containers/image](https://github.com/containers/image/blob/main/docs/containers-policy.json.5.md#signedby), and `MatchRepositoryDigestOrExact` is a little long as a variable name :)

signature/src/policy.rs
jialez0: This actually refers to GPG public key ring, so `GPGKeys` is more appropriate.

signature/src/signatures/sigstore.rs
jialez0: This is a good idea, but I think this optimization can be implemented through a new commit after the current version is merged, because it will lead to a lot of code changes.

signature/src/signatures/verify.rs
```diff
@@ -0,0 +1,299 @@
...
+// is consistant with the keyID recorded in the signature itself.
+#[derive(Default)]
+struct ValidateKeyID {
+    pub trusted_key_id: Vec<u8>,
+    pub sig_info_key_id: Vec<u8>,
+}
+
+impl ValidateKeyID {
+    pub fn validate(&self) -> Result<()> {
+        if self.sig_info_key_id == self.trusted_key_id[12..].to_vec() {
```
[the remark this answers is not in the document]
jialez0: How did you test that it would crash here? When I run `cargo test`, it can work normally here.

signature/src/policy.rs
wainersm: ah, ok. Could you add this link somewhere on this module? So reviewers like me realize this module implements a standard.

signature/src/signatures/verify.rs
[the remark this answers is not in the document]
wainersm: @jialez0 in case `trusted_key_id` is empty then it will crash?

signature/src/policy.rs
jialez0: OK, I have add the link at the comment of `Policy` struct :)

jialez0:
@sameo @wainersm @jodh-intel @fitzthum Thanks a lot for reviewing! I have resolved the above comments and you can take a look. Do you have any other comments on this PR?

signature/fixtures/sigstore_config/test_case_3/2.yaml
```diff
@@ -0,0 +1,5 @@
+default-docker:
+  sigstore: file:///default/sigstore
+docker:
+  example2.com:
+    sigstore: file:///var/lib/containers/sigstore
```
wainersm, line 5: Could we add a dummy `sigstore-staging` (https://github.com/containers/image/blob/main/docs/containers-registries.d.5.md) key to ensure the parser will ignore it properly?

signature/src/signatures/sigstore.rs
```diff
@@ -163,4 +163,137 @@ pub fn get_sigs_from_specific_sigstore(sigstore_uri: url::Url) -> Result<Vec<Vec
...
+                reference: Reference::try_from("example1.com/mylib/busybox:latest").unwrap(),
+                sigstore_base_url: "file:///var/lib/containers/sigstore",
+            },
+            TestData {
+                reference: Reference::try_from("example3.com/mylib/busybox:latest").unwrap(),
+                sigstore_base_url: "file:///default/sigstore",
+            }
+        ];
+
+        if !Path::new(SIGSTORE_CONFIG_DIR).exists() {
```
wainersm, line 218: `SIGSTORE_CONFIG_DIR` value is `/etc/containers/registries.d`, meaning the test should be run by a privileged user (otherwise it can fail to create the directory). So shouldn't you instead skip it if not superuser?

signature/src/signatures/sigstore.rs
```diff
@@ -163,4 +163,137 @@ pub fn get_sigs_from_specific_sigstore(sigstore_uri: url::Url) -> Result<Vec<Vec
...
+            sigstore_base_url: &'a str,
+        }
+
+        let tests = &[
+            TestData {
+                reference: Reference::try_from("example1.com/mylib/busybox:latest").unwrap(),
+                sigstore_base_url: "file:///var/lib/containers/sigstore",
+            },
+            TestData {
+                reference: Reference::try_from("example3.com/mylib/busybox:latest").unwrap(),
```
wainersm, line 213: Here `sigstore_base_url` is the default base url because there isn't match with `example3.com` namespace. It took me a while to understand it...maybe worth a comment?

signature/src/signatures/sigstore.rs
```diff
@@ -163,4 +163,137 @@ pub fn get_sigs_from_specific_sigstore(sigstore_uri: url::Url) -> Result<Vec<Vec
...
+
+        let tests_expect = &[
+            TestData {
+                sigstore_config_path: "./fixtures/sigstore_config/test_case_3",
+                merged_res_path: "./fixtures/sigstore_config/res.yaml",
+            }
+        ];
+
+        for case in tests_unexpect.iter() {
+            println!("{:?}", SigstoreConfig::new_from_configs(case));
```
wainersm, line 265: I suppose this `println` call was added to debug?

wainersm:
Hi @jialez0 , I finally had a chance to look at the tests. Left only a few comments. btw, nice tests!

sameo (approved):
Thanks a lot @jialez0 for addressing all those comments.
One last ask from me: Can you please squash the last 2 commits into the relevant previous commits, to have a clean git history?

jialez0:
@sameo Thanks! I have cleaned the git history :) btw, @wainersm @fitzthum @jodh-intel Do you have any other comments on this PR?

sameo (approved):
LGTM, thanks @jialez0 !

jodh-intel (approved):
Thanks @jialez0.

lgtm

wainersm (approved):
hey @jialez0 , thanks! LGTM.

fitzthum (approved):
Epic effort. Big thanks to everyone who reviewed as well.

PR #6, opened 2022-02-16
Modify security design document

jialez0:
There are multiple image signing and verification protocols/solutions available now, and there’s no standardization work for image signing yet. So in the design of confidential containers, we should support different signature mechanisms with modular implementation.

### DIFF
--- a/docs/ccv1_image_security_design.md
+++ b/docs/ccv1_image_security_design.md
@@ -154,55 +154,32 @@ From its `layers` field, we can see the changes we stated above.
 
 ## Image signing
 
-The container image signing mechanism used by the image-rs of confidential containers V1 
-is consistent with the signing mechanism used in [containers/image](https://github.com/containers/image). 
-At present, the only supported cryptographic signature format in this mechanism is OpenPGP([RFC 4880](https://datatracker.ietf.org/doc/html/rfc4880#section-5.4)). 
+There are multiple image signing and verification protocols/solutions in the field, 
+so an extensible modular architecture should be implemented to support them.
+The owner needs to specify the `scheme` field in the image security [policy file](https://github.com/confidential-containers/image-rs/blob/main/docs/ccv1_image_security_design.md#policy) distributed to image-rs. 
+when verifying the signature, image-rs can select the appropriate scheme for signature verification according to this field.
 
-The signing process will not modify any content of the container image (or an encrypted image), including layer, config, manifest, etc. 
+In CC V1, we start by supporting the [Red Hat simple signing format](https://www.redhat.com/en/blog/container-image-signing).
+This is a simple and direct signature system, 
+which uses the key of OpenPGP([RFC 4880](https://datatracker.ietf.org/doc/html/rfc4880#section-5.4)) 
+and a simple signature [payload format](https://github.com/containers/image/blob/main/docs/containers-signature.5.md).
 
-The container signature is essentially a blob, which consists of two parts:
+In the future, confidential containers will gradually support a variety of image signature schemes. 
+The differences between these schemes are as follows:
 
-- JSON payload, its format is defined in 
-[containers-signatures](https://github.com/containers/image/blob/main/docs/containers-signature.5.md)
+1. Signature payload format.
 
-- OpenPGP cryptographic signature of JSON payload
+2. Signature algorithm.
 
-The JSON payload contains the manifest digest and identity information (reference) of the container image. Examples are as follows:
+3. Signature storage and signature distribution.
 
-```json
-{
-    "critical": {
-        "type": "atomic container signature",
-        "image": {
-            "docker-manifest-digest": "sha256:...."
-        },
-        "identity": {
-            "docker-reference": "docker.io/mylib/example:latest"
-        }
-    },
-    "optional": {
-        "creator": "some software package v1.0.1-35",
-        "timestamp": 1483228800,
-    }
-}
-```
-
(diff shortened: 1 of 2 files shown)


docs/ccv1_image_security_design.md
```diff
@@ -154,55 +154,33 @@ From its `layers` field, we can see the changes we stated above.
 
 ## Image signing
 
-The container image signing mechanism used by the image-rs of confidential containers V1 
-is consistent with the signing mechanism used in [containers/image](https://github.com/containers/image). 
-At present, the only supported cryptographic signature format in this mechanism is OpenPGP([RFC 4880](https://datatracker.ietf.org/doc/html/rfc4880#section-5.4)). 
+There are multiple image signing and verification protocols/solutions available now,
```
jiangliu, line 157: There are multiple image signing and verification protocols/solutions in the field, so an extensible modular architecture should be implemented to support them.

docs/ccv1_image_security_design.md
```diff
@@ -231,27 +208,29 @@ The steps are as follows:
...
 
 ## Policy
 
-Policy file (`policy.json`) is a necessary configuration file for image pulling in CCv1,
-the spec of it is defined in 
-[container/image policy.json](https://github.com/containers/image/blob/main/docs/containers-policy.json.5.md).
+The policy file here comes from the signature verification [policy configuration file](https://github.com/containers/image/blob/main/docs/containers-policy.json.5.md)
+introduced in the implementation of red hat simple signing in the [containers/image](https://github.com/containers/image) project. 
+This policy file can not only be used for red hat simple signing In the secret container design, 
+we will continue to customize and expand the content of the policy file as
```
jodh-intel, line 224: Nit: Missing period after "signing" and before "In"?

docs/ccv1_image_security_design.md
```diff
@@ -341,44 +328,31 @@ he just need to change the policy of the member matching the registry under the
...
-```
+Different signature schemes generally use different signature storage locations. 
+When image-rs need to verify the signature, it first need to obtain the signature of the image. 
+The storage location may be:
+
+1. Local file system
+
+2. Container registry
+
+3. Customized remote server
```
jodh-intel, line 339: Can you give examples? Could this be an https web server for example?

jodh-intel (commented):
Thanks @jialez0. A few comments...

docs/ccv1_image_security_design.md
jodh-intel: Even if this doc doesn't list them all, is it worth collecting the solutions we're aware of and plan to recognise in an issue?

docs/ccv1_image_security_design.md
```diff
@@ -231,27 +208,29 @@ The steps are as follows:
...
 ## Policy
 
-Policy file (`policy.json`) is a necessary configuration file for image pulling in CCv1,
-the spec of it is defined in 
-[container/image policy.json](https://github.com/containers/image/blob/main/docs/containers-policy.json.5.md).
+The policy file here comes from the signature verification [policy configuration file](https://github.com/containers/image/blob/main/docs/containers-policy.json.5.md)
+introduced in the implementation of red hat simple signing in the [containers/image](https://github.com/containers/image) project. 
+This policy file can not only be used for red hat simple signing In the secret container design, 
+we will continue to customize and expand the content of the policy file as 
+a general signature verification configuration file compatible with different signature schemes.
```
fitzthum, line 225: Are we assuming that all signature schemes are going to have a policy file of some sort? Is there any reason to allow schemes to request additional files?

fitzthum (approved):
I think this is a good idea. In the future we might want to expand this even beyond signature verification. In addition to verifying the container signature, maybe the guest owner wants to scan the container image or perform more complex checks before running the workload. Eventually we might provide a hook for any trusted verification process. For now I think supporting multiple signature schemes is a good idea. Pending a bit of cleanup and maybe a few examples I think this is good.

docs/ccv1_image_security_design.md
jialez0: Maybe it is a better idea to use a unified policy file format in image-rs. I think there are the following reasons:

1. The policy file is distributed by the owner from the attestation-agent, which is a good channel for image-rs to accept the owner's control and configuration. The use of a unified policy file format enables the owner to determine which image use which scheme for signature verification through the policy file. In addition, this is conducive to playing more roles of the policy file on the image-rs side in the future.

2. The policy file format we currently use provides a strong configuration mechanism and has flexible scalability. We can support different signature schemes by customizing the current policy file.

3. Using the unified policy file format can make the architecture of image-rs verification signature function not too loose. Maybe this will enable us to do more optimization in the future.

docs/ccv1_image_security_design.md
jialez0: @jodh-intel Done: #7

docs/ccv1_image_security_design.md
fitzthum: Ok, I think it's fine to start with this assumption. We might need to revisit if we encounter a signature scheme with totally different parameters.

jiangliu (approved):

jialez0:
@sameo Maybe we can get this PR merged :)

jodh-intel:
Can someone merge this? I don't have privs but it seems to have been sitting here approved for a long time... ;)

sameo (approved):
LGTM thanks @jialez0 !

PR #13, opened 2022-03-03
tests: Add integration test for KBS encrypted OCI image 

arronwy:
Current unit tests in pull module only covered the tests for container
    image encrypted by a public key file.

    The integration test verify the container images encrypted by an KBS.
    How to create KBS encrypted container image is documented in:
    test_data/generate_test_data.md

    It will first launch an attestation-agent daemon build
    with sample KBS support. Then it will try to pull an KBS encrypted
    container image to verify pull_image API.

### DIFF
--- a/scripts/build_attestation_agent.sh
+++ b/scripts/build_attestation_agent.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+#
+# Copyright (c) 2022 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+[ -n "${BASH_VERSION:-}" ] && set -o errtrace
+[ -n "${DEBUG:-}" ] && set -o xtrace
+
+source $HOME/.cargo/env
+
+SCRIPT_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+AA_DIR=$SCRIPT_DIR/attestation_agent
+
+pushd $SCRIPT_DIR
+git clone --depth 1 "https://github.com/confidential-containers/attestation-agent.git" $AA_DIR
+pushd $AA_DIR
+make KBC=sample_kbc
+make DESTDIR="$SCRIPT_DIR" install
+popd
+
+cleanup() {
+  rm -rf "$AA_DIR"
+}
+trap cleanup EXIT
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -1,0 +1,53 @@
+// Copyright (c) 2022 Intel Corporation
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#[cfg(test)]
+mod tests {
+    use anyhow::Result;
+    use image_rs::image::ImageClient;
+    use std::path::Path;
+    use std::process::{Child, Command};
+
+    fn start_attestation_agent() -> Result<Child> {
+        let script_dir = format!("{}/{}", std::env!("CARGO_MANIFEST_DIR"), "scripts");
+        let aa_path = format!("{}/{}", script_dir, "attestation-agent");
+
+        if !Path::new(&aa_path).exists() {
+            let script_path = format!("{}/{}", script_dir, "build_attestation_agent.sh");
+            Command::new(script_path)
+                .output()
+                .expect("Failed to build attestation-agent");
+        }
+
+        Ok(Command::new(aa_path)
+            .args(&["--keyprovider_sock"])
+            .args(&["127.0.0.1:48888"])
+            .spawn()?)
+    }
+
+    #[tokio::test]
+    async fn test_image_rs() {
+        let mut aa = start_attestation_agent().expect("Failed to start attestation agent!");
+
+        let manifest_dir = std::env!("CARGO_MANIFEST_DIR");
+        let image = "docker.io/arronwang/busybox_kbs_encrypted";
+
+        let keyprovider_config =
+            format!("{}/{}", manifest_dir, "test_data/ocicrypt_keyprovider.conf");
+        std::env::set_var("OCICRYPT_KEYPROVIDER_CONFIG", keyprovider_config);
+        let decrypt_config = "provider:attestation-agent:sample_kbc::null";
+
+        let work_dir = tempfile::tempdir().unwrap();
+        std::env::set_var("CC_IMAGE_WORK_DIR", &work_dir.path());
+        let bundle_dir = tempfile::tempdir().unwrap();
+
+        let mut image_client = ImageClient::default();
+        assert!(image_client
+            .pull_image(image, bundle_dir.path(), &None, &Some(decrypt_config))
+            .await
+            .is_ok());
+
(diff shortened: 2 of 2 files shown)


jialez0 (approved):
LGTM. After these test code is merged, we can add security validation tests.

scripts/build_attestation_agent.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
```
jodh-intel, line 2: Missing copyright header?

scripts/build_attestation_agent.sh
```diff
@@ -0,0 +1,16 @@
...
+set -x
```
jodh-intel, line 3: You could add the standard Kata shell script "boiler plate" code here:

```suggestion
set -o errexit
set -o nounset
set -o pipefail

[ -n "$BASH_VERSION" ] && set -o errtrace
[ -n "${DEBUG:-}" ] && set -o xtrace
```

scripts/build_attestation_agent.sh
```diff
@@ -0,0 +1,16 @@
...
+source $HOME/.cargo/env
+
+SCRIPT_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+AA_DIR=$SCRIPT_DIR/attestation_agent
+
+pushd $SCRIPT_DIR
+git clone --depth 1 "https://github.com/confidential-containers/attestation-agent.git" $AA_DIR
+pushd $AA_DIR
+make KBC=sample_kbc
+cp target/release/attestation-agent $SCRIPT_DIR/
```
jodh-intel, line 13:
```suggestion
install target/release/attestation-agent $SCRIPT_DIR/
```

tests/integration.rs
```diff
@@ -0,0 +1,53 @@
...
+            .args(&["127.0.0.1:48888"])
+            .spawn()?)
+    }
+
+    #[tokio::test]
+    async fn test_image_rs() {
+        let mut aa = start_attestation_agent().expect("Failed to start attestation agent!");
+
+        let manifest_dir = std::env!("CARGO_MANIFEST_DIR");
+        let image = "docker.io/arronwang/busybox_kbs_encrypted";
```
jodh-intel, line 34: To avoid using your own docker hub user, how about creating a https://hub.docker.com/u/confidential-containers or similar ([like we have for Kata]( https://hub.docker.com/u/katadocker))?

jodh-intel (commented):
Thanks @arronwy - a few comments.

scripts/build_attestation_agent.sh
arronwy: Yes, thanks, will fix.

scripts/build_attestation_agent.sh
arronwy: good suggestion, will update this part.

tests/integration.rs
arronwy: Thanks, good suggestion, I'll try to create an confidential-containers repo in docker hub.

jodh-intel: Ack - the name might be a problem (as admittedly `confidentialcontainers` is pretty hard to read) but it might also be safer long term to have a CC specific account.

wainersm: Hi @arronwy @jodh-intel , I would even avoid docker.io as some developers have complaint about hitting the limits to pull images. So I would rather use quay.io. In fact we already have a repository for CC images: https://quay.io/repository/kata-containers/confidential-containers

scripts/build_attestation_agent.sh
```diff
@@ -0,0 +1,27 @@
...
+source $HOME/.cargo/env
+
+SCRIPT_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+AA_DIR=$SCRIPT_DIR/attestation_agent
+
+pushd $SCRIPT_DIR
+git clone --depth 1 "https://github.com/confidential-containers/attestation-agent.git" $AA_DIR
+pushd $AA_DIR
+make KBC=sample_kbc
+install target/release/attestation-agent $SCRIPT_DIR/
```
wainersm, line 24: `make DESTDIR="$SCRIPT_DIR" install`

scripts/build_attestation_agent.sh
```diff
@@ -0,0 +1,27 @@
...
+popd
+
+rm -rf $AA_DIR
```
wainersm, line 27: This command won't be called if something goes wrong before. I suggest you to create a clean up function to be called on shell's exit. Example:

```bash
cleanup() {
  rm -rf "$AA_DIR"
}
trap cleanup EXIT
```

tests/integration.rs
```diff
@@ -0,0 +1,53 @@
...
+
+#[cfg(test)]
+mod tests {
+    use anyhow::Result;
+    use image_rs::image::ImageClient;
+    use std::path::Path;
+    use std::process::{Child, Command};
+
+    fn start_attestation_agent() -> Result<Child> {
+        let script_dir = format!("{}/{}", std::env!("CARGO_MANIFEST_DIR"), "scripts");
```
wainersm, line 13: Not a rust expert but I would use https://doc.rust-lang.org/std/path/struct.Path.html to manipulate paths. Not a big deal, it is up to you take this suggestion :)

tests/integration.rs
```diff
@@ -0,0 +1,53 @@
...
+            .args(&["--keyprovider_sock"])
+            .args(&["127.0.0.1:48888"])
+            .spawn()?)
+    }
+
+    #[tokio::test]
+    async fn test_image_rs() {
+        let mut aa = start_attestation_agent().expect("Failed to start attestation agent!");
+
+        let manifest_dir = std::env!("CARGO_MANIFEST_DIR");
```
wainersm, line 33: nit: you could have a `fn manifest_dir()` at test module level so that the function could be called here and in `start_attestation_agent()`

wainersm:
Hi @arronwy , I gave just a couple suggestions. Overall it looks great!

scripts/build_attestation_agent.sh
arronwy: Thanks, much better.

tests/integration.rs
wainersm: cool, thanks!

wainersm (approved):
Thanks for addressing some of my comments @arronwy !

tests/integration.rs
arronwy: seems [quay.io](http://quay.io/) don't support encrypted container image format：
``` 
Error uploading manifest kbs_encrypted_unsigned to [quay.io/kata-containers/confidential-containers](http://quay.io/kata-containers/confidential-containers): manifest invalid: manifest invalid
Unknown media type during manifest conversion: "application/vnd.oci.image.layer.v1.tar+gzip+encrypted
```

wainersm: hi @arronwy, hmm... I will ask people about that. However I don't want to block his PR; so it's okay to me use docker.io temporarily

sameo (approved):

PR #20, opened 2022-04-21
Refine create_runtime_config()

jiangliu:
Refine create_runtime_config() by:
1) add more comments
2) fix error in generating annotations
3) add support for missed annotations

### DIFF
--- a/src/bundle.rs
+++ b/src/bundle.rs
@@ -14,7 +14,10 @@ pub const BUNDLE_ROOTFS: &str = "rootfs";
 pub const BUNDLE_HOSTNAME: &str = "image-rs";
 
 const ANNOTATION_OS: &str = "org.opencontainers.image.os";
+const ANNOTATION_OS_VERSION: &str = "org.opencontainers.image.os.version";
+const ANNOTATION_OS_FEATURES: &str = "org.opencontainers.image.os.features";
 const ANNOTATION_ARCH: &str = "org.opencontainers.image.architecture";
+const ANNOTATION_VARIANT: &str = "org.opencontainers.image.variant";
 const ANNOTATION_AUTHOR: &str = "org.opencontainers.image.author";
 const ANNOTATION_CREATED: &str = "org.opencontainers.image.created";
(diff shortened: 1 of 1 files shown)


sameo (approved):

sameo (approved):
LGTM, let's wait for @arronwy review before merging.

src/bundle.rs
```diff
@@ -14,66 +14,134 @@ pub const BUNDLE_ROOTFS: &str = "rootfs";
...
+/// runtime configuration".
 pub fn create_runtime_config(
     image_config: &ImageConfiguration,
     bundle_path: &Path,
 ) -> Result<PathBuf> {
     let mut spec = Spec::default();
+    let mut annotations: HashMap<String, String> = HashMap::new();
+    let mut labels = HashMap::new();
 
+    // Update the default hostname of `youki`
```
arronwy, line 44: a small typo of hostname fo `image-rs`

src/bundle.rs
```diff
@@ -14,66 +14,134 @@ pub const BUNDLE_ROOTFS: &str = "rootfs";
...
             args.extend(entrypoint.clone());
         }
-
         if let Some(cmd) = config.cmd() {
             args.extend(cmd.clone());
         }
-
         if !args.is_empty() {
             process.set_args(Some(args));
         }
```
arronwy, line 79: There have a bug in original code, could help fix it together with this PR?
we need update the `process` in `spec` which is missing in original code :-(
```
spec.set_process(Some(process));
```

arronwy (commented):
Thanks @jiangliu , LGTM, only some minor changes maybe needed.

src/bundle.rs
jiangliu: Yes, I also found this issue, but missed it.

jodh-intel (approved):
Nice PR @jiangliu - thanks !

src/bundle.rs
jiangliu: Fixed

PR #23, opened 2022-04-22
[WIP]: enhance security with safe-path 

jiangliu:


src/bundle.rs
```diff
@@ -88,31 +157,59 @@ pub fn create_runtime_config(
...
     }
 
     spec.set_annotations(Some(annotations));
+    // TODO
     let bundle_config = bundle_path.join(BUNDLE_CONFIG);
     spec.save(&bundle_config)?;
 
-    Ok(bundle_config)
+    PinnedPathBuf::new(bundle_path, BUNDLE_CONFIG)
+        .context("The path of generated config.json file has changed, possible attack!")
```
jodh-intel, line 212: - Could you use `BUNDLE_CONFIG` rather than hard-coding `config.json`.
- I wonder if we want to use a standard "eye catcher" / tag for scenarios like this (maybe use a "SECURITY" prefix in the context message?) That would make errors like this more `grep(1)`-able.

jodh-intel (commented):
Thanks @jiangliu - A few comments...

src/bundle.rs
jodh-intel: Any update on this?

dcmiddle:
Cleaning up old issues / PRs .. feel free to reopen if this is still active.

PR #24, opened 2022-04-22
decrypt: improve performance and reduce memory consumption

jiangliu:
Currently the flow to handle encrypted and compressed is as below:
1) create a decryptor
2) decrypt compressed data to a buffer
3) decompress from the data buffer to a file

The design has two issues:
1) comsume too much memory
2) not performance optimal

So change the flow to
1) create a decryptor
2) decompress from the decryptor into a file in stream mode

Signed-off-by: Jiang Liu <gerry@linux.alibaba.com>

### DIFF
--- a/src/decrypt.rs
+++ b/src/decrypt.rs
@@ -2,7 +2,10 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
+use std::io::{Cursor, Read};
+
 use anyhow::{anyhow, Result};
+use oci_distribution::manifest::{self, OciDescriptor};
 use ocicrypt_rs::config::CryptoConfig;
 use ocicrypt_rs::encryption::decrypt_layer;
 use ocicrypt_rs::helpers::create_decrypt_config;
@@ -10,11 +13,6 @@ use ocicrypt_rs::spec::{
     MEDIA_TYPE_LAYER_ENC, MEDIA_TYPE_LAYER_GZIP_ENC, MEDIA_TYPE_LAYER_NON_DISTRIBUTABLE_ENC,
     MEDIA_TYPE_LAYER_NON_DISTRIBUTABLE_GZIP_ENC,
 };
-
-use oci_distribution::manifest;
-use oci_distribution::manifest::OciDescriptor;
-
-use std::io::Read;
 
 #[derive(Default, Clone)]
 pub struct Decryptor {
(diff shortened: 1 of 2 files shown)


src/decrypt.rs
```diff
@@ -60,53 +58,38 @@ impl Decryptor {
     ///           - \<filename>:fd=\<filedescriptor> \
     ///           - \<filename>:\<password> \
     ///           - provider:<cmd/gprc>
-    pub async fn get_plaintext_layer(
+    pub fn get_plaintext_layer(
```
jodh-intel, line 61: Could you add a few unit tests for the two functions in this file (even if you just test a few of the error paths)?

jodh-intel (commented):
Thanks @jiangliu - One suggestion...

src/decrypt.rs
```diff
@@ -60,53 +58,38 @@ impl Decryptor {
...
-
-        let cc = create_decrypt_config(vec![decrypt_config.to_string()], vec![])?;
-        let descript = descriptor.clone();
-
-        // ocicrypt-rs keyprovider module will create a new runtime to talk with
-        // attestation agent, to avoid startup a runtime within a runtime, we
-        // spawn a new thread here.
-        let handler = tokio::task::spawn_blocking(move || {
-            decrypt_layer_data(&encrypted_layer, &descript, &cc)
-        });
```
arronwy, line 85: This is still needed, since the `unwrap_key` in `ocicrypt-rs` keyprovider already created a new runtime, else it will have below error:
```
thread 'tests::test_image_rs' panicked at 'Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.', /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.18.0/src/runtime/enter.rs:39:9
```

The root cause is unwrap key operation for keyprovider is async, we may need seperate it as a pubic API to avoid block the whole decrypt processing.

arronwy (commented):
Thanks @jiangliu , I verified the decrypt+decompress time can be decreased ~3.6204% with JWE keywrapper, for keyprovide keywrapper the change will have issue as I commented inline.

jodh-intel:
Any update on this @jiangliu?

jodh-intel:
Hi @jiangliu - do you think you'll get a chance to update this PR?

dcmiddle:
I think this was eventually subsumed by https://github.com/confidential-containers/guest-components/pull/96/files.
Can we close this?
@jiangliu @arronwy @jodh-intel

arronwy:
Yes, I think we can close this pr @jiangliu

dcmiddle:
Closing per discussion. Feel free to reopen if we have misjudged this.

Issue #7, opened 2022-02-17
Support different signature schemes.


jialez0:
As mentioned in #6 , There are multiple image signing and verification protocols/solutions available now, and there’s no standardization work for image signing yet. So in the image-rs of confidential containers, we should support different signature mechanisms with modular implementation.

In order to achieve this goal, we should first reconstruct the current signature verification function (#5 ) and modularize it to provide scalability for supporting more signature schemes. Specifically, the functions that need modularization include:

1. Getting signature: In different signature schemes, the storage and distribution methods of signature are different.
2. Parsing signature payload: Different signature schemes may use different signature payload formats.
3. Cryptography signature verification: Different signature schemes may use different signature algorithms.

After modularizing the above functions, we should add more signature schemes to support the verification of images signed by different schemes, such as:

1. Image signed with [Red Hat simple signing scheme](https://www.redhat.com/en/blog/container-image-signing).
2. Images signed with [cosign](https://github.com/sigstore/cosign) signature tool.
3. Image signed with [Alibaba cloud ACR image signing scheme](https://partners-intl.aliyun.com/help/en/doc-detail/159642.htm).
4. ETC.

bodzhang:
One more signature scheme to consider - [Notary V2/Notation](https://github.com/notaryproject/notation)

Issue #14, opened 2022-03-23
Simple signing scheme: now only support `GPGKeys` key type.


jialez0:
Following the simple signing golang repo: https://github.com/containers/image

ariel-adam:
@jialez0 can you close this issue?

PR #31, opened 2022-05-13
[Signature] Reconstruct to support different signature schemes.

jialez0:
Reconstruct the signature library to support different signature schemes.

Fix: #7


signature/src/image/mod.rs
```diff
@@ -75,53 +77,25 @@ impl Image {
         }
     }
 
-    pub fn signatures(&mut self) -> Result<Vec<Vec<u8>>> {
+    pub fn signatures(&mut self, scheme: &str) -> Result<Vec<Vec<u8>>> {
```
jodh-intel, line 80: This is a rather odd method as `scheme` is ignored if it's already been called. It might be clearer to have a "setter method" that accepts the `scheme` and then this can remain parameter free (and error if the setter hasn't been called?

signature/src/image/mod.rs
```diff
@@ -75,53 +77,25 @@ impl Image {
...
-            let sigstore_config =
-                signatures::SigstoreConfig::new_from_configs(signatures::SIGSTORE_CONFIG_DIR)?;
-            if let Some(base_url) = sigstore_config.base_url(&self.reference)? {
-                self.set_sigstore_base_url(base_url)?;
+    fn internal_signatures(&mut self, scheme: &str) -> Result<Vec<Vec<u8>>> {
+        match crate::SignatureScheme::from_str(scheme) {
+            Ok(SignatureScheme::SimpleSigning) => {
+                self.signatures = mechanism::simple::get_signatures(self)?;
+            }
+            // TODO: Add more signature mechanism.
```
jodh-intel, line 93: - Is there a known list that should be added?
- A GitHub issue URL would be good here.

signature/src/mechanism/simple/mod.rs
```diff
@@ -0,0 +1,210 @@
...
+                e,
+                &signedby_req.key_path
+            )
+        })?
+    };
+
+    let sigs = image.signatures(&signedby_req.scheme)?;
+    let mut reject_reason: Vec<anyhow::Error> = Vec::new();
+
+    for sig in sigs.iter() {
```
jodh-intel, line 59: ooi, is the order of the signatures important?

signature/src/mechanism/simple/mod.rs
```diff
@@ -0,0 +1,210 @@
...
+            image.set_sigstore_base_url(base_url)?;
+        }
+    }
+
+    let sigstore = format!(
+        "{}/{}",
+        &image.get_sigstore_base_url_string(),
+        &sigstore_name
+    );
+    let sigstore_uri = url::Url::parse(&sigstore[..])
```
jodh-intel, line 140: I think you can remove the `[..]` here?

signature/src/policy/mod.rs
```diff
@@ -27,6 +28,8 @@ pub enum ErrorInfo {
     ErrUnknowPolicyReqType,
     #[strum(to_string = "The reference match policy type name is Unknown.")]
     ErrUnknownMatchPolicyType,
+    #[strum(to_string = "The signature scheme is Unsupported.")]
+    ErrUnknownScheme,
```
jodh-intel, line 32: This is slightly confusing as the error text doesn't really align with the error ("unknown" compared to "unsupported").

signature/src/policy/policy_requirement.rs
```diff
@@ -160,166 +152,49 @@ impl PolicyRequirement for PolicyReqReject {
...
-                    return Ok(());
-                }
-                Err(e) => {
-                    reject_reason.push(e);
-                }
+        match SignatureScheme::from_str(&self.scheme) {
+            Ok(SignatureScheme::SimpleSigning) => {
+                return mechanism::simple::judge_signatures_accept(self.clone(), image);
             }
+            // TODO: Add more signature mechanism.
```
jodh-intel, line 192: Again, a GH issue URL would be good to avoid forgetting about this.

jodh-intel (commented):
Thanks @jialez0 - a few comments...

signature/src/mechanism/simple/mod.rs
jialez0: The order of signature files is not important. The logic here is that as long as there is a signature that can be verified correctly, the image can be accepted.

signature/src/image/mod.rs
jialez0: Here, we probably shouldn't store `signatures` and `sigstore_base_url` as the member fields of image, because an image may be required to verify the signatures under different schemes. I updated the code accordingly.

jialez0: Here, we probably shouldn't store `signature` and `sigstore_base_url` as the member fields of `image`, because an image may be required to verify the signatures under different schemes. I updated the code accordingly.

signature/src/mechanism/simple/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+            anyhow!(
+                "Read SignedBy keyPath failed: {:?}, path: {}",
+                e,
+                &signedby_req.key_path
+            )
+        })?
+    };
+
+    let sigs = image.signatures(&signedby_req.scheme)?;
+    let mut reject_reason: Vec<anyhow::Error> = Vec::new();
```
jodh-intel, line 57: Nit: This should strictly be plural as it could contain >1 reason:

```suggestion
    let mut reject_reasons: Vec<anyhow::Error> = Vec::new();
```

jodh-intel (approved):
Thanks @jialez0.

lgtm

jodh-intel:
Anyone else in the community able to tal at this PR?

signature/src/mechanism/simple/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+}
+
+pub fn judge_single_signature(
+    image: &image::Image,
+    signed_identity: Option<&Box<dyn policy::PolicyReferenceMatcher>>,
+    pubkey_ring: Vec<u8>,
+    sig: Vec<u8>,
+) -> Result<()> {
+    // Verify the signature with the pubkey ring.
+    let sig_payload = verify::verify_sig_and_extract_payload(pubkey_ring, sig)?;
```
arronwy, line 93: The verifier may also need support different key types, but it's ok for now since we only support GPGKey.

arronwy (approved):
LGTM, Thanks @jialez0

signature/src/mechanism/simple/mod.rs
jialez0: Yes, I mentioned this in issue #14.

PR #36, opened 2022-06-22
Pin a specific version of ocicrypt-rs

fidencio:
Instead of using ocicrypt-rs from `main`, let's pin a specific version,
which is known to be working, and move to newer commits as it becomes
required / desired, instead of keep following ocicrypt-rs' main.

Signed-off-by: Fabiano Fidêncio <fabiano.fidencio@intel.com>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ libc = "0.2"
 nix = "0.23.0"
 oci-distribution = { git = "https://github.com/krustlet/oci-distribution", rev = "1ba0d94a900a97aa1bcac032a67ea23766bcfdef" }
 oci-spec = { git = "https://github.com/containers/oci-spec-rs" }
-ocicrypt-rs = { git = "https://github.com/confidential-containers/ocicrypt-rs" }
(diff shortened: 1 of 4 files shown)


fidencio:
I've updated the PR with an attempt to fix the warnings that are not related to my PR, raised by the nightly CI.

fidencio:
We have a small issue here that I'm not exactly sure how to solve.

The autogenerated code for the `protos/getresource.proto` file fails to build with the `nightly` CI due to "you are deriving `PartialEq` and can implement `Eq`".  This, most likely, will end up fixed in the `prost` side at some point, but not sure how we should proceed here.

fidencio:
And the issue has been reported to prost: https://github.com/tokio-rs/prost/issues/670

fidencio:
Okay, it seems good to be reviewed.

fidencio:
> Okay, it seems good to be reviewed.

Spoke too soon! :-)

fidencio:
Now it's finally good to review! :-)

arronwy (approved):
Thanks @fidencio , lgtm.

jialez0 (approved):
LGTM.

PR #37, opened 2022-07-01
snapshot: Add occlum/unionfs.rs to adapt to pulling images in the LibOS Occlum environment

zhiwei-intel-h:
1 Add occlum/unionfs.rs only for Occlum adaption.
2 Add overlay_feature and occlum_feature for differentiate usage.
3 Mount the host path to the specified path on Occlum.
4 Move the pulled image to the specified path on Occlum.
5 Create the required environment for Occlum.

Signed-off-by: Zhiwei Han <zhiwei1.han@intel.com>

### DIFF
--- a/src/snapshots/occlum/unionfs.rs
+++ b/src/snapshots/occlum/unionfs.rs
@@ -1,0 +1,266 @@
+// Copyright (c) 2022 Intel Corporation
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// This unionfs file is used for occlum only
+
+use std::fs;
+use std::path::{Path, PathBuf};
+use std::sync::atomic::AtomicUsize;
+
+use anyhow::{anyhow, Result};
+use dircpy::CopyBuilder;
+use fs_extra;
+use fs_extra::dir;
+use nix::mount::MsFlags;
+
+use crate::snapshots::{MountPoint, Snapshotter};
+
+const LD_LIB: &str = "ld-linux-x86-64.so.2";
+
+#[derive(Debug)]
+pub struct Unionfs {
+    pub data_dir: PathBuf,
+    pub index: AtomicUsize,
+}
+
+fn clear_path(mount_path: &Path) -> Result<()> {
+    let mut from_paths = Vec::new();
+    let paths = fs::read_dir(
+        mount_path
+            .to_str()
+            .ok_or(anyhow!("mount_path does not exist"))?,
+    )?;
+    for path in paths {
+        from_paths.push(path?.path());
+    }
+    fs_extra::remove_items(&from_paths)?;
+
+    Ok(())
+}
+
+fn create_dir(create_path: &PathBuf) -> Result<()> {
+    if !create_path.exists() {
+        fs::create_dir_all(create_path.as_path())?;
+    }
+
+    Ok(())
+}
+
+fn create_environment(mount_path: &Path) -> Result<()> {
+    let mut from_paths = Vec::new();
+    let mut copy_options = dir::CopyOptions::new();
+    copy_options.overwrite = true;
+
+    // copy the libs required by occlum to the mount path
+    let path_lib64 = mount_path.join("lib64");
+    create_dir(&path_lib64)?;
+
+    let lib64_libs = vec![LD_LIB];
+    let ori_path_lib64 = Path::new("/lib64");
+    for lib in lib64_libs.iter() {
+        from_paths.push(ori_path_lib64.join(lib));
+    }
+
+    // if ld-linux-x86-64.so.2 as symlink exist in ${path_lib64},
+    // copy ld-linux-x86-64.so.2 from occlum to ${path_lib64} failed (file exists).
+    // so firstly remove it.
+    let ld_lib = path_lib64.join(LD_LIB);
+    if fs::symlink_metadata(ld_lib.as_path()).is_ok() {
+        fs::remove_file(ld_lib)?;
+    }
+
+    fs_extra::copy_items(&from_paths, &path_lib64, &copy_options)?;
+    from_paths.clear();
+
+    let path_opt = mount_path
+        .join("opt")
+        .join("occlum")
+        .join("glibc")
+        .join("lib");
+    fs::create_dir_all(&path_opt)?;
+
+    let occlum_lib = vec![
+        "libc.so.6",
+        "libdl.so.2",
+        "libm.so.6",
+        "libpthread.so.0",
+        "libresolv.so.2",
+        "librt.so.1",
+    ];
+
+    let ori_occlum_lib_path = Path::new("/")
+        .join("opt")
+        .join("occlum")
+        .join("glibc")
+        .join("lib");
+    for lib in occlum_lib.iter() {
+        from_paths.push(ori_occlum_lib_path.join(lib));
+    }
+    fs_extra::copy_items(&from_paths, &path_opt, &copy_options)?;
+    from_paths.clear();
+
+    let sys_path = vec!["dev", "etc", "host", "lib", "proc", "root", "sys", "tmp"];
+    for path in sys_path.iter() {
+        create_dir(&mount_path.join(path))?;
+    }
+
+    Ok(())
+}
+
+impl Snapshotter for Unionfs {
+    fn mount(&mut self, layer_path: &[&str], mount_path: &Path) -> Result<MountPoint> {
+        // From the description of https://github.com/occlum/occlum/blob/master/docs/runtime_mount.md#1-mount-trusted-unionfs-consisting-of-sefss ,
+        // the source type of runtime mount is "unionfs".
+        let fs_type = String::from("unionfs");
+        let source = Path::new(&fs_type);
+
+        if !mount_path.exists() {
+            fs::create_dir_all(mount_path)?;
+        }
+
+        // store the rootfs in different places according to the cid
+        let cid = mount_path
+            .parent()
+            .ok_or(anyhow!("parent do not exist"))?
+            .file_name()
+            .ok_or(anyhow!("Unknown error: file name parse fail"))?;
+        let sefs_base = Path::new("/images").join(cid).join("sefs");
+        let unionfs_lowerdir = sefs_base.join("lower");
+        let unionfs_upperdir = sefs_base.join("upper");
+
+        // For mounting trusted UnionFS at runtime of occlum,
+        // you can refer to https://github.com/occlum/occlum/blob/master/docs/runtime_mount.md#1-mount-trusted-unionfs-consisting-of-sefss.
+        // "c7-32-b3-ed-44-df-ec-7b-25-2d-9a-32-38-8d-58-61" is a hardcode key used to encrypt or decrypt the FS currently,
+        // and it will be replaced with dynamic key in the near future.
+        let options = format!(
+            "lowerdir={},upperdir={},key={}",
+            unionfs_lowerdir.display(),
+            unionfs_upperdir.display(),
+            "c7-32-b3-ed-44-df-ec-7b-25-2d-9a-32-38-8d-58-61"
+        );
+
+        let flags = MsFlags::empty();
+
+        nix::mount::mount(
+            Some(source),
+            mount_path,
+            Some(fs_type.as_str()),
+            flags,
+            Some(options.as_str()),
+        )
+        .map_err(|e| {
+            anyhow!(
+                "failed to mount {:?} to {:?}, with error: {}",
+                source,
+                mount_path,
+                e
+            )
+        })?;
+
+        // clear the mount_path if there is something
+        clear_path(mount_path)?;
+
+        // copy dirs to the specified mount directory
+        let mut layer_path_vec = layer_path.to_vec();
+        let len = layer_path_vec.len();
+        for _i in 0..len {
+            let layer = layer_path_vec
+                .pop()
+                .ok_or(anyhow!("Pop() failed from Vec"))?;
+            CopyBuilder::new(layer, &mount_path).overwrite(true).run()?;
+        }
+
+        // create environment for Occlum
+        create_environment(mount_path)?;
+
+        nix::mount::umount(mount_path)?;
+
+        Ok(MountPoint {
+            r#type: fs_type,
+            mount_path: mount_path.to_path_buf(),
+            work_dir: self.data_dir.to_path_buf(),
+        })
+    }
+
+    fn unmount(&self, mount_point: &MountPoint) -> Result<()> {
+        nix::mount::umount(mount_point.mount_path.as_path())?;
+
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::fs::File;
+    use std::io::Write;
+    use tempfile;
+
+    #[test]
+    fn test_create_dir() {
+        let tempdir = tempfile::tempdir().unwrap();
+
+        let foo_dir = tempdir.path().join("foo");
+        assert!(!foo_dir.exists());
+
+        create_dir(&foo_dir).unwrap();
+        assert!(foo_dir.exists());
+    }
+
+    #[test]
+    fn test_clear_path() {
+        let tempdir = tempfile::tempdir().unwrap();
+
+        let file = tempdir.path().join("foo.txt");
+        let mut f = File::create(file.as_path()).unwrap();
+        f.write_all(b"Hello, world!").unwrap();
+        assert!(file.exists());
+
+        clear_path(tempdir.path()).unwrap();
+        assert!(!file.exists());
+    }
+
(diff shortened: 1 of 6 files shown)


jialez0:
@zhiwei-intel-h Hi, can you rebase this PR to the latest version of upstream?

src/snapshots/occlum/unionfs.rs
```diff
@@ -0,0 +1,167 @@
...
+        .join("glibc")
+        .join("lib");
+    create_dir(&path_opt)?;
+
+    from_paths.push(PathBuf::from("/opt/occlum/glibc/lib/libc.so.6"));
+    from_paths.push(PathBuf::from("/opt/occlum/glibc/lib/libdl.so.2"));
+    from_paths.push(PathBuf::from("/opt/occlum/glibc/lib/libm.so.6"));
+    from_paths.push(PathBuf::from("/opt/occlum/glibc/lib/libpthread.so.0"));
+    from_paths.push(PathBuf::from("/opt/occlum/glibc/lib/libresolv.so.2"));
+    from_paths.push(PathBuf::from("/opt/occlum/glibc/lib/librt.so.1"));
```
jodh-intel, line 75: Could you use a loop for this to both reduce the number of lines but also only have a single occurence of `/opt/occlum` (could that ever change btw?)

src/snapshots/occlum/unionfs.rs
```diff
@@ -0,0 +1,167 @@
...
+            .ok_or(anyhow!("Unknown error: file name pasre fail"))?;
+        let sefs_base = Path::new("/images").join(cid).join("sefs");
+        let unionfs_lowerdir = sefs_base.join("lower");
+        let unionfs_upperdir = sefs_base.join("upper");
+
+        let options = format!(
+            "lowerdir={},upperdir={},key={}",
+            unionfs_lowerdir.display(),
+            unionfs_upperdir.display(),
+            "c7-32-b3-ed-44-df-ec-7b-25-2d-9a-32-38-8d-58-61"
```
jodh-intel, line 116: Could you make this a global `const` with a comment explaining what this actually is (random uuid, or a magic/special value?)

src/snapshots/occlum/unionfs.rs
```diff
@@ -0,0 +1,167 @@
...
+    from_paths.clear();
+
+    create_dir(&mount_path.join("host"))?;
+    create_dir(&mount_path.join("proc"))?;
+    create_dir(&mount_path.join("dev"))?;
+    create_dir(&mount_path.join("etc"))?;
+    create_dir(&mount_path.join("lib"))?;
+    create_dir(&mount_path.join("root"))?;
+    create_dir(&mount_path.join("sys"))?;
+    create_dir(&mount_path.join("tmp"))?;
```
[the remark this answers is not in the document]
zhiwei-intel-h: Ok, will use loop to implement it.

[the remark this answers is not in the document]
zhiwei-intel-h: > * Unsorted list.
> * Why not use a loop over a string slice for this?

Ok, will use loop to implement it.

src/snapshots/unionfs.rs
```diff
@@ -0,0 +1,148 @@
...
+        .join("lib");
+    create_dir(&path_opt)?;
+
+    from_paths.push(PathBuf::from("/opt/occlum/glibc/lib/libc.so.6"));
+    from_paths.push(PathBuf::from("/opt/occlum/glibc/lib/libdl.so.2"));
+    from_paths.push(PathBuf::from("/opt/occlum/glibc/lib/libm.so.6"));
+    from_paths.push(PathBuf::from("/opt/occlum/glibc/lib/libpthread.so.0"));
+    from_paths.push(PathBuf::from("/opt/occlum/glibc/lib/libresolv.so.2"));
+    from_paths.push(PathBuf::from("/opt/occlum/glibc/lib/librt.so.1"));
+    fs_extra::copy_items(&from_paths, &path_opt, &copy_options).unwrap();
```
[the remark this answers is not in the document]
zhiwei-intel-h: > These operations are occlum specific, we may avoid use the generic `unionfs` module, maybe `unionfs_occlum`?

Aligned with you, will add occlum dir to include all implementations based on occlum.

src/snapshots/occlum/unionfs.rs
```diff
@@ -0,0 +1,167 @@
...
+    for path in paths {
+        from_paths.push(path?.path());
+    }
+    fs_extra::remove_items(&from_paths)?;
+
+    Ok(())
+}
+
+fn create_dir(create_path: &PathBuf) -> Result<()> {
+    if !create_path.exists() {
```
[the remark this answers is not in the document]
zhiwei-intel-h: > You could remove the check as `fs::create_dir_all()` is equivalent to `mkdir -p` I think?

Yes ,  `fs::create_dir_all()` is equivalent to `mkdir -p`.  But I think it is logical to create a directory only after checking it does not exist.

src/snapshots/occlum/unionfs.rs
zhiwei-intel-h: > Could you use a loop for this to both reduce the number of lines but also only have a single occurence of `/opt/occlum` (could that ever change btw?)

I could use a loop for this part.  The benefit is that only add an item in the string slice in the future if a new lib have to be added required by occlum.

src/snapshots/occlum/unionfs.rs
```diff
@@ -0,0 +1,176 @@
...
+        "libpthread.so.0",
+        "libresolv.so.2",
+        "librt.so.1",
+    ];
+    let ori_occlum_lib_path = Path::new("/opt/occlum/glibc/lib");
+    for lib in occlum_lib.iter() {
+        from_paths.push(ori_occlum_lib_path.join(lib));
+    }
+    fs_extra::copy_items(&from_paths, &path_opt, &copy_options)?;
+    from_paths.clear();
```
mythi, line 86: Currently, `shim-rune` combines Occlum parts with the app using an overlayfs mount. It sounds this copying is redundant.

qzheng527: From Occlum pespective, the default glibc libraries in the app container layers couldn't be directly used in Occlum enviroment for some reason. Instead, Occlum provides its own modified glibc libraries. Thus, to generate a runnable RootFS in Occlum, the original glibc libraries need to be replaced with the ones Occlum provided.

mythi: > Thus, to generate a runnable RootFS in Occlum, the original glibc libraries need to be replaced with the ones Occlum provided.

Do they really need to be replaced if the libOS/container just sets `LD_LIBRARY_PATH` correctly? Right now Occlum libc is both copied (and encrypted) AND overlay mounted from the host into the container rootfs.

src/snapshots/occlum/unionfs.rs
```diff
@@ -0,0 +1,188 @@
...
+        if !mount_path.exists() {
+            fs::create_dir_all(mount_path)?;
+        }
+
+        // store the rootfs in different places according to the cid
+        let cid = mount_path
+            .parent()
+            .ok_or(anyhow!("parent do not exist"))?
+            .file_name()
+            .ok_or(anyhow!("Unknown error: file name pasre fail"))?;
```
mythi, line 124:
```suggestion
            .ok_or(anyhow!("Unknown error: file name parse fail"))?;
```

src/snapshots/occlum/unionfs.rs
qzheng527: > Thus, to generate a runnable RootFS in Occlum, the original glibc libraries need to be replaced with the ones Occlum provided.

The "replaced" in above sentence is not that "physically replaced". The original glibc libraries are still in the final rootfs image. Copying the Occlum glibc libraries to the path "/opt/occlum/glibc/lib" makes the LD could find the correct glibc libraries in the first order. The only "physically replaced" file is "/lib64/ld-linux-x86-64.so.2". Occlum uses this LD to find correct libraries in its defined order. (For glibc, firstly it is path "/opt/occlum/glibc/lib").

src/snapshots/occlum/unionfs.rs
zhiwei-intel-h: Yes, fixed.

src/snapshots/occlum/unionfs.rs
mythi: > Copying the Occlum glibc libraries to the path "/opt/occlum/glibc/lib" makes the LD could find the correct glibc libraries in the first order.

my original comment was why copy since `shim-rune` puts an overlayfs mount for them. Is there another reason why these lines needed?

qzheng527: > > Copying the Occlum glibc libraries to the path "/opt/occlum/glibc/lib" makes the LD could find the correct glibc libraries in the first order.
> 
> my original comment was why copy since `shim-rune` puts an overlayfs mount for them. Is there another reason why these lines needed?

@zhiwei-intel-h Can you help answer this question? My undertanding is that the `shim-rune` doesn't have these Occlum glibc libraries, thus the the copy is required in agent enclave stage.

zhiwei-intel-h: Now shim-rune do not combines Occlum libs with the app container using overlayfs mount.
From shim-rune design docs https://github.com/confidential-containers/enclave-cc/blob/main/docs/shim-rune.md#occlum,  
shim-rune overlay the boot instance bundle and encrypted unionfs image to generate the final bundle of the app container.
image

The encrypted unionfs image includes the pulled image and some libs required by Occlum.
image

The boot instance bundle mainly contains Occlum boot template instance.

When launching the app container, the boot instance bundle will find the path of the encrypted unionfs image , 
mount  app image and boot application.

mythi: > The encrypted unionfs image includes the pulled image and some libs required by Occlum.

@zhiwei-intel-h OK, thanks for the clarification! The way I've used gramine included both "boot instance" and the libs in the same mount so I assumed the same here too. Would this work with Occlum too? As I commented, copying and encrypting these onto the unionfs image sounds inefficient.

zhiwei-intel-h: > > The encrypted unionfs image includes the pulled image and some libs required by Occlum.
> 
> @zhiwei-intel-h OK, thanks for the clarification! The way I've used gramine included both "boot instance" and the libs in the same mount so I assumed the same here too. Would this work with Occlum too? As I commented, copying and encrypting these onto the unionfs image sounds inefficient.

As I know, currently  there is no way for Occlum to simply combine glibc libraries from Occlum through shim-rune in the stage of running app enclave.

mythi: I would have imagined this is possible using something like Occlum's `hostfs` mounts just like you are doing it with `sefs`.

qzheng527: @mythi `hostfs` mount is not a secure way.  To keep whole-flow security for the solution, I don't think `hostfs` should be used to mount "occlum glibc libraries". Moreover, besides "occlum glibc libraries", for some container images, "occlum musl-libc libraries" should be used as well. For shim-rune,  it doesn't know which one should be choosed, the "glibc" or "musl-libc". Thus, let the agent-enclave to do the choice makes sense.

mythi: @qzheng527 OK, maybe `hostfs` is not the right Occlum term I'm using. I mean that ideally the app enclave with Occlum would be able to just verify the integrity of those Occlum glibc libraries to ensure security and still make the app to use them.

YangLiang3:
@mythi @arronwy @jodh-intel @jialez0  Any other comments to the PR?

src/snapshots/occlum/unionfs.rs
```diff
@@ -0,0 +1,188 @@
...
+
+    Ok(())
+}
+
+fn create_environment(mount_path: &Path) -> Result<()> {
+    let mut from_paths = Vec::new();
+    let mut copy_options = dir::CopyOptions::new();
+    copy_options.overwrite = true;
+
+    #[cfg(target_arch = "x86_64")]
```
jodh-intel, line 53: ooi, do we need something like this in the codebase somewhere?

```rust
#[cfg(not(all(feature = "occlum_feature", target_arch = "x86_64")))]
compile_error!("occlum only available on Intel x86_64");
```

jodh-intel (commented):
Thanks @zhiwei-intel-h - a few comments...

src/snapshots/occlum/unionfs.rs
zhiwei-intel-h: Good comment. I think the best place is `src/snapshots/occlum/mod.rs`. What do you think of it?

jodh-intel: I think what we actually want is the following (but please test it ;)...

```rust
#[cfg(all(feature = "occlum_feature", not(target_arch = "x86_64")))]
compile_error!("occlum only available on Intel x86_64");
```

mythi:
> @mythi @arronwy @jodh-intel @jialez0 Any other comments to the PR?

@YangLiang3 I'm planning to experiment with the integrity protected mount for the app enclave to see can we skip copying Occlum's glibc to that encypted mount.

src/snapshots/occlum/unionfs.rs
zhiwei-intel-h: Ok, get it.

src/snapshots/occlum/unionfs.rs
```diff
@@ -0,0 +1,188 @@
...
+        let lib64_libs = vec!["ld-linux-x86-64.so.2"];
+        let ori_path_lib64 = Path::new("/lib64");
+        for lib in lib64_libs.iter() {
+            from_paths.push(ori_path_lib64.join(lib));
+        }
+
+        // if ld-linux-x86-64.so.2 as symlink exist in ${path_lib64},
+        // copy ld-linux-x86-64.so.2 from occlum to ${path_lib64} failed (file exists).
+        // so firstly remove it.
+        let ld_lib = path_lib64.join("ld-linux-x86-64.so.2");
```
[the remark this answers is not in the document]
zhiwei-intel-h: > Nit: Duplicated string `"ld-linux-x86-64.so.2"`. I'd consider a global `const` for this.

Fixed, thanks

src/snapshots/occlum/unionfs.rs
```diff
@@ -0,0 +1,188 @@
...
+
+use crate::snapshots::{MountPoint, Snapshotter};
+
+#[derive(Debug)]
+pub struct Unionfs {
+    pub data_dir: PathBuf,
+    pub index: AtomicUsize,
+}
+
+fn clear_path(mount_path: &Path) -> Result<()> {
```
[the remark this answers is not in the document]
zhiwei-intel-h: > Please add some unit tests for this function.

Yes, fixed.

src/snapshots/occlum/unionfs.rs
```diff
@@ -0,0 +1,188 @@
...
+    )?;
+    for path in paths {
+        from_paths.push(path?.path());
+    }
+    fs_extra::remove_items(&from_paths)?;
+
+    Ok(())
+}
+
+fn create_dir(create_path: &PathBuf) -> Result<()> {
```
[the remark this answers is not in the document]
zhiwei-intel-h: Yes, fixed

src/snapshots/occlum/unionfs.rs
```diff
@@ -0,0 +1,188 @@
...
+
+    let occlum_lib = vec![
+        "libc.so.6",
+        "libdl.so.2",
+        "libm.so.6",
+        "libpthread.so.0",
+        "libresolv.so.2",
+        "librt.so.1",
+    ];
+    let ori_occlum_lib_path = Path::new("/opt/occlum/glibc/lib");
```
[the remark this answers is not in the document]
zhiwei-intel-h: Get it, thanks

src/snapshots/occlum/unionfs.rs
```diff
@@ -0,0 +1,188 @@
...
+    let sys_path = vec!["dev", "etc", "host", "lib", "proc", "root", "sys", "tmp"];
+    for path in sys_path.iter() {
+        create_dir(&mount_path.join(path))?;
+    }
+
+    Ok(())
+}
+
+impl Snapshotter for Unionfs {
+    fn mount(&mut self, layer_path: &[&str], mount_path: &Path) -> Result<MountPoint> {
```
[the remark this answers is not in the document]
zhiwei-intel-h: Get it, thanks

arronwy (approved):
Thanks @zhiwei-intel-h LGTM

jodh-intel (approved):
Thanks @zhiwei-intel-h.

lgtm

PR #47, opened 2022-08-17
Add cosign signature verification support

Xynnn007:
This pr handles 1,2,3 in https://github.com/confidential-containers/image-rs/issues/46, including the following
* Implemention of cosign signature verification
* Documentation of the mechanism behind cosign and the `policy.json` format of cosign
* Unit tests for pulling cosign-signed images from different registries, now including `dockerhub`, `quay.io` and `aliyun.com`


signature/src/mechanism/cosign/README.md
```diff
@@ -0,0 +1,211 @@
...
+## Policy Format
+
+Cosign is verified due to the container's policy configuration file.
+
+A Policy Requirement of Cosign should be like this
+
+```json
+{
+    "type": "signedBy",
+    "scheme": "cosign",
```
jialez0, line 16: Should this document be updated and consistent with the code?

Xynnn007: Sorry, forgot to update. Fixed and force pushed

Xynnn007:
Do we need an integration test to verify a cosign-signed image?

signature/src/mechanism/cosign/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+Here, 
+* The `type` field must be `sigstoreSigned`, showing that this image is signed by `cosign`.
+* The `keyData` field includes the pubkey's content in base64.
+* The `keyPath` field indicates the pubkey's URL.
+* `signedIdentity` includes a JSON object, refer to [signedIdentity](https://github.com/containers/image/blob/main/docs/containers-policy.json.5.md#sigstoreSigned) for details.
+Because of the mechanism of Cosign, only `matchRepository` and `exactRepository`
+can be used to accept an image.
+
+**WARNING**: Must specify either `keyData` or `keyPath`, and must not both.
```
jodh-intel, line 30: You might want to use a block comment to make the warning stand out, as we do for [Kata](https://github.com/kata-containers/kata-containers/blob/main/docs/Documentation-Requirements.md#warnings-and-other-admonitions):

```suggestion
> **Warning**: Must specify either `keyData` or `keyPath`, but not both.
```

Which renders as:

> **Warning**: Must specify either `keyData` or `keyPath`, but not both.

Oh nice! GH now sees this as a warning and adds a warning symbol too! :smile:

signature/src/mechanism/cosign/README.md
```diff
@@ -0,0 +1,207 @@
...
+        "type": "cosign container image signature"
+    },
+    "optional": null
+}
+```
+
+Due to this, only `matchRepository` and `exactRepository`
+can be used to accept an image in Cosign.
+
+Now, if a client gets the signature and the `Payload`, he/she
```
jodh-intel, line 81: Might be easier to read if you just say "they":

```suggestion
Now, if a client gets the signature and the `Payload`, they
```

signature/src/mechanism/cosign/README.md
```diff
@@ -0,0 +1,207 @@
...
+- Download the image's manifest and digest.
+- Due to the digest, calculate the the reference for the signature "image".
+- Download the signature "image".
+- Extract signature and `Payload` from the downloaded signature "image".
+- Cryptographically verify the signature and the `Payload`.
+- Check the `signedIdentity` rules for the reference in it and the
+reference in `Payload`.
+
+**WARNING**: Only `matchRepository` and `exactRepository` can be use
+for Cosign in a `signedIdentity`.
```
jodh-intel, line 207:
```suggestion
> **Warning**: Only `matchRepository` and `exactRepository` can be used for Cosign in a `signedIdentity`.
```

signature/Cargo.toml
```diff
@@ -19,6 +19,7 @@ tokio = { version = "1.0", features = [ "rt", "macros", "rt-multi-thread" ] }
 tonic = "0.5"
 prost = "0.8"
 strum_macros = "0.24.2"
+sigstore_rs = { version = "0.3.3", package = "sigstore" }
```
jodh-intel, line 22: I notice that you've contributed to this crate. Could you get https://crates.io/crates/sigstore updated to show the repository URL? Currently, that is not showing anywhere so you need to search for it (I'm assuming it's https://github.com/sigstore/sigstore-rs :)

signature/src/mechanism/cosign/README.md
```diff
@@ -0,0 +1,207 @@
...
+in our defined interface for a signing scheme.
+
+## Work Flow
+
+Let's quickly see how Cosign works. For example, there's an image
+to be signed, named `172.17.0.3:5000/alpine:latest`.
+So as a reference, its
+* `registry`: `172.17.0.3:5000`, which is a test registry.
+* `repository`: `alpine`
+* `tag`: `latest`
```
jodh-intel, line 44: Nice example and explanation!

signature/src/mechanism/cosign/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+There will be three steps for the signing:
+- Signing the `Payload`
+- Wrap the `Payload` and signature into an image bundle
+- Upload the image bundle to the registry
+
+#### Signature Generate
+
+When we sign this image, we use a cosign-generated key pair.
+The digital signature algorithm is by default ECDSA-P256.
```
jodh-intel, line 56: ooi, how would I check that and how would I know if the default had been changed somehow?

Hmm, I've just found this, from https://github.com/sigstore/cosign/blob/main/specs/SIGNATURE_SPEC.md#signature-schemes:

> Implementations must support at least the following schemes:
> 
> - ECDSA-P256
> 
> No information about the signature scheme is included in the object. Clients must determine the signature scheme out-of-band during the verification process.

Why is the signature scheme not included though? Does that imply the verifier needs to attempt to verify against all the schemes it supports. If it fails to verify, it moves to the next scheme, _etc_? I might be missing something obvious, but if this is the way it works, it doesn't sound very efficient.

jodh-intel (commented):
Thanks @Xynnn007 - a few comments...

signature/src/mechanism/cosign/README.md
Xynnn007: Ha XD, brilliant!

signature/Cargo.toml
Xynnn007: Yes it is https://github.com/sigstore/sigstore-rs, and I'll update the url

signature/src/mechanism/cosign/README.md
Xynnn007: AFAIK, now cosign only uses `ECDSA-P256` to sign and verify (from the source code), and it does not provide any signature scheme option.

signature/src/mechanism/cosign/mod.rs
```diff
@@ -0,0 +1,346 @@
...
+        }
+
+        Ok(())
+    }
+
+    fn resource_manifest(&self) -> HashMap<&str, &str> {
+        let mut manifest = HashMap::new();
+        if let Some(key_path) = &self.key_path {
+            if !Path::new(key_path).exists() {
+                manifest.insert(COSIGN_KEY_KBS, &key_path[..]);
```
jodh-intel, line 77: I'm rather confused by this. If the key doesn't exist, we add it's path. But what if the file does exist? But also we're effectively doing `insert("key", path)`, which might be confusing as you may well expect the code to do `insert("key", key)`.

signature/src/mechanism/cosign/mod.rs
```diff
@@ -0,0 +1,346 @@
...
+        }
+
+        Ok(())
+    }
+}
+
+impl CosignParameters {
+    /// Check whether this Policy Request Match Type (i.e., signed identity
+    /// check type) for the reference
+    /// is MatchRepository or ExactRepository.
```
jodh-intel, line 107: Nit: Looks like these lines could be joined.

signature/src/mechanism/cosign/mod.rs
```diff
@@ -0,0 +1,346 @@
...
+    /// Because cosign-created signatures only contain a repository,
+    /// so only matchRepository and exactRepository can be used to accept them.
+    /// Other types are all to be denied.
+    /// If it is neither of them, return `Error`. Otherwise, return `Ok()`
+    fn check_reference_rule_types(&self) -> Result<()> {
```
jodh-intel, line 112: Could you add a unit test for all the rule scenarios?

signature/src/mechanism/cosign/mod.rs
```diff
@@ -0,0 +1,346 @@
...
+    }
+
+    /// Verify the cosign-signed image. There will be three steps:
+    /// * Get the pub key.
+    /// * Download the cosign-signed image's manifest and its digest. Calculate its
+    /// signature's image.
+    /// * Download the signature image, gather the signatures and verify them
+    /// using the pubkey.
+    /// If succeeds, the payloads of the signature will be returned.
+    async fn verify_signature_and_get_payload(&self, image: &mut Image) -> Result<Vec<SigPayload>> {
```
jodh-intel, line 130: Could you add a unit test for this method?

signature/src/mechanism/cosign/mod.rs
```diff
@@ -0,0 +1,346 @@
...
+                Ok(payloads)
+            }
+            Err(SigstoreVerifyConstraintsError {
+                unsatisfied_constraints,
+            }) => Err(anyhow!("{:?}", unsatisfied_constraints)),
+        }
+    }
+}
+
+async fn read_key_from_url(url: &str) -> Result<Vec<u8>> {
```
jodh-intel, line 175: Could you unit test some of the failure paths for this function (empty url, invalid url, invalid scheme, etc)?

jodh-intel (commented):
Thanks @Xynnn007 - a few comments...

signature/src/mechanism/cosign/README.md
```diff
@@ -0,0 +1,206 @@
...
+{
+    "type": "sigstoreSigned",
+    "keyData": "<PUBKEY-DATA-IN-BASE64>",
+    "keyPath": "<URL-TO-THE-PUBKEY>",
+    "signedIdentity": <JSON-OBJECT>,
+},
+```
+
+Here, 
+* The `type` field must be `sigstoreSigned`, showing that this image is signed by `cosign`.
```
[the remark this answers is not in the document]
Xynnn007: Yes, I'll fix it

signature/src/mechanism/cosign/README.md
```diff
@@ -0,0 +1,206 @@
...
+to be signed, named `172.17.0.3:5000/alpine:latest`.
+So as a reference, its
+* `registry`: `172.17.0.3:5000`, which is a test registry.
+* `repository`: `alpine`
+* `tag`: `latest`
+
+### Signing
+
+There will be three steps for the signing:
+- Signing the `Payload`
```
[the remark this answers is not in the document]
Xynnn007: Yes thanks, I'll fix it

signature/src/mechanism/cosign/README.md
```diff
@@ -0,0 +1,206 @@
...
+- Wrap the `Payload` and signature into an image bundle
+- Upload the image bundle to the registry
+
+#### Signature Generate
+
+When we sign this image, we use a cosign-generated key pair.
+The digital signature algorithm is by default ECDSA-P256.
+The signed object, i.e. `Payload`, is [simple signing payload](https://github.com/sigstore/cosign/blob/main/specs/SIGNATURE_SPEC.md#simple-signing).
+The reference for the signed image will be included in the `Payload`.
+However, ranther than the full reference, only `registry` and
```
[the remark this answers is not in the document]
Xynnn007: Thanks for the deep dive!

signature/src/mechanism/cosign/README.md
```diff
@@ -0,0 +1,206 @@
...
+
+Now, let's store the signature and the `Payload` in the repository.
+
+There will be a new entry to be stored in the repository.
+This new entry includes the signature and the `Payload`.
+
+Due to [OCI Distribution Spec](https://github.com/opencontainers/distribution-spec/blob/main/spec.md),
+an image is stored in the following way in a image registry:
+* There must be a `manifest` file to show the components that make up a container image.
+* Each component (e.g., filesystem layers) is stored in a sha256-addressed
```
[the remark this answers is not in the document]
Xynnn007: Thanks, will fix it

signature/src/mechanism/cosign/README.md
```diff
@@ -0,0 +1,206 @@
...
+Then, follow the steps to verify a Cosign-signed image.
+- Download the image's manifest and digest.
+- Due to the digest, calculate the the reference for the signature "image".
+- Download the signature "image".
+- Extract signature and `Payload` from the downloaded signature "image".
+- Cryptographically verify the signature and the `Payload`.
+- Check the `signedIdentity` rules for the reference in it and the
+reference in `Payload`.
+
+> **WARNING**: Only `matchRepository` and `exactRepository` can be use for Cosign in a `signedIdentity`.
```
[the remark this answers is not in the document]
Xynnn007: Oh, cool!

signature/src/mechanism/cosign/README.md
```diff
@@ -0,0 +1,206 @@
...
+
+Let's see how the "image" is stored:
+- The image to be signed is `172.17.0.3:5000/alpine:latest`
+- Resolve the image reference tag to the digest `sha256:9b2a28eb47540823042a2ba401386845089bb7b62a9637d55816132c4c3c36eb`. Here
+the sha256 digest has the same function as a `tag`. They both can be used
+to find a specific image. And this sha256 digest is for example, not true.
+- Encoding the digest into a new name `sha256-9b2a28eb47540823042a2ba401386845089bb7b62a9637d55816132c4c3c36eb.sig`. The rule is
+    - Replace `:` character with a `-`
+    - Append the `.sig` suffix
+- Store the "image" in `172.17.0.3:5000/alpine:sha256-9b2a28eb47540823042a2ba401386845089bb7b62a9637d55816132c4c3c36eb.sig`
```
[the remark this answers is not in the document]
Xynnn007: Yes, you're right. The `latest` example here is not good, because as you've mentioned, the image tagged with `latest` will change. For other tags, the image and its relative sha256 digest will not change.

I'll fix the docs to use a new tag to discuss.

signature/src/mechanism/cosign/README.md
```diff
@@ -0,0 +1,206 @@
...
+We wrap the [rust implementation](https://github.com/sigstore/sigstore-rs) for sigstore to fit
+in our defined interface for a signing scheme.
+
+## Work Flow
+
+Let's quickly see how Cosign works. For example, there's an image
+to be signed, named `172.17.0.3:5000/alpine:latest`.
+So as a reference, its
+* `registry`: `172.17.0.3:5000`, which is a test registry.
+* `repository`: `alpine`
```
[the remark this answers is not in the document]
Xynnn007: Ok. The original IP, digests I used is real in a local docker repo I set up in a simple experiment. If I change it, the digest can be inaccurate. 

Well, I'll change it for better reading. Hope that nobody will check the digest is accurate :)

signature/src/mechanism/cosign/README.md
```diff
@@ -0,0 +1,206 @@
...
+        },
+        "type": "cosign container image signature"
+    },
+    "optional": null
+}
+```
+
+#### store "image" in the registry
+
+Let's see how the "image" is stored:
```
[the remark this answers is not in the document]
Xynnn007: Now I do not know it. IMO this case will be handled by the cosign signer rather than the registry. There may be some resident service to monitor this case.

signature/src/mechanism/cosign/mod.rs
Xynnn007: I'm not sure what confused you. Here `key_path` means `path-to-the-key`, e.g., `/etc/keys/cosign.pub`. It's not `/etc/keys`.

So I think here may be right. If there is no file `/etc/keys/cosign.pub`, the hashmap will be inserted `("Cosign Key", "/etc/keys/cosign.pub")`. Here `"Cosign Key"` tells the KBC to resource name to be requested. `"/etc/keys/cosign.pub"` declares when the resource bit stream is received, it will be stored into `"/etc/keys/cosign.pub"`

signature/src/mechanism/cosign/mod.rs
Xynnn007: Yea, thanks!

signature/src/mechanism/cosign/README.md
Xynnn007: Sure!

signature/src/mechanism/cosign/mod.rs
Xynnn007: Yes, will add it the next day

signature/src/mechanism/cosign/mod.rs
Xynnn007: I have refactored this function to treat the url as a path-to-filesystem. And more url schemes will be support in the issue https://github.com/confidential-containers/image-rs/issues/50 which may be continued after the first release. So now maybe we do not need unit test for this function?

signature/src/mechanism/cosign/mod.rs
Xynnn007: Added

signature/src/mechanism/cosign/mod.rs
Xynnn007: Added

Xynnn007:
Thanks for your deep dive @jodh-intel . I've add some unit tests and use `rstest` crate to keep the code dry. Please go on reviewing it, thanks!

signature/src/mechanism/cosign/mod.rs
```diff
@@ -0,0 +1,380 @@
...
+    // TODO: Do we need define a new URL scheme
+    // named `kbs://` to indicate that the key
```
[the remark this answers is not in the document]
Xynnn007: Yes, thanks for the link. I changed the function name to avoid confusing. We may take some effort to support more source of the key, like different urls, or a directly path to local file, etc.

signature/src/mechanism/cosign/mod.rs
Xynnn007: sry for the confusing name, I've changed it

bodzhang:
@Xynnn007 , does exactRepository support a full docker_reference_value of repository followed by the docker_manifest_digest  (as in the cosign payload)?

bodzhang:
Thanks, @Xynnn007 

The context of my question is about how a Pod owner can specify exact containers allowed to be loaded to the protected guest. A container Policy Config file that lists the trusted container manifest signing key(s) will accept any containers signed by the trusted key(s). I was wondering whether `exactRepository` policy supports the capability to specify allowed containers by their manifest digest. Looks like it's not the design intention of `exactRepository` policy.

I will submit an issue for design discussion about the need to specify exact containers in the Policy Config file.

Xynnn007:
@jodh-intel Hi, could you please take another review? Or can this PR be merged?

jialez0 (approved):
LGTM.

signature/src/mechanism/cosign/mod.rs
arronwy: maybe change the Hashmap name from `manifest` to `manifest_from_kbs` can be less confusing?

arronwy (approved):
Thanks @Xynnn007 , LGTM! only few comments.

signature/src/mechanism/cosign/README.md
```diff
@@ -0,0 +1,213 @@
...
+- Store the "image" in `example.com/alpine:sha256-9b2a28eb47540823042a2ba401386845089bb7b62a9637d55816132c4c3c36eb.sig`
+
+### Verification
+
+When a Policy Requirement with `type` set `sigstoreSigned`, the relative public key will be read
+due to `keyPath` or `keyData` field.
+
+Then, follow the steps to verify a Cosign-signed image.
+- Download the image's manifest and digest.
+- Due to the digest, calculate the the reference for the signature "image".
```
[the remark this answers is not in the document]
Xynnn007: Thanks, fixed it

Issue #54, opened 2022-08-24
Reduce pull code complexity


jodh-intel:
The `async pull_layers()` function is overly long and getting too complex imho. Break it into smaller functions to reduce complexity and legibility, and to allow those new parts to be unit tested.

PR #55, opened 2022-08-24
pull: Refactor pull_layers()

jodh-intel:
Split the `pull_layers()` method which now calls `handle_layer()` to
reduce the length and complexity of the original function.

Also implemented `TryFrom` for `Compression` (and added some unit tests)
to further simplify the code.

> **Note:**
>
> This PR also contains a new `assert_result!()` test macro that was
> originally written by me as `assert_log_level!()` for Kata Containers
> on:
>
> - https://github.com/kata-containers/kata-containers/pull/65
>
> ... but which now exists in Kata Containers as `assert_result!()` here:
>
> - https://github.com/kata-containers/kata-containers/blob/2b5dc2ad39b9b077246e5bc17c91cc2578c22497/src/libs/test-utils/src/lib.rs#L68

Fixes: #54.

### DIFF
--- a/src/pull.rs
+++ b/src/pull.rs
@@ -5,8 +5,7 @@
 use anyhow::{anyhow, Result};
 use futures_util::future;
 use oci_distribution::manifest::{OciDescriptor, OciImageManifest};
-use oci_distribution::{manifest, secrets::RegistryAuth, Client, Reference};
-use oci_spec::image::MediaType;
+use oci_distribution::{secrets::RegistryAuth, Client, Reference};
 use sha2::Digest;
 use std::convert::TryFrom;
 use std::fs;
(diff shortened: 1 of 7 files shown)


jodh-intel:
Note that this PR conflicts with #52. If that one lands first, I'll rebase to add the fix on that PR to the new `handle_layer()` method on this PR...

jodh-intel:
ftr, I think the pull code could / should be simplfied further. This is just the start ;)

src/pull.rs
```diff
@@ -83,108 +85,138 @@ impl PullClient {
...
+            return Err(e);
+        }
+
+        layer_meta.store_path = destination.display().to_string();
+
+        Ok(layer_meta)
+    }
+}
+
+impl TryFrom<&str> for Compression {
```
arronwy, line 192: Can we move the implementation and tests of `TryFrom` for `Compression` to https://github.com/confidential-containers/image-rs/blob/main/src/decoder/mod.rs ?

arronwy (commented):
Thanks @jodh-intel LGTM, just one suggestion inline.

arronwy:
> Note that this PR conflicts with #52. If that one lands first, I'll rebase to add the fix on that PR to the new `handle_layer()` method on this PR...

Yes, #52 is just merged, please help rebase with that fix.

src/pull.rs
jodh-intel: Good call - done.

jodh-intel:
@arronwy - I've updated the PR by:

- Creating a local crate for the `assert_result!()` macro so it can be used easily from different parts of the code.
- Added some tests for `handle_layer()`.

Note the single cached `PullClient()` in `test_handle_layer()` to avoid continually hitting the network.

arronwy (approved):
Thanks @jodh-intel , LGTM!

jodh-intel:
Resolved conflicted manifest file.

PR #57, opened 2022-08-31
test: add integration test using offline-fs-kbc

Xynnn007:
Close https://github.com/confidential-containers/image-rs/issues/44

- refactor integration test module for different kbcs
- add a new test image encrypted with `offline-fs-kbs`: `docker.io/xynnn007/busybox:encrypted`
- add integration test using offline-fs-kbc
- update docs for the integration test module


scripts/install_offline_fs_kbc_files.sh
```diff
@@ -0,0 +1,21 @@
...
+#
+# Copyright (c) 2022 Alibaba Cloud
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+script_dir="$(dirname $(readlink -f $0))"
+resource_json_name="aa-offline_fs_kbc-resources.json"
+test_resource_record="${script_dir}/../test_data/offline-fs-kbc/${resource_json_name}"
+target_file_path="/etc/${resource_json_name}"
```
jodh-intel, line 11: This line assumes the script will be run as `root`.

scripts/install_offline_fs_kbc_files.sh
```diff
@@ -0,0 +1,21 @@
...
+
+if [ $1 == "install" ]; then
+    cp "${test_resource_record}" "${target_file_path}"
```
jodh-intel, line 14: I'd consider:

- Removing one of the equals signs (as it's not required).
- Using `sudo` to allow this to run for non-`root` users who wish to run the tests.
- Being explicit about the perms and ownership

... resulting in something like:

```suggestion
if [ "${1:-}" = "install" ]; then
    sudo install --owner=root --group=root --mode=0640 "${test_resource_record}" "${target_file_path}"
```

scripts/install_offline_fs_kbc_files.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+#
+# Copyright (c) 2022 Alibaba Cloud
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
```
jodh-intel, line 7: Nit: I'd recommend adding this "boiler-plate" code at the top of the script:

```bash
set -o errexit
set -o nounset
set -o pipefail
set -o errtrace

```

scripts/install_offline_fs_kbc_files.sh
```diff
@@ -0,0 +1,21 @@
...
+resource_json_name="aa-offline_fs_kbc-resources.json"
+test_resource_record="${script_dir}/../test_data/offline-fs-kbc/${resource_json_name}"
+target_file_path="/etc/${resource_json_name}"
+
+if [ $1 == "install" ]; then
+    cp "${test_resource_record}" "${target_file_path}"
+
+elif [ $1 == "clean" ]; then
+    rm "${target_file_path}"
+
```
jodh-intel, line 18:
```suggestion
elif [ "${1:-}" = "clean" ]; then
    sudo rm "${target_file_path}"

```

scripts/install_offline_fs_kbc_files.sh
```diff
@@ -0,0 +1,21 @@
...
+else
+    echo "Wrong or missing argument"
```
jodh-intel, line 20: This is an error so I'd consider doing something like:

```suggestion
    echo >&2 "ERROR: Wrong or missing argument: '${1:-}'"
```

tests/README.md
```diff
@@ -41,3 +41,7 @@ any entry (`transports/docker`) in the policy.json. Now it refers to
 repository `quay.io/kata-containers/confidential-containers` concretely.
 * `allow/deny`: Whether the image to be pulled in this test should be allowed
 or denied.
+
+In `signature_verification.rs`, the tests are organized due different kinds
+of KBCs, which means for each given KBC, all the 4 tests mentioned will be
```
jodh-intel, line 46: Nit: Numbers less than ten are better written as words:

```suggestion
of KBCs, which means for each given KBC, all four tests mentioned will be
```

tests/signature_verification.rs
```diff
@@ -6,74 +6,72 @@
...
-
 struct TestItem<'a, 'b, 'c> {
     image_ref: &'a str,
     allow: bool,
     signing_scheme: &'b str,
     description: &'c str,
 }
 
+/// Four test cases.
+const TESTS: [TestItem; 4] = [
```
jodh-intel, line 28: Do these tests need to run as `root`? If so, we should consider using the macros defined below to skip if not running as that user:

- https://github.com/confidential-containers/image-rs/blob/main/src/snapshots/occlum/unionfs.rs#L235

jodh-intel (commented):
Thanks @Xynnn007 - a few comments...

scripts/install_offline_fs_kbc_files.sh
Xynnn007: Thanks for the suggestions and suggested change!

tests/README.md
Xynnn007: Thanks, sorry for non-native English user :-)

Xynnn007:
@jodh-intel Thanks for your suggestions. Have updated following them

jodh-intel (approved):
Thanks @Xynnn007.

lgtm

sameo (approved):

PR #60, opened 2022-09-21
features: support native signature client

HaokunX-intel:
Enclave-cc plans to replace the gRPC AA (attestation agent) with an native AA in the [issue 21](https://github.com/confidential-containers/enclave-cc/issues/21) . Currently, the signature sub-crate in this repo only supports communicating AA via gRPC socket. This PR adds an native signature client that uses an native AA to communicate with the remote KBS.

The PR uses the environment variable "ENABLE_SECURITY_VALIDATE" to enable image signature verification and loads the config file of the ocicrypt-rs to determine whether the AA is native.

Signed-off-by: haokun.xing <haokun.xin@intel.com>

### DIFF
--- a/.github/workflows/ci_eaa_kbc.yml
+++ b/.github/workflows/ci_eaa_kbc.yml
@@ -1,0 +1,60 @@
+name: image-rs eaa_kbc build
+on:
+  pull_request:
+    paths:
+      - 'signature/**'
+
+jobs:
+  ci:
+    if: github.event_name == 'pull_request' || github.event_name == 'push'
+    name: Check
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        rust:
+          - stable
+          - beta
+          - nightly
+    # Run all steps in the compilation testing containers
+    container:
+      image: runetest/compilation-testing:ubuntu18.04
+      env:
+        LD_LIBRARY_PATH: /usr/local/lib/rats-tls
+
+    steps:
+      - name: Update cargo home
+        run: |
+          apt-get update && apt-get install -y cargo
+          cp -r /root/.cargo /github/home/.cargo
+
+      - name: Install Rust toolchain (${{ matrix.rust }})
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: ${{ matrix.rust }}
+          override: true
+          components: rustfmt, clippy
+
+      - name: Install nettle-sys building dependence
+        run: |
+          apt-get install -y clang llvm pkg-config nettle-dev protobuf-compiler libprotobuf-dev
+
+      - name: Build and install rats-tls
+        run: | 
+          apt-get install -y libcurl4-openssl-dev
+          git clone https://github.com/inclavare-containers/rats-tls
+          cd rats-tls
+          cmake -DBUILD_SAMPLES=on -H. -Bbuild
+          make -C build install
+
+      - name: Code checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 1
+
+      - name: Run cargo build
+        uses: actions-rs/cargo@v1
+        with:
+          command: build
+          args: --features occlum_feature
--- a/signature/Cargo.toml
+++ b/signature/Cargo.toml
@@ -20,6 +20,8 @@ tonic = "0.5"
 prost = "0.8"
 strum_macros = "0.24.2"
 sigstore_rs = { git = "https://github.com/sigstore/sigstore-rs", rev = "6dd3281c25270f0d82550368abfb399ed3da7b41", package = "sigstore" }
+ocicrypt-rs = { git = "https://github.com/confidential-containers/ocicrypt-rs", optional = true }
+attestation_agent = { git = "https://github.com/confidential-containers/attestation-agent", rev = "3b4716dd3d8bbf0d5f8cec7bc0d528421f00fd06", optional = true }
 
 [build-dependencies]
 tonic-build = "0.5"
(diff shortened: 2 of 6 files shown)


signature/Cargo.toml
```diff
@@ -19,7 +19,9 @@ tokio = { version = "1.0", features = [ "rt", "macros", "rt-multi-thread" ] }
 tonic = "0.5"
 prost = "0.8"
 strum_macros = "0.24.2"
+ocicrypt-rs = { git = "https://github.com/confidential-containers/ocicrypt-rs" }
+attestation_agent = { git = "https://github.com/confidential-containers/attestation-agent" }
```
Xynnn007, line 23: Can we add a commit rev here? This can avoid potential problems caused by API changes in `AA`. The same as `ocicrypt-rs`.

signature/src/agent.rs
```diff
@@ -91,14 +105,29 @@ impl Agent {
...
-            kbc_name: self.kbc_name.clone(),
-            kbs_uri: self.kbc_uri.clone(),
-            resource_description,
-        });
-        let res = self.client.get_resource(req).await?;
-
-        fs::write(path, res.into_inner().resource).await?;
+        let res = if self.enable_native_aa {
+            AttestationAgent::new()
+                .download_confidential_resource(
```
Xynnn007, line 110: A quick question: 
every time we call `get_resource()` will create a new `AttestationAgent` object. What if we keep the `AA` object as a member of `Agent`, s.t.
```rust
pub struct Agent {
    /// Get Resource Service gRPC client
    client: Option<GetResourceServiceClient<Channel>>,
    aa: Option<AttestationAgent>,
    /// Enable Get Rserouce native AA client
    enable_native_aa: bool,
    kbc_name: String,
    kbs_uri: String,
}
```

Or an enum for `GetResourceServiceClient<Channel>` and `AttestationAgent`, for extensibility., i.e.
```rust
enum Client {
    GRPC(GetResourceServiceClient<Channel>),
    AA(AttestationAgent),
}
pub struct Agent {
    /// Get Resource Service gRPC client
    client: Client,
    enable_native_aa: bool,
    kbc_name: String,
    kbs_uri: String,
}
```

In both ways every time we call `get_resource()` will not do duplicated thins as creations and dropping of the aa object. How do you think about this?

Xynnn007 (commented):
Thanks. Overall, LGTM. Only a few small comments

signature/Cargo.toml
HaokunX-intel: Sure. It is meaningful.

signature/src/agent.rs
HaokunX-intel: Sure. We choose to use the enum to wrap the real client. Thanks for your great comments!

HaokunX-intel:
Need to wait for the [issue](https://github.com/confidential-containers/ocicrypt-rs/pull/32) in ocicrypt-rs to be merged to support `eaa_kbc` AA.

HaokunX-intel:
> Need to wait for the [issue](https://github.com/confidential-containers/ocicrypt-rs/pull/32) in ocicrypt-rs to be merged to support `eaa_kbc` AA.

The PR in ocicrypt-rs has been merged.

.github/workflows/ci_eaa_kbc.yml
```diff
@@ -0,0 +1,57 @@
+name: image-rs eaa_kbc build
+on: [push, pull_request, create]
```
arronwy, line 2: This CI is testing the build for `signature` module with `attestation-agent/eaa_kbc` , we can use path filter to run when `signature` module is changed: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore

src/config.rs
```diff
@@ -32,11 +32,18 @@ impl Default for ImageConfig {
         let work_dir = PathBuf::from(
             std::env::var(CC_IMAGE_WORK_DIR).unwrap_or_else(|_| DEFAULT_WORK_DIR.to_string()),
         );
-
         ImageConfig {
             work_dir,
             default_snapshot: SnapshotType::Overlay,
-            security_validate: false,
+            security_validate: std::env::var(ENABLE_SECURITY_VALIDATE).map_or_else(
```
arronwy, line 38: security validate is a key config to enable signature verification, first should we enable it by default? second, environment variable is not measured, the host may inject the envrionment variable to guest to disable our verification @sameo @jialez0 @Xynnn007 any suggestions?

arronwy: https://github.com/kata-containers/kata-containers/issues/4888

intchr: currently image-rs itself does not allow control variables to be passed down during pull_image(), and the ocicrypt-rs control reads its config separately inside image-rs instance, so my quick suggestion is as bellow:

1. with confidential containers cases, default to enable image verification and we have a switch in agent config(both Kata agent and enclave-cc agent) to provide a second choice for user with their serious consideration to disable it
2. image-rs to allow configuration to be sent down during pull_image() to allow caller to tune the exact behavior of image-rs in run time

arronwy: Hi @hairongchen @HaokunX-intel , with @stevenhorsman 's PR in `kata-agent`: https://github.com/kata-containers/kata-containers/pull/5552 , we can set `enable_signature_verification` config in kernel param of kata config file, can this meet your requirement?

hairongchen: Hi @arronwy thanks for the updates, we will update this PR to follow this method, i.e.: enclave-agent will read config from agent config file and set enable_signature_verification accordingly

arronwy (approved):
Thanks @HaokunX-intel , LGTM！

jialez0 (approved):

PR #61, opened 2022-09-29
Allow multiple bundles to be created from the same image

anakrish:
Earlier ImageClient::pull_image used to return early if the image had already been pulled. In this case, the second bundle created using an existing image would not get populated.

In this PR, if the image as already been pulled, then prior to returning early, the bundle is setup.

Signed-off-by: Anand Krishnamoorthi <anakrish@microsoft.com>

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -148,8 +148,23 @@ impl ImageClient {
         let (image_manifest, image_digest, image_config) = client.pull_manifest().await?;
 
         let id = image_manifest.config.digest.clone();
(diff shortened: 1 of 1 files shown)


arronwy (approved):
Thanks @anakrish LGTM!

arronwy:
@anakrish , please help fix the rust lint check reported by CI.

anakrish:
Thanks for the quick review, @arronwy. Fixed.

PR #67, opened 2022-10-24
Avoid multiple in-memory copies of image layers during pull

### LINKED ISSUE #66: Multiple copies of image layers maintained in memory during image pull
Currently during an image pull, multiple copies of each layer are maintained in memory before it is written out (possibly to secure storage):

1. Each layer's blob is first pulled into an in memory vector. https://github.com/confidential-containers/image-rs/blob/c98c4916cbe8d2c4f8933b343d20b7fdb338f50e/src/pull.rs#L91-L95
2. If the blob is encrypted, then it is decrypted to another in-memory vector. https://github.com/confidential-containers/image-rs/blob/c98c4916cbe8d2c4f8933b343d20b7fdb338f50e/src/pull.rs#L125-L127
3. Then if the layer is compressed, it is decompressed to another in memory vector. https://github.com/confidential-containers/image-rs/blob/c98c4916cbe8d2c4f8933b343d20b7fdb338f50e/src/pull.rs#L166-L168
4. The `out` vector which contains the layer's blob processed via the above steps is finally unpacked to destination folder: https://github.com/confidential-containers/image-rs/blob/c98c4916cbe8d2c4f8933b343d20b7fdb338f50e/src/pull.rs#L198

Pulling `docker.io/library/python:latest` using an image-rs test program on a linux machine shows the following memory consumption under heaptrack
```
peak heap memory consumption:
1.5GB after 11.733s
peak RSS (including heaptrack overhead):
888.8MB
total memory leaked:
50.9MB (10.6kB suppressed)
```

Given a kata-VM's default memory configuration of 2GB, 1.5GB memory consumption during image pull may cause slow performance and possibly OOM.

anakrish:
New flow:
blob ->  download to secure storage file -> optional decrypt to secure storage file -> optional decompress to secure storage file -> unpack to layer folder

With this PR, pulling `docker.io/library/python:latest` using a test image-rs program under heaptrack reports:
```
peak heap memory consumption:
53.2MB after 04.025s
peak RSS (including heaptrack overhead):
72.6MB
total memory leaked:
50.9MB (10.6kB suppressed)
```

fixes #66 

Signed-off-by: Anand Krishnamoorthi <anakrish@microsoft.com>

### DIFF
--- a/src/decrypt.rs
+++ b/src/decrypt.rs
@@ -14,7 +14,7 @@ use ocicrypt_rs::spec::{
 use oci_distribution::manifest;
 use oci_distribution::manifest::OciDescriptor;
 
-use std::io::Read;
+use std::io;
 
 #[derive(Default, Clone)]
 pub struct Decryptor {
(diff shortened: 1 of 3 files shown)


anakrish:
@arronwy Is this PR useful?

arronwy:
> @arronwy Is this PR useful?

Hi @anakrish , sure, this PR is very nice to reduce memory consumption, the only concern is use a `File` cache the layers, it may have performance issue with extra write/read from disk, also in cloud scenario the storage backend  may network base which have io rate limits.

dcmiddle:
I think this was eventually subsumed in https://github.com/confidential-containers/guest-components/pull/96/files.
Can we close this?
@anakrish @arronwy

PR #72, opened 2022-10-31
Add support for container registry credential retrievement

Xynnn007:
#### New Features
- Added container-auth support
- Added unit tests and integration test of container-auth

#### Change Log
- Refactored lib `signature` into a `mod`
- Added new parameter `RegistryAuth` to `SignScheme`, which is helpful when the signing scheme needs to interact with the registry (such as cosign to access a private registry where the signature stores)
- Added a switch boolean `auth` in `ImageConfig` to enable container-auth feature, which is disabled by default.
- Refactored logic in `pull_image`. 
    - Auth logic:
        - If `auth_info` is given, use it as the credential (authentication), regardless of `auth`.
        - If `auth_info` is not given and `auth` is enabled, check/get `auth.json` from Attestation-Agent and look up the credential due to the image reference
    - SecureChannel (used to communicate with AA) logic
        - When at least one of `auth` and `signature_verification` in `ImageConfig` is enabled, a `SecureChannel` will be establish due to the given `decrypt_config` (contents like `provider:attestation-agent:<kbc-name>:<kbs-url>`)
- Added container-auth configuration file `/run/image-security/auth.json`, and the related `ResourceDescriptor` used to request from KBC is **hard-coded** `Credential`
- Refactored sync to async functions in integration test module.

#### TODO List
- Finish `test_credential_from_auth_config` unit test when upstream is prepared (`oci_distribution` and `ocicrypt-rs`)
- Support credential helpers
- Support flexible `ResourceDescriptor` of `auth.json`

#### Related Issues
https://github.com/confidential-containers/image-rs/issues/63
https://github.com/kata-containers/kata-containers/issues/4601

#### Considerations
There are two related rust implementations for container auth:
* https://github.com/keirlawson/docker_credential: This does not support directly customerized `auth.json` path. The path can only be conveyed by default `~/.docker/` or by environment variable `DOCKER_CONFIG`, both of which are not safe. In CoCo we hope that the `auth.json` is located in `/run`, s.t. encrypted ram. Also this repo stopped updating for almost four years.
* https://github.com/krustlet/oci-distribution/pull/39: This PR aims to support container auth. Same disadvantages as the first one. Also still WIP.

So we choose to implement it. Althrough this pr does not support `CredentialHelpers`, we can implement the feature __lazy__ and __in need__.
cc @stevenhorsman @liudalibj


Xynnn007:
@HaokunX-intel @arronwy @hairongchen Hi, after https://github.com/confidential-containers/image-rs/pull/60 is merged, I will rebase and update `SecureChannel` related code in this PR.

tests/README.md
```diff
@@ -55,3 +55,17 @@ Here
...
+
+Registry Credential Retrievement will do the following steps to test this feature.
+- Get `auth.json` from the Attestation Agent. The test KBC is `Offline-Fs-Kbc`.
+- Try to pull images from private registry using the matched credential in `auth.json`
+
+The test cases are
+| Image Reference | Related credential|
+|---|---|
+|`docker.io/liudalibj/public-busy-box` |`bGl1ZGFsaWJqOlBhc3N3MHJkIXFhego=`|
+|`quay.io/liudalibj/public-busy-box`|`bGl1ZGFsaWJqOlBhc3N3MHJkIXFhego=`|
```
liudalibj, line 71: Should we use `private-busy-box` at here for these two images?

Xynnn007: Yea, thanks! fixed it

stevenhorsman:
Hey @Xynnn007  - this is looking great, can I just check a couple of things to make sure I understood them correctly when reading the code and description:
- In order to enable pulling the authentication info from the secure channel, we need to set `ImageConfig.auth` to be true (and it defaults to false). If this is correct then I'm trying to work out the best way to integrate this into kata-containers for the e2e flow. The options I can think of:
  1. Hard-code `ImageConfig.auth=true` in the kata-agent's image_rpc.rs. This is obviously the simplest, but I'm guessing that if we aren't using the attestation-agent then thinks might go wrong, so probably not the best idea
  2. Set `ImageConfig.auth=true` in the kata-agent's image_rpc.rs, if we are using an attestation-agent KBC (ie `aa_kbc_params` is set). This should mean that the secure channel is always available, but we might make extra calls that aren't nescessary - is there an big issues with this?
  3. Also the user to set something like `enable_authenticated_pull` in the kata-containers configuration. This *should* always be correct (if we trust users and get the default right [I guess it should be off], but we've already added a bunch of user config and I'd like to avoid adding more where possible.

  To me 2 seems the clear winner, but I might be missing something.
- The extra PR in `ocicrypt-distribution` to enable comparisons of `RegistryAuth` for the unit tests isn't blocking this PR being merged and we can create a separate issue to address this later?

Thanks so much!

src/secure_channel.rs
```diff
@@ -0,0 +1,101 @@
...
+    /// Get Resource Service client.
+    client: GetResourceServiceClient<Channel>,
+    kbc_name: String,
+    kbc_uri: String,
+}
+
+impl SecureChannel {
+    /// Create a new [`SecureChannel`], the input parameter:
+    /// * `aa_kbc_params`: s string with format `<kbc_name>::<kbs_uri>`.
+    pub async fn new(aa_kbc_params: &str) -> Result<Self> {
```
anakrish, line 58: In the overall flow where is  `aa_kbc_params` specified? Is it included in hardware measurement of the kata-cc vm?

Xynnn007:
Hey @stevenhorsman , firstly thanks for your questions!

> * To me 2 seems the clear winner, but I might be missing something.

I agreed with way 2. 

If we use way 3, the more flexibility provided than way 2 is that an user can set `ImageConfig.auth=false` while `aa_kbc_params` is set. This case can only happen when an user wants other resources to be used, like `CosignVerificationKey`, `Policy`, etc., at the same time he doesn't want images in his private registry to be pulled using his credential. 

This is a contradiction, for the image that `image-rs` aims to pull is assigned by the user himself (let's assume the security tunnel for user to manipulate). How can an user say "I want pull image A from my private registry" and "DO NOT use my credential" at the same time?

> * The extra PR in `ocicrypt-distribution` to enable comparisons of `RegistryAuth` for the unit tests isn't blocking this PR being merged and we can create a separate issue to address this later?

Yes, I agreed. For we do not know when `oci_distribution` will accept this PR. I'd like to open the issue to track this.

Another quick sync, this PR doesn't support `CredentialHelpers` now ([related link](https://pkg.go.dev/github.com/google/go-containerregistry/pkg/authn#section-readme)). I'll open an issue then to record, and if there is surely need, we will go on this then.

src/secure_channel.rs
Xynnn007: Hi, here is a brief call routine
`SecureChannel::new(aa_kbc_params)` in `mage-rs` <- `ImageClient::pull_image()` in `mage-rs` <- `AGENT_CONFIG` in [`kata-agent`](https://github.com/confidential-containers/kata-containers-CCv0/blob/CCv0/src/agent/src/image_rpc.rs#L249) <- [From Guest commandline](https://github.com/confidential-containers/kata-containers-CCv0/blob/CCv0/src/agent/src/main.rs#L85)

For the second question, sry for not deep dive into this now, but [this issue](https://github.com/kata-containers/kata-containers/issues/5178) may help you

stevenhorsman:
Great - I'll write up way 2 later then

> > 2. This should mean that the secure channel is always available, but we might make extra calls that aren't nescessary - is there an big issues with this?
> 
> I'm not sure what the unnecessary calls here. Are you referring to the calls which checks `aa_kbc_params` is `Some` or `None`?

So the potentially unnecessary call here that I'm thinking of is the case when a user wan't to use an encrypted, or signed image from a public registry. They will need `aa_kbc_params` set, for the encryption/signature, so we'd turn the `ImageConfig.auth=true` and therefore try and get the credentials using the attestation-agent's secure channel, but there won't be any, so we'd fall back to using Anonymous credentials, so that request wouldn't have be strictly required to do. I don't think it should be a significant overhead though, so am not sure it's worth worrying about.

I hope that helps explain my thinking a little bit more?

Xynnn007:
> So the potentially unnecessary call here that I'm thinking of is the case when a user wan't to use an encrypted, or signed image from a public registry. They will need `aa_kbc_params` set, for the encryption/signature, so we'd turn the `ImageConfig.auth=true` and therefore try and get the credentials using the attestation-agent's secure channel, but there won't be any, so we'd fall back to using Anonymous credentials, so that request wouldn't have be strictly required to do. I don't think it should be a significant overhead though, so am not sure it's worth worrying about.

Got it clearly, thanks. I'd fix this in this PR.

src/secure_channel.rs
```diff
@@ -0,0 +1,101 @@
...
+        }
+    }
+}
+
+/// SecureChannel to connect with KBS
+pub struct SecureChannel {
+    /// Get Resource Service client.
+    client: GetResourceServiceClient<Channel>,
+    kbc_name: String,
+    kbc_uri: String,
```
HaokunX-intel, line 52: If I understand correctly, the `kbc_uri` is same as the `KBS_URI` in the [docs](https://github.com/confidential-containers/attestation-agent/blob/main/docs/IMPLEMENTATION.md) of the attestation-agent. Should we follow the convention, converting the  `kbc_uri`  to `kbs_uri`?

Xynnn007: Yes, thanks! Fixed it

src/secure_channel.rs
anakrish: Thanks for the description, @Xynnn007. One thing that is not clear is whether which KBS kata-agent/image-rs talks to is measured.

jialez0: Hi, @anakrish , in the current deployment model, KBS is controlled by the tenant and deployed in the security environment outside the cloud by the tenant, so KBS does not need to be measured. In other words, KBS actually plays the role of "rely party and verifier".

liudalibj (approved):
LGTM

stevenhorsman (approved):
This looks great to me, but I'm not an image-rs expert, so would like @jialez0, and/or @arronwy to review it if possible before it is merged?

Xynnn007:
Hi @HaokunX-intel , I have rebased the code you've pushed but still involve some modification. Would you please help to check the two things if they follow your initial impl? 
* features in `Cargo.toml`
* `NaiveAA` in `secure_channel.rs`

Thanks a lot!

arronwy (approved):
Thanks @Xynnn007 , LGTM! The only concern is current API between `secure_channel` to `attestation-agent`/`attestation-service`  without identity info which don't support multiple users, then these credentials are not isolated for different user.

jialez0 (approved):

PR #77, opened 2022-11-23
fix: pulling error when image has duplicated layers

Xynnn007:
When an image to be pulled has duplicated layers, the 2nd layer will write the same layer data and raise layer. This is because the time to update the database that stores layer metadata is after pulling all layers of an image. This commit
will try to construct a new layer list to be pulled via deduplicating the layers of the OciManifest.

Close https://github.com/confidential-containers/image-rs/issues/76

Signed-off-by: Xynnn007 <xynnn@linux.alibaba.com>

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -8,7 +8,7 @@ use oci_distribution::secrets::RegistryAuth;
 use oci_distribution::Reference;
 use oci_spec::image::{ImageConfiguration, Os};
 use serde::Deserialize;
-use std::collections::HashMap;
+use std::collections::{BTreeSet, HashMap};
 use std::convert::TryFrom;
 use std::path::Path;
 use std::sync::atomic::AtomicUsize;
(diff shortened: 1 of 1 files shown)


src/image.rs
```diff
@@ -282,13 +282,19 @@ impl ImageClient {
             ));
         }
 
+        let mut layers = Vec::new();
```
arronwy, line 285: Maybe use `unique_layers` will be more straightforward

arronwy (approved):
Thanks @Xynnn007 LGTM!

src/image.rs
Xynnn007: Thanks, much clearer! fixed it

Xynnn007:
The nightly error occurs because of crate `rstest`, and it is being handled by the crate's owner now. I think we can fix it some days later in another pr.

arronwy:
> The nightly error occurs because of crate `rstest`, and it is being handled by the crate's owner now. I think we can fix it some days later in another pr.

Yes, nightly CI error is not related to this PR: https://github.com/la10736/rstest/issues/169

jialez0 (approved):

PR #79, opened 2022-11-24
image: Update layer db with all success pulled layers

arronwy:
Some layers may failed to pull with network issues, orchectration framwork like k8s will trigger repull, we can skip layers pulled successfully last time.

Signed-off-by: Wang, Arron <arron.wang@intel.com>


Xynnn007 (approved):
LGTM, thanks

jialez0 (approved):
LGTM.

PR #81, opened 2022-11-25
Exclude the sigstore_rs library when building on s390x

mattarnoatibm:
Exclude the sigstore_rs library when building on s390x and do not support cosign image signing on s390x.

Fixes: confidential-containers#80

Signed-off-by: Matthew Arnold <mattarno@uk.ibm.com>


mattarnoatibm:
Hi Hendrick @hbrueckner . Could you check this PR for excluding the sigstore-rs crate and cosign image signing when building on s390x.

src/signature/mechanism/cosign/mod.rs
```diff
@@ -173,8 +177,18 @@ impl CosignParameters {
...
         }
     }
+
+    #[cfg(target_arch = "s390x")]
+    async fn verify_signature_and_get_payload(
+        &self,
+        _image: &mut Image,
+        _auth: &RegistryAuth,
+    ) -> Result<Vec<SigPayload>> {
+        Err(anyhow!("Cosign is not supported on IBM s390x"))
```
hbrueckner, line 187: minor wording: _"Cosign is not supported on the s390x architecture"_

hbrueckner, line 187: Could you add a comment with referencing to the issue to let the reader understand why and that there is work in progress to support it (see ring discussions).

src/signature/mechanism/cosign/mod.rs
```diff
@@ -356,6 +378,78 @@ mod tests {
...
+    #[cfg(target_arch = "s390x")]
+    #[rstest]
+    #[case(
+        r#"{
+            "type": "sigstoreSigned",
+            "keyPath": "test_data/signature/cosign/cosign2.pub"
+        }"#,
+        "registry.cn-hangzhou.aliyuncs.com/xynnn/cosign:latest",
+        false,
+        "Cosign is not supported on IBM s390x"
```
hbrueckner, line 394: see above.. and there will be few more below.

hbrueckner (changes requested):
Hi @mattarnoatibm ,

many thanks for looking into this. So far it looks nice to me. I Have just a few minor remarks. In addition, I think we also need revert [kata-containers/#5600](https://github.com/kata-containers/kata-containers/pull/5600/commits/b4d3a79b56244388b3805418bb456687d6f47b25) to include the latest `image-rs` version into the agent.

arronwy:
Hi @mattarnoatibm @Xynnn007 , it is possible to build different signature schemes as features and `kata-agent` can use platform specific dependeicies: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#platform-specific-dependencies or enable workaround in `image-rs` to build default features depends on different hardware: https://stackoverflow.com/questions/67373380/can-i-set-cargo-projects-default-features-depending-on-the-platform

src/signature/mechanism/cosign/mod.rs
mattarnoatibm: I added a comment about big-endian architectures being not currently supported by ring, with a link to the ring issue for adding support for big-endian architectures including s390x.

mattarnoatibm: I changed the wording a bit more because following Aaron's feedback cosign is a feature, so you can built image-rs without support for cosign on, for example, the amd64 architecture if you build image-rs with non-default feature flags which exclude the cosign feature.

mattarnoatibm:
Hi @arronwy . Thanks for your feedback.

I've got the cosign signing building conditionally as a feature. On s390x architecture machine the following commands run successfully:

```
cargo build --no-default-features --features overlay_feature,ocicrypt-rs,attestation_agent/default,ocicrypt-rs/default
cargo test --no-default-features --features overlay_feature,ocicrypt-rs,attestation_agent/default,ocicrypt-rs/default
```

On amd64 architecture machine build and test with default features using

```
cargo build
cargo test
```
works.

BbolroC:
lgtm, thanks for the PR. :wink:

src/signature/mechanism/cosign/mod.rs
hbrueckner: Thanks you!

hbrueckner: The new approach looks much cleaner for me! Thanks.

hbrueckner (approved):
@mattarnoatibm 

many thanks for the rework. You can mark my review comments above as resolved.


> I've got the cosign signing building conditionally as a feature. On s390x architecture machine the following commands run successfully:
> 
> ```
> cargo build --no-default-features --features overlay_feature,ocicrypt-rs,attestation_agent/default,ocicrypt-rs/default
> cargo test --no-default-features --features overlay_feature,ocicrypt-rs,attestation_agent/default,ocicrypt-rs/default
> ```
> 
> On amd64 architecture machine build and test with default features using
> 
> ```
> cargo build
> cargo test
> ```

@mattarnoatibm @arronwy Based on the fact that `cargo` cannot disable single features (being discussed at [ Feature Request: Ability to disable single default features #3126 ](https://github.com/rust-lang/cargo/issues/3126)), I wonder whether we should flip and build `amd64` specifically with the `cosign` feature? Wdyt?

However, above question should not hold back PR merging.

/lgtm

Xynnn007:
Hi, thanks for the PR.  For `Cosign` is a separate mod and I prefer to add 
```rust
#[cfg(feature = "cosign")]
```
to all the cosign related imports and sentences, s.t.
* `pub mod cosign;` in `src/signature/mechanism/mod.rs`
* `Cosign(CosignParameters),` in `src/signature/policy/policy_requirement.rs`
* `PolicyReqType::Cosign(inner) => inner.allows_image(image, auth).await,` in `src/signature/policy/policy_requirement.rs`
* `PolicyReqType::Cosign(scheme) => Some(scheme as &dyn SignScheme),` in `src/signature/policy/policy_requirement.rs`

In this way we can avoid unnecessary conditional test code in `cosign/mod.rs` and it seems more clean.

As for the `simple_signing.rs`, I'm going to make a new PR to define the counter part in `image-rs`. Then we can directly avoid add more definitions of SimpleSigning Payload.

Xynnn007:
Hi, please rebase the newest commit and feel free to add
```
#[cfg(feature = "cosign")]
```
to
```
impl From<sigstore_rs::simple_signing::SimpleSigning> for SigPayload {
```

mattarnoatibm:
> Hi, thanks for the PR. For `Cosign` is a separate mod and I prefer to add
> 
> ```rust
> #[cfg(feature = "cosign")]
> ```
> 
> to all the cosign related imports and sentences, s.t.
> 
>     * `pub mod cosign;` in `src/signature/mechanism/mod.rs`
> 
>     * `Cosign(CosignParameters),` in `src/signature/policy/policy_requirement.rs`
> 
>     * `PolicyReqType::Cosign(inner) => inner.allows_image(image, auth).await,` in `src/signature/policy/policy_requirement.rs`
> 
>     * `PolicyReqType::Cosign(scheme) => Some(scheme as &dyn SignScheme),` in `src/signature/policy/policy_requirement.rs`
> 
> 
> In this way we can avoid unnecessary conditional test code in `cosign/mod.rs` and it seems more clean.
> 
> As for the `simple_signing.rs`, I've made a new PR #83 to define the counterpart in `image-rs`. Then we can directly avoid add more definitions of SimpleSigning Payload.

Hi @Xynnn007 , thanks for your feedback.

I've conditionally included all cosign related modules and imports in `mechanism/mod.rs`, `payload/simple_signing.rs` and `policy/policy_requirement.rs`.

arronwy (approved):
Thanks @mattarnoatibm LGTM!

src/signature/policy/policy_requirement.rs
```diff
@@ -40,14 +48,49 @@ pub enum PolicyReqType {
     // Refer to issue: https://github.com/confidential-containers/image-rs/issues/7
 }
 
+/// Copy cosign parameters struct from mechansim/cosign/mod.rs when image-rs isn't
+/// built with the cosign module
+#[cfg(not(feature = "cosign"))]
+#[derive(Deserialize, Debug, Eq, PartialEq, Serialize, Default)]
+pub struct CosignParameters {
```
Xynnn007, line 55: IMOP, if we decided not to support cosign in a feature, we do not need to define extra cosign-related structures to make deserialize successful and tell the user `image-rs was built without support for cosign image signing`, but directly an error that can be raised by `serde` when deserializing the `policy.json`. I'm considering duplicated code her, but still not sure about this. What is your opinion?

src/signature/policy/policy_requirement.rs
```diff
@@ -40,14 +48,49 @@ pub enum PolicyReqType {
...
+        // support compiling on big-endian targets. There is an issue open to add
+        // big-endian support to ring here: https://github.com/briansmith/ring/issues/1555
         match self {
             PolicyReqType::Accept => Ok(()),
             PolicyReqType::Reject => Err(anyhow!(r#"The policy is "reject""#)),
             PolicyReqType::SimpleSigning(inner) => inner.allows_image(image, auth).await,
+            #[cfg(feature = "cosign")]
             PolicyReqType::Cosign(inner) => inner.allows_image(image, auth).await,
+            #[cfg(not(feature = "cosign"))]
+            PolicyReqType::Cosign(inner) => Err(anyhow!(
```
Xynnn007, line 91: Same as mentioned

Xynnn007 (commented):
Thanks a lot for taking the suggestions, only a little question

src/signature/policy/policy_requirement.rs
mattarnoatibm: The error I got in the `signature_verification_one_kbc::case_1` test on a s390x machine when the `CosignParameters` struct is not included in `policy_requirement.rs` and the `Cosign(CosignParameters)` enumeration is conditionally taken out of the `PolicyReqType` enum when building on s390x is 
```
Signing scheme: None, Err(Security validate failed: unknown variant `sigstoreSigned`, expected one of `insecureAcceptAnything`, `reject`, `signedBy`
```

which as an error doesn't indicate that image-rs has been built without support for cosign image signing.

It may be possible to remove the duplicated `CosignParameters` struct in `policy_requirement.rs` if the `CosignParameters` struct can be imported from `cosign/mod.rs` without also including any part of the sigstore_rs create.

mattarnoatibm: Including `cosign::CosignParameters` on s390x from in `policy_requirement.rs` requires removing `#[cfg(feature = "cosign")]` from

```
#[cfg(feature = "cosign")]
pub mod cosign;
```

which requires we can import from the sigstore_rs crate:

```
   Compiling image-rs v0.1.0 (/root/image-rs)
error[E0433]: failed to resolve: use of undeclared crate or module `sigstore_rs`
  --> src/signature/mechanism/cosign/mod.rs:14:5
   |
14 | use sigstore_rs::{
   |     ^^^^^^^^^^^ use of undeclared crate or module `sigstore_rs`
```

Based on that, I think the options are to duplicate `CosignParameters` in `policy_requirement.rs` and have the error that tells the user that image-rs was built without support for cosign, and have the duplicated code, or have the error from failed deserialisation.

I'll double check what happens in the test when the error comes from the failed deserialisation.

mattarnoatibm:
I've currently got a work in progress for making the tests/signature_verification.rs tests work when the `Cosign(CosignParameters)` is removed from the `PolicyReqType` enum in `policy_requirements.rs` on s390x.

mattarnoatibm:
Hi @Xynnn007 ,

I refactored my PR to remove all Cosign structures from the image-rs code if we're building image-rs without the cosign feature.

This did have a side effect that I needed to make some changes in the test module to make the tests in `signature_verification.rs` test do the right thing for the test data in the case where image-rs is built without cosign support.  Because if image-rs is built without cosign a policy with any policies that use `sigstoreSigned` will trigger the parse error, even when the image being pulled is not signed using cosign.

If that change is not good for the testing module could we revert this PR to commit [6758fe3](https://github.com/confidential-containers/image-rs/pull/81/commits/6758fe38c72233aa2a2d5dcbfabbc2c27c31203b) and return the error for image-rs being built without support for cosign?

Xynnn007 (approved):
Hi @mattarnoatibm , this PR LGTM with different policies for enable-cosign and disable-cosign. As for the integration test module, I'd like do some little refactor later. Thanks a lot.

mattarnoatibm:
Thanks @Xynnn007 for all you help and time reviewing this PR.

PR #99, opened 2023-01-04
Prepare to reduce dependency

jiangliu:
The image-rs has too heavy dependency, about 500 crates get compiled. So let's start to reduce dependencies.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,9 +37,21 @@ jobs:
           command: build
           args: --features default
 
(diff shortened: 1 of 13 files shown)


arronwy (approved):
Thanks @jiangliu LGTM!

jodh-intel:
Thanks @jiangliu. Please could you reference atleast one issue in a "Fixes #nnn" comment though in future. That helps explain the changes, helps with future "digital archaeology", and also help to avoid duplicated effort with multiple people working on the same changes (like #103).

PR #104, opened 2023-01-06
Second part to simplify dependency chain

jiangliu:
Second part to simplify dependency chain of image-rs.

### DIFF
--- a/src/secure_channel.rs
+++ b/src/secure_channel.rs
@@ -3,24 +3,22 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-//! Secure-Channel is a module used to connect KBS (Relying Party)
-//! for confidential resources
+//! Fetch confidential resources from KBS (Relying Party).
 
 use std::collections::HashMap;
-use std::result::Result::Ok;
 
-use anyhow::*;
+use anyhow::{anyhow, bail, Result};
+#[cfg(feature = "keywrap-native")]
 use attestation_agent::AttestationAPIs;
-use attestation_agent::AttestationAgent;
-use ocicrypt_rs::config::{OcicryptConfig, OCICRYPT_ENVVARNAME};
 use serde::{Deserialize, Serialize};
 use tokio::fs;
-use tonic::transport::Channel;
 
+#[cfg(feature = "keywrap-grpc")]
 use self::get_resource::{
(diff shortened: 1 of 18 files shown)


src/secure_channel.rs
```diff
@@ -52,45 +50,81 @@ impl ResourceDescription {
...
         &mut self,
-        kbc_name: &str,
-        kbs_uri: &str,
-        resource_description: String,
+        _kbc_name: &str,
+        _kbs_uri: &str,
+        _resource_description: String,
     ) -> Result<Vec<u8>> {
         match self {
+            Self::None => Err(anyhow!("no mechanism to fetch resources")),
```
Xynnn007, line 75: IMO a better way is to raise the error when create the instance of `Client`. We can leave as what this looks now, and I'd refactor the logic in another pr.

src/signature/policy/policy_requirement.rs
```diff
@@ -7,17 +7,14 @@ use anyhow::{anyhow, Result};
...
-    mechanism::{simple::SimpleParameters, SignScheme},
-};
+use crate::signature::image::Image;
+#[cfg(feature = "signature-cosign")]
+use crate::signature::mechanism::cosign::CosignParameters;
+#[cfg(feature = "signature-simple")]
+use crate::signature::mechanism::simple::SimpleParameters;
+use crate::signature::mechanism::SignScheme;
+#[cfg(all(feature = "signature-cosign", feature = "signature-simple"))]
+compile_error!("both signature-cosign and signature-simple are enabled");
```
Xynnn007, line 17: We surely allow both of the features are enabled, although currently some of the scenarios only use one of them.

Xynnn007 (commented):
This pr is very important for deducing the compiling time and binary size. Thanks a lot.

src/signature/policy/policy_requirement.rs
arronwy: BTW, should we make feature `keywrap-native` and `keywrap-grpc` mutually exclusive since it may used by different runtime like `kata-cc` and `enclave-cc`?

arronwy (commented):
Thanks @jiangliu , LGTM!, only minor suggestions.

src/signature/policy/policy_requirement.rs
jiangliu: Fixed

Issue #102, opened 2023-01-05
Swith error handling to thiserror


jodh-intel:
We're using the [`anyhow`](https://github.com/dtolnay/anyhow) crate for error handling but a better approach would be to use [`thiserror`](https://github.com/dtolnay/thiserror) crate as it encourages defining a set of errors, rather than sprinkling the code with lots of custom errors.

I don't have privs to assign this issue to myself but I'm happy to tal if folk agree with this idea?

jodh-intel:
@arronwy, @jiangliu - Since #103 was related to error handling, but may not now be needed (?), I plan to work on this issue. **Please** let me know if you also intend to so we can avoid duplicated effort again.

PR #106, opened 2023-01-10
[RFC/WIP/DNM]: errors: Switch from anyhow to thiserror

jodh-intel:
Convert the top level code to use concrete error type (using the `thiserror` crate) rather than creating errors using the `anyhow` crate ("all errors are strings").

Fixes: #102.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,7 @@ tonic = "0.8"
 strum_macros = "0.24"
 sigstore_rs = { git = "https://github.com/sigstore/sigstore-rs", rev = "6dd3281c25270f0d82550368abfb399ed3da7b41", package = "sigstore",  optional = true}
 url = "2.2.2"
+thiserror = "1.0.38"
 
 [build-dependencies]
 tonic-build = "0.8"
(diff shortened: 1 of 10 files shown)


jodh-intel:
This is a RFC/POC PR only to swithc to `thiserror` errors, which is designed for libraries.

Some limitations and issues:

- I've only converted the "top-level `src/` directory only.
- I haven't converted the tests to use the new error types (so the tests will currently fail).
- Some of the context is not maintained
  For example, in `stream_processing()`, we now return the error, but do not annotate it with the digest type (sha256/sha512). I could add that to all the `StreamError`'s, but that doesn't seem like the best solution. Ideas?

wdyt @arronwy, @jiangliu?

jodh-intel:
btw, who can I speak to about getting perms to add labels in this repo? :smile:

jodh-intel:
:bricks: :bricks: Is there anybody... out there? :bricks: :bricks:

sameo:
> btw, who can I speak to about getting perms to add labels in this repo? 😄

You should have those perms now, @jodh-intel

sameo (commented):
LGTM, thanks @jodh-intel

arronwy:
Thanks @jodh-intel , LGTM, thanks for this enhancement!

PR #113, opened 2023-01-29
feat: add feature to control secure_channel

Xynnn007:
For non-confidential container scenarios, `secure_channel` is not supported. This commit brings a feature `getresource` to control whether to enable `secure_channel`.

Also, add flexibly configurable for paths to support both confidential scenarios and non-confidential scenarios.
Close https://github.com/confidential-containers/image-rs/issues/110

Signed-off-by: Xynnn007 <xynnn@linux.alibaba.com>

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -12,17 +12,15 @@ use std::collections::{BTreeSet, HashMap};
 use std::convert::TryFrom;
 use std::path::Path;
 use std::sync::Arc;
-use tokio::fs;
+
 use tokio::sync::Mutex;
 
-use crate::auth::credential_for_reference;
 use crate::bundle::{create_runtime_config, BUNDLE_ROOTFS};
 use crate::config::ImageConfig;
 use crate::decoder::Compression;
 use crate::meta_store::{MetaStore, METAFILE};
 use crate::pull::PullClient;
 
-use crate::secure_channel::SecureChannel;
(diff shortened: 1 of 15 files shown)


Xynnn007:
The eaakbc test fails because the test container environment is not up-to-date, let me fix this.

Xynnn007:
related pr https://github.com/confidential-containers/image-rs/pull/114

Xynnn007:
cc @jiangliu @arronwy

src/image.rs
```diff
@@ -183,11 +179,12 @@ impl ImageClient {
 
         // If one of self.config.auth and self.config.security_validate is enabled,
         // there will establish a secure channel between image-rs and Attestation-Agent
+        #[cfg(feature = "getresource")]
```
jiangliu, line 182: Seems it may help readability by grouping code between line 182-263 as a helper function?

Xynnn007: As `secure_channel` of line 182 will also be used in line 289, and between them there should be other irrelevant code. If a function, there should be two different functions. One of them return a tuple `(SecureChannel, Auth)`, and the other return `Auth` controlled by feature, which will be more confusing. I also have not thought of a better way to refactor : (

jiangliu (approved):

jialez0 (approved):

PR #117, opened 2023-02-14
feat: use ndyus to speed up the image deployment process in image-rs

taoohong:
Integrate nydus related services into image-rs initially. Only nydus FUSE service is supported right now.

### DIFF
--- a/image-rs/src/config.rs
+++ b/image-rs/src/config.rs
@@ -34,6 +34,9 @@ pub const AUTH_FILE_PATH: &str = "kbs:///default/credential/test";
 /// Default max concurrent download.
 pub const DEFAULT_MAX_CONCURRENT_DOWNLOAD: usize = 3;
 
+/// Path to the configuration file to generate ImageConfiguration
+pub const CONFIGURATION_FILE_PATH: &str = "/var/lib/image-rs/config.json";
+
 /// `image-rs` configuration information.
 #[derive(Clone, Debug, Deserialize)]
 pub struct ImageConfig {
@@ -60,6 +63,10 @@ pub struct ImageConfig {
     ///
     /// This defaults to [`DEFAULT_MAX_CONCURRENT_DOWNLOAD`].
     pub max_concurrent_download: usize,
+
+    /// Nydus services configuration
+    #[serde(rename = "nydus")]
+    pub nydus_config: Option<NydusConfig>,
 }
 
 /// This function used to parse from string. When it is an
@@ -91,6 +98,10 @@ impl Default for ImageConfig {
             auth: false,
             file_paths: Paths::default(),
             max_concurrent_download: DEFAULT_MAX_CONCURRENT_DOWNLOAD,
+            #[cfg(feature = "nydus")]
+            nydus_config: Some(NydusConfig::default()),
+            #[cfg(not(feature = "nydus"))]
+            nydus_config: None,
         }
     }
 }
@@ -106,8 +117,35 @@ impl TryFrom<&Path> for ImageConfig {
         let file = File::open(config_path)
             .map_err(|e| anyhow!("failed to open config file {}", e.to_string()))?;
 
-        serde_json::from_reader::<File, ImageConfig>(file)
-            .map_err(|e| anyhow!("failed to parse config file {}", e.to_string()))
+        match serde_json::from_reader::<File, ImageConfig>(file) {
+            Ok(image_config) => {
+                if image_config.validate() {
+                    Ok(image_config)
+                } else {
+                    Err(anyhow!("invalid configuration"))
+                }
+            }
+            Err(e) => Err(anyhow!("failed to parse config file {}", e.to_string())),
+        }
+    }
+}
+
+impl ImageConfig {
+    /// Validate the configuration object.
+    pub fn validate(&self) -> bool {
+        if let Some(nydus_cfg) = self.nydus_config.as_ref() {
+            if !nydus_cfg.validate() {
+                return false;
+            }
+        }
+
+        true
+    }
+
+    pub fn get_nydus_config(&self) -> Result<&NydusConfig> {
+        self.nydus_config
+            .as_ref()
+            .ok_or_else(|| anyhow!("no configuration information for nydus"))
     }
 }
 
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -45,6 +45,12 @@ url = { version = "2.2.2", optional = true }
 ttrpc = { version = "0.7.1", features = [ "async" ], optional = true }
 walkdir = "2"
 zstd = "0.11"
+fuse-backend-rs = "0.10.1"
+lazy_static = "1"
+
+nydus-api = "0.2.1"
+nydus-app = "0.3.2"
```
jiangliu, line 52: It would be better to avoid dependency on `nydus-app`, we may provide a fake BuildTimeInfo for image-rs.

taoohong: Got it.

jepio:
Is it possible to add a `nydus` feature so that this support is optional?

taoohong:
> Is it possible to add a `nydus` feature so that this support is optional?

Yes, that sounds good. I am currently completing the support of the nydus erofs driver and doing some performance tests, 
the `nydus feature` will also be launched when I finish these.

jiangliu:
The PR shows development history, could you please help to fold the commit history?

taoohong:
> The PR shows development history, could you please help to fold the commit history?

done

image-rs/src/config.rs
```diff
@@ -133,6 +171,141 @@ impl Default for Paths {
...
+    }
+
+    /// Validate the configuration object.
+    pub fn validate(&self) -> bool {
+        if self.driver_type != "fuse" && self.driver_type != "fscache" {
+            return false;
+        }
+
+        if !self.is_fuse() && !self.is_fscache() {
+            return false;
```
jiangliu, line 240: Should we check that both fuse and fscache are enabled?

image-rs/src/config.rs
```diff
@@ -133,6 +171,141 @@ impl Default for Paths {
...
+            fuse_config: Some(FuseConfig::default()),
+            fscache_config: None,
+        }
+    }
+}
+
+impl NydusConfig {
+    /// Check whether the service type is `fuse`
+    pub fn is_fuse(&self) -> bool {
+        self.fuse_config.is_some()
```
jiangliu, line 213: Check `driver_type == `fuse`" instead?

image-rs/src/config.rs
jiangliu: Or
```
if self.is_fuse() && self.fuse_config.is_none() {
    return false;
}
```

image-rs/src/config.rs
```diff
@@ -133,6 +171,141 @@ impl Default for Paths {
...
+#[derive(Clone, Debug, Deserialize)]
+pub struct FuseConfig {
+    /// FUSE server failover policy
+    pub fail_over_policy: String,
+
+    /// Number of worker threads to serve FUSE I/O requests
+    pub fuse_threads: u32,
+
+    /// Mounts FUSE filesystem in rw mode
+    pub writable: bool,
```
jiangliu, line 257: Seems we should only support `ro` mode so this flag is redundant?

image-rs/src/config.rs
```diff
@@ -133,6 +171,141 @@ impl Default for Paths {
...
+
+    /// Path to the `localfs` working directory, which also enables the `localfs` storage backend
+    pub localfs_dir: Option<PathBuf>,
+
+    /// Path to the directory to be shared via the `passthroughfs` FUSE driver
+    pub shared_dir: Option<PathBuf>,
```
jiangliu, line 263: Seems no need to support `passthroughfs` for `image-rs`.

image-rs/src/nydus/utils.rs
```diff
@@ -0,0 +1,55 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+
+use nydus_api::BuildTimeInfo;
+use oci_distribution::manifest;
+
+pub fn is_nydus_data_layer(desc: &manifest::OciDescriptor) -> bool {
+    desc.annotations
+        .as_ref()
+        .unwrap()
```
jiangliu, line 11: It would be better to avoid `unwrap()` when possible:)
```
desc.annotations
        .as_ref()
        .map(|v| v.contains_key(super::NYDUS_META_LAYER))
        .unwrap_or_default()
```

image-rs/src/image.rs
```diff
@@ -336,10 +353,119 @@ impl ImageClient {
...
+        }
+
+        let reference = Reference::try_from(image_data.reference.clone())?;
+        let nydus_config = self
+            .config
+            .get_nydus_config()
+            .expect("Nydus configuration not found");
+        let work_dir = self.config.work_dir.clone();
+
+        let image_id = service::start_nydus_service(
```
arronwy, line 405: Do we need start nydus service for each `pull_image` request? or create one nydus service daemon for each `ImageClient` instance?

image-rs/src/nydus/utils.rs
```diff
@@ -0,0 +1,55 @@
...
+    get_nydus_bootstrap_desc(image_manifest).is_some()
+}
+
+pub fn get_nydus_bootstrap_desc(
+    image_manifest: &manifest::OciImageManifest,
+) -> Option<manifest::OciDescriptor> {
+    let layers = &image_manifest.layers;
+    if !layers.is_empty() {
+        let desc = &layers[layers.len() - 1];
+        if (desc.media_type == manifest::IMAGE_LAYER_GZIP_MEDIA_TYPE
```
jiangliu, line 32: Should we also supported uncompressed metablob in addition to compressed one?

image-rs/src/nydus/service.rs
```diff
@@ -0,0 +1,377 @@
...
+        DAEMON_CONTROLLER.set_singleton_mode(false);
+        DAEMON_CONTROLLER.shutdown();
+    });
+    let image_id = create_nydus_bundle(image_data, bundle_dir)?;
+
+    Ok(image_id)
+}
+
+#[allow(clippy::too_many_arguments)]
+pub fn precess_fuse_daemon(
```
arronwy, line 131: typo `precsee` -> `process`

image-rs/src/nydus/utils.rs
```diff
@@ -0,0 +1,55 @@
...
+            || desc.media_type == manifest::IMAGE_DOCKER_LAYER_GZIP_MEDIA_TYPE)
+            && desc.annotations.is_some()
+            && desc.annotations.as_ref().unwrap()[super::NYDUS_META_LAYER] == "true"
```
jiangliu, line 35: use `is_nydus_meta_layer()` instead?

image-rs/src/image.rs
taoohong: One nydus daemon for one image reference. If we want to use image-rs to download a new image, then we need to reconfigure the nydus daemon and start a new one. If we download the same image, there is no need to start a new daemon. One nydus daemon for all images is unsupported.

jiangliu (approved):

arronwy (approved):
Thanks @taoohong LGTM!

PR #119, opened 2023-02-26
Bring KBS Resource URI into Image-rs

Xynnn007:
This PR mainly aims to bring KBS Resource URI scheme into image-rs. Concrete contents of this PR includes
- Let the following files be configurable by a uri in `ImageClient.config.file_paths` before doing pulling. Related codes were refactored
    - `sigstore_config`:  Signature storage configuration file for simple signing, which describes where an image's signature is stored
    - `policy_path`: image security policy JSON file, s.t. `policy.json`
    - `auth_file`: credential file to access a private image registry, s.t. `auth.json`
- Refactored and extended `secure_channel` module into `resource` module which gives an public api `get_resource(uri)`. Here uri supports two schemes for now and can be extended appropriately
    - `kbs://...`: retrieve the resource from KBS.
    - `file://...`: read the resource from local file system. By default if no scheme is given, try to read using `file://`
- Changed the trait for signature scheme. Deleted the method `resource_manifest`. For now, when a resource is needed, function `crate::resource::get_resource(uri)` can be called to get lazily. The code logic was simplified.
- Some promotes for simple signing scheme. Refactoring some logic.
- Fixed related unit tests to fit in with KBS URI scheme
- Fixed integration tests to fit in with KBS URI scheme. Introduced a new test image for image decryption `docker.io/xynnn007/busybox:encrypted-uri-key` which is made using [cc-kbc's key provider](https://github.com/confidential-containers/attestation-agent/tree/main/sample_keyprovider/src/enc_mods/cc_kbc_enc) as it shares the same format of AnnotationPacket and KBS URI.

> *warning* please do not merge this pr before https://github.com/confidential-containers/attestation-agent/pull/129 is approved and merged. Because this pr includes personal repo of AA for integration test.

Close https://github.com/confidential-containers/image-rs/issues/116


Xynnn007:
enclave-cc test error is due to different attestation-agent version in ocicrypt-rs and image-rs. When https://github.com/confidential-containers/attestation-agent/pull/129 is merged, AA's rev should be updated in ocicrypt-rs and then in this PR.

Xynnn007:
Hi @stevenhorsman @arronwy 

This PR introduces three more parameters in `ImageClient` to specify the uri of `auth` file, `image security policy` file and `sigstore configuration for simplesigning` (if simple signing will be used)

I think these parameters should be input from `kata-agent` as who is the caller of image-rs.

Still, some questions comes. For test we can fix these fields like `kbs://example.com/...` for now.

Potentially, we need to think about how we provide these uris to `kata-agent` as they should be specified by the tenant. A naive thinking is still from commandline, which require us to update kata-agent's related code and AS can verify whether the `policy.json`'s uri is as expected

jialez0 (approved):
@Xynnn007 Thanks a lot for this important work! Although we use the method of overwriting the resource URI to use `aa_kbc_params` correctly, this is a suitable method at present.

src/signature/mod.rs
```diff
@@ -8,124 +8,41 @@ pub mod mechanism;
...
+) -> Result<()> {
+    use crate::{resource, signature::image::Image};
+
+    let reference = oci_distribution::Reference::try_from(image_reference)?;
+    let mut image = Image::default_with_reference(reference);
+    image.set_manifest_digest(image_digest)?;
+
+    // Read the set of signature schemes that need to be verified
+    // of the image from the policy configuration.
+    let policy_json_string = resource::get_resource(&file_paths.policy_path).await?;
```
jialez0, line 35: **BUG**: Here `&file_paths.policy_path` is `/run/image-security/security_policy.json`, but `get_resource` parameter needs KBS resource URI like `kbs://...`, right?

src/auth/mod.rs
```diff
@@ -26,53 +26,15 @@ pub struct DockerAuthConfig {
...
 ) -> Result<RegistryAuth> {
-    // if Policy config file does not exist, get if from KBS.
-    if !Path::new(auth_file_path).exists() {
-        secure_channel
-            .lock()
-            .await
-            .get_resource(RESOURCE_DESCRIPTION, HashMap::new(), auth_file_path)
-            .await?;
-    }
+    let auth = crate::resource::get_resource(auth_file_path).await?;
```
jialez0, line 35: **BUG**: Now `auth_file_path` is `/run/image-security/auth.json`, not a KBS resource URI.

src/config.rs
```diff
@@ -108,33 +105,21 @@ impl TryFrom<&Path> for ImageConfig {
 
 #[derive(Clone, Debug, Deserialize)]
 pub struct Paths {
+    /// sigstore config file for simple signing
+    pub sigstore_config: String,
+
     /// Path to `Policy.json`
     pub policy_path: String,
```
jialez0, line 112: Here, `policy_path` should be `policy_uri`, right? So it should be fixed as `kbs://example.org/default/image_policy/latest` or some else.

So as `auth_file`.

arronwy:
> Hi @stevenhorsman @arronwy
> 
> This PR introduces three more parameters in `ImageClient` to specify the uri of `auth` file, `image security policy` file and `sigstore configuration for simplesigning` (if simple signing is to be used)
> 
> I think these parameters should be input from `kata-agent` as who is the caller of image-rs.
> 
> Still, a questions comes. For test we can fix these fields like `kbs://example.com/...` for now.
> 
> Potentially, we need to think about how we provide these uris to `kata-agent` as they should be specified if default value for test is not used. A naive thinking is still from commandline, which requires us to update kata-agent's related code and AS can verify whether the `policy.json`'s uri is as expected

Thanks @Xynnn007 , to support the flexiblity for different senarios  and easy to use for end user, we can expose these interfaces with default value which can backword compatibale previous release, and let the end user to config them if needed.

arronwy (approved):

src/signature/mod.rs
Xynnn007: Yes, thanks! fixed!

src/auth/mod.rs
Xynnn007: Thanks for pointing out. Fixed

src/config.rs
Xynnn007: Yes. it is set to `"kbs:///default/security-policy/test"`, same as defined in AA

Xynnn007:
@arronwy  Yes, now the three parameters are set to
- `kbs:///default/credential/test`
- `kbs:///default/sigstore-config/test`
- `kbs:///default/security-policy/test`

These are defined in attestation-agent https://github.com/confidential-containers/attestation-agent/blob/main/src/kbc_modules/mod.rs#L152

PR #120, opened 2023-02-27
fix: delete broken layer when duplication detected

Xynnn007:
When a layer is duplicated but not recorded in layer_db, it should be treated as "broken" and be deleted. Then the newly pulled layer data can be written, instead of raising an error and exit.

cc @arronwy


arronwy (approved):
Thanks @Xynnn007 LGTM!

Xynnn007:
Do we need to wait after the code freeze phase?

PR #122, opened 2023-03-06
fix: use ttrpc instead of grpc as default protocol

Xynnn007:
Close https://github.com/confidential-containers/image-rs/issues/121


arronwy (approved):
Thanks @Xynnn007 LGTM!

Xynnn007:
Hi @arronwy As the new definition of default feature will influence the CI behavior, I've updated ci.yaml. Please take another look to check whether it is ok

PR #126, opened 2023-03-20
cargo: Update oci-distribution to support manifest lists

arronwy:
Latest oci-distribution have fixed faile to pull `ubuntu` image from docker.io: https://github.com/krustlet/oci-distribution/pull/67

Restrict sigstore-rs to use the same `oci-distribuiton` with `image-rs` to fix the build issue.


Xynnn007 (approved):
LGTM, thanks! much tidier

PR #140, opened 2023-04-14
deps: Bump ocicrypt-rs to v0.5.1

fidencio:
I've missed one dependency bump while releasing ocicrypt-rs v0.5.0, which lead to a new v0.5.1 release, which we should use here.


Xynnn007 (approved):
lgtm thanks

PR #144, opened 2023-05-09
Cargo.toml: add occlum cckbc feature

### LINKED ISSUE #143: Get rid of `jwe` for ocicrypt-rs
Now we only use `keyprovider` for image decryption. `jwe` is not used. We should disable `jwe` in kata-cc and enclave-cc's features, for `jwe` is based on `josekit`, and `josekit` is based on `openssl`.

Xynnn007:
This new feature is used inside encalve-cc to integrate cc-kbc and generic-kbs

add "ocicrypt-rs/block-cipher" for feature `encryption`

delete "keywrap-jwe" for ocicrypt-rs

Close #143

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,23 +53,23 @@ jobs:
         run: |
           sudo -E PATH=$PATH -s cargo test --all --features default
 
-      - name: Run cargo test - kata-cc (rust-tls version) with keywrap-grpc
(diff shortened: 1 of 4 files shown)


arronwy (approved):
Thanks @Xynnn007 , LGTM!
BTW, do we need test case for `enclave-cc-cckbc-*` ?

Xynnn007:
@arronwy It will be great to have one but hard to implement in image-rs. We will have an test in enclave-cc to cover the whole process. We can left this in future if needed

Issue #145, opened 2023-05-10
Got IncompleteBody error, when pull a big image which is 1.93G.


BaoshunFang:
My container image is 1.93 Gbytes, when using kata-agent to pull image, it always failed with below error:

May 10 07:52:03 k8s-master containerd[1153]: time="2023-05-10T07:52:03.418712368Z" level=info msg="PullImage \"docker.io/test/test:latest\""
May 10 07:52:03 k8s-master containerd[1153]: time="2023-05-10T07:52:03.418798917Z" level=info msg="TaskManager get ImageService succeed." id=45399c080ecae7bf8800aae9509e960061d0bf3bdfece9b57d7daa95b86f4f4c
May 10 07:54:38 k8s-master containerd[1153]: time="2023-05-10T07:54:38.754185654Z" level=error msg="agent pull image err. rpc error: code = Internal desc = failed to handle layer: hasher sha256 : channel: read failed Custom { kind: Other, error: reqwest::Error { kind: Body, source: hyper::Error(Body, Custom { kind: UnexpectedEof, error: IncompleteBody }) } }" name=containerd-shim-v2 pid=375818 sandbox=45399c080ecae7bf8800aae9509e960061d0bf3bdfece9b57d7daa95b86f4f4c source=virtcontainers subsystem=kata_agent
May 10 07:54:38 k8s-master containerd[1153]: time="2023-05-10T07:54:38.754316393Z" level=error msg="kata runtime PullImage err. rpc error: code = Internal desc = failed to handle layer: hasher sha256 : channel: read failed Custom { kind: Other, error: reqwest::Error { kind: Body, source: hyper::Error(Body, Custom { kind: UnexpectedEof, error: IncompleteBody }) } }" name=containerd-shim-v2 pid=375818 sandbox=45399c080ecae7bf8800aae9509e960061d0bf3bdfece9b57d7daa95b86f4f4c source=containerd-kata-shim-v2
May 10 07:54:38 k8s-master containerd[1153]: time="2023-05-10T07:54:38.754529642Z" level=error msg="PullImage \"docker.io/test/test:latest\" failed" error="rpc error: code = Internal desc = failed to handle layer: hasher sha256 : channel: read failed Custom { kind: Other, error: reqwest::Error { kind: Body, source: hyper::Error(Body, Custom { kind: UnexpectedEof, error: IncompleteBody }) } }"

fitzthum:
[...]

PR #146, opened 2023-05-16
pull: Add max concurrent download limit support

arronwy:
Improve pull performance for image with multiple layers and small CPU num without needless schedule overhead.

Fixes: #145


mythi:
> this looks to be problematic for enclave-cc since `std::thread::available_parallelism()` gives what's available on host and that triggers the similar resource issues as with #145 originally

So we would need to be able to set upper limit and this is equivalent to containerd's CRI config:

```
  [plugins."io.containerd.grpc.v1.cri"]
    ...
    max_concurrent_downloads = 3
```

arronwy:
> 

Hi @mythi , could you help review this PR: https://github.com/confidential-containers/image-rs/pull/153

PR #147, opened 2023-05-22
gh: add dependency bot

Xynnn007:


.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: cargo
+    directory: "/"
+    schedule:
+      interval: daily
```
arronwy, line 6: Do we need daily? or is weekly will be ok?

Xynnn007: I'm not sure. Now PRs in KBS & AS are daily too, so I choose daily here

https://github.com/confidential-containers/kbs/pull/89/files#diff-dd4fbda47e51f1e35defb9275a9cd9c212ecde0b870cba89ddaaae65c5f3cd28R6

PR #153, opened 2023-05-25
config: Add max_concurrent_download for pull operation

arronwy:
`std::thread::available_parallelism()` will get all available cpu num in host side for enclave-cc, export the config for enclave/kata agent to decide.

Thanks @mythi for point this out.

DEFAULT_MAX_CONCURRENT_DOWNLOAD is set to 3 which is the same default config for dockerd.


mythi (approved):
Looks good to me and tested with:
```
-image-rs = { git = "https://github.com/confidential-containers/image-rs.git", default-features = false, rev = "44f27c2" }
+image-rs = { git = "https://github.com/arronwy/image-rs.git", default-features = false, branch = "config" }
```

in enclave-cc and it helps to fix an image pull that was failing earlier.

Xynnn007 (approved):

PR #158, opened 2023-06-05
3 Repo merge

dcmiddle:
Experiment merging attestation-agent, image-rs, and ocicrypt-rs

In each repo I moved each project into a subdirectory of its own name and then merged them into the image-rs repo.
This preserved the commit history from each repo.
It also makes this PR look like it has 346 commits. In actuality only the most recent ~4 commits are changes.

I created a workspace at the top level of the repo as suggested in https://github.com/confidential-containers/community/issues/89
Attestation-agent has its own workspace. This creates a little issue since cargo does not support nested workspaces. 
I pulled each of the attestation-agent workspace members into the top level. I'm not very confident that was the right resolution.

image-rs also had a workspace but the members seemed unused so I commented it out.  

I'm looking for feedback in particular on this new workspace layout. 
A different approach would be not to have a top level Cargo file and let each sub-project have one if it needed.
The downside of that is that you no longer have common dependency version enforcement.

todo:
- [x] Reconcile the dependency sections of each package with the top level workspace. 
- [x] Add a README
- [x] Update all the CI workflows


ocicrypt-rs/Cargo.toml
```diff
@@ -12,14 +12,14 @@ edition = "2018"
...
-attestation_agent = { git = "https://github.com/confidential-containers/attestation-agent.git", rev = "aa1d3c5", optional = true  }
+attestation_agent = { path = "../attestation-agent/lib", optional = true  }
 base64 = "0.13"
 base64-serde = { version = "0.6", optional = true }
 cfg-if = "1.0.0"
 ctr = { version = ">=0.9", optional = true }
 hmac = { version = ">=0.12", optional = true }
 josekit = { version = ">=0.7", optional = true }
-kbc = { git = "https://github.com/confidential-containers/attestation-agent.git", rev = "aa1d3c5", optional = true  }
+########## kbc = { path = "../attestation-agent/lib", optional = true  }
```
Xynnn007, line 22: This should be `../attestation-agent/kbc`

Xynnn007 (commented):
Thanks for bringing all these up! The code structure looks great to me, while it should also be cared that the files of `.github/workflows` cover different parts inside the workspace

attestation-agent/MAINTAINERS
fitzthum, line 1: Is github smart enough to understand nested maintainers files? We are probably going to have to have the same maintainers across the whole repo.

.github/workflows/basic.yml
fitzthum, line 1: As @Xynnn007 mentioned I think we'll need to rework these tests a bit

.github/workflows/basic.yml
```diff
@@ -0,0 +1,73 @@
...
+          sudo apt-get install -y libtdx-attest-dev
+
+      - name: Install TPM dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libtss2-dev
+
+      - name: Build and install with default features
+        run: |
+          make && make install
```
fitzthum, line 46: Gotta run these make commands in the AA dir, for instance.

.github/workflows/cc_kbc.yml
```diff
@@ -0,0 +1,69 @@
+name: CC kbc build CI
+on:
+  push:
+    paths:
+      - 'kbc/cc_kbc/**'
```
fitzthum, line 5: And these paths will need to be updated

attestation-agent/README.md
```diff
@@ -0,0 +1,140 @@
+# Attestation Agent
+[![FOSSA Status](https://app.fossa.com/api/projects/git%2Bgithub.com%2Fconfidential-containers%2Fattestation-agent.svg?type=shield)](https://app.fossa.com/projects/git%2Bgithub.com%2Fconfidential-containers%2Fattestation-agent?ref=badge_shield)
```
fitzthum, line 2: We may need to consolidate to one FOSSA thing as well? I think we should probably have a top level README and maybe that could include the scanner (with the link to the image-rs one).

fitzthum (commented):
Pulling all the code together looks good, but I have a few notes on various dependencies.

dcmiddle:
@fitzthum @Xynnn007 I have *changed* the CI scripts (it feels too bold to claim that I *fixed* them :-D )
https://github.com/confidential-containers/image-rs/pull/158/commits/da4914024d80bcf9c495c58d947b2248f4320c2f

I also tried to make the naming more clear.
https://github.com/confidential-containers/image-rs/pull/158/commits/01fca88e455cbe164f6c17cff36495feaf962bc9

.github/workflows/release.yml
```diff
@@ -23,7 +23,7 @@ jobs:
         uses: docker/build-push-action@v4
         with:
           context: .
-          file: ./docker/Dockerfile.keyprovider
+          file: ./attestation-agent/docker/Dockerfile.keyprovider
```
dubek: I think this Dockerfile.keyprovider should be fixed; currently it has this:

```
COPY . .
RUN cd coco_keyprovider && cargo install --path .
```
but this probably won't work if we copy _only_ the `attestation-agent` dir, because it wouldn't be able to find `../image-rs` etc.

attestation-agent/MAINTAINERS
dubek: I guess we should replace MAINTAINERS with the GitHub-supported CODEOWNERS.

According to [github docs](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners), CODEOWNERS must be located at:

> To use a CODEOWNERS file, create a new file called `CODEOWNERS` in the root, `docs/`, or `.github/` directory of the repository, in the branch where you'd like to add the code owners.

So we should merge all the files and adapt the paths in them, something like:

```
# Global owner for changes not matched by more specific rules
* @sameo

/attestation-agent/ @jialez0
/image-rs/ @sameo @jiangliu @arronwy @lumjjb @jialez0 @Xynnn007
/ocicrypt-rs/ @sameo @jiangliu @arronwy @lumjjb
```

(I guess we can add others to `/attestation-agent/`, and/or refer to github teams names.)

----

Also, maybe we can remove `ocicrypt-rs/CODE-OF-CONDUCT.md` because we already have one in the community repo.

.github/workflows/release.yml
dcmiddle: @dubek thanks for the catch. Pushed a fix.

attestation-agent/MAINTAINERS
dcmiddle: thanks for writing that for me @dubek! 
@fitzthum this also resolves the fossa cruft and other meta files. https://github.com/confidential-containers/image-rs/pull/158/commits/b856cf5b190b31fbb8c2847acd99d90204b6d49f

.github/workflows/ocicrypt_rs_build.yml
```diff
@@ -54,112 +54,113 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: build
-          args: --all-features
+          args: -p ocicrypt-rs --all-features
```
Xynnn007, line 57: We might also need to add `paths` to this yaml file.

Xynnn007 (approved):

fitzthum (approved):
I'm sure there are still some issues here, but I think we're probably in a place where we can merge and then see what breaks.

arronwy (approved):
Thanks a lot @dcmiddle ， LGTM. one thing missing is `ci_eaa_kbc.yml` is removed for `image-rs`

jiangliu (approved):

Xynnn007:
Hi Dan, could you please rebase this PR and then let's merge?

mythi:
Perhaps squash the Cargo workspace fixup commits in one?

jialez0 (approved):

sameo (approved):
Thanks a lot for that PR @dcmiddle !

PR #159, opened 2023-06-13
Add some robustness fixes to integration tests

mkulke:
- If the AA process for integration tests fails to come up properly the tests do not abort.
- On failed tests the AA process is not killed, because the teardown fn is not called.
- Garbage collection of tempfile::tempdir doesn't work for overlayfs mounted bundles

The changes probe for an unexpected exit, instruct tokio to kill the process on `drop()` of the `Command`, and umount the rootfs on a successful pull

```bash
running 1 test
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: transport error

Caused by:
    0: error creating server listener: Address already in use (os error 98)
    1: Address already in use (os error 98)', app/src/main.rs:31:37
test test_decrypt_layers::case_1 ... FAILED

failures:

---- test_decrypt_layers::case_1 stdout ----
thread 'test_decrypt_layers::case_1' panicked at 'Attestation Agent failed to start', tests/common/mod.rs:108:9


failures:
    test_decrypt_layers::case_1

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.05s
```

### DIFF
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -111,6 +111,11 @@ pub async fn start_attestation_agent() -> Result<Child> {
     Ok(aa)
 }
 
+pub fn umount_bundle(bundle_dir: &tempfile::TempDir) {
+    let rootfs_path = bundle_dir.path().join("rootfs");
+    nix::mount::umount(&rootfs_path).expect("failed to umount rootfs");
+}
+
 pub async fn clean_configs() -> Result<()> {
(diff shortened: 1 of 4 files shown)


Xynnn007 (approved):
Thanks for fixing this. LGTM!

jiangliu (approved):

PR #160, opened 2023-06-14
Add cc_kbc e2e tests

mkulke:
## Rationale

This code wants to cover an e2e scenario for image decryption and repository cred retrieval in a cc_kbc + kbs setup. It could detect unintended breakage in those components, but maybe also provide spec + documentation.

It's two tests that are ignored by default, because they are rather costly to run (> 20m without cache, ~4m w/ populated cache) and it requires a bit of setup ceremony. 

## Implementation

The decryption test encrypts an image with a random key using aa's `coco_keyprovider` and skopeo. The encrypted image is stored in a local registry (to keep the test self-contained).

Attestation Agent is started using the sample attester (could use a matrix to also run it on real TEEs). A kbs instance (bundled w/ Attestion Service) is also started locally. The test then retrieves the encrypted image from the local registry, triggering a remote attestation procedure via attestation-agent. After a successful attestation ocicrypt-rs is decrypting the layers via the KBS-released key.

The credentials test creates a local registry with auth and stores an image in it. A dockerconfig auth is put into the kbs repo. On image pull the attestation-agent then requests the credentials from kbs which is then used to access the local registry.

I'd suggest a good schedule for the trigger would be merges to main.

## Example

[example run](https://github.com/mkulke/image-rs/actions/runs/5390050114)

### DIFF
--- a/.github/workflows/cc_kbc-e2e.yaml
+++ b/.github/workflows/cc_kbc-e2e.yaml
@@ -1,0 +1,89 @@
+name: cc_kbc-e2e
+
+on:
+  push:
+    branches: [ "main" ]
+
+# Self-hosted runners do not set -o pipefail otherwise
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  e2e-test:
+    runs-on: 'ubuntu-22.04'
+    env:
+      SKOPEO_VERSION: "1.12.0"
+    strategy:
+      matrix:
+        test: [credentials-e2e.sh, decryption-e2e.sh]
+
+    steps:
+    - uses: actions-rs/toolchain@v1
+      with:
+        profile: minimal
+        toolchain: stable
+
+    - uses: actions/setup-go@v4
+      with:
+        go-version: stable
+
+    - name: Install dependencies
+      env:
+        SGX_URL: https://download.01.org/intel-sgx/sgx_repo/ubuntu
+      run: |
+        curl -L "${SGX_URL}/intel-sgx-deb.key" | sudo apt-key add -
+        echo "deb [arch=amd64] ${SGX_URL} jammy main" \
+          | sudo tee /etc/apt/sources.list.d/intel-sgx.list
+        sudo apt-get update
+        sudo apt-get install -y \
+          build-essential \
+          clang \
+          libassuan-dev \
+          libbtrfs-dev \
+          libdevmapper-dev \
+          libgpgme-dev \
+          libsgx-dcap-quote-verify-dev \
+          libtdx-attest-dev \
+          libtss2-dev \
+          openssl \
+          pkg-config \
+          protobuf-compiler \
+          wait-for-it \
+          wget
+
+    - uses: actions/checkout@v3
+
+    - uses: actions/checkout@v3
+      with:
+        repository: confidential-containers/kbs
+        path: kbs
+
+    - name: Set up rust build cache
+      uses: actions/cache@v3
+      continue-on-error: false
+      with:
+        path: |
+          ~/.cargo/registry/index/
+          ~/.cargo/registry/cache/
+          target
+          kbs/target
+        key: rust-${{ hashFiles('**/Cargo.toml') }}
+
+    - name: Set up skopeo build cache
+      uses: actions/cache@v3
+      continue-on-error: false
+      with:
+        path: /usr/local/bin/skopeo
+        key: skopeo-${{ env.SKOPEO_VERSION }}
+
+    - name: Build skopeo
+      run: |
+        if [ -f /usr/local/bin/skopeo ]; then exit 0; fi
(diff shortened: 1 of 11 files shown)


mkulke:
@Xynnn007: was this PR closed intentionally? I suppose this was a typo, since [this commit](https://github.com/confidential-containers/image-rs/commit/426b464586012321f94b3fea640e98554ad8343b) seems unrelated

mkulke:
> > @Xynnn007: was this PR closed intentionally? I suppose this was a typo, since [this commit](https://github.com/confidential-containers/image-rs/commit/426b464586012321f94b3fea640e98554ad8343b) seems unrelated
> 
> Oh I'm sorry, I think I closed this one by mistake. Let me reopen. ( I do not remember I have clicked the Close button, weird

no worries, I think it was auto-closed by the commit message in the commit I linked

Xynnn007:
> > > @Xynnn007: was this PR closed intentionally? I suppose this was a typo, since [this commit](https://github.com/confidential-containers/image-rs/commit/426b464586012321f94b3fea640e98554ad8343b) seems unrelated
> > 
> > 
> > Oh I'm sorry, I think I closed this one by mistake. Let me reopen. ( I do not remember I have clicked the Close button, weird
> 
> no worries, I think it was auto-closed by the commit message in the commit I linked

Oh yes. It's a commit from Attestation Agent. Similiar errors would happpen to other sub crates.

.github/workflows/cc_kbc-e2e.yaml
```diff
@@ -0,0 +1,89 @@
...
+          libsgx-dcap-quote-verify-dev \
+          libtdx-attest-dev \
+          libtss2-dev \
+          openssl \
+          pkg-config \
+          protobuf-compiler \
+          wait-for-it \
+          wget
+
+    - uses: actions/checkout@v3
```
fidencio, line 55: Please, add a comment that this is related to the guest-components checkout.

.github/workflows/cc_kbc-e2e.yaml
```diff
@@ -0,0 +1,89 @@
...
+
+    - uses: actions/checkout@v3
+      with:
+        repository: confidential-containers/kbs
+        path: kbs
```
fidencio, line 60: Are we fine testing against `main` kbs?
I am mostly wondering here whether we should pin a version here and make sure that we only updated when needed.

.github/workflows/cc_kbc-e2e.yaml
```diff
@@ -0,0 +1,89 @@
...
+    - name: Set up rust build cache
+      uses: actions/cache@v3
+      continue-on-error: false
+      with:
+        path: |
+          ~/.cargo/registry/index/
+          ~/.cargo/registry/cache/
+          target
+          kbs/target
+        key: rust-${{ hashFiles('**/Cargo.toml') }}
```
fidencio, line 71: Hmm.  I think Kata Containers builds would benefit from this, a lot.

.github/workflows/cc_kbc-e2e.yaml
```diff
@@ -0,0 +1,89 @@
...
+        path: /usr/local/bin/skopeo
+        key: skopeo-${{ env.SKOPEO_VERSION }}
+
+    - name: Build skopeo
+      run: |
+        if [ -f /usr/local/bin/skopeo ]; then exit 0; fi
+        wget -qO- "https://github.com/containers/skopeo/archive/refs/tags/v${SKOPEO_VERSION}.tar.gz" | tar xz
+        cd "skopeo-${SKOPEO_VERSION}"
+        DISABLE_DOCS=1 make bin/skopeo
+        sudo cp bin/skopeo /usr/local/bin
```
fidencio, line 86: I'd prefer having this as part of its own script.  It'd make things simpler when changing either the action itself or the content here.

.github/workflows/cc_kbc-e2e.yaml
```diff
@@ -0,0 +1,89 @@
...
+          libdevmapper-dev \
+          libgpgme-dev \
+          libsgx-dcap-quote-verify-dev \
+          libtdx-attest-dev \
+          libtss2-dev \
+          openssl \
+          pkg-config \
+          protobuf-compiler \
+          wait-for-it \
+          wget
```
fidencio, line 53: I'd like to have this as part of a script, as it'll make things simpler for us, in the future, in case we need to change the content installed.

mkulke:
returning this to draft to rework a bit. I don't think we need to involve kbs/cc_kbc for the image/layer tests here, since we have an e2e test in kbs now. it will probably be fine to use offlline_fs kbc here.

mkulke:
closing this one, the scenario should be covered by higher level e2e tests in kata or CAA.

PR #167, opened 2023-06-22
aa: Rename Occlum attester to SGX attester and add Gramine support to it

mythi:


attestation-agent/attester/src/sgx_dcap/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+                match handler.generate_quote(
+                    occlum_quote.as_mut_ptr(),
+                    report_data_bin.as_ptr() as *const sgx_report_data_t,
+                ) {
+                    Ok(_) => occlum_quote,
+                    Err(e) => bail!("generate quote: {e}"),
+                }
+            }
+            SgxLibOsType::Gramine => {
+                std::fs::write("/dev/attestation/user_report_data", report_data_bin)?;
```
Xynnn007, line 74: Do we also need `occlum_dcap` dependency if gramine is the libos? I mean if not, we can avoid  bringing `occlum_dcap` to decrease the footprint.

mythi: we don't need it but I did not want to introduce new features but keep only one "SGX DCAP" feature.

Xynnn007: Ok. Let's handle this in future when footprint is actually an obstacle for enclave-cc.

Xynnn007:
Something went wrong here https://github.com/confidential-containers/guest-components/actions/runs/5343957102/jobs/10178767536#step:7:292

mythi:
> Something went wrong here https://github.com/confidential-containers/guest-components/actions/runs/5343957102/jobs/10178767536#step:7:292

I just rebased to fix merge conflicts. Let's see the results now.

Xynnn007 (approved):
lgtm, thanks!

mythi:
> Hi @mythi Can you rebase this PR?

done

PR #169, opened 2023-06-22
attester: az_snp_vtpm: Check cpuid bits in detect_platform()

jepio:
It took a while to realize that we can check cpuid registers exposed by Hyper-V to determine whether we are in an SNP VM that can use the az-snp-vtpm attester. Rework detect_platform() to perform these checks.

Closes: https://github.com/confidential-containers/attestation-agent/pull/220


fitzthum:
What happened with this PR?

dcmiddle:
Looks like this was superseded by https://github.com/confidential-containers/guest-components/pull/399

PR #170, opened 2023-06-22
Add Azure TDX (preview) support

jepio:
These changes were needed to get the TDX attestation to work on Azure.
Based on https://github.com/intel/amber-client/commit/c590bdea577a28a0143b0634ced6eeb1688d192f.

Closes: https://github.com/confidential-containers/attestation-agent/pull/221 (@mythi started reviewing)

### DIFF
--- a/attestation-agent/attester/Cargo.toml
+++ b/attestation-agent/attester/Cargo.toml
@@ -10,8 +10,11 @@ anyhow.workspace = true
 async-trait.workspace = true
 az-snp-vtpm = { git = "https://github.com/kinvolk/azure-cvm-tooling", rev = "2c2e411", default-features = false, features = ["attester"], optional = true }
 base64.workspace = true
+ioctl-sys = { version = "0.8.0", optional = true }
 log.workspace = true
 occlum_dcap = { git = "https://github.com/occlum/occlum", rev = "dbe404f", optional = true }
+raw-cpuid = { version = "11", optional = true }
+reqwest = { version = "0.11", features = ["json"], optional = true }
 serde.workspace = true
 serde_json.workspace = true
 sev = { git = "https://github.com/virtee/sev", rev = "3dca05d2c93388cb00534ad18f5928fd812e99cc", optional = true }
@@ -25,7 +28,7 @@ tokio = { version = "1", features = ["macros"] }
 default = ["all-attesters"]
 all-attesters = ["tdx-attester", "occlum-attester", "az-snp-vtpm-attester", "snp-attester"]
 
-tdx-attester = ["tdx-attest-rs"]
+tdx-attester = ["tdx-attest-rs", "dep:reqwest", "dep:ioctl-sys", "dep:raw-cpuid"]
 occlum-attester = ["occlum_dcap"]
 az-snp-vtpm-attester = ["az-snp-vtpm"]
 snp-attester = ["sev"]
--- a/attestation-agent/attester/Cargo.toml
+++ b/attestation-agent/attester/Cargo.toml
@@ -7,6 +7,7 @@ edition = "2021"
 
 [dependencies]
 anyhow.workspace = true
+async-trait.workspace = true
 az-snp-vtpm = { git = "https://github.com/kinvolk/azure-cvm-tooling", rev = "2c2e411", default-features = false, features = ["attester"], optional = true }
 base64.workspace = true
 log.workspace = true
@@ -16,6 +17,9 @@ serde_json.workspace = true
 sev = { git = "https://github.com/virtee/sev", rev = "3dca05d2c93388cb00534ad18f5928fd812e99cc", optional = true }
 strum.workspace = true
 tdx-attest-rs = { git = "https://github.com/intel/SGXDataCenterAttestationPrimitives", rev = "cc582e8be0c9010295c66fb58c59f74744017600", optional = true }
+
+[dev-dependencies]
+tokio = { version = "1", features = ["macros"] }
 
 [features]
 default = ["all-attesters"]
--- a/attestation-agent/attester/src/tdx/az.rs
+++ b/attestation-agent/attester/src/tdx/az.rs
@@ -1,0 +1,105 @@
+// Copyright (c) Microsoft Corporation.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use anyhow::*;
+use ioctl_sys::ioctl;
+use raw_cpuid::{CpuId, Hypervisor};
+use serde::{Deserialize, Serialize};
+use std::fs::OpenOptions;
+use std::io::Error;
+use std::os::unix::io::AsRawFd;
+
+use crate::Attester;
+
+// Length of the REPORTDATA used in TDG.MR.REPORT TDCALL
+const TDX_REPORTDATA_LEN: usize = 64;
+
+// Length of TDREPORT used in TDG.MR.REPORT TDCALL
+const TDX_REPORT_LEN: usize = 1024;
+
+#[repr(C)]
+#[derive(Debug)]
+pub struct TdxReportReq {
+    reportdata: [u8; TDX_REPORTDATA_LEN],
+    tdreport: [u8; TDX_REPORT_LEN],
+}
+
+impl TdxReportReq {
+    pub fn new(reportdata: [u8; TDX_REPORTDATA_LEN]) -> Self {
+        Self {
+            reportdata,
+            tdreport: [0; TDX_REPORT_LEN],
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize)]
+struct AzTdQuoteRequest {
+    report: String,
+}
+#[derive(Serialize, Deserialize, Debug)]
+struct AzTdQuoteResponse {
+    quote: String,
+}
+
+impl Default for TdxReportReq {
+    fn default() -> Self {
+        Self {
+            reportdata: [0; TDX_REPORTDATA_LEN],
+            tdreport: [0; TDX_REPORT_LEN],
+        }
+    }
+}
+
+ioctl!(readwrite tdx_cmd_get_report0 with b'T', 0x01; TdxReportReq);
+
+pub(super) async fn get_hyperv_tdx_evidence(report_data: &[u8]) -> Result<String> {
+    let file = OpenOptions::new().write(true).open("/dev/tdx_guest")?;
+    let fd = file.as_raw_fd();
+    let mut tdx_req = TdxReportReq::new(report_data.try_into()?);
+    unsafe {
+        let err = tdx_cmd_get_report0(fd, &mut tdx_req);
+        if err != 0 {
+            bail!("TDX Attester: ioctl failed: {}", Error::last_os_error());
+        }
+    }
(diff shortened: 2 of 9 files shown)


attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -9,10 +9,24 @@ use serde::{Deserialize, Serialize};
...
 const CCEL_PATH: &str = "/sys/firmware/acpi/tables/data/CCEL";
 
 pub fn detect_platform() -> bool {
-    Path::new("/dev/tdx-attest").exists() || Path::new("/dev/tdx-guest").exists()
+    // Upstream kernel exposes a /dev/tdx_guest device but Intel's lib expects
+    // /dev/tdx-guest
+    let paths = [
+        Path::new("/dev/tdx-attest"),
+        Path::new("/dev/tdx-guest"),
+        Path::new("/dev/tdx_guest"),
```
mythi, line 22: Getting back to the cpuid topic one more time ;-) Would it be sufficient to check `TD guest` cpuid bit here or we want to explicitly check the device nodes are present?

jepio: In this spot we could check the `TD guest` bit instead, i don't mind doing it that way

attestation-agent/attester/src/tdx/az.rs
```diff
@@ -0,0 +1,108 @@
...
+            return Err(anyhow!(
+                "TDX Attester: ioctl failed: {}",
+                Error::last_os_error()
+            ));
+        }
+    }
+    let report = base64::encode_config(tdx_req.tdreport, base64::URL_SAFE_NO_PAD);
+    let tdquotereq = AzTdQuoteRequest { report };
+    let req = reqwest::Client::new()
+        .post("http://169.254.169.254/acc/tdquote")
```
Xynnn007, line 74: Seems that the conversion from report to quote doesn't follow the standard way, s.t. call to host. Why?

attestation-agent/attester/src/tdx/mod.rs
jepio: Do you have a pointer to the cpuid TD guest bit? All I can find cpuid leaf 0x21 which returns the "IntelTDX" string after reassembling registers.

attestation-agent/attester/src/tdx/az.rs
jepio: Unfortunately it doesn't look like there is a single standard way for TDX, see this https://lore.kernel.org/lkml/cover.1684048511.git.sathyanarayanan.kuppuswamy@linux.intel.com/:
> Although attestation software can use communication methods like TCP/IP or vsock to send the TDREPORT to QE, not all platforms support these communication models. So TDX GHCI specification [1] defines a method for Quote generation via hypercalls.

It looks like TCP/IP using the IMDS service as a proxy to QE was the most best way to implement the communication for preview. It might look different for GA.

Xynnn007: Is the QE on the same platform/host? I mean the conversion from report to quote rely on the verification of the MAC of the report, and the key can only be generated on the same platform via EGETKEY.

jepio: I think yes, the QE VM is on the same node as the TD VM.

attestation-agent/attester/Cargo.toml
```diff
@@ -7,21 +7,28 @@ edition = "2021"
...
 [dependencies]
 anyhow.workspace = true
+async-trait.workspace = true
 az-snp-vtpm = { git = "https://github.com/kinvolk/azure-cvm-tooling", rev = "2c2e411", default-features = false, features = ["attester"], optional = true }
 base64.workspace = true
+ioctl-sys = { version = "0.8.0", optional = true }
 log.workspace = true
 occlum_dcap = { git = "https://github.com/occlum/occlum", rev = "dbe404f", optional = true }
+reqwest = { version = "0.11", features = ["json"], optional = true }
+raw-cpuid = { version = "11", optional = true }
```
Xynnn007, line 17:
```suggestion
raw-cpuid = { version = "11", optional = true }
reqwest = { version = "0.11", features = ["json"], optional = true }
```

attestation-agent/attester/src/tdx/az.rs
```diff
@@ -0,0 +1,108 @@
+// Copyright (c) 2022 Alibaba Cloud
```
Xynnn007, line 1: I think it should be `Microsoft`, right?

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -24,18 +38,32 @@ struct TdxEvidence {
...
+        Box::<TdxAttester>::default()
+    }
+}
+
+fn convert_report_data(report_data: String) -> Result<Vec<u8>> {
+    let mut report_data_bin = base64::decode(report_data)?;
+    if report_data_bin.len() != 48 {
+        return Err(anyhow!(
+            "TDX Attester: Report data should be SHA384 base64 String"
+        ));
```
Xynnn007, line 54: `anyhow::bail!` macro helps to make it shorter
```suggestion
        bail!("TDX Attester: Report data should be SHA384 base64 String");
```

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -9,10 +9,24 @@ use serde::{Deserialize, Serialize};
...
+        Path::new("/dev/tdx-attest"),
+        Path::new("/dev/tdx-guest"),
+        Path::new("/dev/tdx_guest"),
+    ];
+    for path in paths {
+        if path.exists() {
+            return true;
+        }
+    }
+    false
```
Xynnn007, line 29: There is a rust way I have ever read. Just fYI : )
```rust
pub fn detect_platform() -> bool {
    ["/dev/tdx-attest", "/dev/tdx-guest", "/dev/tdx_guest"]
        .iter()
        .any(|p| Path::new(p).exists())
}
```

attestation-agent/attester/Cargo.toml
jepio: done

attestation-agent/attester/src/tdx/az.rs
jepio: yes, thank you

attestation-agent/attester/src/tdx/mod.rs
jepio: i didn't know this, but i like it 👍

attestation-agent/attester/src/tdx/mod.rs
jepio: thanks for the hint.

i've reworked this part to use cpuid like Mikko suggested. Can you look and tell me what you think about it?

attestation-agent/attester/src/tdx/mod.rs
jepio: i'm now using the cpuid leaf 0x21

attestation-agent/attester/Cargo.toml
```diff
@@ -7,21 +7,28 @@ edition = "2021"
...
 
+[dev-dependencies]
+tokio = { version = "1", features = ["macros"] }
+
 [features]
 default = ["all-attesters"]
 all-attesters = ["tdx-attester", "occlum-attester", "az-snp-vtpm-attester", "snp-attester"]
 
-tdx-attester = ["tdx-attest-rs"]
+tdx-attester = ["tdx-attest-rs", "dep:reqwest", "dep:ioctl-sys", "dep:raw-cpuid"]
```
Xynnn007, line 31: I'm thinking about this: if more CSPs are going to add their attester code under `tdx`, like `tdx/az`, `tdx/ali`, `tdx/another-company`. The all features shares a same feature here, but must introduce some dependencies they will not use.

The advantage is the community version binary of attestation-agent can be easy to build (only one!)
The disadvantage is big footprint, which will influence the performance and cost.

I do not have a clear solution here, and maybe we should bring a more robust verified build pipeline (CD) of the components of CoCo Community to solve the trustiness here.

attestation-agent/attester/src/tdx/az.rs
```diff
@@ -0,0 +1,108 @@
...
+    let file = OpenOptions::new().write(true).open("/dev/tdx_guest")?;
+    let fd = file.as_raw_fd();
+    let mut tdx_req = TdxReportReq::new(report_data.try_into()?);
+    unsafe {
+        let err = tdx_cmd_get_report0(fd, &mut tdx_req);
+        if err != 0 {
+            return Err(anyhow!(
+                "TDX Attester: ioctl failed: {}",
+                Error::last_os_error()
+            ));
```
Xynnn007, line 68:
```suggestion
            bail!("TDX Attester: ioctl failed: {}", Error::last_os_error());
```

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -5,14 +5,20 @@
...
+
 const CCEL_PATH: &str = "/sys/firmware/acpi/tables/data/CCEL";
 
 pub fn detect_platform() -> bool {
-    Path::new("/dev/tdx-attest").exists() || Path::new("/dev/tdx-guest").exists()
+    const TDX_CPUID_LEAF: u32 = 0x21;
+    let c = cpuid!(TDX_CPUID_LEAF, 0);
+    let tdbytes = [c.ebx, c.edx, c.ecx].map(|x| x.to_le_bytes()).concat();
+    let tdstring = String::from_utf8_lossy(&tdbytes);
+    tdstring == "IntelTDX    "
```
Xynnn007, line 21: looks good to me

Xynnn007 (approved):
Overall, LGTM, and I'm looking forward to seeing Azure to embrace CoCo as an attester. There is a thought left that I want to hear your opinions

attestation-agent/attester/src/tdx/az.rs
jepio: thanks, i like this

attestation-agent/attester/Cargo.toml
jepio: I think for community builds we should support as many different attesters as possible, so that it is easy to try out coco.

But we will need to have ways to disable/enable things that pull in a lot of dependencies so that companies don't need to build in code that will never be used in their environment. We'll add them over time.

Of these dependencies `reqwest` is the largest, but it is also already used by `kbs_protocol`.

> I do not have a clear solution here, and maybe we should bring a more robust verified build pipeline (CD) of the components of CoCo Community to solve the trustiness here.

👍

attestation-agent/attester/src/tdx/mod.rs
mythi: > Do you have a pointer to the cpuid TD guest bit? All I can find cpuid leaf 0x21 which returns the "IntelTDX" string after reassembling registers.

This is what I meant. Sorry for the confusion about that "bit" which was a misleading word.

mythi (approved):
My comments were addressed so LGTM

jepio:
Let's not merge this - it was a nice experiment but the TDX interfaces will change from those exposed during the preview.

Will revisit in the coming months.

PR #173, opened 2023-06-26
Adjusted the handling of URLs to support more types of hosts

jialez0:
Currently, our default host URL format for KBS is http://IP:PORT, But in productization, we may give host URL more forms. For example: http://my-kbs.com , http://my-kbs.com/user-alice Therefore, it is necessary to expand the supported KBS host URL format, More precisely, it should no longer be referred to as the "host URL", but rather as the "root URL".

### DIFF
--- a/attestation-agent/kbs_protocol/src/lib.rs
+++ b/attestation-agent/kbs_protocol/src/lib.rs
@@ -17,7 +17,7 @@ pub mod types;
 const KBS_REQ_TIMEOUT_SEC: u64 = 60;
 const KBS_GET_RESOURCE_MAX_ATTEMPT: u64 = 3;
 
-pub const KBS_URL_PREFIX: &str = "kbs/v0";
+pub const KBS_PREFIX: &str = "/kbs/v0";
 
 #[async_trait]
 pub trait KbsRequest {
@@ -93,10 +93,10 @@ impl KbsProtocolWrapper {
         &mut self.http_client
     }
 
-    async fn attestation(&mut self, kbs_host_url: String) -> Result<String> {
+    async fn attestation(&mut self, kbs_root_url: String) -> Result<String> {
         let challenge = self
             .http_client()
-            .post(format!("{kbs_host_url}/{KBS_URL_PREFIX}/auth"))
+            .post(format!("{kbs_root_url}/{KBS_PREFIX}/auth"))
             .header("Content-Type", "application/json")
             .json(&Request::new(self.tee().to_string()))
             .send()
@@ -107,7 +107,7 @@ impl KbsProtocolWrapper {
 
         let attest_response = self
             .http_client()
-            .post(format!("{kbs_host_url}/{KBS_URL_PREFIX}/attest"))
+            .post(format!("{kbs_root_url}/{KBS_PREFIX}/attest"))
             .header("Content-Type", "application/json")
             .json(&self.generate_evidence()?)
             .send()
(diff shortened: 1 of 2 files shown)


attestation-agent/kbs_protocol/src/lib.rs
```diff
@@ -136,19 +136,16 @@ impl KbsProtocolWrapper {
...
     async fn http_get(&mut self, url_string: String) -> Result<Vec<u8>> {
-        let url = Url::parse(&url_string)
+        let _ = Url::parse(&url_string)
             .map_err(|e| anyhow!("Invalid Request URL {url_string}: {e}"))?;
 
-        // Use default port of KBS: 8080
-        let port = url.port().unwrap_or(8080);
-        let host_str = url.host_str().ok_or_else(|| anyhow!("No KBS host given"))?;
-        let kbs_url = format!("{}://{host_str}:{port}", url.scheme());
+        let root_url = url_string.split(KBS_URL_PREFIX).collect::<Vec<&str>>()[0].to_string();
```
Xynnn007, line 142: If the given URL does not have a `KBS_URL_PREFIX` const, is it still as expected? s.t.
```rust
let root_url = "google.com/not-kbs".split("/kbs/v0").collect::<Vec<&str>>();
// root_url is "google.com/not-kbs"
```

jialez0: This type of URL input without a `KBS_URL_PREFIX` const is illegal, that situation will not occur in the normal call scenario of this function

jialez0:
cc @sameo

attestation-agent/kbs_protocol/src/lib.rs
```diff
@@ -17,7 +17,7 @@ pub mod types;
 const KBS_REQ_TIMEOUT_SEC: u64 = 60;
 const KBS_GET_RESOURCE_MAX_ATTEMPT: u64 = 3;
 
-pub const KBS_URL_PREFIX: &str = "kbs/v0";
+pub const KBS_URL_PREFIX: &str = "/kbs/v0";
```
arronwy, line 20: maybe `KBS_RESOURCE_PREFIX` is better?

attestation-agent/kbs_protocol/src/lib.rs
```diff
@@ -93,10 +93,10 @@ impl KbsProtocolWrapper {
         &mut self.http_client
     }
 
-    async fn attestation(&mut self, kbs_host_url: String) -> Result<String> {
+    async fn attestation(&mut self, kbs_root_url: String) -> Result<String> {
```
arronwy, line 96: maybe use `kbs_url` directly which contains the `scheme://domain-name[:port][/user]`

arronwy (approved):
Thanks @jialez0 LGTM! Just some minor naming suggestion.

attestation-agent/kbs_protocol/src/lib.rs
jialez0: The prefix here is not only used to obtain resources, but also to obtain tokens in the passport model. Perhaps we can just call it `KBS_PREFIX`.

attestation-agent/kbs_protocol/src/lib.rs
jialez0: In many domain name allocation scenario, the root domain name assigned to KBS may have a more complex form. I believe we should leave enough flexibility for the form of the root domain name. Therefore, for now, I suggest not defining and resolving the format and content of the root domain name.

jialez0: > maybe use `kbs_url` directly which contains the `scheme://domain-name[:port][/user]`

If we use `kbs_url` here, it may be confused with the whole resource url.

PR #264, opened 2023-07-04
Optimize implementation of image-rs/stream

jiangliu:
Optimize implementation of image-rs/stream:
1) unify digest algorithm prefixes
2) avoid unnecessary memory copy
3) simplify function stream_processing()
4) get rid of flume crate to reduce dependency


arronwy (approved):
Thanks @jiangliu LGTM!

PR #266, opened 2023-07-07
Optimize implementation of image-rs/pull

jiangliu:
Optimize implementation of image-rs/pull by:
1) simplify implementation of async_pull_layers()
2) remove pull_layers()
3) simplify definition of pull_bootstrap()

### DIFF
--- a/image-rs/src/pull.rs
+++ b/image-rs/src/pull.rs
@@ -273,72 +273,58 @@ impl<'a> PullClient<'a> {
         ms: Arc<Mutex<MetaStore>>,
     ) -> Result<LayerMeta> {
         let layer_db = &ms.lock().await.layer_db;
-
         if let Some(layer_meta) = layer_db.get(&layer.digest) {
             return Ok(layer_meta.clone());
         }
 
-        let mut layer_meta = LayerMeta::default();
-        let mut media_type_str: &str = layer.media_type.as_str();
-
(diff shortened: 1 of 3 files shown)


image-rs/src/pull.rs
```diff
@@ -19,19 +16,13 @@ use crate::decrypt::Decryptor;
...
 use crate::meta_store::MetaStore;
 use crate::stream::stream_processing;
-use crate::unpack::unpack;
-
-const DIGEST_SHA256: &str = "sha256";
-const DIGEST_SHA512: &str = "sha512";
 
 const ERR_NO_DECRYPT_CFG: &str = "decrypt_config is None";
-const ERR_BAD_UNCOMPRESSED_DIGEST: &str = "unsupported uncompressed digest format";
-const ERR_BAD_COMPRESSED_DIGEST: &str = "unsupported compressed digest format";
```
arronwy, line 29: These two const str still needed by the `test_async_handle_layer`

jiangliu: It would be better use a single definition instead of using two from both stream.rs and pull.rs.
So moved `ERR_BAD_UNCOMPRESSED_DIGEST` into lib.rs

arronwy (approved):
Thanks @jiangliu LGTM! all CI passed.

PR #270, opened 2023-07-10
image-rs: add image block device dm-verity and mount

ChengyuZhu6:
Implement image block device integrity check using dm-verity and mount the verity device in image-rs.
Notes: the kernel should be upgrade to >=6.4 for mounting the verity device with erofs tarfs.

### DIFF
--- a/image-rs/Cargo.toml
+++ b/image-rs/Cargo.toml
@@ -15,6 +15,7 @@ async-trait.workspace = true
 attestation_agent = { path = "../attestation-agent/lib", optional = true }
 base64.workspace = true
 cfg-if = { workspace = true, optional = true }
+devicemapper = "0.33.5"
 dircpy = { version = "0.3.12", optional = true }
 flate2 = "1.0"
 fs_extra = { version = "1.2.0", optional = true }
@@ -24,6 +25,7 @@ hex = { version = "0.4.3", optional = true }
 lazy_static = { workspace = true, optional = true }
 libc = "0.2"
 log = "0.4.14"
+loopdev ="0.4.0"
 nix = { version = "0.26", optional = true }
 oci-distribution = { git = "https://github.com/krustlet/oci-distribution.git", rev = "f44124c", default-features = false, optional = true }
 oci-spec = "0.5.8"
--- a/image-rs/src/image.rs
+++ b/image-rs/src/image.rs
@@ -4,6 +4,7 @@
 
 use anyhow::{anyhow, bail, Result};
 use log::warn;
+use nix::mount::MsFlags;
 use oci_distribution::manifest::{OciDescriptor, OciImageManifest};
 use oci_distribution::secrets::RegistryAuth;
 use oci_distribution::Reference;
@@ -21,6 +22,7 @@ use crate::config::{ImageConfig, CONFIGURATION_FILE_PATH};
 use crate::decoder::Compression;
 use crate::meta_store::{MetaStore, METAFILE};
 use crate::pull::PullClient;
+use crate::verity;
 
 #[cfg(feature = "snapshot-unionfs")]
 use crate::snapshots::occlum::unionfs::Unionfs;
@@ -432,6 +434,38 @@ impl ImageClient {
     }
 }
 
+/// mount_image_block_with_integrity creates a mapping backed by image block device <source_device_path> and
+/// decoding <verity_options> for in-kernel verification. And mount the verity device
+/// to <mount_path> with <mount_type>.
+/// It will return the verity device path if succeeds and return an error if fails .
+pub fn mount_image_block_with_integrity(
+    verity_options: &str,
+    source_device_path: &Path,
+    mount_path: &Path,
+    mount_type: &str,
+) -> Result<String> {
+    let parsed_data = verity::decode_verity_options(verity_options)?;
+    let verity_device_path = verity::create_verity_device(&parsed_data, source_device_path)?;
+
+    nix::mount::mount(
+        Some(verity_device_path.as_str()),
+        mount_path,
+        Some(mount_type),
+        MsFlags::MS_RDONLY,
+        None::<&str>,
+    )?;
+    Ok(verity_device_path)
+}
+/// umount_image_block_with_integrity umounts the filesystem and closes the verity device named verity_device_name.
+pub fn umount_image_block_with_integrity(
+    mount_path: &Path,
+    verity_device_name: String,
+) -> Result<()> {
+    nix::mount::umount(mount_path)?;
+    verity::close_verity_device(verity_device_name)?;
+    Ok(())
+}
+
 /// Create image meta object with the image info
 /// Return the image meta object, oci descriptors of the unique layers, and unique diff ids.
 fn create_image_meta(
@@ -507,6 +541,9 @@ fn create_bundle(
 #[cfg(test)]
 mod tests {
     use super::*;
+    use base64::Engine;
+    use std::fs;
+    use std::process::Command;
 
     #[tokio::test]
     async fn test_pull_image() {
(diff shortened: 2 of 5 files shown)


ChengyuZhu6:
> ` the kernel should be upgrade to >=6.4 for mounting the verity device with erofs.` This is not accurate, only erofs tarfs support was landed in Linux 6.4. Otherwise, if you just mount exist erofs images much like ext4 with dm-verity, there is no such requirement (5.4, 5.10, 5.15, 6.1 LTS are all supported. )

Thanks, I have updated.

image-rs/src/image.rs
```diff
@@ -39,7 +41,6 @@ use crate::nydus::{service, utils};
 /// The reason for using the `/run` directory here is that in general HW-TEE,
 /// the `/run` directory is mounted in `tmpfs`, which is located in the encrypted memory protected by HW-TEE.
 pub const IMAGE_SECURITY_CONFIG_DIR: &str = "/run/image-security";
-
```
jiangliu, line 42: Unrelated change?

image-rs/src/image.rs
```diff
@@ -421,6 +422,33 @@ impl ImageClient {
 
         Ok(image_id)
     }
+    /// set_image creates a mapping backed by image block device <source_device_path> and
```
jiangliu, line 425: Add a blank line above?

image-rs/src/verity.rs
```diff
@@ -0,0 +1,144 @@
...
+    Ok(decoded_data.hash)
+}
+/// Calculates and permanently stores hash verification data for data_device.
+/// It will return the root hash.
+pub fn create_hash_device(
+    verity_options: &mut DmVerityOption,
+    data_device_path: &str,
+) -> Result<String> {
+    let veritysetup_path: &str = veritysetup_exists()?;
+    let output = Command::new(veritysetup_path)
```
jiangliu, line 88: How about using https://docs.rs/devicemapper/latest/devicemapper/struct.DM.html to get rid of the dependency on veritysetup binary?

ChengyuZhu6: > I haven't used devicemapper, but @jiangliu's comment seems right to me. Related to this: it looks like `verity_option` doesn't get sanity-checked or sanitized anywhere, so its values get plugged directly into the args to `versitysetup` when issuing the `Command`. Using a library call (@jiangliu's suggestion) and/or sanity-checking those values might make error handling/checking easier.

Yes, I agree with @jiangliu's comment and I am working on using the devicemapper library to handle/check the options.

image-rs/src/verity.rs
```diff
@@ -0,0 +1,102 @@
+use anyhow::Result;
```
jiangliu, line 1: How about adding a copyright header?

image-rs/Cargo.toml
```diff
@@ -24,6 +25,7 @@ hex = { version = "0.4.3", optional = true }
 lazy_static = { version = "1.4.0", optional = true }
 libc = "0.2"
 log = "0.4.14"
+loopdev ="0.1.0"
```
jiangliu, line 28: `loopdev = "0.4.0"`?

image-rs/src/verity.rs
```diff
@@ -0,0 +1,102 @@
...
+    pub hash: String,
+}
+
+/// Creates a mapping with <name> backed by data_device <source_device_path>
+/// and using hash_device for in-kernel verification.
+/// It will return the verity block device Path "/dev/mapper/<name>"
+/// Notes: the data device and the hash device are the same one.
+pub fn create_verity_device(
+    verity_option: &DmVerityOption,
+    source_device_path: &str,
```
jiangliu, line 29: How about using `source_device_path: &Path` here and `source_device_path.display()` below?
So the caller will get simpler.

image-rs/src/verity.rs
```diff
@@ -0,0 +1,102 @@
...
+) -> Result<String> {
+    let dm = DM::new()?;
+    let verity_name = DmName::new(&verity_option.hash)?;
+    let id = DevId::Name(verity_name);
+    let opts = DmOptions::default().set_flags(DmFlags::DM_READONLY);
+    let device_num_blocks = verity_option.blocknum.parse::<u64>().unwrap();
```
jiangliu, line 35: Please return error instead of `unwrap()` here.

jiangliu: The data source is untrusted and unverified, so we can't unwrap() here to avoid possible attack.

image-rs/src/verity.rs
```diff
@@ -0,0 +1,102 @@
...
+use devicemapper::{DevId, DmFlags, DmName, DmOptions, DM};
+use serde::{Deserialize, Serialize};
+use serde_json;
+
+#[derive(Debug, Deserialize, Serialize)]
+pub struct DmVerityOption {
+    /// Hash algorithm for dm-verity.
+    pub hashtype: String,
+    /// Used block size for the data device.
+    pub blocksize: String,
```
jiangliu, line 12: how about using `u64` instead of `String`?

image-rs/src/verity.rs
```diff
@@ -0,0 +1,102 @@
...
+) -> Result<String> {
+    let dm = DM::new()?;
+    let verity_name = DmName::new(&verity_option.hash)?;
+    let id = DevId::Name(verity_name);
+    let opts = DmOptions::default().set_flags(DmFlags::DM_READONLY);
+    let device_num_blocks = verity_option.blocknum.parse::<u64>().unwrap();
+    let device_block_size = verity_option.blocksize.parse::<u64>().unwrap();
+    let hash_block_size = verity_option.hashsize.parse::<u64>().unwrap();
+    let hash_start_block: u64 =
+        (device_num_blocks * device_block_size + hash_block_size - 1) / hash_block_size;
```
jiangliu, line 39: Please help to verify:
- device_block_size and hash_block_size is power of two.
- hash_block_size is not zero

jiangliu: BTW, we have `verity_option.offset` to calculate `hash_start_block`?

image-rs/src/verity.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+pub fn get_verity_device_name(verity_options: &str) -> Result<String> {
+    let parsed_data = decode_verity_options(verity_options)?;
+    Ok(parsed_data.hash)
+}
+
+pub fn decode_verity_options(verity_options: &str) -> Result<DmVerityOption> {
+    let decoded = base64::decode(verity_options)?;
+    let parsed_data = serde_json::from_slice::<DmVerityOption>(&decoded)?;
```
jiangliu, line 90: Please help to validate parsed DmVerityOption here.

image-rs/src/verity.rs
ChengyuZhu6: Thanks, I'll update it.

image-rs/src/verity.rs
ChengyuZhu6: Thanks, I'll update it.

image-rs/src/verity.rs
ChengyuZhu6: Thanks, I'll update it.

image-rs/src/verity.rs
ChengyuZhu6: Thanks, I'll update it.

image-rs/src/verity.rs
ChengyuZhu6: Thanks, I'll update it.

image-rs/src/verity.rs
ChengyuZhu6: Thanks, I'll add a copyright.

image-rs/src/image.rs
```diff
@@ -146,7 +148,33 @@ impl Default for ImageClient {
         }
     }
 }
-
+/// set_image creates a mapping backed by image block device <source_device_path> and
+/// decoding <verity_options> for in-kernel verification. And mount the verity device
+/// to <mount_path> with <mount_type>.
+/// It will return the verity device path if succeeds and return an error if fails .
+pub fn mount_image_block_with_integrity(
```
jiangliu, line 155: Do we need the counterpart `unmount_image_block_with_integrity()` to clean up all resources?

PR #271, opened 2023-07-10
ci: Fix integration test for image-rs

Xynnn007:
After merging attestation-agent into image-rs, we do not need to pull attestation-agent separately when doing the integration test. It is only needed to build the attestation-agent inside the current workspace.

cc @taoohong


Xynnn007:
There are still some errors. let me fix this tomorrow

arronwy (approved):
Thanks @Xynnn007 LGTM!

PR #274, opened 2023-07-11
image-rs: improve AA build and output

katexochen:
Improving output when building and starting AA.
Placing gRPC and ttRPC binaries in different directories, so both can be cached and it is clear what feature the binary supports.

Test itself was already fixed in #271.

### DIFF
--- a/image-rs/scripts/build_attestation_agent.sh
+++ b/image-rs/scripts/build_attestation_agent.sh
@@ -9,11 +9,17 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-parameters="KBC=offline_fs_kbc"
+parameters=("KBC=offline_fs_kbc")
 
 [ -n "${BASH_VERSION:-}" ] && set -o errtrace
 [ -n "${DEBUG:-}" ] && set -o xtrace
-[ -n "${TTRPC:-}" ] && parameters+=" ttrpc=true"
+if [[ -n "${TTRPC:-}" ]]; then
+    parameters+=("ttrpc=true")
+    dest_dir_suffix="ttrpc"
+else
+    dest_dir_suffix="grpc"
+fi
+
 source $HOME/.cargo/env
 
 SCRIPT_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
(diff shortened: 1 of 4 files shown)


image-rs/scripts/build_attestation_agent.sh
```diff
@@ -9,18 +9,26 @@ set -o errexit
...
+parameters=("KBC=offline_fs_kbc")
 
 [ -n "${BASH_VERSION:-}" ] && set -o errtrace
 [ -n "${DEBUG:-}" ] && set -o xtrace
-[ -n "${TTRPC:-}" ] && parameters+=" ttrpc=true"
+if [[ -n "${TTRPC:-}" ]]; then
+    parameters+=("ttrpc=true")
+    dest_dir_suffix="ttrpc"
+else
+    dest_dir_suffix="grpc"
```
fitzthum, line 20: Will we also need changes to the kata initrd building scripts and peer pods tooling to account for the new location?

image-rs/tests/common/mod.rs
```diff
@@ -59,47 +59,59 @@ pub async fn clean() {
...
                 .kill_on_drop(true)
                 .args(&[
                     "--keyprovider_sock",
                     "unix:///run/confidential-containers/attestation-agent/keyprovider.sock",
                     "--getresource_sock",
                     "unix:///run/confidential-containers/attestation-agent/getresource.sock"
                     ])
-                .spawn()?;
+                .spawn()
+                .expect("Failed to start ttrpc attestation-agent");
```
fitzthum, line 103: Nothing against this more verbose error but note that currently the `--help` flag will tell you if the AA was built with grpc or ttrpc.

katexochen: That can be helpful for local development, but won't help for debugging CI runs.

arronwy (approved):
Thanks @katexochen  LGTM!

image-rs/scripts/build_attestation_agent.sh
katexochen: I'm not sure I understand that question. I think this build script is only ever called by the image-rs tests. Everything else should use the attestation-agents Makefile. The small modification to the attestation-agent Makefile should only change behavior in edge cases.

PR #275, opened 2023-07-11
ci: refactor workflows

katexochen:
Some refactoring of workflows:

- Remove invalid filter on create trigger. This trigger doesn't allow filters.
- Only use push trigger on main (wasn't used before anyway, as there was an if statement on most jobs).
- Adding path filters to image-rs and ocicrypt-rs workflows.
- Removing if statements on jobs. The workflow triggers control when a job run, no need to filter with an if statement on a single job.
- Trying to unify naming a bit, fixing white space issues.
- Adding `workflow_dispatch` trigger to every test workflow, so they can be triggered manually where needed (like on forks).

### DIFF
--- a/.github/workflows/aa_cc_kbc.yml
+++ b/.github/workflows/aa_cc_kbc.yml
@@ -1,4 +1,5 @@
-name: CC kbc build CI
+name: attestation-agent cc_kbc tests
+
 on:
   push:
     branches:
--- a/.github/workflows/aa_cc_kbc.yml
+++ b/.github/workflows/aa_cc_kbc.yml
@@ -1,6 +1,8 @@
 name: CC kbc build CI
 on:
   push:
+    branches:
+      - 'main'
     paths:
       - 'attestation-agent/kbc/cc_kbc/**'
       - 'attestation-agent/kbs_protocol/**'
--- a/.github/workflows/aa_cc_kbc.yml
+++ b/.github/workflows/aa_cc_kbc.yml
@@ -14,6 +14,7 @@ on:
       - 'attestation-agent/kbs_protocol/**'
       - '.github/workflows/aa_cc_kbc.yml'
   create:
+  workflow_dispatch:
 
 jobs:
   cc_kbc_ci:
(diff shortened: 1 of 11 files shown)


.github/workflows/image_rs_build.yml
```diff
@@ -1,9 +1,20 @@
 name: image-rs build
-on: [push, pull_request, create]
+on:
+  push:
+    branches:
+      - 'main'
+    paths:
+      - 'image-rs/**'
+      - '.github/image_rs_build.yml'
```
portersrc, line 8: Hey @katexochen , I'm not strong with github workflows, but should this be `.github/workflow/image_rs_build.yml`? Similar thing for `.github/workflow/ocicrypt_rs_build.yml`

katexochen: Yes, thanks for catching this. :)

.github/workflows/aa_cc_kbc.yml
```diff
@@ -1,6 +1,8 @@
 name: CC kbc build CI
 on:
   push:
+    branches:
+      - "main"
```
Xynnn007, line 6: I found others are `'main'`. Does this matter? s.t. `'` and `"`

.gitignore
```diff
@@ -4,3 +4,4 @@ Cargo.lock
 .DS_Store
 
 image-rs/scripts/attestation-agent
+shell.nix
```
Xynnn007, line 7: Is this something left when test?

Xynnn007 (approved):
Much thanks for this @katexochen ! The CI now looks much better. Only some personal questions

.github/workflows/aa_cc_kbc.yml
katexochen: I don't think it makes a difference, but I updated it anyway for consistency.

.gitignore
katexochen: This file is part of my local development setup, so I added to .gitignore.

arronwy (approved):
Thanks @katexochen LGTM!

Xynnn007:
Can you rebase this PR? @katexochen

PR #279, opened 2023-07-12
Tidy up the warnings from integration tests

portersrc:
When looking at the "Files changed" in recent PRs, there are many extraneous warnings in image-rs/tests (e.g. "unused import"). The two main causes seem to be:
(1) the appropriate annotation for an integration test is not always applied to the respective import statement, variable, etc.
(2) not all integration tests use all of the common support from image-rs/tests/common/mod.rs

This PR tries to resolve the warnings. Annotations are added to solve (1). For (2), I've added `pub mod common` to the integration tests, which could lead to actual dead code that won't get caught by warnings. This seems to be a reasonable workaround, however, due to [this issue](https://github.com/rust-lang/rust/issues/46379).

### DIFF
--- a/image-rs/tests/common/mod.rs
+++ b/image-rs/tests/common/mod.rs
@@ -71,7 +71,7 @@ pub async fn start_attestation_agent() -> Result<Child> {
                     .await
                     .expect("Failed to build attestation-agent");
             } else {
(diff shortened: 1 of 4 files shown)


arronwy (approved):
Thanks @portersrc LGTM!

PR #280, opened 2023-07-13
docs: update generate_test_data guide link

Xynnn007:
With the help of @katexochen , there is a document bug found in https://github.com/confidential-containers/guest-components/actions/runs/5538239881/jobs/10107974050


arronwy (approved):
Thanks @Xynnn007 LGTM!

Issue #283, opened 2023-07-13
Fix the Implementations of `Attester`


Xynnn007:
Now the `Attester` has defined the following trait:
```rust
pub trait Attester {
    fn get_evidence(&self, report_data: String) -> Result<String>;
}
```

Which means directly the input will be included inside the report without any assumption. However, all the implementations firstly decode the input with base64-standard. My idea is to bring this logic out of all attester implementations. At the meantime, we should add more notes about the API to tell the user the limited length of the input.

PR #284, opened 2023-07-17
attestation-agent/Attesters: refactor the trait of Attester

Xynnn007:
Change the API of `get_evidence` function to just performing getting evidence via calling the underlying hardware drivers.

Fixes: #283

Please do not merge this until the v0.8.0 cycle.

cc @jialez0 @fitzthum @mkulke please take a look if these changes are good for the architecture/concrete attester implementation.

### DIFF
--- a/attestation-agent/attester/src/snp/mod.rs
+++ b/attestation-agent/attester/src/snp/mod.rs
@@ -25,17 +25,15 @@ struct SnpEvidence {
 pub struct SnpAttester {}
 
 impl Attester for SnpAttester {
-    fn get_evidence(&self, report_data: String) -> Result<String> {
-        let mut report_data_bin = base64::decode(report_data)?;
-
-        if report_data_bin.len() != 48 {
-            bail!("Malformed SNP Evidence");
+    fn get_evidence(&self, mut report_data: Vec<u8>) -> Result<String> {
+        if report_data.len() > 64 {
+            bail!("SNP Attester: Report data must be no more than 64 bytes");
         }
 
-        report_data_bin.extend([0; 16]);
+        report_data.resize(64, 0);
 
         let mut firmware = Firmware::open()?;
-        let data = report_data_bin.as_slice().try_into()?;
+        let data = report_data.as_slice().try_into()?;
 
         let (report, certs) = firmware
             .get_ext_report(None, Some(data), Some(0))
--- a/attestation-agent/attester/src/az_snp_vtpm/mod.rs
+++ b/attestation-agent/attester/src/az_snp_vtpm/mod.rs
@@ -28,10 +28,9 @@ struct Evidence {
 }
 
 impl Attester for AzSnpVtpmAttester {
-    fn get_evidence(&self, report_data: String) -> Result<String> {
+    fn get_evidence(&self, report_data: Vec<u8>) -> Result<String> {
         let report = vtpm::get_report()?;
-        let report_data_bin = base64::decode(report_data)?;
-        let quote = vtpm::get_quote(&report_data_bin)?;
+        let quote = vtpm::get_quote(&report_data)?;
         let certs = imds::get_certs()?;
         let vcek = certs.vcek;
 
--- a/attestation-agent/attester/src/lib.rs
+++ b/attestation-agent/attester/src/lib.rs
@@ -58,7 +58,10 @@ impl Tee {
 }
 
 pub trait Attester {
-    fn get_evidence(&self, report_data: String) -> Result<String>;
+    /// Call the hardware driver to get the Hardware specific evidence.
+    /// The parameter `report_data` will be used as the user input of the
+    /// evidence to avoid reply attack.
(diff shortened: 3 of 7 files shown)


attestation-agent/attester/src/snp/mod.rs
```diff
@@ -25,17 +25,15 @@ struct SnpEvidence {
...
 impl Attester for SnpAttester {
-    fn get_evidence(&self, report_data: String) -> Result<String> {
-        let mut report_data_bin = base64::decode(report_data)?;
-
-        if report_data_bin.len() != 48 {
-            bail!("Malformed SNP Evidence");
+    fn get_evidence(&self, mut report_data: Vec<u8>) -> Result<String> {
+        if report_data.len() > 64 {
+            bail!("SNP Attester: Report data must be no more than 64 bytes");
         }
```
mkulke, line 31: We might want to keep the signature open to accept slices. that'll save use a `try_into()?` conversion, too. 

```suggestion
    fn get_evidence(&self, mut report_data: &[u8]) -> Result<String> {
        if report_data.len() > 64 {
            bail!("SNP Attester: Report data must be no more than 64 bytes");
        }
        
        let mut data = [0u8; 64];
        data[..64].copy_from_slice(&report_data);
```

attestation-agent/attester/src/snp/mod.rs
```diff
@@ -25,17 +25,15 @@ struct SnpEvidence {
...
 
-        report_data_bin.extend([0; 16]);
+        report_data.resize_with(64, || 0);
```
mkulke, line 33: `report_data.resize(64, 0)` (if we want to keep the vec in the signature, we can just call)

mkulke (commented):
a small remark about the signature, but I agree it makes a lot of sense to move the encoding to the caller site.

attestation-agent/attester/src/snp/mod.rs
Xynnn007: If we use `Vec<u8>` here, it can explicitly let the caller to handle the memory allocation (by `new()` behind `.into()`). If a slice is accepted, the caller may use
```rust
// memory allocation 1
let report_data = get_report_data();

// memory allocation 2 inside the call
let ev = obj.get_evidence(&report_data).await?;
```

But with `Vec<u8>`
```rust
// memory allocation 1
let report_data = get_report_data();

// the lint check will prevent the caller to use `report_data.clone()`
// here to avoid an extra memory allocation on heap.
let ev = obj.get_evidence(report_data).await?;
```

attestation-agent/attester/src/snp/mod.rs
Xynnn007: Nice suggestion! Let me fix this

attestation-agent/attester/src/snp/mod.rs
mkulke: > If we use `Vec<u8>` here, it can explicitly let the caller to handle the memory allocation (by `new()` behind `.into()`). If a slice is accepted, the caller may use
> 
> ```rust
> // memory allocation 1
> let report_data = get_report_data();
> 
> // memory allocation 2 inside the call
> let ev = obj.get_evidence(&report_data).await?;
> ```
> 
> But with `Vec<u8>`
> 
> ```rust
> // memory allocation 1
> let report_data = get_report_data();
> 
> // the lint check will prevent the caller to use `report_data.clone()`
> // here to avoid an extra memory allocation on heap.
> let ev = obj.get_evidence(report_data).await?;
> ```

are you sure there is an additional allocation required in `get_evidence`? fixed-size arrays shouldn't allocate:

```rust
#![no_std]

#![allow(dead_code)]
fn get_evidence(report_data: &[u8]) {
   let mut data = [0u8, 64];
   data[..64].copy_from_slice(report_data);
   get_quote(&data);
}

fn get_quote(_report_data: &[u8]) {
    unimplemented!();
}
```

imo, it's a good practice to default to slices as arguments [when possible](https://doc.rust-lang.org/std/vec/struct.Vec.html#slicing). If our intent is to consume the input data and prevent further use, it's fine I guess.

Xynnn007: > are you sure there is an additional allocation required in `get_evidence`? fixed-size arrays shouldn't allocate:

I thought yes. But we can avoid using heap, but stack with a sized slice like your suggestion to decrease consumption.
```rust
let mut data = [0u8, 64];
```

> If our intent is to consume the input data and prevent further use, it's fine I guess.

Yes. I think we consumes it. Overall, the consumption helps avoids extra memory usage. Now, I'm ok with both designs, as it seems that we can only save at most 64 bytes memory.

sameo (approved):

Xynnn007:
Rebased upstream to resolve conflicts.

fitzthum (approved):
LGTM. Hopefully this won't cause any mismatch with the validation in the AS. I don't think it will.

mkulke (approved):

PR #285, opened 2023-07-17
Unify common deps to the same version in Cargo.toml of the worksppace

Xynnn007:
This PR unifies common deps to the same version via `*.workspace = true`.

This will help to maintain the dependencies of the whole workspace without fix them case by case.


jiangliu:
Any introduction about the `workspace` attribute in dependency?
What happens when publishing a crate with dep.workspeace attribute?

Xynnn007:
@jiangliu The `workspace = true` will use the same version as the workspace's `Cargo.toml`

I tried with the command `cargo publish --dry-run -p coco_keyprovider`

and it will do all the steps to publish a crate to `crates.io` except uploading that. This will also geenrate a source code package to be uploaded in `target/package/coco_keyprovider-0.1.0`, whose `Cargo.toml` is like
```toml
...
[dependencies.aes-gcm]
version = "0.10.2"

[dependencies.anyhow]
version = "1.0"

[dependencies.base64]
version = "0.13.0"
...
```

That means all the `XXX.workspace = true` is translated into a version of that package in `crates.io`

Why I use `.workspace = true` is that ocicrypt-rs/image-rs and attestation-agent stack are built in a same final binary. Use same version dep will help to decrease the built binary and easy to maintain.

jiangliu (approved):

Issue #101, opened 2023-01-04
Add unit test case for unencrypted images


jiangliu:
Currently there's a unit test case for encrypted image at `tests/image_decryption.rs`, it would be better to add a test case for unencrypted images too.

portersrc:
Hi, this can be closed (see [PR 287](https://github.com/confidential-containers/guest-components/pull/287))

PR #287, opened 2023-07-19
Add unit test case for unencrypted images

portersrc:
Addresses [issue 101](https://github.com/confidential-containers/guest-components/issues/101)
I added an unencrypted test case.

Also, I removed the previous encrypted test case, which was pulling from here:
    docker.io/xynnn007/busybox:encrypted-uri-key
And I replaced it with an encrypted test case, which pulls from here:
    ghcr.io/confidential-containers/test-container:encrypted
Reasoning: I assume we want to tie this to the CC project on GHCR while we're here and move it off of @Xynnn007's image.


Xynnn007 (approved):
LGTM, thanks. I think it is a work that we should do but always did not.

PR #289, opened 2023-07-21
aa: occlum-attester: update occlum_dcap to a tagged version

mythi:


Xynnn007:
@mythi The error in CI seems to be brought by the new version of occlum. As a former PR does not has dependency `devicemapper-sys` https://github.com/confidential-containers/guest-components/actions/runs/5663723895/job/15345917558

mythi:
> The error in CI seems to be brought by the new version of occlum.

Let me take a look. It could also be because of #270

Xynnn007:
> > The error in CI seems to be brought by the new version of occlum.
> 
> Let me take a look. It could also be because of #270

Oh yes. There are some dep installations https://github.com/confidential-containers/guest-components/pull/270/files#diff-77d1cbd67b7597a3eb86382e96a9a71581274ba88db14b2a31415f856f895667R59-R62 which were not applied in `aa_basic.yml`

Xynnn007 (approved):
lgtm, thanks Mikko!

PR #290, opened 2023-07-24
[RFC] tdm: Trusted Device Manager architecture definition

sameo:
This is the initial commit for the Trusted Device Manager (TDM). It only contains an architecture document, no implementation is provided yet.


sameo:
cc @cclaudio

larrydewey:
This is great work @sameo! Thanks for putting it together! I have some questions about a few of the approaches, but I think you have successfully captured the vision!

tdm/docs/architecture.md
```diff
@@ -0,0 +1,222 @@
...
+
+1. The **TDM Core** runs the main TDM flows and managing confidential guest
+   assignment with the support of the detection, ABI plugins and attestation
+   modules.
+2. The **Device Detection** module traps device discovery, assignment and
+   removal from the confidential guest PCIe emulated bus.
+3. The **Linux CoCo ABI plugins** abstracts the vendor specific CoCo Linux
+   kernel ABIs in order expose a vendor-agnostic internal interface for the TDM
+   core to consume.
+4. The **Device Attestation** is actually a relying party implementation, for
```
jialez0, line 142: If we use local attestation within the guest, how can we set the attestation policy and reference values? Does this mean they need to be built into the rootfs image?

jialez0 (commented):
@sameo This is an exciting field, thanks for putting them together!

tdm/docs/architecture.md
```diff
@@ -0,0 +1,222 @@
...
+#### TDI Security Attributes Check
+
+#### Device Attestation
+
+##### Local Attestation
+
+TDM is a Relying Party.
+TDM to use part of the Attestation Service crates
+
+##### Remote Attestation
```
jialez0, line 222: For remote attestation, I have a design suggestion: Device attestation module of TDM can be responsible for call `get_token` API of Attestation-Agent (We are about to implement this API) and pass the device evidence to AA to perform remote attestation. Then TDM only need to verify the signature of the attestation result token. This aligns with AA's now role in guest components.

tdm/docs/architecture.md
```diff
@@ -0,0 +1,222 @@
...
+
+                                   ┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
+                                   ┃                            Confidential Guest ┃
+                                   ┃           ╔═══════════════════════╗           ┃
+                                   ┃           ║Trusted Device Manager ║           ┃
+                                   ┃           ║                       ║           ┃
+                                   ┃           ║┌─────────────────────┐║           ┃
+                                   ┃           ║│    Relying Party    │║           ┃
+                                   ┃      ┌───▷║│(Attestation Service)│║           ┃
+                                   ┃      │    ║└─────────────────────┘║           ┃
```
fitzthum, line 64: It's a bit confusing to me to have the Relying Party / AS drawn inside the guest here. Should this be the AA, which is a trusted proxy to those services? I assume the idea is that the TDM can get trusted configuration or secrets from the client?

Also, should we differentiate between client provided trusted information and other artifacts that might come from the manufacturer (with signatures)?

tdm/docs/architecture.md
```diff
@@ -0,0 +1,222 @@
...
+                                   ┃      │    ║                       ║           ┃
+                                   ┃      │    ║ ╔════════╗ ╔════════╗ ║           ┃
+                                   ┃      │    ╚═╣  TEE   ╠═╣  TEE   ╠═╝           ┃
+                                   ┃      │      ║ Plugin ║ ║ Plugin ║             ┃
+                                   ┃     udev    ╚════════╝ ╚════════╝             ┃
```
fitzthum, line 69: So the TDM will be modular at the level of the TEE? What about different devices? Will they all be attested with the same code?

fitzthum (commented):
I guess some of my questions are sort of addressed later in the doc

tdm/docs/architecture.md
bodzhang: @sameo , is the Relying Party block in the diagram specific for "local attestation" scheme, where the Trusted Device Manager performs the TEE-IO device Attestation REPORT verification?

tdm/docs/architecture.md
```diff
@@ -0,0 +1,222 @@
...
+   guest TCB, the TDM must use the Linux kernel CoCo ABI to verify the following
+   security attributes from the TSM:
+   * Device link state
+   * TDI state
+   * TDI MMIO mappings
+4. Attest of the assigned device trustworthiness. As part of the overall device
+   verification process, the TDM must run a local or remote attestation of the
+   device attestation evidence. To support local attestation, the TDM
+   practically becomes a Relying Party.
+5. Notify the TSM of it TDI acceptance decision. Here as well, this relies on
```
bodzhang, line 128: Assuming the proposal here is to perform TEE-IO device attestation evidence verification before accepting the device assignment to the confidential guest,  an alternative for at least some of the CoCo usages is to capture the TEE-IO device attestation evidence, supply it to a remote Relying Party, together with the confidential guest attestation evidence (including containers loaded or allowed to be loaded), to be verified at the time of secret provisioning from the remote Relying Party.

tdm/docs/architecture.md
sameo: When running local attestation, I think that the guest can be provisioned with attestation policies and reference values in two ways:

1. Statically: The guest image is built with these pieces of data.
2. Dynamically: The guest could use an attestation token to retrieve the policy and RV from a relying party and run a local attestation service.

The TDM should support both attestation methods (local and remote) through a TDM configuration file.

tdm/docs/architecture.md
sameo: @jialez0 Would the TDM provide the attestation agent with the device attestation evidence? And then let the AA run remote attestation to finally get an attestation results token? If so, that makes sense to me, and I started to add something like that in this document.

tdm/docs/architecture.md
sameo: Sorry the diagram was specific to local attestation. In my mind, the device attestation can either be local or remote and both paths should be handled through the attestation module. For local attestation, this module would expect to be provided (statically or dynamically) with attestation policies and reference values. For remote attestation, the module would use the AA API and use it as an attestation proxy. Note that this creates a dependency between the TDM and the AA (i.e. the TDM could not be used outside of the confidential-containers context without pulling the AA dependency), which is fine with me.

tdm/docs/architecture.md
sameo: Yes, I see the TEE interface as the ABI beween the guest and the TSM. Today we use it mostly for fetching attestation evidences, but I think it should/will be extended to also support the TEE-IO flows.
All assigned devices will be managed by the TSM, on behalf of the untrusted VMM, so the TSM will be the single interface for getting any device attestation evidence, and for guests to accept devices into their TCBs.
Does that make sense?

tdm/docs/architecture.md
sameo: > Assuming the proposal here is to perform TEE-IO device attestation evidence verification before accepting the device assignment to the confidential guest, an alternative for at least some of the CoCo usages is to capture the TEE-IO device attestation evidence, supply it to a remote Relying Party, together with the confidential guest attestation evidence (including containers loaded or allowed to be loaded), to be verified at the time of secret provisioning from the remote Relying Party.

The evidence verification must happen before the guest accepts the assigned device, because once it's accepted the device is trusted and can directly access confidential memory. With that in mind, combining the device and guest attestation evidence together is appealing but has the following drawbacks:
* It can only happen when running guest attestation, i.e. device hotplug is not really supported (Kata relies on device hotplug for direct assignment)
* The remote relying party is able to verify both the guest evidence and the device one. I can see how vendors may want to maintain and provision those services separately because a device attestation evidence is basically orthogonal to a guest one. On the other hand, there may be cases where workload owners may want to support only certain combinations of guest stacks and devices (e.g. I would only want kernel version A with GPU driver version B to work with GPU firmware version C) and we need to allow for supporting that use case as well.

tdm/docs/architecture.md
```diff
@@ -0,0 +1,262 @@
...
+
+1. The **TDM Core** runs the main TDM flows and managing confidential guest
+   assignment with the support of the detection, ABI plugins and attestation
+   modules.
+2. The **Device Detection** module traps device discovery, assignment and
+   removal from the confidential guest PCIe emulated bus.
+3. The **Linux CoCo ABI plugins** abstracts the vendor specific CoCo Linux
+   kernel ABIs in order expose a vendor-agnostic internal interface for the TDM
+   core to consume.
+4. The **Device Attestation** module is responsible for verifying a device
```
sameo, line 129: Note to myself (Thanks @jyao1): The attestation module may have to support vendor specific attestation results, so it may need to be modularized as well.

zvonkok:
@sameo Some PCIe devices need additional functionality beyond attestation. In the case of the GPU e.g. if attestation succeeds, the GPU is still in a "Not Ready" state. A Relying Party has to set the GPU explicitly into a "Ready" state. Are we planning to add vendor-specific "hook" or "callbacks" to the TDM to enable such behavior?

tdm/docs/architecture.md
zvonkok: We may leverage here: https://github.com/confidential-containers/confidential-containers/issues/135 
- Interoperability outside the CoCo flows - for example to reuse the AR computed by CoCo in subsequent TLS exchanges involving the CoCo attester,
- Ease the task of defining and computing authorization policies by relying parties due to the reliance on AR4SI's "trustworthiness vector" to present a normalized view of the evaluation results. An example showing how to use [OPA](https://www.openpolicyagent.org/) to evaluate a EAR in just a few lines of Rego can be found [here](https://play.openpolicyagent.org/p/fBydqNm0G0) (click on the Evaluate button.)
-  Code reuse via existing open-source libraries,
- Normalised and attester-agnostic result format, which allows to accommodate all currently supported formats as well as any other kinds of attester / CC technologies in the future.

sameo: Oh, I'm all for it for sure. There may be cases where we would not talk to KBS but to vendor specific attestation services that return non EAR formatted attestation results.

sameo:
> @sameo Some PCIe devices need additional functionality beyond attestation. In the case of the GPU e.g. if attestation succeeds, the GPU is still in a "Not Ready" state. A Relying Party has to set the GPU explicitly into a "Ready" state. 

How does the RP sets the GPU into this state? Is that documented anywhere?

> Are we planning to add vendor-specific "hook" or "callbacks" to the TDM to enable such behavior?

We may have to, and that's partly why I started this as an RFC, to collect feedback and build a flexible enough design.

jyao1:
@sameo , good idea to have a dedicate TDM!
I will review and comment with more detail info.

tdm/docs/architecture.md
```diff
@@ -0,0 +1,262 @@
...
+
+The TDM core module interacts with three other TDM components:
+
+1. The device detection module. It monitors udev events to detect PCI devices
+   addition and removal from the guest emulated PCI bus. This module calls into
+   the TDM core to notify it about both event types. The TDM core then relies on
+   the following two modules to verify, attest and accept newly detected
+   devices. Only TDM-accepted TDIs are visible and available to the rest of the
+   guest software stack.
+2. The ABI plugins module. This is an abstraction over the Linux CoCo
```
jyao1, line 186: I feel it might bring some confusing if we call "ABI plugins module".
Maybe just name it as its functionality, e.g.
* Device evidence collection module. (cert chain + measurement)
* TDI state management module.
* TDI Report validation module.

tdm/docs/architecture.md
```diff
@@ -0,0 +1,262 @@
...
+## Glossary
+
+| Term | Acronym | Definition |
+|:-----|:-------:|------------|
+| Confidential Computing                 | CoCo   | The protection of data in use by performing computation in a Hardware-based TEE.
+| Confidential VM                        | CVM    | A confidential computing virtual machine. Same as a TVM.
+| Device Interface                       | DI     | TDISP term, Device Interface. Same as a TDI. A DI can be a Virtual Function (VF) or a Physical Function (PF).
+| Device Security Manager                | DSM    | A DSM is a logical entity on a TEE-IO device that enforces the TDISP security policies, attributes and states.
+| Integrity and Data Encryption          | IDE    | Extended PCIe capability for integrity, confidentiality and replay protection of PCIe Transport Layer Packets (TLP).
+| Secure Protocol and Data Model         | SPDM   | A DMTF defined specification for exchanging messages between devices over a variety of transports and physical media. SPDM is used to exchange TDISP and IDE Key Management messages over PCIe DOE mailboxes.
```
jyao1, line 256: SPDM = **Security** Protocol and Data Model

tdm/docs/architecture.md
```diff
@@ -0,0 +1,262 @@
...
+
+##### Remote Attestation
+
+## Glossary
+
+| Term | Acronym | Definition |
+|:-----|:-------:|------------|
+| Confidential Computing                 | CoCo   | The protection of data in use by performing computation in a Hardware-based TEE.
+| Confidential VM                        | CVM    | A confidential computing virtual machine. Same as a TVM.
+| Device Interface                       | DI     | TDISP term, Device Interface. Same as a TDI. A DI can be a Virtual Function (VF) or a Physical Function (PF).
```
jyao1, line 253: What is the difference between "DI" and "TDI"?
If they are same, why we need to define "DI" ?

tdm/docs/architecture.md
```diff
@@ -0,0 +1,262 @@
...
+## Theory of Operations
+
+### TEE-IO Acceptance Flow
+
+#### Detection
+
+#### TDI Security Attributes Check
+
+#### Device Attestation
+
```
jyao1, line 239: We have different attestation model, according to RATS rfc9334 (https://datatracker.ietf.org/doc/rfc9334/).

Do we want to support different patter, such as Passport Model, Background-Check Model?

zvonkok:
@sameo It needs to be done via a specific call to `nvidia-smi`, currently we do local-gpu-attestation with bundled golden-measurement RIMs and if this succeeds we do `nvidia-smi conf-compute -srs 1` (srs = set ready state). 
It is all done inside the VM so that a GPU workload container does not see anything. If attestation fails we do not set it "Ready" and a user will have a GPU in the Container but not in the preferred state. A Container cannot set "Ready" state.

zvonkok:
@sameo @fitzthum @fidencio @stevenhorsman One thing we're missing is, I think crucial, how to provide a channel to the user for important messages that he can react upon. 
What if the Relying Party is the user creating the Pod or the Container, all the other parts can do their job KBS/KBC etc, but the user wants to do e.g. 
```
echo "Ready" > /run/..../whatever 
```
Or read from a dedicated channel that e.g. attetation failed without going through a ton of journald messages
```
cat /run/.../whataver
```

Also other entities could read like operators or side-car containers to provide important messages to Prometheus/Kubelet/3rd party logging on which decision could be made.

tdm/docs/architecture.md
sameo: What about TDI management module?

sameo:
> @sameo It needs to be done via a specific call to `nvidia-smi`, currently we do local-gpu-attestation with bundled golden-measurement RIMs and if this succeeds we do `nvidia-smi conf-compute -srs 1` (srs = set ready state). It is all done inside the VM so that a GPU workload container does not see anything. If attestation fails we do not set it "Ready" and a user will have a GPU in the Container but not in the preferred state. A Container cannot set "Ready" state.

Thanks for the clarification. It seems to me that  `nvidia-smi conf-compute -srs 1` is equivalent to accepting the device into the TVM TCB. Is that a stretch?

Also, do you support remote device attestation? If so, would the flow be similar i.e. the guest VM would need to call `nvidia-smi` based on the retrieved attestation results? What would be the attestation results format, if that path is supported?

tdm/docs/architecture.md
jyao1: Sure, any concrete name is better than "plug in module".

tdm/docs/architecture.md
jyao1: I treat [EAR](https://datatracker.ietf.org/doc/draft-fv-rats-ear/) to be **one** possible option.
But I am not sure if that is the **only** option.

Do we want to allow different format?

tdm/docs/architecture.md
sameo: Thanks, fixed.

tdm/docs/architecture.md
sameo: Removed, thanks.

tdm/docs/architecture.md
jialez0: @sameo Now AA has two interfaces: Get Evidence and Get Token
I think a better design would be:
- In local attestation mode: TDM Attestation Module is responsible for calling AA's Get Evidence API to obtain Evidence and verifying evidence with locally integrated AS.
- In remote attestation mode: TDM Attestation Module is responsible for calling AA's Get Token API to perform remote attestation and ultimately obtain an attestation results token.

In both modes, the collection of Attestation Evidence is completed by the integrated attester drivers in AA. So that, to support device attestation we only need to add a new attester for AA and a new verifier for AS. Just like we support a new type of TEE.

tdm/docs/architecture.md
jialez0: @sameo @fitzthum I think a better approach may be to unify the TEE plugin here with the existing attester plugin in AA, both of which are at the same level as the `attester` clause module. Here, we can import the `attester` crate to obtain evidence or call AA's `get_evidence` API to obtain evidence.

jialez0:
@sameo I have the following design suggestions for the attestation:

In the current design, the more accurate name of TDM's attestation module should be *evidence verification module*, but the complete attestation process should include two stages: obtaining evidence and verifying evidence.
Therefore, in TDM, there should be two corresponding modules:

1. Attester module: for evidence acquisition.
2. Verifier module: for evidence verification.

### TDM Attester Module
This module is responsible for collecting device attestation evidence, and **it will only be used in local attestation mode**.
The TDM Attester Module here should provide two ways to use it:

1. Integration: Directly integrate the `attester` library crate to obtain device attestation evidence. The evidence collection logic plugin that is specific to TEE should have the same status as the plugins of other CPU TEEs, implemented as a plugin of attester crate.
2. Call AA: Call the API of the AA gRPC/ttRPC service to obtain device attestation.

### TDM Verifier Module
This module is responsible for verifying device attestation evidence and has two modes:
1. Local attestation: In this mode, the TDM Verifier module actually becomes a Reply Party, which integrates `attestation-service` crate and RVPS for evidence verification.
2. Remote attestation: In this mode, TDM no longer needs to collect device attestation evidence. Instead, the TDM Verifier module directly calls the API of the AA gRPC/ttRPC service, where AA is responsible for collecting device attestation evidence and sending it to the remote AS for verification. After verification, a signed Attestation Results Token will be returned to TDM.

What do you think?

tdm/docs/architecture.md
```diff
@@ -0,0 +1,261 @@
...
+                                   ┃      │      ║ Plugin ║ ║ Plugin ║             ┃
+                                   ┃     udev    ╚════════╝ ╚════════╝             ┃
+                                   ┃ (PCIe device                △                 ┃
+                                   ┃  detection)                 │                 ┃
+                                   ┃      │               Device Evidence          ┃
+                                   ┃      │                Accept/Reject           ┃
+                                   ┃      │                      │                 ┃
+                                   ┃      │                      ▽                 ┃
+                                   ┃      │                  ┌───────┐             ┃
+                                   ┃      │  ┌───────────────┤TEE ABI├────┐        ┃
```
djbw, line 61: I accept that there are cases where userspace needs an override for the evidence acceptance flow especially in these early days where there might not be unification across devices on evidence formats, but I want to advocate for and end goal where the kernel only needs a certificate chain to be able to make acceptance decisions. I am reluctant to sign-off on a design direction that makes the kernel beholden to userspace for device usage as I worry about scenarios like error recovery and power management where userspace upcalls are not feasible.

tdm/docs/architecture.md
bodzhang: I think an unverified device can be accepted by the TEE-protected guest (itself also unverified until remote attestation), as long as the device identity/configuration can be captured in Attestation Evidence as "tamper-proof" against potential attack from the unverified device itself. 

This requirement rules out an Attestation Evidence design where the SW running inside the CPU-TEE-protected guest simply appends the TEE-IO device attestation evidence stored in guest memory to the CPU-TEE-produced confidential guest attestation evidence. A malicious device once attached would be able to modify the memory content. But a design that captures the TEE-IO device attestation evidence in CPU-TEE-produced attestation evidence, for example, in TDX RTMRs, can make sure Relying Party detects untrusted TEE-IO device attached to the environment during Remote Attestation. Designs not relying on RTMR-type capability (as some CPU-TEEs don't have RTMR-type capability) requires more exploration. Maybe SEV-SNP HostData or TDX MRConfigID can be used to capture the expected device evidence at the guest creation time. 

With regard to Relying Party verifying both the guest evidence and device evidence, I agree that the flexibility of maintaining separate service/infrastructure for CPU and Devices verification is valuable. Combining the device and guest attestation evidence together does not force any specific implementation of monolithic CPU and Device verification service. A Relying Party verifying a combined device and guest evidence can consult CPU and Device verification service separately.

tdm/docs/architecture.md
sameo: Fixed then.

tdm/docs/architecture.md
sameo: > I think an unverified device can be accepted by the TEE-protected guest (itself also unverified until remote attestation), as long as the device identity/configuration can be captured in Attestation Evidence as "tamper-proof" against potential attack from the unverified device itself.
> 
> This requirement rules out an Attestation Evidence design where the SW running inside the CPU-TEE-protected guest simply appends the TEE-IO device attestation evidence stored in guest memory to the CPU-TEE-produced confidential guest attestation evidence. A malicious device once attached would be able to modify the memory content. 

Not only the malicious device could tamper with confidential memory, but the host VMM could use an attached device to access and modify confidential memory as well. Keep in mind that the host VMM emulates the guest PCI bus, and thus controls the MMIO mappings between the guest physical addresses and the host physical ones. TDISP, through the TDI interface report, allows guests to verify that those mappings are correctly set by the TSM, i.e. that MMIO GPAs do map to the actual device, not to other devices or to the host controlled memory.
Moreover, the device directly accesses the TVM private memory through IOMMU controlled translation tables, and it can do so at any point in time, i.e. not only when the TVM is actually running.
This leads to a bunch of threat vectors that are different from plain confidential computing ones, as assigning and accepting a device into a TVM allows for untrusted host software components to access private memory indirectly (through the device MMIO and DMA). FWIW We tried to capture those threats here: https://github.com/riscv-non-isa/riscv-ap-tee-io/blob/main/specification/security_model.adoc#threats

Device attestation, either captured through CPU attestation evidence or run as a separate attestation flow, can not mitigate those threats alone. A device attestation evidence result will tell the guest about a device trustworthiness, but it won't tell about the host emulated PCI mappings and DMA translation tables being valid or not. In other words, your relying party could very well show that a perfectly trustworthy device is attached to your TVM, but the host software could have programmed the actual IO (MMIO and/or DMA) in such a way that the guest believes its talking to this trusted device while part or the whole IO traffic is captured by the host. This is a fundamental difference between guest and device attestation and the reason why a guest must verify the device before accepting it. And by verifying we mean checking with the TSM about I/O mappings, the state of the physical link, the device identity and finally the device trustworthiness (i.e. the actual attestation). Skipping any of these steps could allow for the guest or the device itself to trick the guest and leak confidential data. And they should all be completed before the device is accepted (i.e. the IO mappings are enabled) and the guest can use it.

This is a fundamental difference between guest and device attestation. The former can be done asynchronously, before secrets are delivered to the guest, and is sufficient to guarantee that confidential data is protected. The latter only concludes the device verification process and is not by itself sufficient for guaranteeing confidential data protection against malicious devices or host components that could then access it even while the guest is not running. 

> But a design that captures the TEE-IO device attestation evidence in CPU-TEE-produced attestation evidence, for example, in TDX RTMRs, 

I'm not sure I understand how this is different than appending to the CPU attestation evidence? The device attestation evidence is integrity protected as it's signed by a device manufacturer endorsed key. Maybe I'm not getting your point, apologies if that's the case.

sameo:
> @sameo I have the following design suggestions for the attestation:
> 
> In the current design, the more accurate name of TDM's attestation module should be _evidence verification module_, but the complete attestation process should include two stages: obtaining evidence and verifying evidence. Therefore, in TDM, there should be two corresponding modules:
> 
> 1. Attester module: for evidence acquisition.
> 2. Verifier module: for evidence verification.
> 
> ### TDM Attester Module
> This module is responsible for collecting device attestation evidence, and **it will only be used in local attestation mode**. The TDM Attester Module here should provide two ways to use it:
> 
> 1. Integration: Directly integrate the [`attester` library crate](https://github.com/confidential-containers/guest-components/tree/main/attestation-agent/attester) to obtain device attestation evidence. The evidence collection logic plugin that is specific to device TEE should have the same status as the plugins of other CPU TEEs, implemented as a plugin of attester crate.
> 2. Call AA: Call the API of the AA gRPC/ttRPC service to obtain device attestation evidence.
> 
> ### TDM Verifier Module
> This module is responsible for verifying device attestation evidence and has two modes:
> 
> 1. Local attestation: In this mode, the TDM Verifier module actually becomes a Reply Party, which integrates `attestation-service` crate and RVPS for evidence verification.
> 2. Remote attestation: In this mode, TDM no longer needs to collect device attestation evidence. Instead, the TDM Verifier module directly calls the API of the AA gRPC/ttRPC service, where AA is responsible for collecting device attestation evidence and sending it to the remote AS for verification. After verification, a signed Attestation Results Token will be returned to TDM.
> 
> What do you think?

I think it makes a lot of sense. Let me try to summarize how the TDM attestation module should behave (Regardless of it being split in 2 separate modules):

* If configured to run **local attestation**
   1. Either call into the AA gRPC API or directly use the `attester` crate to collect the device attestation evidence.
   2. Use the `attestation-service` crate to verify the collected attestation evidence.
* If configured to run **remote attestation**
   1. Use the AA gRPC API to get an attestation result token.

If that's correct, this simplifies things quite a bit. I'd like for the TDM to not depend on an attestation agent to be running as I think this would make it usable outside of the confidential-containers context. If we want to fully support that, we'd need to be able to use the AA as a standalone crate. I think we're not too far from being able to do that.

jialez0:
> If that's correct, this simplifies things quite a bit. I'd like for the TDM to not depend on an attestation agent to be running as I think this would make it usable outside of the confidential-containers context. If we want to fully support that, we'd need to be able to use the AA as a standalone crate. I think we're not too far from being able to do that.

@sameo This is exactly. As you mentioned, our AA can now provide both gRPC API for remote calls and be integrated as a standalone crate. Therefore, whether it is in local or remote mode, we can freely choose the gRPC or native integration mode of AA according to the configuration.

sameo:
> > If that's correct, this simplifies things quite a bit. I'd like for the TDM to not depend on an attestation agent to be running as I think this would make it usable outside of the confidential-containers context. If we want to fully support that, we'd need to be able to use the AA as a standalone crate. I think we're not too far from being able to do that.
> 
> @sameo This is exactly. As you mentioned, our AA can now provide both gRPC API for remote calls and be integrated as a standalone crate. Therefore, whether it is in local or remote mode, we can freely choose the gRPC or native integration mode of AA according to the configuration.

Thanks @jialez0 . I'll update the diagrams accordingly.

tdm/docs/architecture.md
bodzhang: By capturing the TEE-IO device attestation evidence in TDX RTMRs (for example, extending a RTMR with the measurement of the device attestation evidence) _**before**_ accepting the device in the guest VM, the guest makes sure the device, if malicious or compromised, cannot alter the attestation evidence without detection (by Relying Party). The device attestation evidence can be appended to the TEE-CPU produced guest attestation evidence and bound to the RTMR value. If the attack from a malicious or compromised device alters the guest memory (code/data) to spoof a legit Device attestation evidence, it won't match the RTMR value and will be detect when the Relying Party verifies the guest and device attestation evidence. If the TEE-IO attestation evidence is instead simply stored in guest memory, but not bound with "tamper-proof" RTMRs, the potential alternation mentioned above won't be detected by the Relying Party. 

Before the guest accepts the device, it should perform many checks using static code logic, for example, IO mapping must 
match TSM provided info. Capturing the TSM provided info, I/O mapping info, etc as part of device attestation evidence to be verified by Relying Party later only moves the static checking logic to the Relying Party, without the flexibility benefit. For checks using "dynamic" reference data, such as device cert revocation list, expected firmware measurement and configuration values, etc., capturing the relevant info provided by TSM in the device attestation evidence and performing   the check by the Relying Party during remote attestation allows much more flexibility on the mechanism to acquire/update the "dynamic" reference data used in verification, compared with hardcoding the reference data or provisioning the reference data to the guest to perform the "local" check.

Sorry for not including more details in my previous post.

sameo:
@jialez0 @djbw @larrydewey @zvonkok @jyao1 I updated the PR to move to an hybrid approach, where the kernel would rely on the TDM for the device attestation part.

The TDM <-> kernel interaction is now synchronous, and the TDI ownership model is clearer: The kernel owns the TDI, and optionally decides to call into the TDM for attesting to it. This would happen after the kernel detects the TDI, and before it probes it.

KB5201314:
@sameo The new diagrams are great.

There seems to be a problem in the flow.

> I think it makes a lot of sense. Let me try to summarize how the TDM attestation module should behave (Regardless of it being split in 2 separate modules):
> 
> * If configured to run **local attestation**
>   
>   1. Either call into the AA gRPC API or directly use the `attester` crate to collect the device attestation evidence.
>   2. Use the `attestation-service` crate to verify the collected attestation evidence.
> * If configured to run **remote attestation**
>   
>   1. Use the AA gRPC API to get an attestation result token.
> 

In the previous discussion, the **attester crate** is responsible for collecting the device attestation evidence.

> 7. The **guest kernel** requests the previously registered TDM to attest the
>   assigned device. The device attestation evidence is added to the attestation
>   request message sent by the guest kernel to the TDM.

And in the new "synchronous flow", the device attestation evidence is sent by guest kernel to the TDM (I don't know the exact way, I'm assuming it's via polling an ioctl fd). This looks confusing since it seems to imply that device evidence will be passed on to the **attester crate**, rather than letting the **attester crate** get it from the guest kernel.

Please correct me if there are any problems in my understanding. Thanks.

PR #294, opened 2023-07-27
image-rs: refine implementation of dm-verity

jiangliu:
Refine implementation of dm-verity.


arronwy (approved):
Thanks @jiangliu LGTM!

PR #299, opened 2023-07-27
Made Attester trait's get_evidence() async

### LINKED ISSUE #298: attestation-agent: Attester trait should be async
# Attester trait should be async

At the moment the `Attester` trait in the `attester` package is not async:

https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/attester/src/lib.rs#L60

## Motivation

TEE-specific Implementations of the trait's `get_evidence()` fn usually perform some form of IO, which we currently implement using sync calls to the low-level platform APIs. However, there are TEE platforms which, as part of their evidence retrieval routine, involve roundtrips to external resources like a CSP's IMDS/Metadata service (Azure's TDX preview being an example).

It's possible to work around this, but possibly not pretty (nesting async runtimes, resorting to special libraries, etc). Since we expect async work to be performed, the trait's function should be marked as async, I think.   
 
## Impact

The consumer of `Atttester` trait objects is currently `KbsProtocolWrapper#attestation()`, which is already `async` so the impact of that change should be confined to the `kbs_protocol` and `attester` packages.

mkulke:
fixes #298

### DIFF
--- a/attestation-agent/attester/Cargo.toml
+++ b/attestation-agent/attester/Cargo.toml
@@ -7,6 +7,7 @@ edition = "2021"
 
 [dependencies]
 anyhow.workspace = true
+async-trait.workspace = true
 az-snp-vtpm = { git = "https://github.com/kinvolk/azure-cvm-tooling", rev = "2c2e411", default-features = false, features = ["attester"], optional = true }
 base64.workspace = true
 log.workspace = true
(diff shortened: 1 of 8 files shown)


Xynnn007 (approved):
lgtm

Xynnn007:
@mkulke the error is caused by the pr.

Xynnn007:
hi @mkulke plz rebase this

PR #301, opened 2023-07-28
image pull tests: replace image ref

Xynnn007:
We use CoCo's ghcr.io repo instead of personal repo for the test images.


arronwy (approved):
Thanks a lot @Xynnn007 , this will make our CI more stable.

We will retrigger the CI after new test image is changed from private to public.

Xynnn007:
Let me retrigger the CI when all the images are made public

PR #306, opened 2023-08-01
image-rs: Fix the flaky CI with assert_retry

arronwy:
The pull operation will interactive with remote registry which may have network or rate limits issue, add the retry operation for related to avoid trigger the rebuild for CI with these issues.

### DIFF
--- a/image-rs/libs/test-utils/src/lib.rs
+++ b/image-rs/libs/test-utils/src/lib.rs
@@ -26,3 +26,37 @@ macro_rules! assert_result {
         }
     };
 }
+
+// Parameters:
+//
+// $limit: the maximum number of retries
+// $delay: the delay between retries (in seconds)
+// $obj: the object (struct instance)
+// $method: an identifier representing the method name
+// $(,$arg:expr)*: zero or more expressions representing the method arguments
+#[macro_export]
+macro_rules! assert_retry {
+    ($limit:expr, $delay:expr, $obj:expr, $method:ident $(,$arg:expr)*) => {{
+        let mut last_err = None;
+        for i in 0..$limit {
+            if let Err(e) = $obj.$method($($arg),*).await {
+                println!(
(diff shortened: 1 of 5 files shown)


Xynnn007 (approved):
lgtm

arronwy:
I retriggered the CI three times and are passed:
image

image-rs/libs/test-utils/src/lib.rs
```diff
@@ -26,3 +26,37 @@ macro_rules! assert_result {
         }
     };
 }
+
+// Parameters:
+//
+// $limit: the maximum number of retries
+// $deplay: the delay between retries (in seconds)
```
fitzthum, line 33: nit: `$deplay` -> `$delay`

fitzthum (commented):
I guess we want to test with multiple registries, so we need some way to handle rate limiting, but I am a bit worried that this approach could mask other intermittent issues that are not related to rate limits. Is there a way to check more specifically the cause of the failure before we retry?

image-rs/libs/test-utils/src/lib.rs
arronwy: Thanks, fixed now :)

arronwy:
> I guess we want to test with multiple registries, so we need some way to handle rate limiting, but I am a bit worried that this approach could mask other intermittent issues that are not related to rate limits. Is there a way to check more specifically the cause of the failure before we retry?

current `image-rs` use `oci-distribution` crate to pull image which will use `reqwest` as HTTP client, by default `reqwest` don't support retry policy and the related test case may fragile when multiple CI piplelines request the same image, the test case in `oci-distribution` also do similar operation to avoid broke the CI.


Add retry policy in `oci-distribution` maybe one of potential options, currently we depend on the Kubernetes image pull retry policy to handle this issue.

fitzthum (approved):

PR #307, opened 2023-08-01
kbs_protocol: use rusttls when rust-crypto feature is enabled

mythi:
reqwest uses the platform's native TLS stack by default, also when rusttls is enabled.

Force reqwest Client to use rustls TLS stack when rust-crypto feature is enabled. This is what users most likely expect to get.


jialez0:
cc @Xynnn007 . This should also be added to your PR #304 .

PR #308, opened 2023-08-02
image-rs: change fallback kbs_uri from localhost to http://localhost

mkulke:
Eventually `localhost` will fail in this constructor, because it's not a valid URL.

https://github.com/confidential-containers/guest-components/blob/3eb25b1df25418f99140d11f9855e9ab5d0b0a13/attestation-agent/kbc/src/cc_kbc/mod.rs#L55-L57


Xynnn007 (approved):
great

arronwy (approved):
The nightly CI failure is not related to this PR and already fixed in https://github.com/confidential-containers/guest-components/commit/94bdead6a835bcd08181e9abadea687c635161a4

Issue #296, opened 2023-07-27
Roadmap to develop Confidential DataHub


Xynnn007:
As proposed in https://github.com/confidential-containers/confidential-containers/issues/136, confidential datahub helps to access different resources/data. This roadmap aims to provide a roadmap to track on this. There are two main goals:
1. Bring in confidential datahub as a guest service
2. Refactor attestation-agent so that only attestation-related APIs are provided

As attestation-agent now has some important APIs like `get_resource()` etc, we will try not to break current mechanism in such a way:
1. Add KBS-client related features to CDH firstly so that the first version of CDH can replace the attestation-agent and will not break current system.
2. Do the refactoring and add more features.

#### Stage 1

- [ ] CDH repo
- [ ] KMS trait definition
- [ ] Support KBS Client to access resources from KBS
- [ ] Support GetResource API
- [ ] Support KeyProvider(Image Decryption) API
- [ ] Support Aliyun KMS client as an example implementation for KMS
- [ ] Refactor Attestation-Agent
- [ ] https://github.com/confidential-containers/guest-components/issues/295

#### Stage 2
- [ ] Tool to generate sealed secret
- [ ] Tool to sign image & sealed secret

The roadmap & task list can be updated then.

arronwy:
@LindaYu17 @Lu-Biao

Issue #295, opened 2023-07-27
Roadmap to support Sealed Secret


Xynnn007:
Sealed Secret was proposed in https://github.com/confidential-containers/confidential-containers/issues/136. This issue is used to track the development of sealed secret.

CDH:
- [x] Define the format of sealed secret
- [x] Impl the UnsealSecret API in Confidential DataHub
- [x] Tool to generate sealed secret

Kata runtime & agent:
- [x] Sealed Secret use as env & unit test 
- [x] Sealed Secret use as volume file & unit test 
- [x] Call CDH to unseal with Ttrpc client
- [x] Kata E2E Test

Kata Rootfs 
- [x] Makefile for CDH
- [ ] put CDH to guest image as systemd service
- [ ] put CDH to guest image as initrd script
- [x] Startup order

arronwy:
@LindaYu17 @Lu-Biao

arronwy:
[CC | add sealed secret support in Kata](https://github.com/kata-containers/kata-containers/pull/7570#top) 
ci: Add test cases for CoCo sealed secret feature  https://github.com/kata-containers/tests/pull/5754

arronwy:
Current CDH & AA is launched by Kata-agent to support runtime with rootfs and initrd with one code base.

PR #309, opened 2023-08-03
Confidential DataHub Part 2: KMS support and unseal secret with KMS

Xynnn007:
This PR:
- Defines the KMS API
- Impl Aliyun KMS as an example
- Given a more detailed `Annotation` for plugin KMS/Sealed Secret
- Impl Unseal secret API
- Add seacret sealing tool

Some work are left to be done:
- Define Vault API and give an example impl
- Verify JWS signature before unseal the secret

related to : 
- https://github.com/confidential-containers/guest-components/issues/296
- https://github.com/confidential-containers/guest-components/issues/295


Xynnn007:
Having tidied the commits. Let me mark this as ready for review.

confidential-datahub/kms/src/api.rs
```diff
@@ -0,0 +1,74 @@
...
+//!
+//! ## APIs
+//! - `Decryptor`: KMS's decrypt API.
+//! - `Encryptor`: KMS's encrypt API.
+//! - `Getter`: Vault's get secret API.
+//! - `Setter`: Vault's set secret API.
+//!
+//! The rationality to distinguish these four different traits:
+//! - `Decryptor` and `Getter` are used in-pod, while `Encryptor` and `Setter`
+//! are used userside. They do not need to be implemented by a same object.
```
fitzthum, line 30: Do we plan to implement a secret sealing tool somewhere for the client? Will it share any of this code?

confidential-datahub/kms/src/plugins/mod.rs
```diff
@@ -0,0 +1,22 @@
...
+
+/// Create a new [`Decryptor`] by given provider name and [`ProviderSetting`]
+pub async fn new_decryptor(
+    provider: &str,
+    provider_setting: ProviderSetting,
+) -> Result<Box<dyn Decryptor>> {
+    match provider {
+        "ali" => Ok(Box::new(
+            aliyun::AliyunKmsClient::from_provider_setting(&provider_setting).await?,
+        ) as Box<dyn Decryptor>),
```
fitzthum, line 19: Do we want a feature flag around this condition?

confidential-datahub/kms/src/api.rs
```diff
@@ -0,0 +1,74 @@
...
+//!
+//! To establish a connection between the client and the KMS/Vault server, two
+//! kinds of parameters are required:
+//! - Public parameters: like Region Id of the KMS/Vault, Instance Id of the KMS,
+//! etc. They are not confidential and can be seen by anyone. [`ProviderSetting`]
+//! is a json object. It is to include all the public parameters. The
+//! hashmap style makes it flexible for different kinds of KMSes/Vaults. There
+//! should be a function that takes a [`ProviderSetting`] as a input parameter
+//! and returns a client instance.
+//! - Private parameters: like the credential to access (e.g. access key).
```
fitzthum, line 16: What about the URI of a KMS? That isn't private, but it should be provided via new, right?

confidential-datahub/kms/src/plugins/aliyun/client.rs
```diff
@@ -0,0 +1,355 @@
...
+    fn read_kms_instance_cert(cert_pem: &[u8]) -> Result<Certificate> {
+        let kms_instance_ca_cert = Certificate::from_pem(cert_pem)
+            .map_err(|e| Error::AliyunKmsError(format!("read kms instance ca cert failed: {e}")))?;
+        Ok(kms_instance_ca_cert)
+    }
+
+    pub fn new(
+        client_key: &str,
+        kms_instance_id: &str,
+        pswd: &str,
```
fitzthum, line 50: nit: `pswd` -> `pwd`

confidential-datahub/kms/src/plugins/aliyun/client.rs
```diff
@@ -0,0 +1,355 @@
...
+
+        Ok(Self {
+            http_client,
+            kms_instance_id: kms_instance_id.to_string(),
+            credential,
+            endpoint,
+        })
+    }
+
+    /// This new function is used by a in-pod client. The side-effect is to read the
```
fitzthum, line 71: What do you mean by an "in-pod client?"

confidential-datahub/kms/src/plugins/aliyun/client.rs
```diff
@@ -0,0 +1,355 @@
...
+        Ok((encrypt_response.ciphertext_blob, operation_context))
+    }
+}
+
+#[async_trait]
+impl Decryptor for AliyunKmsClient {
+    async fn decrypt(
+        &mut self,
+        ciphertext: &[u8],
+        keyid: &str,
```
fitzthum, line 181: nit: `keyid` -> `key_id` (As it is in the above function)

confidential-datahub/kms/src/error.rs
```diff
@@ -0,0 +1,17 @@
...
+//
+
+use thiserror::Error;
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+#[derive(Error, Debug)]
+pub enum Error {
+    #[error("Aliyun KMS error: {0}")]
+    AliyunKmsError(String),
```
fitzthum, line 13: Should this be behind feature flag? I guess it doesn't really matter.

confidential-datahub/kms/src/plugins/aliyun/example_credential/PrivateKmsCA_kst-shh64702cf2jvc_____.pem
fitzthum, line 1: Can we get some info about the certificates in the README?

fitzthum (commented):
Nice work. Review covers 70 percent of the code. 

Main notes: 
- Not sure about the additional annotations parameters. Parameter names seem unclear and I'm not sure how extensible it is.
- Not sure why we are getting the private parameters from the FS. Isn't the idea of the CDH to get this stuff from the KBS?
- Could use more specific documentation about how to prepare secrets and what the certificates are for.
- Do we have a client for encrypting the secrets?

Overall though code looks pretty good. Should be able to merge pretty soon I think.

confidential-datahub/hub/src/hub.rs
```diff
@@ -0,0 +1,49 @@
...
+
+#[async_trait]
+impl DataHub for Hub {
+    async fn unseal_secret(&self, secret: Vec<u8>) -> Result<Vec<u8>> {
+        // TODO: verify the jws signature using the key specified by `kid`
+        // in header. Here we directly get the JWS body
+        let body = secret
+            .split(|c| *c == b'.')
+            .nth(1)
+            .ok_or_else(|| Error::UnsealSecret("illegal input sealed secret (not a JWS)".into()))?;
```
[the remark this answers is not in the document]
Xynnn007: I think the hub receives the part after prefix `sealed.`. I will sync with Arron offline about this to make the sender and the receiver here are consistent

confidential-datahub/docs/SEALED_SECRET.md
```diff
@@ -41,27 +41,35 @@ The format of the KMS type Sealed Secret is
...
 	"annotations": {
-		"region_id": "cn-hangzhou",
-		"endpoint": "xxx",
-		"signature_method": "xxx"
+		"operation_context": {
+			"algorithm": "xxx"
+		},
+		"provider_setting": {
+			"kms_instance_id": "xxx"
+		}
```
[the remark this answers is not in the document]
Xynnn007: Ah, yes. At first I want to put all the information about the kms information into the annotation field.

Soon when we did the coding things, we found that there are two different parts of information we need:
Type 1. Parameters to init the KMS client.
Type 2. Parameters of this enc/decryption

The logic is as following:
1. user's client tool side:
a. use Type 1 parameter as the public init parameters of the client
b. the user provides private parameters (like credentials) of the client
c. create the client object
d. use the client to do encryption operation. The encryption will return the extra parameters for the encryption (`operation_context`), like `iv`. (some KMS will generate `iv` server side, and some we need to manually generate in the impl of `encryption` randomly and send to KMS)
e. the client can export its information (`provider_settings`), which will help the cdh to init a client.
f. wrap `operation_context` and `provider_settings` into the annotations and generate the sealed secret

2. in-pod (CDH)
a. when getting a sealed secret, it parses the `operation_context` and `provider_settings` in annotations.
b. use the `provider_settings` as the public init parameters of the client
c. read the hardcoded path to find the private parameters (like credentials) of the client. (Not like the userside where we can directly provide the private parameters directly in the commandline para, in-pod we should read from a safe place. Here we choose the in-memory fs path. The question is how we put these files in that? It can be fetches from KBS. In the [architecture ](https://github.com/confidential-containers/guest-components/blob/main/confidential-datahub/docs/Architecture.svg), we have an `auth layer`. The components in this layer is to get the credentials that is used by ResourceClients like KMS. At first, we can impl KBS to be this one -- KBS client fetch the private parameters from KBS and put them to the specified in-memory fs path)
d. create the client object
e. use the client object to do decryption with the `operation_context` in `annotations`

So, we are actually classifying the fields in annotations for logic decoupling, which is potentially helpful for use cases beyond coco (yes, `kms` it self can be used as a crate). A `serde_json::Map<String, Value>` leaves the extensibility for the two different kinds of parameters.

About the name:
`encryption_context` is more accurate for KMS. But when taking `vault` into consideration, it does not have "encryption". So to make it more generic, the name `operation_context` is chosen to refer to either the context of a KMS enc/decryption or a vault setting/getting operation.

`provider_settings` is the same reason.

confidential-datahub/docs/SEALED_SECRET.md
```diff
@@ -41,27 +41,35 @@ The format of the KMS type Sealed Secret is
...
 	}
 }
 ```
 Here,
-- `version`: indicates the format version of the Sealed Secret. Currently `0.1.0`.
-- `type`: MUST be `envelope`, indicating this is a Envelope type Sealed Secret
-- `provider`: indicates the provider of the __sealing key__. This field determines
+- `version`: **REQUIRED**. indicates the format version of the Sealed Secret. Currently `0.1.0`.
+- `type`: **REQUIRED**. MUST be `envelope`, indicating this is a Envelope type Sealed Secret
+- `provider`: **REQUIRED**. indicates the provider of the __sealing key__. This field determines
```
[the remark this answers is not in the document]
Xynnn007: Nice idea. let me add a commit about aliyun's for this

confidential-datahub/kms/src/api.rs
```diff
@@ -0,0 +1,74 @@
...
+//! These parameters should be captured inside the logic of `new()` rather
+//! than the input parameter. it is strongly recommended that private parameters
+//! be read from the encrypted filesystem, e.g. `/run/*` which is in TEE's
```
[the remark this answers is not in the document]
Xynnn007: please see https://github.com/confidential-containers/guest-components/pull/309#discussion_r1284924833

As CDH is also a component inside pod, we finally should get the private parameters from somewhere else. This sentence is confusing as it does not make it clear "who puts the files in the encrypted fs". It's by "auth layer" of CDH

confidential-datahub/kms/src/api.rs
Xynnn007: Yes. That tool will leverage `Encrypter` trait. We will put that in later. The secret sealing tool is some higher level codes which calls the `kms` crate.

For aliyun, it will also use the `AliyunKmsClient`.

confidential-datahub/kms/src/plugins/mod.rs
Xynnn007: Oh, yes! Thanks for pointing out!

confidential-datahub/kms/src/api.rs
Xynnn007: Yes! The URI of a KMS is a "public parameter", so it should be put in `provider_settings`. Taking aliyun as example, the URL of the KMS is derived by the `kms_instance_id`, so we do not directly have URI of a KMS to `new()`, but derive it inside `new()`

confidential-datahub/kms/src/plugins/aliyun/client.rs
Xynnn007: ok, it is actually password, let me rename it then

confidential-datahub/kms/src/plugins/aliyun/client.rs
Xynnn007: The client definitions in `kms` crate are generic things. They can be used by a client tool to generate sealed secret (user side). Also they can be used by CDH to unseal secrets (so-called in-pod client, as it works in the pod)

confidential-datahub/kms/src/error.rs
Xynnn007: makes sense, let me fix it then

confidential-datahub/kms/src/plugins/aliyun/example_credential/PrivateKmsCA_kst-shh64702cf2jvc_____.pem
Xynnn007: Ok, I will add a separate README for aliyun kms. As this cert file is only for aliyun

Xynnn007:
@fitzthum Thanks for the very careful review! I really need the these review views to make the code/doc more clear and easy to understand to more people.

confidential-datahub/docs/SEALED_SECRET.md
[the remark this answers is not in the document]
Xynnn007: Ok. Actually I'm not sure about the name, as docker has a `dockerhub` without `-` ...  I'm ok to change the name, and maybe we can do that in a separate PR?

Xynnn007:
Rebased.

Xynnn007:
Fixed an error cause by auto-merge

confidential-datahub/kms/src/plugins/aliyun/client.rs
fitzthum: When you say "in-pod" do you mean in-guest? To me in-pod sounds like it is running in a container.

confidential-datahub/docs/SEALED_SECRET.md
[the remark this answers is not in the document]
fitzthum: If we have `-` after `confidential`, we should also have one after `data`.

confidential-datahub/kms/src/plugins/aliyun/client.rs
```diff
@@ -0,0 +1,355 @@
...
+}
+
+pub struct AliyunKmsClient {
+    http_client: reqwest::Client,
+    credential: Credential,
+    kms_instance_id: String,
+    endpoint: String,
+}
+
+const IN_POD_DEFAULT_KEY_PATH: &str = "/run/kms-credential/aliyun";
```
portersrc, line 38: Maaaybe there's a better place for this string (if not now , then later)? Will there be more plugins/clients in the future that reuse `/run/kms-credential`?

confidential-datahub/kms/src/plugins/aliyun/protobuf/dkms_api.proto
```diff
@@ -0,0 +1,236 @@
...
+}
+
+message HashRequest {
+  bytes Message = 1;
+  string Algorithm = 2;
+}
+
+message HashResponse {
+  bytes  Digest = 1;
+  string RequestId = 3;
```
portersrc, line 123: Maybe you wanted `string RequestId = 2;` here

confidential-datahub/secret/src/secret/mod.rs
```diff
@@ -24,12 +26,20 @@ pub struct Secret {
     pub r#type: SecretContent,
 }
 
+impl Secret {
+    pub async fn unseal(&self) -> Result<Vec<u8>> {
+        match &self.r#type {
+            SecretContent::Envelope(env) => env.unseal().await,
+            SecretContent::Vault(_) => todo!(),
+        }
```
portersrc, line 34: What about checking the `version` here? I think it's a good field to have. It looks to me like `unseal_secret` will deserialize everything, create the `Secret` struct, and then invoke `unseal`. But you could end up with all the right fields but a subtle difference due to version change, in which case you would want to catch it.

confidential-datahub/hub/src/hub.rs
```diff
@@ -0,0 +1,49 @@
...
+
+use crate::{DataHub, Error, Result};
+
+pub struct Hub {}
+
+#[async_trait]
+impl DataHub for Hub {
+    async fn unseal_secret(&self, secret: Vec<u8>) -> Result<Vec<u8>> {
+        // TODO: verify the jws signature using the key specified by `kid`
+        // in header. Here we directly get the JWS body
```
portersrc, line 18: This TODO makes sense, but is this a typo? Should it be, "Here we directly get the JWT body"?

portersrc (commented):
Hey @Xynnn007 nice work. I'm not sure I follow all of it just yet, but I tried to leave some helpful comments.

portersrc:
Follow-up question which I'm not sure is appropriate to this PR, but where in the code (if anywhere yet) is the following enforced: Attestation must happen before unsealing secrets.

confidential-datahub/docs/SEALED_SECRET.md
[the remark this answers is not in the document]
fitzthum: `operation` is pretty vague. What about calling it `secret_settings` or even just `settings`?

confidential-datahub/docs/SEALED_SECRET.md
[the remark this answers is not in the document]
Xynnn007: Ok, let me add a separate commit to do this.

confidential-datahub/hub/src/hub.rs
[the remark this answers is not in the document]
Xynnn007: see https://github.com/kata-containers/kata-containers/pull/7570/files#diff-f87c1c212bba3abd0b07bafe69d5b01ad00dfe2c01bed585e2a75fcbb6442b89R793-R795

confidential-datahub/kms/src/plugins/aliyun/client.rs
Xynnn007: Oh. Ok, let me use `in-guest` then. Fixed this in the code notes.

confidential-datahub/kms/src/plugins/aliyun/client.rs
Xynnn007: Yes. Thanks for pointing out this. My idea is all KMSes should be placed in a common parent directory. To follow up the ttrpc sock used by attestation-agent https://github.com/kata-containers/kata-containers/blob/CCv0/src/agent/src/image_rpc.rs#L31, I think `/run/confidential-containers/cdh/kms-credential` might be a good place?

different kms plugins can use different sub directories.

confidential-datahub/kms/src/plugins/aliyun/protobuf/dkms_api.proto
Xynnn007: I'm not sure about this actually. The protos are given by the official aliyun kms sdk. Also I found some message definitions are not used at all. Let me delete them.

confidential-datahub/secret/src/secret/mod.rs
Xynnn007: Yes, thanks! Let me fix this

confidential-datahub/hub/src/hub.rs
Xynnn007: Ah, ok. My idea is the claims included in a JWS. From RFC it should be called `payload` https://www.rfc-editor.org/rfc/rfc7515.html#section-3. Let me fix this.

Xynnn007:
> Follow-up question which I'm not sure is appropriate to this PR, but where in the code (if anywhere yet) is the following enforced: Attestation must happen before unsealing secrets.

Nice question! The unsealing operation implies that the credentials to connect to the kms is prepared. The credentials are the confidential resources for the tenants. We cannot embed the credentials into the guest image because the host can see the plaintext. It means that the credentials are injected into the guest at some time from the tenant.  It must be after the remote attestation for the tenant to inject the credentials, because in this way the tenant can ensure the environment is as expected.

To say more, the `auth` layer design in CDH https://github.com/confidential-containers/guest-components/blob/main/confidential-datahub/docs/Architecture.svg is responsible for getting the credentials. Soon after this PR we will add a plugin which calls KBS client in `auth`, which functionally will get the credentials or other things and is based on the remote attestation.

confidential-datahub/docs/SEALED_SECRET.md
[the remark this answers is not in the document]
Xynnn007: How about `secret_setting`. Actually I do not have a good name in my mind either.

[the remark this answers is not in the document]
arronwy: > These fields are a bit confusing to me. I don't think they were in the original RFC. Will all KMSes need the same annotations? Why separate into `operation_context` and `provider_setting` (should be `provider_settings`?)

Yes, `provider_settings` is needed to initialize an KMS client, we can move to upper level as mandatory, and `annotations` can only save optional data like `secret_settings` or `operattion_context` which can be vendor specific for different KMS implementation.  @Xynnn007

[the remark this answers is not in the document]
Xynnn007: > > These fields are a bit confusing to me. I don't think they were in the original RFC. Will all KMSes need the same annotations? Why separate into `operation_context` and `provider_setting` (should be `provider_settings`?)
> 
> Yes, `provider_settings` is needed to initialize an KMS client, we can move to upper level as mandatory, and `annotations` can only save optional data like `secret_settings` or `operattion_context` which can be vendor specific for different KMS implementation. @Xynnn007

Sounds rationale to me. Let me apply this change to the code and plz take another look if it is straightforward and acceptable.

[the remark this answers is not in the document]
Xynnn007: force pushed. I think @arronwy gets the core issue. We cannot create a kms client with only `provider` field, but also some other metadata is required. Thus an extra `provider_settings` key-value map is introduced in the sealed secret format. We do not use `secret_settings` anymore, instead we put all other things into the origin optional `annotations` field.

fitzthum (approved):
LGTM

confidential-data-hub/docs/SEALED_SECRET.md
```diff
@@ -148,4 +154,10 @@ stringData:
 
 When this sealed secret is transferred to TEE, Confidential DataHub will help to unseal.
 
+## Supported Providers
+
+| Provider Name      | README                                                              | Maintainer                |
+| ------------------ | ------------------------------------------------------------------- | ------------------------- |
+| ali       	     |  [ali](kms-providers/alibaba.md)                              	   | Alibaba                   |
```
arronwy, line 161: should we use `aliyun` as the provider name to consistent with the KMS build feature?

confidential-data-hub/docs/kms-providers/alibaba.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Spec
+
+### Consts & Layouts
+
+Here are the consts for Aliyun KMS.
+
+| Name               | Value       |
+| ------------------ | ----------- |
+| `provider`       	 | `ali`       |
```
arronwy, line 11: same as above use `aliyun`?

arronwy (approved):
Thanks @Xynnn007 LGTM!

confidential-data-hub/docs/SEALED_SECRET.md
Xynnn007: great. Let me fix all the typos.

PR #315, opened 2023-08-08
Kbs protocol fix cargo toml

Xynnn007:
This newly refactored `kbs_protocol` crate by default activate all attesters when background-check is enabled. This PR fixes this.

related to https://github.com/confidential-containers/enclave-cc/pull/180


arronwy (approved):
Thanks @Xynnn007 LGTM!

PR #325, opened 2023-08-17
Initial implementation rest api server for CoCo

arronwy:
Verified with latest CDH code https://github.com/confidential-containers/guest-components/pull/322

```
$ ./api-server-rest
Starting API server on 127.0.0.1:8006
API Server listening on http://127.0.0.1:8006

$ curl http://127.0.0.1:8006/cdh/resource/default/key/1
12345678901234567890123456xxxx

$ curl http://127.0.0.1:8006/aa/evidence\?runtime_data\=xxxx
{"svn":"1","report_data":"eHh4eA=="}

$ curl http://127.0.0.1:8006/aa/token\?token_type\=kbs
{"token":"eyJhbGciOiJFi...","tee_keypair":"-----BEGIN... "}
```


api-server-rest/openapi/api.yaml
```diff
@@ -0,0 +1,148 @@
+{
```
Xynnn007, line 1: should the file name be `-.json`?

api-server-rest/build.rs
```diff
@@ -0,0 +1,102 @@
+use ttrpc_codegen::{Codegen, Customize, ProtobufCustomize};
```
Xynnn007, line 1: do you need SPDX-License-Identifier here?

api-server-rest/src/aa.rs
```diff
@@ -0,0 +1,115 @@
+use crate::router::ApiHandler;
```
Xynnn007, line 1: do you need SPDX-License-Identifier here?

api-server-rest/src/router.rs
```diff
@@ -0,0 +1,101 @@
...
+        self.routes.insert(route.to_string(), handler);
+    }
+
+    pub async fn route(
+        &mut self,
+        remote_addr: SocketAddr,
+        req: Request<Body>,
+    ) -> Result<Response<Body>> {
+        if let Some((root_path, url_path)) = split_nth_slash(req.uri().path(), 2) {
+            println!("root_path {}, url_path {}", root_path, url_path);
```
Xynnn007, line 91: A little question from my side: I saw several `println!` . Is it a better choice to use `log` module?

Xynnn007 (approved):
Cool! Thanks @arronwy   let's discuss more about this in the weekly meeting

api-server-rest/openapi/api.yaml
arronwy: Thanks @Xynnn007 , good catch, fixed now：）

api-server-rest/build.rs
arronwy: Thanks @Xynnn007 , good catch, fixed now：）

danmihai1:
Sharing this information just in case it helps:

This is how the OPA REST API works now, in the main branch:

1. Get the value of foo from the Policy:

curl -X POST http://localhost:8181/v1/data/agent_policy/foo

2. In case foo has a set of rules in the Policy, returning different values based on some query inputs/parameters:

curl -X POST http://localhost:8181/v1/data/agent_policy/foo --data-binary @foo-inputs.json

api-server-rest/src/aa.rs
```diff
@@ -0,0 +1,120 @@
...
+                        .await
+                        .unwrap_or_else(|e| e.to_string().into());
+                    return self.octet_stream_response(results);
+                }
+                None => return self.bad_request(),
+            },
+            AA_EVIDENCE_URL => match params.get("runtime_data") {
+                Some(runtime_data) => {
+                    let results = self
+                        .get_evidence(&runtime_data.clone().into_bytes())
```
fitzthum, line 71: This can only be exposed to the guest in particular circumstances (when the evidence is not generic). This endpoint needs to be disabled for many workloads or we will have a huge security gap. I don't see any configuration for enabling/disabling endpoints in this PR.

api-server-rest/src/aa.rs
fitzthum, line 1: Why connect directly to the AA from the proxy? It seems like it might be wiser to add a `GetEvidence` endpoint to the CDH and expose that via this proxy. That could help with configuration issue mentioned below.

api-server-rest/src/cdh.rs
```diff
@@ -0,0 +1,89 @@
...
+
+#[async_trait]
+impl ApiHandler for CDHClient {
+    async fn handle_request(
+        &self,
+        remote_addr: SocketAddr,
+        url_path: &str,
+        req: Request<Body>,
+    ) -> Result<Response<Body>> {
+        if !remote_addr.ip().is_loopback() {
```
fitzthum, line 38: What is the purpose of this check?

fitzthum (commented):
I have some concerns. See comments.

api-server-rest/src/aa.rs
arronwy: Yes, agree, we plan to launch `api_server_rest` daemon only when kata agent config to enable `rest_api`, maybe we can add more config in kata agent like 
```
agent.rest_api=aa
agent.rest_api=cdh
agent.rest_api=all
```

api-server-rest/src/aa.rs
arronwy: Current CDH is not export AA `GetEvidence/Token` interface yet, maybe we can switch to use CDH's interface once they are ready.

api-server-rest/src/cdh.rs
arronwy: Current `api-server-rest` listen on local host which only can be accessed by internal signed container workload, we also support export external server if needed, add this check in current `cdh` and `aa` endpoint is to avoid the info leak if we switch to listen on public IP address.

api-server-rest/src/aa.rs
arronwy: I added the feature flag for `api-server-rest` which allows enable different endpoints during launch time, and enable `resource` as default feature

fitzthum: Ok, I am a little bit wary of adding more configs to the kernel params. I think these configs will be scoped to the runtime class, so we could potentially use a build-time configuration that is invoked when we setup the initrd, or a config file that lives in the initrd.

arronwy: Yes, agree, since kernel cmdline will impact the measurement evidence, for runtime enablement to decide whether to launch api server, finally we may need use init data(maybe part of the CoCo policy file) with integrity protection.
for build enablement, we will use a feature flag to decide whether to integrate api server binary into the rootfs.

fitzthum: Ok, i think configuring via cmdline is fine as a first step. Let's make sure we add a README somewhere for how to setup all these new options and let's think about how to refine the configuration.

fitzthum (approved):

Issue #333, opened 2023-08-29
Cargo Vendor fails due to duplicate version of sev


emanuellima1:
When executing `cargo vendor` on `guest-components`, the command fails with the following message:

```bash
error: failed to sync

Caused by:
  found duplicate version of package `sev v1.2.0` vendored from two sources:

        source 1: registry `crates-io`
        source 2: [https://github.com/virtee/sev](https://github.com/virtee/sev?rev=3dca05d2c93388cb00534ad18f5928fd812e99cc#3dca05d2)
```

This is due to the sev dependency coming from two different places on the same project.

PR #334, opened 2023-08-29
Remove git reference for sev

emanuellima1:
Having the sev dependency come from GitHub instead of crates.io was causing errors when vendoring. This change fixes it, as it makes Cargo look for sev on crates.io.

Fixes [issue 333](https://github.com/confidential-containers/guest-components/issues/333).


beraldoleal:
@fitzthum any reason for using the virtee/sev here?

fitzthum (approved):
I think this is probably just a remnant from the switch to the virtee repo. Should be fine to change.

PR #335, opened 2023-08-30
Fix enclave-cc dep

Xynnn007:
Current image-rs will bring `sev` dep to enclave-cc while it is not used.


mythi:
I've tested it works for me as expected (no tdx/sev dependencies are pulled anymore after this and the build succeeds).

arronwy (approved):
Thanks @Xynnn007 LGTM!

PR #349, opened 2023-09-10
CDH add unwrapkey API

Xynnn007:
Support decrypting image for CDH.

Let's continue this PR in v0.9.0 cycle.


confidential-data-hub/hub/protos/api.proto
```diff
@@ -18,10 +18,22 @@ message GetResourceResponse {
...
 service SealedSecretService {
     rpc UnsealSecret(UnsealSecretInput) returns (UnsealSecretOutput) {};
 }
 
 service GetResourceService {
     rpc GetResource(GetResourceRequest) returns (GetResourceResponse) {};
 }
+
+service KeyProviderService {
+    rpc UnWrapKey(KeyProviderKeyWrapProtocolInput) returns (KeyProviderKeyWrapProtocolOutput) {};
```
fitzthum, line 38: unwrap is one word, so `W` should probably be lowercase, but I don't think it's worth changing

confidential-data-hub/hub/src/bin/confidential-data-hub/server/mod.rs
```diff
@@ -0,0 +1,153 @@
...
+        let mut reply = GetResourceResponse::new();
+        reply.Resource = resource;
+        debug!("send back the resource");
+        Ok(reply)
+    }
+}
+
+#[async_trait]
+impl KeyProviderService for Server {
+    async fn un_wrap_key(
```
fitzthum, line 100: It always annoyed me that this wasn't `unwrap_key` in the AA but I guess it is auto-generated.

confidential-data-hub/hub/src/bin/confidential-data-hub/server/mod.rs
```diff
@@ -0,0 +1,153 @@
...
+        &self,
+        _ctx: &TtrpcContext,
+        req: KeyProviderKeyWrapProtocolInput,
+    ) -> ::ttrpc::Result<KeyProviderKeyWrapProtocolOutput> {
+        debug!("get new GetResource request");
```
fitzthum, line 105: Is this the correct debug message?

confidential-data-hub/image/src/annotation_packet/v2.rs
```diff
@@ -0,0 +1,140 @@
...
+    pub iv: Option<String>,
+
+    /// Wrap type to specify encryption algorithm and mode. Only used when
+    /// provider is `"kbs"`
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub wrap_type: Option<String>,
+
+    /// extra information to create a client
+    #[serde(default = "Map::default")]
+    pub provider_settings: ProviderSettings,
```
fitzthum, line 48: So does this mean that I would have to modify my image if any of the provider settings change? Similar to the KBS URI, I feel like this should maybe be decoupled from the workload?

fitzthum (commented):
A few comments. Looks pretty good overall. I had expected this PR to remove support from the AA, but I guess that will be done separately?

confidential-data-hub/hub/src/bin/confidential-data-hub/server/message.rs
```diff
@@ -0,0 +1,89 @@
...
+//
+
+use anyhow::*;
+use base64::Engine;
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use std::str;
+use std::vec::Vec;
+
+const ANNOTATION_KEY_NAME: &str = "attestation-agent";
```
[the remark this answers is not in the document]
Xynnn007: hah. Yes, the legacy term already used in current encrypted images. Another way is to rename directly as `confidential-containers` for newly produced images, which can be done in upcoming coco keyprovider pr. For CDH, we support both new and old name. But for coco keyprovider, we stop producing images with legacy `attestation-agent` name.

confidential-data-hub/hub/protos/api.proto
Xynnn007: Well, the api name is defined by ocicrypt here https://github.com/containers/ocicrypt/blob/ffc163c18e970386fcd1b35acf2d350c8665d080/utils/keyprovider/keyprovider.proto#L16C1-L16C1

confidential-data-hub/hub/src/bin/confidential-data-hub/server/mod.rs
Xynnn007: same as mentioned. Coinincidentally, when I coded this I also thought it should be

confidential-data-hub/image/src/annotation_packet/v2.rs
Xynnn007: Yes. This is somehow the same as sealed secret. Up to now when we generate an encrypted image we fix the origin of the KEK. So it bounds to the workload.

You enlight me that we can also move the ＂information of KEK＂ part out of the image, e.g. annotations from kubernetes yaml.

The pros are more flexible.
The cons are complexer for user and we should think about how to attest this. (if in image, it can be covered by image signature)

Anyway, it is not implemented now. We can discuss about this if we need.

Xynnn007:
> A few comments. Looks pretty good overall. I had expected this PR to remove support from the AA, but I guess that will be done separately?

Yes. I want to put a final PR to remove all useless things when all are ready.

confidential-data-hub/hub/protos/api.proto
fitzthum: 😞 oh well

confidential-data-hub/image/src/annotation_packet/v2.rs
fitzthum: Yeah there are some tradeoffs here. We are basically taking the same approach with sealed secrets, but sealed secrets have somewhat different properties than images. For instance there is no such thing as a generic sealed secret. We expect users to create sealed secrets when they use CoCo. It's not very hard to do that so I don't feel bad about asking users to bind a secret to particular KMS provider and re-seal things if the KMS provider changes.

Images, on the other hand, are a bit more generic. Ideally people could run existing encrypted images with CoCo or to decrypt a CoCo image locally for debugging. We also probably don't want to have to change the image if the KMS provider changes.

One simple option would be to support one KMS provider per guest. Obviously this would be somewhat limiting, but it would allow us to configure the KMS provider similarly to how we currently configure the AA.

I'm not sure about the measurement question here. I don't know if the provider settings have to be measured at all. Connecting to a KMS requires the correct parameters from the auth layer.

confidential-data-hub/hub/src/bin/confidential-data-hub/server/message.rs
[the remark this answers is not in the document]
fitzthum: Yeah, we might want to transition to a new name at some point.

confidential-data-hub/image/src/annotation_packet/v2.rs
Xynnn007: A problem needed to be thought over is how to define the `keyid` field in the AnnotationPacket. Different KMS has its own different `keyid` format. For example, 
- aliyun kms: `key-shh64705d53mly8ubfxxx`
- [azure key vault](https://learn.microsoft.com/en-us/rest/api/keyvault/keys/decrypt/decrypt?tabs=HTTP#decrypt-example): `sdktestkey/4eb68492b5f6421e835d961ad2be3155`

if we use a common `keyid` field inside the image, we need some mechanism to map the same keyid into different key ids due to the given KMS type. However which is hard, as the keyid is often generated by the KMS service not the user. If we use KBS there will not be any problems as we now use a common `ResourceUri` as keyid for all different kbc plugins (coco-KBS, simple-KBS, offline-fs-kbs, etc.).

> I'm not sure about the measurement question here. I don't know if the provider settings have to be measured at all. Connecting to a KMS requires the correct parameters from the auth layer.

A basic idea is that the guest is a finite-state machine, any input from the host will influence the behavior and state of the guest, thus they need to be measured.

confidential-data-hub/hub/src/bin/confidential-data-hub/server/message.rs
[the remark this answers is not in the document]
Xynnn007: Agreed. I think it is a good time to transit during the reviewing and merging of this PR. If we keep the backward compability, we won't break any current applications.

confidential-data-hub/image/src/annotation_packet/v2.rs
fitzthum: Hm yeah specifying the key name is tricky. As you say, this problem goes away when we use the KBS because the KBS can map ResourceURI onto just about anything. If we don't want to use a KBS for that the mapping has to happen inside of the guest or even earlier when the container is encrypted. I wonder if there is any way to create a resource mapping system in the CDH that wouldn't be complicated to configure.

> A basic idea is that the guest is a finite-state machine, any input from the host will influence the behavior and state of the guest, thus they need to be measured

I'm not sure I agree with this. We've said in the past that the AA_KBC_PARAMS don't need to be measured. Let's forget about the KBS URI for now, but the KBC type almost certainly doesn't need to be measured. If the host sets this to the wrong platform, the AA won't be able to get valid HW evidence. I think the KMS provider probably doesn't need to be measured either. Similarly the configurations for the auth layer don't seem sensitive. At least, I think it's possible to have a secure workload without measuring these things. We might still want to measure things in order to generate a more complete view of the environment.

Xynnn007: > I wonder if there is any way to create a resource mapping system in the CDH that wouldn't be complicated to configure.

In principle, what we need is exactly the mapping `keyid in AnnotationPacket => keyid in KMS`. So a configuretion file of cdh would help. Part of that in my mind would be (json or toml are both ok)
```json
{
    "kms": "aliyun",
    "key_mappings": [
        "keyid-1": "key-shh64705d53mly8ubfxxx",
        "keyid-2": "key-3mly8ubfxxxshh64705"
    ],
    "credentials": [
        "CLIENT_KEY": "kbs:///default/credential/aliyun-client-key",
        ...
    ],
    "aa_kbc_params": "cckbc::http://1.2.3.4:8080"
}
```

Here,
- `kms`: specify the KMS plugin in the whole lifetime of the CDH will use.
- `key_mappings`: specify the mapping from `keyid in AnnotationPacket` onto `keyid of the KMS`
- `credentials`: current code in upstream we store the credential in guest memfs. you've mentioned that we'd better to store them in the memory of CDH. One way for this is to store them in the env of the CDH. This `credential` list will tell the env key names and the value which should be fetched from KBS. We need to refactor the KMS clients to read the credentials they need from env of current process instead of guest memfs.

For comparation, if we use [`eHSM-KMS`](https://github.com/confidential-containers/guest-components/pull/359/files), it can be
```
{
    "kms": "aliyun",
    "key_mappings": [
        "keyid-1": "key1",
        "keyid-2": "key2"
    ],
    "credentials": [
        "APPID": "kbs:///default/credential/ehsm-appid",
        "APIKEY": "kbs:///default/credential/ehsm-apikey",
        ...
    ],
    "aa_kbc_params": "cckbc::http://1.2.3.4:8080"
}
```

> Let's forget about the KBS URI for now, but the KBC type almost certainly doesn't need to be measured. If the host sets this to the wrong platform, the AA won't be able to get valid HW evidence.

The configurations for the auth layer are not sensitive, but their integraty needs to be protected. Let us temporary forget the configuration format I've mentioned and go back to current one. If we specify an extra parameter `cdh.kbs_resources=kbs:///default/key/unexpected::/run/temp1`, the resource `kbs:///default/key/unexpected` from KBS will be accessed but unexpected. The KBC type also needs to be measured as we now have guest images that includes `/etc/aa-offline-fs-kbc-keys.json` and `/etc/aa-offline-fs-kbc-resources.json`. change of KBC from others into `offlinefs` would cause unexpected behaviors that malicious host can leverage to finish attacks in edge case.

> We might still want to measure things in order to generate a more complete view of the environment.
Agree with this. The view of the environment should also include the configurations of AA, CDH, ...

fitzthum: It's going to be annoying for a user to generate these mappings. I wonder if they could be automatically created by the KeyProvider used for encryption. Then the question becomes how you get the mappings into the guest easily. One interesting option would be to have the key mappings travel with the image, possibly in an additional layer. Would this layer need to be signed? This brings us to the second question. 

There is a fundamental tradeoff regarding how much stuff is measured. On one hand you want to measure as much of the workload as possible. This feels more secure and it allows a more complete record of the workload. On the other hand, you want to measure as little as possible, because the more things you measure the more difficult it is to validate the measurement. Kubernetes is a very dynamic environment and it isn't easy to reconcile this with our security goals.

I understand the instinct to try to measure as much as we can, but I think it might be a mistake at times and I do think there are things that don't need to be measured. For example, if we had this key mapping file somewhere, I don't think that would need to be measured. If the host changed this file, they could get the guest to request an unexpected resource from a KMS. Most likely this resource wouldn't exist and nothing would happen. If the resource did exist, it wouldn't be the correct decryption key and image decryption would fail. I think this is roughly equivalent to the KMS provider field.

There are some corner cases we would need to be careful about. If we don't configure logging correctly a host could potentially manipulate the mappings to determine which resource names exist and which don't.

Xynnn007: Sry I'm a little confused about the mappings inside the image layer. Let's go back to the original question.

In this PR, I suppose the `kid` field in the AnnotationPacket will tell which KMS is used and what `keyid` is, e.g. `aliyun://key-shh64705d53mly8ubfxxx`.

You pointed out that if so, the image can only be decrypted by aliyun KMS. If the user wants build one encrypted image and decrypt that in different KMSes after inserting the decryption key into the KMS.

I then proposed that we can make the kid inside AnnotationPacket easy, like `keyid-1`. Also there should be a map inside CDH. When CDH receives a request to index a key with `keyid`, CDH will look up the value of the key `keyid` in the map. For example `keyid-1` is mapped to `key-3mly8ubfxxxshh64705`. In this way, the `keyid` field inside an encrypted image can be mapped to different kid, based on the KMS type chosen of this guest launch.

This map should be configured by the user, because it is the user who inserts the key into KMS with BYOK. If we store it as a layer with the image, the config will be static together with the image, which will be somehow the same as the initial proposal that the information of the key to encrypt the image is static and fixed in the image.

Let's go on. If we want a flexible mapping, it can be stored inside the KBS and be fetched by CDH. We can get rid of the mapping from the launch configuration of CDH.

Xynnn007: > I understand the instinct to try to measure as much as we can, but I think it might be a mistake at times and I do think there are things that don't need to be measured. For example, if we had this key mapping file somewhere, I don't think that would need to be measured. If the host changed this file, they could get the guest to request an unexpected resource from a KMS. Most likely this resource wouldn't exist and nothing would happen. If the resource did exist, it wouldn't be the correct decryption key and image decryption would fail. I think this is roughly equivalent to the KMS provider field.

There can be two ways to ensure the things inside the guest are as expected. One is measurement, and the other is to filter that with a policy. Certainly there should be a three option, to fetch from KBS.

So,
- Measurement. Is suitable to those configurations helps to connect to KBS (if we do not connect to KBS, we cannot leverage way 3) and check the base TCB of the guest.
- Policy. Is suitable to those dynamic input from the host, like kata-agent API callings, mount operations, etc.
- KBS. Is suitable for other files, configs which are used before the guest to response the result of the first `CreateContainer` request, etc. like container policy, image auth, etc.

My idea is that all configurations from the host should be classified as one of the above three. If we want to keep the attestation simpler, try to squeeze things in way 1 into way 3 with only a ResourceId pointing to the expected one.

Xynnn007:
@fitzthum Can we merge this before we think of a better way to hide the concrete KMS type used to encrypt the image?

fitzthum:
> Can we merge this before we think of a better way to hide the concrete KMS type used to encrypt the image?

Yeah probably. I will take another quick look at this PR.

fitzthum (approved):
LGTM. Might want to wait a few days to merge in case we still need to cut v0.8.1

PR #350, opened 2023-09-11
ci: install DCAP packages from Jammy repo

mythi:
The workflows use ubuntu-22.04 (ubuntu-latest) but install DCAP packages from focal.

While fixing that, also move away from deprecated apt-key usage and install recommended packages only.

### DIFF
--- a/.github/workflows/aa_basic.yml
+++ b/.github/workflows/aa_basic.yml
@@ -24,7 +24,7 @@ jobs:
     defaults:
       run:
         working-directory: ./attestation-agent
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


mythi:
> and install recommended packages only.

this does not seem to make any difference so can be dropped if that's preferred

arronwy:
current Kata rootfs seems still use 20.04/focal as default config: https://github.com/kata-containers/kata-containers/blob/main/tools/osbuilder/rootfs-builder/ubuntu/config.sh#L7

@fidencio  Do we have plan to upgrade ubuntu to 22.04 as default rootfs image?

mythi:
this PR is still valid and safe to be merged, AFAICS. Rebased and force pushed

arronwy (approved):
Thanks @mythi LGTM! The failed CI is not related to this PR.

PR #356, opened 2023-09-12
workflows: Bump to rust 1.72

stevenhorsman:
Now we've gone to v1.72 in kata-containers, we can bump the versions we test against to stay in sync. At the moment there is a bit of duplication as 1.72 is the stable version, but that will change soon

### DIFF
--- a/.github/workflows/image_rs_build.yml
+++ b/.github/workflows/image_rs_build.yml
@@ -24,7 +24,7 @@ jobs:
       fail-fast: false
       matrix:
         rust:
-          - 1.69.0
+          - 1.72.0
           - stable
           - nightly
     steps:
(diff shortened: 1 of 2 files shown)


surajssd (approved):

Xynnn007:
btw, do we need to revert https://github.com/confidential-containers/guest-components/pull/337?

arronwy (approved):
CCv0 have upgraded to 1.72.0: https://github.com/kata-containers/kata-containers/pull/7918

stevenhorsman:
> btw, do we need to revert #337?

@Xynnn007, Hey Ding, I don't think we want to completely revert it as I think specifying a version is probably better practice for stability (though I'm happy to have that discussion), but I can bump it to pick up a later version if you'd like?

PR #365, opened 2023-09-19
Read agent config from file

stevenhorsman:
Workaround for 0.8 to let peer pod environments read the aa_kbc_params from the agent config toml


attestation-agent/lib/src/token.rs
```diff
@@ -47,3 +55,22 @@ pub(crate) async fn get_kbs_host_from_cmdline() -> Result<String> {
...
+pub(crate) async fn get_kbs_host_from_config_file() -> Result<String> {
+    // We only care about the aa_kbc_params value at the moment
+    #[derive(Debug, Deserialize)]
+    struct AgentConfig {
+        aa_kbc_params: Option<String>,
+    }
+
+    // Hard-code agent config path to "/etc/agent-config.toml" as a workaround
+    let agent_config_str = fs::read_to_string("/etc/agent-config.toml")
+        .expect("Failed to read /etc/agent-config.toml file");
```
Xynnn007, line 68: `expect` here will cause the process panic when error occurs, which is like `unwrap`. We do not encourage panic inside.

We can import `anyhow::Context` to convert any kinds of error into `anyhow` error with a context information.

```suggestion
        .context("Failed to read /etc/agent-config.toml file")?;
```

attestation-agent/lib/src/token.rs
```diff
@@ -47,3 +55,22 @@ pub(crate) async fn get_kbs_host_from_cmdline() -> Result<String> {
...
+
+    let agent_config: AgentConfig =
+        toml::from_str(&agent_config_str).expect("Failed to deserialize /etc/agent-config.toml");
```
Xynnn007, line 71:
```suggestion
        toml::from_str(&agent_config_str).context("Failed to deserialize /etc/agent-config.toml")?;
```

attestation-agent/lib/src/token.rs
```diff
@@ -47,3 +55,22 @@ pub(crate) async fn get_kbs_host_from_cmdline() -> Result<String> {
 
     Ok(kbs_host)
 }
+
+pub(crate) async fn get_kbs_host_from_config_file() -> Result<String> {
+    // We only care about the aa_kbc_params value at the moment
+    #[derive(Debug, Deserialize)]
+    struct AgentConfig {
```
Xynnn007, line 62: I'm not sure whether a partial definition of the config will work. Try it my friend :P

confidential-data-hub/kms/src/plugins/kbs/mod.rs
```diff
@@ -126,3 +133,35 @@ async fn get_aa_params_from_cmdline() -> Result<(String, String)> {
...
+
+    // We only care about the aa_kbc_params value at the moment
+    #[derive(Debug, Deserialize)]
+    struct AgentConfig {
+        aa_kbc_params: Option<String>,
+    }
+
+    // Hard-code agent config path to "/etc/agent-config.toml" as a workaround
+    let agent_config_str = fs::read_to_string("/etc/agent-config.toml")
+        .expect("Failed to read /etc/agent-config.toml file");
```
Xynnn007, line 148:
```suggestion
        .context("Failed to read /etc/agent-config.toml file")?;
```

attestation-agent/lib/src/token.rs
```diff
@@ -16,7 +16,15 @@ struct Message {
 
 pub(crate) async fn get_kbs_token() -> Result<Vec<u8>> {
     let evidence_provider = Box::new(NativeEvidenceProvider::new()?);
-    let kbs_host_addr = get_kbs_host_from_cmdline().await?;
+
+    // Check for /peerpod/daemon.json to see if we are in a peer pod
+    // If so we need to read from the agent-config file, not /proc/cmdline
+    let kbs_host_addr = if Path::new("/peerpod/daemon.json").exists() {
```
Xynnn007, line 22: It is encouraged to add a `const`. s.t.
```rust
const PEER_POD_CONFIG_PATH: &str = "/peerpod/daemon.json";
...
let kbs_host_addr = if Path::new(PEER_POD_CONFIG_PATH).exists() {
```

confidential-data-hub/kms/src/plugins/kbs/mod.rs
```diff
@@ -32,7 +32,14 @@ enum RealClient {
 
 impl RealClient {
     async fn new() -> Result<Self> {
-        let (kbc, _kbs_host) = get_aa_params_from_cmdline().await?;
+
+         // Check for /peerpod/daemon.json to see if we are in a peer pod
+        // If so we need to read from the agent-config file, not /proc/cmdline
+        let (kbc, _kbs_host) = if Path::new("/peerpod/daemon.json").exists() {
```
Xynnn007, line 38: It is encouraged to add a `const`. s.t.
```rust
const PEER_POD_CONFIG_PATH: &str = "/peerpod/daemon.json";
...
let (kbc, _kbs_host) = if Path::new("/peerpod/daemon.json").exists()
```

confidential-data-hub/kms/src/plugins/kbs/mod.rs
```diff
@@ -126,3 +133,35 @@ async fn get_aa_params_from_cmdline() -> Result<(String, String)> {
...
+    struct AgentConfig {
+        aa_kbc_params: Option<String>,
+    }
+
+    // Hard-code agent config path to "/etc/agent-config.toml" as a workaround
+    let agent_config_str = fs::read_to_string("/etc/agent-config.toml")
+        .expect("Failed to read /etc/agent-config.toml file");
+
+    let agent_config: AgentConfig =
+        toml::from_str(&agent_config_str).expect("Failed to deserialize /etc/agent-config.toml");
```
Xynnn007, line 151:
```suggestion
        toml::from_str(&agent_config_str).context("Failed to deserialize /etc/agent-config.toml")?;
```

Xynnn007 (commented):
Hey Steve, thanks for the contribution and I think you did a really good job in Rust. Only some unwritten convention in CoCo from our practice. Please refer in your preferred way :)

attestation-agent/lib/src/token.rs
fitzthum: I might use a match here instead of an if, but it's not a big deal.

attestation-agent/lib/src/token.rs
stevenhorsman: Yeah, I tried this is a dummy test environment and it seems to do fine in ignoring the rest of the toml file 😄

stevenhorsman:
Thanks for the feedback Ding and Tobin, I think it's all addressed now, so I'll wait to get clean tests and then try it out in peer-pods 🤞

stevenhorsman:
Hey Ding, I think I might need some help with the switch to context. My initial attempt was to add `use anyhow::Context;` to mod.rs, but with that I'm hitting errors:
```
error[E0599]: no method named `context` found for opaque type `impl std::future::Future<Output = std::result::Result<std::string::String, std::io::Error>>` in the current scope
   --> confidential-data-hub/kms/src/plugins/kbs/mod.rs:152:10
    |
151 |       let agent_config_str = fs::read_to_string("/etc/agent-config.toml")
    |  ____________________________-
152 | |         .context("Failed to read /etc/agent-config.toml file")?;
    | |         -^^^^^^^ method not found in `impl Future<Output = Result<String, Error>>`
    | |_________|
    |

error[E0277]: `?` couldn't convert the error to `error::Error`
   --> confidential-data-hub/kms/src/plugins/kbs/mod.rs:155:65
    |
155 |         .context("Failed to deserialize /etc/agent-config.toml")?;
    |                                                                 ^ the trait `std::convert::From<anyhow::Error>` is not implemented for `error::Error`
    |
    = note: the question mark operation (`?`) implicitly performs a conversion on the error value using the `From` trait
    = help: the following other types implement trait `std::ops::FromResidual<R>`:
              <std::result::Result<T, F> as std::ops::FromResidual<std::ops::Yeet<E>>>
              <std::result::Result<T, F> as std::ops::FromResidual<std::result::Result<std::convert::Infallible, E>>>
    = note: required for `std::result::Result<(std::string::String, std::string::String), error::Error>` to implement `std::ops::FromResidual<std::result::Result<std::convert::Infallible, anyhow::Error>>`
```
as I think it is not using the anyhow version of `Error`.

If I update to `use anyhow::{Context, Error};` then it clashes with the `use crate::{Annotations, Error, Getter, Result};` and gives more errors. Any suggestions of the best approach to solve this?

fitzthum:
For the first error, it looks like you're using tokio::fs, which is async. So you should probably add an `await` before the `context`. You can see from the error that it is trying to call `context` on a `future` which is usually a pretty good sign to add an `await`.

fitzthum:
The second error is because we are mixing `thiserror` with `anyhow`. This happens because we import result [here](https://github.com/confidential-containers/guest-components/blob/ec47cec5fe9ae2ac59d3fd3bdc2481160b9fd1f0/confidential-data-hub/kms/src/plugins/kbs/mod.rs#L26), which comes from our custom error definitions in `error.rs`.

I'm not really sure I like `thiserror` yet, but you can get around your bug by doing something like [this](https://github.com/confidential-containers/guest-components/blob/ec47cec5fe9ae2ac59d3fd3bdc2481160b9fd1f0/confidential-data-hub/kms/src/plugins/kbs/mod.rs#L82).

confidential-data-hub/kms/src/plugins/kbs/mod.rs
```diff
@@ -126,3 +137,36 @@ async fn get_aa_params_from_cmdline() -> Result<(String, String)> {
...
+    struct AgentConfig {
+        aa_kbc_params: Option<String>,
+    }
+
+    // Hard-code agent config path to "/etc/agent-config.toml" as a workaround
+    let agent_config_str = fs::read_to_string("/etc/agent-config.toml")
+        .context("Failed to read /etc/agent-config.toml file")?;
+
+    let agent_config: AgentConfig = toml::from_str(&agent_config_str)
+        .context("Failed to deserialize /etc/agent-config.toml")?;
```
Xynnn007, line 155:
```suggestion
    let agent_config_str = fs::read_to_string("/etc/agent-config.toml")
        .await
        .map_err(|e| Error::KbsClientError(format!("Failed to read /etc/agent-config.toml file: {e}"))?;

    let agent_config: AgentConfig = toml::from_str(&agent_config_str)
        .map_err(|e| Error::KbsClientError(format!("Failed to deserialize /etc/agent-config.toml: {e}"))?;
```

stevenhorsman:
> Also, I'm not sure that it's necessary to use tokio::fs rather than the std::fs and it's a bit weird to put a use statement inside the method. You might want to move that to the top of the file (unless you really need it to only be in that scope).

I just copied this from the existing method `get_aa_params_from_cmdline`, but I'm happy to make is a std `fs`

stevenhorsman:
I've tested this branch in a peer podvm as part of https://github.com/confidential-containers/cloud-api-adaptor/issues/1424#issuecomment-1729193022 and the CDH hits a different error, so I think this parsing workaround is good enough for now.

fitzthum (approved):
LGTM

Issue #366, opened 2023-09-20
kbs_protocol does not follow the KBS spec


mythi:
In the KBS spec, it says "The hash of the `tee-pubkey` field must be included in the custom field of HW-TEE evidence and signed by HW-TEE hardware."

kbs_protocol hashes only parts of `tee-pubkey`
https://github.com/confidential-containers/guest-components/blob/914251abcb4b8c5a06acf126df5d71cf50451bd2/attestation-agent/kbs_protocol/src/client/rcar_client.rs#L93

mythi:
> kbs_protocol hashes only parts of `tee-pubkey`

the other gap is that since CoCo AS has TEE specific verifiers, it is important for the attesters to generate the corresponding TEE specific quotes/reportdata.

PR #369, opened 2023-09-27
aa: attester: Let attesters to prepare reportdata

mythi:
Fixes: #366

Mostly build-tested only so keeping in draft for now.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -284,6 +284,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sev 1.2.1",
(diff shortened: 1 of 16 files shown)


mythi:
Looks like this old `get_evidence()` is used all over the place. I need to take a closer look what can be changed.

Xynnn007:
I think we couple the report data calculation insde attester is not a good choice. Different users would want to calculate report data with different algorithms. This PR would reduce flexibility.

mythi:
not needed anymore

PR #371, opened 2023-09-29
attestation-agent: fix extraction of peerpod kbs host addr extraction in token code

mkulke:
The current peerpod code path that parses the agent config file does not strip the prefix, hence the kbs_host addr (it's not just a "kbs_host" it's "protocol + host + port") ends up being: `cc_kbc::http://1.2.3.4:8080`.

After this change a pod i'm able to get a token and a secret resource from kbs on a peerpod deployment:

```bash
k exec -it deploy/nginx-caa -- curl http://127.0.0.1:8006/cdh/resource/default/key/1
ohai
```

on kbs:

image


mkulke:
cc @stevenhorsman PTAL

stevenhorsman (approved):
LGTM. Thanks for the fix

fitzthum (approved):
LGTM

PR #372, opened 2023-10-03
image-rs: Support simple signing with X-R-S-S

mattarnoatibm:
Support getting simple signing signatures from
registry that supports X-R-S-S extension.

Refactor into separate tests for XRSS and others.

Fixes: confidential-containers#12


image-rs/src/signature/mechanism/simple/mod.rs
```diff
@@ -10,6 +10,17 @@ use serde::*;
...
+use base64::Engine;
+
+#[cfg(feature = "signature-simple-xrss")]
+use oci_distribution::Reference;
+#[cfg(feature = "signature-simple-xrss")]
+use reqwest::Client;
+
+#[cfg(feature = "signature-simple-xrss")]
+use crate::signature::image::digest::Digest;
+
```
Xynnn007, line 23: There are many `#[cfg(feature = "signature-simple-xrss")]` sentences in this file. If these feature-specific sentences can be put into a separate file, it can be better for reading and maintaining. We only need to mark the `mod xrss;` sentence with `#[cfg(feature = "signature-simple-xrss")]`.

Xynnn007 (commented):
Thanks @mattarnoatibm for the contribution! Could you do some refactoring to the code for better maintainance in future?

image-rs/src/signature/mechanism/simple/mod.rs
```diff
@@ -177,27 +234,142 @@ pub fn judge_single_signature(
...
+                            e
+                        )
+                    })?;
+
+                Ok(oauth_token_response_body.token)
+            }
+        }
+    }
+
+    pub async fn get_signatures_from_registry(
```
arronwy, line 285: maybe we can send the PR in parallel to https://github.com/krustlet/oci-distribution to extend the `Client` API to support pull signatures from registry

image-rs/src/signature/mechanism/simple/mod.rs
mattarnoatibm: Hi @Xynnn007 I moved most of the code in `simple/mod.rs` that was behind the `signature-simple-xrss` feature to a new file, `xrss.rs`. I marked 2 sentences with `#[cfg(feature = "signature-simple-xrss")]` in addition to the `mod xrss;` sentence in `mod.rs`:

If the XRSS feature is used then `crate::resource::get_resource(&config.sigstore_config).await` can return an error, otherwise we cannot allow it to return an error, so I left `if sig_store_config_file.is_err() { return Ok(()); }` marked with `#[cfg(feature = "signature-simple-xrss")]`.

If XRSS is not used then we don't want to create an `xrss::RegistryClient` and try to get signatures from the registry. So I created a code block where the  `xrss::RegistryClient` is created and used to try and get signatures from the registry, and I marked that code block with `#[cfg(feature = "signature-simple-xrss")]`.

image-rs/src/signature/mechanism/simple/mod.rs
```diff
@@ -219,8 +243,13 @@ impl SimpleParameters {
...
 
-        let sigs = sigstore::get_sigs_from_specific_sigstore(sigstore_uri).await?;
+        let mut sigstore_sigs = sigstore::get_sigs_from_specific_sigstore(sigstore_uri).await?;
+        sigs.append(&mut sigstore_sigs);
 
         Ok(sigs)
     }
 }
+
+pub async fn new_sigs_vec() -> Vec<Vec<u8>> {
```
Xynnn007, line 253: Why not we directly use
```rust
let mut sigs: Vec<Vec<u8>> = Vec::new();
```
in `fn get_signatures()`?

image-rs/src/signature/mechanism/simple/xrss.rs
```diff
@@ -0,0 +1,167 @@
...
+        }
+    }
+
+    pub async fn get_signatures_from_registry(
+        &self,
+        image: &Image,
+        digest: &Digest,
+        auth: &RegistryAuth,
+    ) -> Result<Vec<Vec<u8>>> {
+        let mut sigs: Vec<Vec<u8>> = simple::new_sigs_vec().await;
```
Xynnn007, line 106: Also, why not directly use
```rust
let mut sigs: Vec<Vec<u8>> = Vec::new();
```

image-rs/src/signature/mechanism/simple/xrss.rs
```diff
@@ -0,0 +1,167 @@
...
+            .map_err(|e| {
+                anyhow!(
+                    "Failed to query extensions supported by registry v2 endpoint: {:?}",
+                    e
+                )
+            })?;
+
+        let registry_signature_support_header = res.headers().get("x-registry-supports-signatures");
+        if registry_signature_support_header.is_some()
+            && registry_signature_support_header.unwrap() == "1"
```
Xynnn007, line 122: I noticed that here `unwrap()` is used. Will this never cause panic? If so, I suggest that add `.expect("...")` to the end of line 120.

image-rs/src/signature/mechanism/simple/xrss.rs
```diff
@@ -0,0 +1,167 @@
...
+                    ))
+                    .query(&[
+                        ("account", username),
+                        ("scope", &scope),
+                        ("service", &"registry".to_string()),
+                    ])
+                    .basic_auth(username, Some(password))
+                    .send()
+                    .await
+                    .map_err(|e| anyhow!("Failed to fetch oauth token for registry: {:?}", e))?;
```
Xynnn007, line 85: `anyhow` crate provides a simpler way to add error information context
```suggestion
                    .context("Failed to fetch oauth token for registry")?;
```

image-rs/src/signature/mechanism/simple/xrss.rs
```diff
@@ -0,0 +1,167 @@
...
+
+                let oauth_token_response_body: GetOAuthTokenResponse =
+                    res.json::<GetOAuthTokenResponse>().await.map_err(|e| {
+                        anyhow!(
+                            "Unexpected response from fetching oauth token from registry: {:?}",
+                            e
+                        )
+                    })?;
+
```
Xynnn007, line 94:
```suggestion
                    res.json::<GetOAuthTokenResponse>().await.context(
                            "Unexpected response from fetching oauth token from registry",
                    )?;

```

image-rs/src/signature/mechanism/simple/xrss.rs
```diff
@@ -0,0 +1,167 @@
...
+            .client
+            .get(format!("https://{}/v2/", image.reference.registry()))
+            .send()
+            .await
+            .map_err(|e| {
+                anyhow!(
+                    "Failed to query extensions supported by registry v2 endpoint: {:?}",
+                    e
+                )
+            })?;
```
Xynnn007, line 118:
```suggestion
            .context("Failed to query extensions supported by registry v2 endpoint")?;
```

image-rs/src/signature/mechanism/simple/xrss.rs
```diff
@@ -0,0 +1,167 @@
...
+            let res = self
+                .client
+                .get(format_registry_signatures_extension_url(
+                    &image.reference,
+                    digest,
+                ))
+                .bearer_auth(oauth_token)
+                .send()
+                .await
+                .map_err(|e| anyhow!("Failed to get signatures from registry: {:?}", e))?;
```
Xynnn007, line 134:
```suggestion
                .context("Failed to get signatures from registry")?;
```

image-rs/src/signature/mechanism/simple/xrss.rs
```diff
@@ -0,0 +1,167 @@
...
+
+            let signatures_response_body: GetSignaturesResponse =
+                res.json::<GetSignaturesResponse>().await.map_err(|e| {
+                    anyhow!(
+                        "Unexpected response from fetching signatures from registry: {:?}",
+                        e
+                    )
+                })?;
```
Xynnn007, line 142:
```suggestion
                res.json::<GetSignaturesResponse>().await.context(
                        "Unexpected response from fetching signatures from registry"
                )?;
```

image-rs/src/signature/mechanism/simple/xrss.rs
```diff
@@ -0,0 +1,167 @@
...
+
+            for registry_signature in signatures_response_body.signatures.iter() {
+                let signature = base64::engine::general_purpose::STANDARD.decode(&registry_signature.content).map_err(|e| {
+                    anyhow!(
+                        "Error when decoding sognature from registry. Signature is not base64 encoded: {:?}",
+                        e
+                    )
+                })?;
```
Xynnn007, line 150:
```suggestion
                let signature = base64::engine::general_purpose::STANDARD.decode(&registry_signature.content).context(
                        "Error when decoding sognature from registry. Signature is not base64 encoded"
                )?;
```

image-rs/tests/signature_verification.rs
```diff
@@ -93,13 +118,53 @@ const SIGSTORE_CONFIG_URI: &str = "kbs:///default/sigstore-config/test";
...
+                    "aa-offline_fs_kbc-resources-for-icr.json",
+                    auth_info,
+                )
+                .await;
+            }
+            false => {
+                println!("Skipping xrss test cases because the test cases require authentication and no AUTH is set in the environment");
+            }
+        },
+        Err(_e) => {
```
Xynnn007, line 144: if `_e` is not used, we can directly use `_`

image-rs/src/signature/mechanism/simple/xrss.rs
mattarnoatibm: The response returned by `https://<registry_name>/v2/` may not include the `x-registry-supports-signatures` header so `registry_signature_support_header` can be `None`. So if the registry doesn't support the signatures API extension I would expect the code to not panic and return an empty `Vec`.

E.g., quay.io vs. ICR:

```
curl -v https://quay.io/v2/ 2>&1 | grep x-registry-supports-signatures
curl -v https://uk.icr.io/v2/ 2>&1 | grep x-registry-supports-signatures
< x-registry-supports-signatures: 1
```

image-rs/src/signature/mechanism/simple/mod.rs
mattarnoatibm: When I was trying to build the code with my changes before I got an error, and I thought the cause of the error was the async function `allows_image` was returning a `Vec<Vec<u8>>` that wasn't created itself by an async function, so i added the `new_sigs_vec` async function you saw.

I made the change you suggested and my code still builds.

So the error I got before must have been caused by another error and my original 'fix' wasn't necessary.

image-rs/src/signature/mechanism/simple/xrss.rs
mattarnoatibm: I made this change and the code builds OK.

image-rs/src/signature/mechanism/simple/xrss.rs
mattarnoatibm: Changed `.map_err` to `.context`.

image-rs/src/signature/mechanism/simple/xrss.rs
mattarnoatibm: Changed `.map_err` to `.context`.

image-rs/src/signature/mechanism/simple/xrss.rs
mattarnoatibm: Changed `.map_err` to `.context`.

image-rs/src/signature/mechanism/simple/xrss.rs
mattarnoatibm: Changed `.map_err` to `.context`.

image-rs/src/signature/mechanism/simple/xrss.rs
mattarnoatibm: Changed `.map_err` to `.context`.

image-rs/src/signature/mechanism/simple/xrss.rs
mattarnoatibm: Changed `.map_err` to `.context`.

image-rs/tests/signature_verification.rs
mattarnoatibm: Changed `_e` to `_`.

image-rs/tests/credential.rs
```diff
@@ -19,7 +19,7 @@ pub mod common;
 #[tokio::test]
 #[serial]
 async fn test_use_credential(#[case] image_ref: &str, #[case] auth_file_uri: &str) {
-    common::prepare_test().await;
+    common::prepare_test("aa-offline_fs_kbc-resources.json").await;
```
Xynnn007, line 22: I think it would be better to declare this as a `const`, or we add the parameter to `#[case]` macro. How do you like?

image-rs/tests/image_decryption.rs
```diff
@@ -30,7 +30,7 @@ const OCICRYPT_CONFIG: &str = "test_data/ocicrypt_keyprovider_ttrpc.conf";
 #[tokio::test]
 #[serial]
 async fn test_decrypt_layers(#[case] image: &str) {
-    common::prepare_test().await;
+    common::prepare_test("aa-offline_fs_kbc-resources.json").await;
```
Xynnn007, line 33: I think it would be better to declare this as a `const`, or we add the parameter to `#[case]` macro. How do you like?

image-rs/tests/signature_verification.rs
```diff
@@ -93,13 +118,53 @@ const SIGSTORE_CONFIG_URI: &str = "kbs:///default/sigstore-config/test";
 #[tokio::test]
 #[serial]
 async fn signature_verification() {
-    common::prepare_test().await;
+    do_signature_verification_tests(&_TESTS, "aa-offline_fs_kbc-resources.json", &None).await;
```
Xynnn007, line 121: I think it would be better to declare this as a `const`, or we add the parameter to `#[case]` macro. How do you like?

image-rs/tests/signature_verification.rs
```diff
@@ -93,13 +118,53 @@ const SIGSTORE_CONFIG_URI: &str = "kbs:///default/sigstore-config/test";
...
+#[serial]
+async fn signature_verification_xrss() {
+    match std::env::var("AUTH_PASSWORD") {
+        Ok(auth_password) => match !auth_password.is_empty() {
+            true => {
+                let auth = format!("iamapikey:{}", auth_password);
+                let auth_info = &Some(auth.as_str());
+                do_signature_verification_tests(
+                    &_TESTS_XRSS,
+                    "aa-offline_fs_kbc-resources-for-icr.json",
```
Xynnn007, line 135: I think it would be better to declare this as a `const`, or we add the parameter to `#[case]` macro. How do you like?

Xynnn007 (commented):
One last step for the integration test from myside :-)

image-rs/tests/credential.rs
mattarnoatibm: Done. I went with adding some constants to `common/mod.rs` as the signature verification tests do not use the `#[case]` macro.

Xynnn007 (approved):
lgtm. Thanks!

stevenhorsman:
@arronwy - hey Arron, we are keen to try and get this feature into the 0.8 release before `CCv0` gets archived, so are you okay with this being merged now as Matt has created https://github.com/krustlet/oci-distribution/issues/97 to get the opinions of the oci-distribution community about moving it to there, which we can do as a separate follow-up activity? Thanks!

arronwy:
> @arronwy - hey Arron, we are keen to try and get this feature into the 0.8 release before `CCv0` gets archived, so are you okay with this being merged now as Matt has created [krustlet/oci-distribution#97](https://github.com/krustlet/oci-distribution/issues/97) to get the opinions of the oci-distribution community about moving it to there, which we can do as a separate follow-up activity? Thanks!

sure, I though we already code freeze for v0.8.0 and merge after the release, it's ok from my side to merge for v0.8.0

arronwy (approved):
Thanks @mattarnoatibm

PR #374, opened 2023-10-13
cdh/kms/kbs: raise warning when failed to read file for offline-fs-kbc

Xynnn007:
Before this commit, if the given offline-fs-kbc file is not abled to be read when offline-fs-kbc is being initialized, an error will be raised. This will cause the whole CDH process to exit.

In real scenarios, this would require a user to embed an empty aa-offline_fs_kbc-resources.json and aa-offline_fs_kbc-keys.json to the guest image, which is meaningless.

This patch fixes this.

@stevenhorsman


stevenhorsman (approved):
Thanks Ding for responding to the issue so quickly!

PR #375, opened 2023-10-16
Add az-tdx-vtpm attester

mkulke:
This attester is supposed to procure evidence from Azure TDX CVMs, which are currently in limited preview. The attester uses a flow similar to the az_snp_vtpm module, albeit with TDX Quotes.


attestation-agent/attester/src/lib.rs
```diff
@@ -77,6 +82,11 @@ pub fn detect_tee_type() -> Option<Tee> {
         return Some(Tee::Sgx);
     }
 
+    #[cfg(feature = "az-tdx-vtpm-attester")]
+    if az_tdx_vtpm::detect_platform() {
+        return Some(Tee::AzTdxVtpm);
+    }
```
mkulke, line 88: Note: the order is critical here, because AzSnpVtpm has a weak detection heuristic and will report positive if it's able to pull an HCL report, even if it's a tdx one. We need to provide a better heuristic on az-snp-vtpm to detect azure snp cvms,

Xynnn007 (approved):
LGTM. Only a small comment for dep version

Xynnn007 (approved):
lgtm.

fitzthum (approved):
LGTM

PR #381, opened 2023-10-26
Fix CDH & kbs_protocol

Xynnn007:
When doing tests with aliyun kms and current guest components, we found some bugs and this fix will help.

cc @arronwy @mozillazg


fitzthum (approved):
LGTM

PR #385, opened 2023-10-30
Random key generation

piotrpalcz:
This PR is realted to Shim/rune PR https://github.com/confidential-containers/enclave-cc/pull/256 and enclave-cc PR https://github.com/confidential-containers/enclave-cc/pull/265

Creates a randomly generated key to use in unionFS that is then stored in file and passed to be mounted and read by boot enclave to use that unionFS

### DIFF
--- a/image-rs/src/snapshots/occlum/unionfs.rs
+++ b/image-rs/src/snapshots/occlum/unionfs.rs
@@ -14,6 +14,8 @@ use fs_extra;
 use fs_extra::dir;
 use nix::mount::MsFlags;
 
+use ocicrypt_rs::blockcipher::rand::rand_bytes;
+
 use crate::snapshots::{MountPoint, Snapshotter};
 
 const LD_LIB: &str = "ld-linux-x86-64.so.2";
@@ -47,6 +49,21 @@ fn create_dir(create_path: &Path) -> Result<()> {
     Ok(())
 }
 
+// returns randomly generted random 128 bit key
+fn generate_random_key() -> String {
+    let mut key: [u8; 16] = [0u8; 16];
+
+    rand_bytes(&mut key).expect("Random fill failed");
+
+    let formatted_key = key
+        .iter()
+        .map(|byte| format!("{:02x}", byte))
+        .collect::<Vec<String>>()
+        .join("-");
+
+    formatted_key
+}
+
 fn create_environment(mount_path: &Path) -> Result<()> {
     let mut from_paths = Vec::new();
     let mut copy_options = dir::CopyOptions::new();
@@ -69,7 +86,6 @@ fn create_environment(mount_path: &Path) -> Result<()> {
     if fs::symlink_metadata(ld_lib.as_path()).is_ok() {
         fs::remove_file(ld_lib)?;
     }
-
     fs_extra::copy_items(&from_paths, &path_lib64, &copy_options)?;
     from_paths.clear();
 
@@ -112,6 +128,7 @@ impl Snapshotter for Unionfs {
     fn mount(&mut self, layer_path: &[&str], mount_path: &Path) -> Result<MountPoint> {
         let fs_type = String::from("sefs");
         let source = Path::new(&fs_type);
+        let flags = MsFlags::empty();
 
         if !mount_path.exists() {
             fs::create_dir_all(mount_path)?;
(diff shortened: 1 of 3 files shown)


arronwy:
As described in: https://rust-random.github.io/book/guide-rngs.html#not-a-crypto-library
    "It should be emphasised that this is not a cryptography library"
    "It is therefore recommended to use specialized libraries where possible,
    for example openssl, ring and the RustCrypto libraries."

`image-rs` also have two features `encryption-openssl` and ` encryption-ring`  for them.

piotrpalcz:
> As described in: https://rust-random.github.io/book/guide-rngs.html#not-a-crypto-library "It should be emphasised that this is not a cryptography library" "It is therefore recommended to use specialized libraries where possible, for example openssl, ring and the RustCrypto libraries."
> 
> `image-rs` also have two features `encryption-openssl` and ` encryption-ring` for them.

Updated, but I need to explain :
In issues regarding this matter ( https://github.com/confidential-containers/enclave-cc/issues/20 & https://github.com/confidential-containers/enclave-cc/issues/149) there wasnt specified that this key needs to be cryptographically secure and generated with specific library. I just chose the easiest approach at the time.

Now I updated the branch using already existing in Ocicrypt-rs module rand_bytes function, which does exactly what I needed here and is already being drived by `cfg feature` with default value in its cfg. This though needed me to change it's access level from pub(crate) to pub.

piotrpalcz:
Closing, replacing with https://github.com/confidential-containers/enclave-cc/pull/312

ocicrypt-rs/src/blockcipher/mod.rs
```diff
@@ -12,7 +12,7 @@ use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
 mod aes_ctr;
 use aes_ctr::AESCTRBlockCipher;
 
-mod rand;
+pub mod rand;
```
mythi, line 15: we are getting ready with the work. @arronwy can you comment on this part. looks OK?

image-rs/src/snapshots/occlum/unionfs.rs
```diff
@@ -124,18 +143,48 @@ impl Snapshotter for Unionfs {
...
             .ok_or(anyhow!("Unknown error: file name parse fail"))?;
 
-        // For mounting trusted UnionFS at runtime of occlum,
-        // you can refer to https://github.com/occlum/occlum/blob/master/docs/runtime_mount.md#1-mount-trusted-unionfs-consisting-of-sefss.
-        // "c7-32-b3-ed-44-df-ec-7b-25-2d-9a-32-38-8d-58-61" is a hardcode key used to encrypt or decrypt the FS currently,
-        // and it will be replaced with dynamic key in the near future.
+        let random_key = generate_random_key();
+        let new_key_path = Path::new("/new_key");
+        fs::create_dir_all(new_key_path)?;
+        std::fs::write(&PathBuf::from(&new_key_path.join("key.txt")), &random_key)?;
```
mythi, line 149: could this be:
```suggestion
        std::fs::write(&PathBuf::from(&keys_mount_path.join("key.txt")), &random_key)?;
```
and moved to the end to replace `fs::copy()`

mythi (approved):
LGTM and we have enclave-cc CI passing with this

image-rs/src/image.rs
```diff
@@ -526,7 +526,7 @@ mod tests {
             // Azure Container Registry
             "mcr.microsoft.com/hello-world",
             // Docker container Registry
-            "docker.io/i386/busybox",
+            "docker.io/busybox",
```
mythi, line 529: This is from my earlier commit

piotrpalcz: Its cherry picked, so you remain author. Had some github with that particular rebase but now its all fine.

mythi:
@arronwy does this look OK?

image-rs/src/image.rs
piotrpalcz: Fixed git history now

mythi (approved):

ocicrypt-rs/src/blockcipher/mod.rs
arronwy: yes, it's ok from my side.

image-rs/src/snapshots/occlum/unionfs.rs
```diff
@@ -124,18 +141,41 @@ impl Snapshotter for Unionfs {
...
+            )
+        })?;
+
+        let random_key = generate_random_key();
+        std::fs::write(PathBuf::from(&keys_mount_path.join("key.txt")), &random_key)?;
+        nix::mount::umount(keys_mount_path)?;
+
         let options = format!(
             "dir={},key={}",
             Path::new("/images").join(cid).join("sefs/lower").display(),
```
arronwy, line 175: we can also use `key_mount_options` here.·

arronwy (approved):
Thanks @piotrpalcz LGTM! Just a minor suggest.

PR #388, opened 2023-11-02
aa/attester: Update csv-rs dep to rev 9d8882e.

BaoshunFang:
Update csv-rs to be compatible with openssl-sys 0.9.94.
Signed-off-by: fangbaoshun <fangbaoshun@hygon.cn>


Xynnn007 (approved):
Thanks for the fix

PR #392, opened 2023-11-13
AA: Add API to extend measurement register at runtime

jialez0:
cc @arronwy

### DIFF
--- a/attestation-agent/app/build.rs
+++ b/attestation-agent/app/build.rs
@@ -9,9 +9,17 @@ use ttrpc_codegen::{Codegen, Customize, ProtobufCustomize};
 fn main() -> std::io::Result<()> {
     #[cfg(feature = "grpc")]
     {
-        tonic_build::compile_protos("../protos/keyprovider.proto")?;
-        tonic_build::compile_protos("../protos/getresource.proto")?;
-        tonic_build::compile_protos("../protos/attestation-agent.proto")?;
+        tonic_build::configure()
+            .build_server(true)
+            .protoc_arg("--experimental_allow_proto3_optional")
+            .compile(
+                &[
+                    "../protos/keyprovider.proto",
+                    "../protos/getresource.proto",
+                    "../protos/attestation-agent.proto",
+                ],
+                &["../protos"],
+            )?;
     }
 
     #[cfg(feature = "ttrpc")]
--- a/attestation-agent/app/src/rpc/attestation/mod.rs
+++ b/attestation-agent/app/src/rpc/attestation/mod.rs
@@ -20,7 +20,10 @@ pub mod grpc {
     use attestation::attestation_agent_service_server::{
         AttestationAgentService, AttestationAgentServiceServer,
     };
-    use attestation::{GetEvidenceRequest, GetEvidenceResponse, GetTokenRequest, GetTokenResponse};
+    use attestation::{
+        ExtendRuntimeMeasurementRequest, ExtendRuntimeMeasurementResponse, GetEvidenceRequest,
+        GetEvidenceResponse, GetTokenRequest, GetTokenResponse,
+    };
     use std::net::SocketAddr;
     use tonic::{transport::Server, Request, Response, Status};
 
@@ -81,6 +84,35 @@ pub mod grpc {
             debug!("Get evidence successfully!");
 
             let reply = GetEvidenceResponse { evidence };
+
+            Result::Ok(Response::new(reply))
+        }
+
+        async fn extend_runtime_measurement(
+            &self,
+            request: Request<ExtendRuntimeMeasurementRequest>,
+        ) -> Result<Response<ExtendRuntimeMeasurementResponse>, Status> {
+            let request = request.into_inner();
+
+            let attestation_agent_mutex_clone = Arc::clone(&ASYNC_ATTESTATION_AGENT);
+            let mut attestation_agent = attestation_agent_mutex_clone.lock().await;
+
+            debug!("Call AA to extend runtime measurement ...");
+
+            attestation_agent
+                .extend_runtime_measurement(request.events, request.register_index)
+                .await
+                .map_err(|e| {
+                    error!("Call AA to extend runtime measurement failed: {}", e);
+                    Status::internal(format!(
+                        "[ERROR:{}] AA extend runtime measurement failed: {}",
+                        AGENT_NAME, e
+                    ))
+                })?;
+
+            debug!("Extend runtime measurement successfully!");
+
+            let reply = ExtendRuntimeMeasurementResponse {};
 
             Result::Ok(Response::new(reply))
         }
@@ -174,6 +206,34 @@ pub mod ttrpc {
             let mut reply = attestation_agent::GetEvidenceResponse::new();
             reply.Evidence = evidence;
 
+            ::ttrpc::Result::Ok(reply)
+        }
+
+        async fn extend_runtime_measurement(
+            &self,
+            _ctx: &::ttrpc::r#async::TtrpcContext,
+            req: attestation_agent::ExtendRuntimeMeasurementRequest,
+        ) -> ::ttrpc::Result<attestation_agent::ExtendRuntimeMeasurementResponse> {
+            debug!("Call AA to extend runtime measurement ...");
+
+            let attestation_agent_mutex_clone = ASYNC_ATTESTATION_AGENT.clone();
+            let mut attestation_agent = attestation_agent_mutex_clone.lock().await;
+
+            attestation_agent
+                .extend_runtime_measurement(req.Events, req.RegisterIndex)
+                .await
+                .map_err(|e| {
+                    error!("Call AA to extend runtime measurement failed: {}", e);
+                    let mut error_status = ::ttrpc::proto::Status::new();
+                    error_status.set_code(Code::INTERNAL);
+                    error_status.set_message(format!(
+                        "[ERROR:{}] AA extend runtime measurement failed: {}",
+                        AGENT_NAME, e
+                    ));
+                    ::ttrpc::Error::RpcStatus(error_status)
+                })?;
+
+            let reply = attestation_agent::ExtendRuntimeMeasurementResponse::new();
             ::ttrpc::Result::Ok(reply)
         }
     }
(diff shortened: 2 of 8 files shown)


attestation-agent/app/src/rpc/attestation/mod.rs
```diff
@@ -176,6 +208,34 @@ pub mod ttrpc {
...
+
+            attestation_agent
+                .extend_runtime_measurement(req.Events, req.RegisterIndex)
+                .await
+                .map_err(|e| {
+                    error!("Call AA-KBC to extend runtime measurement failed: {}", e);
+                    let mut error_status = ::ttrpc::proto::Status::new();
+                    error_status.set_code(Code::INTERNAL);
+                    error_status.set_message(format!(
+                        "[ERROR:{}] AA-KBC extend runtime measurement failed: {}",
```
Xynnn007, line 230:
```suggestion
                        "[ERROR:{}] AA extend runtime measurement failed: {}",
```

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -65,6 +65,26 @@ impl Attester for TdxAttester {
...
+        _register_index: Option<u64>,
+    ) -> Result<()> {
+        for event in events {
+            match tdx_attest_rs::tdx_att_extend(&event) {
+                tdx_attest_rs::tdx_attest_error_t::TDX_ATTEST_SUCCESS => todo!(),
+                error_code => {
+                    return Err(anyhow!(
+                        "TDX Attester: Failed to extend RTMR. Error code: {:?}",
+                        error_code
+                    ));
```
Xynnn007, line 81: `bail!()` macro can help reduce the complex `return Err(anyhow!(...))`

```suggestion
                    bail!("TDX Attester: Failed to extend RTMR. Error code: {:?}", error_code);
```

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -65,6 +65,26 @@ impl Attester for TdxAttester {
...
     }
+
+    async fn extend_runtime_measurement(
+        &self,
+        events: Vec<Vec<u8>>,
+        _register_index: Option<u64>,
+    ) -> Result<()> {
+        for event in events {
+            match tdx_attest_rs::tdx_att_extend(&event) {
+                tdx_attest_rs::tdx_attest_error_t::TDX_ATTEST_SUCCESS => todo!(),
```
Xynnn007, line 76: Here the `todo!()` will cause the program panic. We'd better add a sentence here, e.g.

```suggestion
                tdx_attest_rs::tdx_attest_error_t::TDX_ATTEST_SUCCESS => debug!("TDX extend runtime measurement succeeded.");
```

attestation-agent/app/build.rs
```diff
@@ -9,9 +9,17 @@ use ttrpc_codegen::{Codegen, Customize, ProtobufCustomize};
...
+        tonic_build::configure()
+            .build_server(true)
+            .protoc_arg("--experimental_allow_proto3_optional")
+            .compile(
+                &[
+                    "../protos/keyprovider.proto",
+                    "../protos/getresource.proto",
+                    "../protos/attestation-agent.proto",
+                ],
+                &["../protos"],
```
Xynnn007, line 21: Is this needed in the PR?

jialez0: This is because we need set `--experimental_allow_proto3_optional` argument.

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -65,6 +65,28 @@ impl Attester for TdxAttester {
...
             .map_err(|e| anyhow!("Serialize TDX evidence failed: {:?}", e))
     }
+
+    async fn extend_runtime_measurement(
+        &self,
+        events: Vec<Vec<u8>>,
+        _register_index: Option<u64>,
+    ) -> Result<()> {
+        for event in events {
+            match tdx_attest_rs::tdx_att_extend(&event) {
```
Xynnn007, line 75: The underlying `tdx_att_extend` function is defined in https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/dcap_1.16_reproducible/QuoteGeneration/quote_wrapper/tdx_attest/tdx_attest.c#L477-L479

Do we need to specify a structured input of this function. In other words, will a random byte string work as an input here?

mythi:
What is the use-case for this API in CoCo?

arronwy:
> What is the use-case for this API in CoCo?

We need support runtime attestation for sidecar containers and FaaS scenarios, they both need measure the workloads like container image.

fitzthum:
> We need support runtime attestation for sidecar containers and FaaS scenarios, they both need measure the workloads like container image.

Won't sidecar images be measured with signatures or dm-verity hashes like any other image? I'm a little unclear on the use case here as well, although I think we might as well add support in the AA. What do you think will call this API?

arronwy:
> > We need support runtime attestation for sidecar containers and FaaS scenarios, they both need measure the workloads like container image.
> 
> Won't sidecar images be measured with signatures or dm-verity hashes like any other image? I'm a little unclear on the use case here as well, although I think we might as well add support in the AA. What component do you think will call this API? Kata agent? image-rs?

Current CoCo supported use cases requires container images are signed by the CoCo user, we also need support scenarios that sidecar/serving containers are from third party or untrusted CSP, CoCo user may only provide the workload or like AI model, then CoCo user will requires add these container image into the measurement evidence. 

We can call this API in `image-rs` when it processing the image manifests.

mythi:
> > > We need support runtime attestation for sidecar containers and FaaS scenarios, they both need measure the workloads like container image.
> > 
> > 
> > Won't sidecar images be measured with signatures or dm-verity hashes like any other image? I'm a little unclear on the use case here as well, although I think we might as well add support in the AA. What component do you think will call this API? Kata agent? image-rs?
> 
> Current CoCo supported use cases requires container images are signed by the CoCo user, we also need support scenarios that sidecar/serving containers are from third party or untrusted CSP, CoCo user may only provide the workload or like AI model, then CoCo user will requires add these container image into the measurement evidence.

AFAIUI, the policy allows users to define what goes into the pod.

arronwy:
> > > > We need support runtime attestation for sidecar containers and FaaS scenarios, they both need measure the workloads like container image.
> > > 
> > > 
> > > Won't sidecar images be measured with signatures or dm-verity hashes like any other image? I'm a little unclear on the use case here as well, although I think we might as well add support in the AA. What component do you think will call this API? Kata agent? image-rs?
> > 
> > 
> > Current CoCo supported use cases requires container images are signed by the CoCo user, we also need support scenarios that sidecar/serving containers are from third party or untrusted CSP, CoCo user may only provide the workload or like AI model, then CoCo user will requires add these container image into the measurement evidence.
> 
> AFAIUI, the policy allows users to define what goes into the pod.

Yes, CoCo user may allow some sidecar containers owned by third party to run in the pod, current image integrity is protected by singing and these images are not signed by CoCo user, in these scenarios, we need also measure these images.

mythi:
> > > > > We need support runtime attestation for sidecar containers and FaaS scenarios, they both need measure the workloads like container image.
> > > > 
> > > > 
> > > > Won't sidecar images be measured with signatures or dm-verity hashes like any other image? I'm a little unclear on the use case here as well, although I think we might as well add support in the AA. What component do you think will call this API? Kata agent? image-rs?
> > > 
> > > 
> > > Current CoCo supported use cases requires container images are signed by the CoCo user, we also need support scenarios that sidecar/serving containers are from third party or untrusted CSP, CoCo user may only provide the workload or like AI model, then CoCo user will requires add these container image into the measurement evidence.
> > 
> > 
> > AFAIUI, the policy allows users to define what goes into the pod.
> 
> Yes, CoCo user may allow some sidecar containers owned by third party to run in the pod, current image integrity is protected by singing and these images are not signed by CoCo user, in these scenarios, we need also measure these images.

The CoCo policy allows users to define the checksums of each layer belonging to the "container group", including the non-signed sidecars. I guess @danmihai1 can say more on what's possible and what is not.

fitzthum:
I am still a bit skeptical about this use case. In theory sidecar images can be validated via public keys. If anything it might make sense to measure the public keys at runtime (at least if we're using signatures). Either way I think it is fine to add underlying support in the AA.

One more question, though. When we update the runtime measurement, what should we do regarding existing tokens/connections?

fitzthum:
Ok, it seems like this might also help us with measuring InitData/Policy for peer pods. We should try to get this merged and then add on some vTPM based backends.

fyi @stevenhorsman @bpradipt @katexochen @mkulke

attestation-agent/attester/src/lib.rs
```diff
@@ -59,6 +59,16 @@ pub trait Attester {
...
     async fn get_evidence(&self, report_data: Vec<u8>) -> Result<String>;
+
+    /// Extend TEE specific dynamic measurement register
+    /// to enable dynamic measurement capabilities for input data at runtime.
+    async fn extend_runtime_measurement(
+        &self,
+        _events: Vec<Vec<u8>>,
+        _register_index: Option<u64>,
+    ) -> Result<()> {
+        bail!("Unimplement")
```
dcmiddle, line 70:
```suggestion
        bail!("Unimplemented")
```

arronwy (approved):
Thanks @jialez0 LGTM!

attestation-agent/attester/src/tdx/mod.rs
bpradipt: @jialez0 I think the commit message should be updated to reflect that this api is implemented only for TDX.

dcmiddle: @Xynnn007 good point, yes this is the expected struct:
https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/5b320f93e16d0bebf700f2ce4bb5c63c4195bf58/QuoteGeneration/quote_wrapper/tdx_attest/tdx_attest.h#L89-L96

Now the event data itself should also follow a format so that we can verify it.
ACON uses an RTMR event log. @binxing can say more on that. 
I don't think we need that event format in place for this PR, but this is a good reminder that we should start specifying it.

binxing: Yeah, I think a structured input is definitely necessary, otherwise the caller wouldn't have any idea what `tdx_att_extend` expects.

Maybe a bit off the topic. Have we ever considered keeping measurement logs for RTMRs? With logs, we can support more usages such as measuring the order in which the containers are started, or upgrading containers at runtime.

PR #396, opened 2023-11-20
Cargo.lock: update dep

Xynnn007:
- csv version was updated in Cargo.toml but not in Cargo.lock.
- image crate is related to cdh's feature image decryption


fitzthum (approved):
LGTM

PR #398, opened 2023-11-28
Attester: Update CSV evidence format

jialez0:
Add a serial number to the evidence content to identify the HYGON chip ID. This field improves the completeness of CSV evidence, allowing for different release logic to be set for different HYGON chips when setting AS policies. This modification comes from the needs of scene users.

cc @BaoshunFang @Xynnn007

### DIFF
--- a/attestation-agent/attester/src/csv/mod.rs
+++ b/attestation-agent/attester/src/csv/mod.rs
@@ -32,6 +32,8 @@ struct CertificateChain {
 struct CsvEvidence {
     attestation_report: AttestationReport,
     cert_chain: CertificateChain,
+    // Base64 Encoded CSV Serial Number (Used to identify HYGON chip ID)
+    serial_number: Vec<u8>,
 }
 
 #[derive(Debug, Default)]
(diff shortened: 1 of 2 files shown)


attestation-agent/attester/src/csv/mod.rs
```diff
@@ -30,8 +31,9 @@ struct CertificateChain {
 
 #[derive(Serialize, Deserialize)]
 struct CsvEvidence {
-    attestation_report: AttestationReport,
+    attestation_report: String,
     cert_chain: CertificateChain,
+    serial_number: Vec<u8>,
```
Xynnn007, line 36: Could you provide more context about this PR? Change `attestation_report` to `String` may help to present the whole evidence as string, but `serial_number` is still `Vec<u8>`

Xynnn007 (approved):

attestation-agent/attester/src/csv/mod.rs
```diff
@@ -32,6 +32,8 @@ struct CertificateChain {
 struct CsvEvidence {
     attestation_report: AttestationReport,
     cert_chain: CertificateChain,
+    // Base64 Encoded CSV Seriail Number (Used to identify HYGON chip ID)
```
fitzthum, line 35: nit: `Seriail` -> `Serial`

fitzthum (approved):
LGTM

PR #399, opened 2023-11-28
attestion-agent: bump az_snp_vtpm attester version

mkulke:
Updated az_snp_vtpm to v0.4 revision, improved detection heuristic, so it won't conflict with other vTPM attesters.


fitzthum (approved):
LGTM

PR #401, opened 2023-11-30
update peerpod daemon.json path

katexochen:
This is moved to /run.

Part of https://github.com/confidential-containers/cloud-api-adaptor/pull/1606


bpradipt:
@jialez0 @sameo are you ok with this change? This is confined to peer-pods only.

fitzthum (approved):
I'm not sure if this peer pod case was intended as a permanent solution, but I guess we do need this update.

Xynnn007 (approved):

PR #402, opened 2023-12-07
Decode key used for image decryption

chendave:
The key should be base64 encoded if it's from a kbs repository, KMS or file system instead of plain text.

This align with other KBC module, suh as `offline_fs_kbc`.

https://github.com/confidential-containers/guest-components/blob/dec7f10ec2ab1e715c49741188d48065ab6c0feb/attestation-agent/kbc/src/offline_fs_kbc/common.rs#L17-L26


chendave:
talked with @Xynnn007 , this is expected, the key is not base64 encoded.

PR #404, opened 2023-12-11
image-rs: fix image layer ordering

### LINKED ISSUE #403: image-rs: concurrent pulls will mix the layer ordering
When using concurrent pull the resulting order of layers can vary from what's specified in the `rootfs.diff_id` of the oci image config:

https://github.com/confidential-containers/guest-components/blob/dec7f10ec2ab1e715c49741188d48065ab6c0feb/image-rs/src/pull.rs#L119

consequently the layers are mounted in the wrong order by kata-agent and this can lead to undesired behaviour, like a container failing to start or changes in layers not being applied.

it was noticed in peerpods because `library/nginx` was [consistently failing to start with rate of ~20%](https://github.com/confidential-containers/cloud-api-adaptor/pull/1605).

mkulke:
Fixes #403

When using concurrent pull the resulting order of layers can vary from what's specified in the `rootfs.diff_id` of the oci image config. Layers might be mounted in the wrong order and containers are not able to run properly (e.g. `library/nginx`).

The change maintains the order of the provided `layer_descs` in the resulting `Vec<LayerMeta>` and thus the layers are mounted in the proper order.

It has been manually tested w/ 50+ coco containers. The provided test here is disabled by default, because we'll run into rate-limits w/ dockerhub and since it's a race condition it's not guaranteed to fail, either.

### DIFF
--- a/image-rs/src/pull.rs
+++ b/image-rs/src/pull.rs
@@ -6,6 +6,7 @@ use anyhow::{anyhow, bail, Result};
 use futures_util::stream::{self, StreamExt, TryStreamExt};
 use oci_distribution::manifest::{OciDescriptor, OciImageManifest};
 use oci_distribution::{secrets::RegistryAuth, Client, Reference};
+use std::collections::BTreeMap;
 use std::convert::TryFrom;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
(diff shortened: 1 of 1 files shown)


Xynnn007 (approved):
LGTM

jialez0 (approved):

arronwy (approved):
Thanks @mkulke nice catch!, LGTM!

PR #406, opened 2023-12-12
AA/kbs_protocol: fix RCAR handshake protocol

### LINKED ISSUE #366: kbs_protocol does not follow the KBS spec

Xynnn007:
Before this commit, the tee-pubkey is not fully integrity-protected by binding the digest into the evidence. The update of this commit is aligned with the KBS side.

Fixes #366

cc @mythi

### DIFF
--- a/attestation-agent/kbs_protocol/src/client/rcar_client.rs
+++ b/attestation-agent/kbs_protocol/src/client/rcar_client.rs
@@ -11,6 +11,7 @@ use kbs_types::{Attestation, Challenge, ErrorInformation, Request, Response};
 use log::{debug, warn};
 use resource_uri::ResourceUri;
 use serde::Deserialize;
+use serde_json::json;
 use sha2::{Digest, Sha384};
 
 use crate::{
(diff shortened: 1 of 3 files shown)


Xynnn007:
The test failure is as expected because the runtime data changes.

attestation-agent/kbs_protocol/src/client/rcar_client.rs
```diff
@@ -133,8 +134,13 @@ impl KbsClient<Box<dyn EvidenceProvider>> {
 
         debug!("get challenge: {challenge:#?}");
         let tee_pubkey = self.tee_key.export_pubkey()?;
-        let materials = vec![tee_pubkey.k_mod.as_bytes(), tee_pubkey.k_exp.as_bytes()];
-        let evidence = self.generate_evidence(challenge.nonce, materials).await?;
+        let runtime_data = json!({
+            "tee-pubkey": tee_pubkey,
+            "nonce": challenge.nonce,
+        });
```
mythi, line 140: is there a reason why to combine them? My expectation was to have it the same way it's implemented in #369

Xynnn007: no matter Nonce or tee pub key, they all are things whose integrity should be protected by binding the evidence. There must be more scenarios except for the two fields.

We use a json map here to have a standard way to organize them

mythi: this contributes to the hashing problem we've talked about. it's not only what hash algorithm is used but also what data is hashed. this won't work well with Intel Trust Authority which makes me think that perhaps we are going to need AS specific cfg flags to decide how the reportdata needs to be generated by the attesters

Xynnn007: see the runtime data spec https://github.com/confidential-containers/kbs/issues/259

This would be a good start to make runtime data usage more standard

mythi: thanks! I saw it earlier but forgot to take a closer look. I'll do it today.

fitzthum (approved):
LGTM. This seems closely tied to https://github.com/confidential-containers/kbs/issues/259 but I guess we can implement that separately.

Xynnn007:
cc @mythi Is this PR now good to you?

mythi (approved):
+1

PR #407, opened 2023-12-12
Makefile: add platform Makefile to quickly build guest component binaries

### LINKED ISSUE #373: Makefile: support build different artifacts due to given `TEE_PLATFORM`
#### Background

Currently, we have added `confidential-data-hub` to guest-components. At the same time, we reconstructed the original KBC abstraction under AA to CDH as the underlying module to implement interfaces such as `get_resource()`. Here, `offline-fs-kbc` is supported by default. Now, after determining the underlying TEE platform, the equipped KBC components can actually be determined. For example, when we select TDX, the KBCs are `cc_kbc_tdx` + `offline-fs-kbc`. The concrete behavior is determined by `aa_kbc_params`. The same is true for `SNP`, `az-vtpm-snp` and `CCA`. Therefore, the goal of this issue is to add a top-level Makefile, replace the original `KBC` parameter used to build AA with `TEE_PLATFORM`. With this, a caller can easily use a TEE platform name to build AA, CDH and ASR under this architecture.

#### Expected behavior

Take TDX for example.
```shell
git clone https://github.com/confidential-containers/guest-components.git && cd guest-components

# build AA
TEE_PLATFORM=tdx make attestation-agent

# build CDH
TEE_PLATFORM=tdx make confidential-data-hub

# build ASR
make api-server-restful
```

#### Rationale

This would be helpful for `rootfs-builder` in kata-containers, s.t. we do not need much coupled conditional expressions.

Xynnn007:
Currently we are building AA/CDH/ASR with different and complex compiling parameters controls. To make it simpler for default cases, this commit adds a Makefile with the following assumption:

test: libc=musl, KBC=offline_fs_kbc
tdx: libc=gnu, KBC=cc_kbc_tdx
sev: libc=musl, KBC=online_sev_kbc

Fixes #373

Hi @stevenhorsman @Alex-Carter01 @fidencio @arronwy Please check if it is as expected.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,56 @@
+TEE_PLATFORM ?= test
+ARCH ?= $(shell uname -m)
+
+DESTDIR ?= /usr/local/bin
+
+LIBC ?= musl
+KBC ?=
+RESOURCE_PROVIDER ?= kbs
+
+ifeq ($(TEE_PLATFORM), test)
+  KBC = offline_fs_kbc
+else ifeq ($(TEE_PLATFORM), tdx)
+  LIBC = gnu
+  KBC = cc_kbc_tdx
+else ifeq ($(TEE_PLATFORM), sev)
+  KBC = online_sev_kbc
+  RESOURCE_PROVIDER = sev
+endif
+# TODO: Add support for SNP, Az-snp-vtpm, CCA, CSV
+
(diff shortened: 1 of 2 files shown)


Alex-Carter01:
The SEV case makes sense

fitzthum (approved):
Nice.

I think we can add SNP support just with `  KBC = cc_kbc_tdx`

PR #411, opened 2023-12-15
Update CI and ttrpc built proto files

Xynnn007:
1. Some dependency updates would make the build cannot work. Thus we need to trigger the ci when Cargo.lock changes. Cargo.lock has the information of the dependency versions. This would help to review PRs.
2. We are now using rust-protobuf 3.3.0 and ttrpc-compiler 0.6.2 to build ttrpc proto files.
3. Rollback `openssl-src` to make `csv` make-able.


jialez0 (approved):

arronwy (approved):
Thanks @Xynnn007 LGTM!

PR #417, opened 2023-12-22
cocokeyprovider: add support for daemonize

### LINKED ISSUE #185: Daemonize CoCo Keyprovider
As an user of CoCo Keyprovider to encrypt container images I would like to start it with the `-D` option so that the process run in daemon mode and this way I don't need to pass bash's `&` to turn it in background.

Xynnn007:
cc @wainersm 

Fixes #185

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -869,6 +869,7 @@ dependencies = [
  "base64 0.21.5",
  "clap 4.2.7",
  "ctr",
+ "daemonize",
  "env_logger 0.10.1",
  "futures",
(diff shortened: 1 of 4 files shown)


jialez0 (approved):
LGTM. Thanks! BTW, do you think AA also need this feature?

Xynnn007:
> LGTM. Thanks! BTW, do you think AA also need this feature?

This depends on the demands. Now no need for this

fitzthum (approved):
LGTM

PR #418, opened 2023-12-22
attester: add evidence_getter binary

Xynnn007:
This tool would help to get evidence in a tee environment, which would help to do tests.


jialez0 (approved):
Looks good. Maybe put this binary in `tools` or `tests` directory or something like that is better?

Xynnn007:
> Looks good. Maybe put this binary in `tools` or `tests` directory or something like that is better?

It is a binary for `attester` and align with https://doc.rust-lang.org/cargo/reference/cargo-targets.html#binaries

PR #425, opened 2024-01-02
Makefile: add more platforms to Makefile

fitzthum:
Adds SNP, AZ-SNP-vTPM, and AZ-TDX-vTPM to the
top-level Makefile.

@mkulke can you double-check that I did the AZ features correctly?

Also changes the name of the offline_fs_kbc flag
to fs rather than test. See https://github.com/confidential-containers/confidential-containers/issues/184 for reasoning behind this change.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-TEE_PLATFORM ?= test
+TEE_PLATFORM ?= fs
 ARCH ?= $(shell uname -m)
 
 DESTDIR ?= /usr/local/bin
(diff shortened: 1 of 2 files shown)


mkulke (approved):
LGTM, should we have a target for the sample tee?

fitzthum:
> LGTM, should we have a target for the sample tee?

It is always build by default. We might want to have a flag for nothing else to be built. I'm not sure if we support that.

fitzthum:
Ok, I will probably create another PR to enable building only the sample attester. Currently we don't quite have the option to do that. To be clear what we can do currently is build the AA with `sample_kbc`, but what we want is to build the `cc_kbc` withe the sample attester only.

fitzthum:
Actually it was a 2-line fix, so I added it here. Now we can build only the cc_kbc with sample attester with `make TEE_PLATFORM=none`

PR #426, opened 2024-01-02
sample: always enable sample attester

fitzthum:
Rather than setting an environment variable to enable the sample attester, always enable it as a fallback.

See https://github.com/confidential-containers/confidential-containers/issues/184 for reasons 

I don't think it makes much sense to enable the sample attester based on an environment variable. Having it always available as a fallback shouldn't cause any security problems. The KBS/AS shouldn't be fooled by the sample evidence in a real deployment.

### DIFF
--- a/attestation-agent/attester/src/lib.rs
+++ b/attestation-agent/attester/src/lib.rs
@@ -77,45 +77,42 @@ pub trait Attester {
 }
 
 // Detect which TEE platform the KBC running environment is.
-pub fn detect_tee_type() -> Option<Tee> {
-    if sample::detect_platform() {
-        return Some(Tee::Sample);
-    }
-
+pub fn detect_tee_type() -> Tee {
     #[cfg(feature = "tdx-attester")]
     if tdx::detect_platform() {
-        return Some(Tee::Tdx);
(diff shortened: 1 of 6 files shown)


Xynnn007 (approved):
LGTM.

attestation-agent/attester/src/lib.rs
```diff
@@ -78,10 +78,6 @@ pub trait Attester {
 
 // Detect which TEE platform the KBC running environment is.
 pub fn detect_tee_type() -> Option<Tee> {
```
Xynnn007, line 80: Maybe we should also change the result to `Tee` rather than `Option` as at least `sample` will be returned.

fitzthum: Got rid of the option. I kept the interface the same at the level of the EvidenceProvider, but adjusted things up to that.

fitzthum:
> At AS side, it might directly use sample verifier without a explicit approval of the user. The user should define a rule in policy on AS side to ban sample verifier claims.

Yes, I think we should setup the AS/KBS so that the default policy rejects the sample attester. Then for tests we can manually enable it.

PR #428, opened 2024-01-04
Clean out-of-date code and documents of AA.

jialez0:
Due to development history, AA has long relied on assuming the responsibility of Unwrap Key and Get Resource, which is actually not in line with AA's positioning. Now we have a CDH that is more suitable for these features, and it already supports the functions that were originally supported by AA. Therefore, this PR will clean up the redundant code and documents in AA, making AA's positioning clearer: **A user space service focused on Attestation itself**.

**Note:** This PR has removed KBC related code and documentation because we no longer need the KBC module, and the KBC related functions have been implemented by CDH using a more elegant architecture.

cc @Xynnn007 

I will fix the CI error after rebasing #427


Xynnn007:
btw, https://github.com/confidential-containers/guest-components/blob/main/Makefile should be also be considered as it defines `KBC` env when build AA.

PR #429, opened 2024-01-08
aa/cdh: make agent-config path configurable by env

mkulke:
relates-to: https://github.com/confidential-containers/cloud-api-adaptor/issues/1637

Since peerpods will template the agent-config.toml with aa_kbc_params and /etc might be on a read-only volume, we need to make this path configurable.

### DIFF
--- a/attestation-agent/lib/src/token.rs
+++ b/attestation-agent/lib/src/token.rs
@@ -3,10 +3,13 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, Context, Result};
 use kbs_protocol::{evidence_provider::NativeEvidenceProvider, KbsClientBuilder};
+use log::debug;
 use serde::{Deserialize, Serialize};
+use std::env;
 use std::path::Path;
+use std::sync::OnceLock;
 use tokio::fs;
 
 const PEER_POD_CONFIG_PATH: &str = "/run/peerpod/daemon.json";
@@ -16,6 +19,8 @@ struct Message {
     token: String,
     tee_keypair: String,
 }
+
+static KATA_AGENT_CONFIG_PATH: OnceLock<String> = OnceLock::new();
 
 pub(crate) async fn get_kbs_token() -> Result<Vec<u8>> {
     let evidence_provider = Box::new(NativeEvidenceProvider::new()?);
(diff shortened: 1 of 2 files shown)


mkulke:
cc @stevenhorsman @bpradipt

attestation-agent/lib/src/token.rs
```diff
@@ -73,15 +77,19 @@ pub(crate) async fn get_kbc_params_from_config_file() -> Result<String> {
...
-    let agent_config_str = fs::read_to_string("/etc/agent-config.toml")
+    // check env for KATA_AGENT_CONFIG_PATH, fall back to default path
+    let path: &String = KATA_AGENT_CONFIG_PATH.get_or_init(|| {
+        env::var("KATA_AGENT_CONFIG_PATH").unwrap_or_else(|_| "/etc/agent-config.toml".into())
+    });
+
+    let agent_config_str = fs::read_to_string(path)
         .await
-        .map_err(|e| anyhow!("Failed to read /etc/agent-config.toml file: {e}"))?;
+        .context(format!("Failed to read ${path}"))?;
```
Xynnn007, line 87: We might not need `$` here?

attestation-agent/lib/src/token.rs
```diff
@@ -73,15 +77,19 @@ pub(crate) async fn get_kbc_params_from_config_file() -> Result<String> {
         aa_kbc_params: Option<String>,
     }
 
-    // Hard-code agent config path to "/etc/agent-config.toml" as a workaround
-    let agent_config_str = fs::read_to_string("/etc/agent-config.toml")
+    // check env for KATA_AGENT_CONFIG_PATH, fall back to default path
+    let path: &String = KATA_AGENT_CONFIG_PATH.get_or_init(|| {
+        env::var("KATA_AGENT_CONFIG_PATH").unwrap_or_else(|_| "/etc/agent-config.toml".into())
```
Xynnn007, line 82: do you think it is good to `info!` or `debug!` here to tell which agent config is used?

Xynnn007 (approved):
This lgtm. However we might want to let CDH to have its own launch configuration file to set `aa_params` etc. https://github.com/confidential-containers/guest-components/issues/431

In this way, both peerpod and kata-cc could share a same interface for CDH.

mkulke:
> This lgtm. However we might want to let CDH to have its own launch configuration file to set `aa_params` etc. #431
> 
> In this way, both peerpod and kata-cc could share a same interface for CDH.

I think we want to look at a consolidated/consistent way of doing configuration in guest-components eventually, but I would wait for the init-data concept to be finalized to start with that. 

This is more like a band-aid for the moment, since a hardcoded path doesn't work with a setup in which /etc isn't writeable. kata-agent on CCv0 is frozen atm and will be for some months probably, env is a way to pass the configuration from kata-agent to CDH & AA without altering kata-agent itself.

attestation-agent/lib/src/token.rs
mkulke: yup, added 👍

fitzthum (approved):

PR #432, opened 2024-01-11
Fixes mount parameter order in CDH/Storage/OSS

Xynnn007:
The previous parameters of the function is not correct, s.t. using `source` as the target mount path. This commit fixes this.
    
Also, this commit leverages the async lib to handle filesystem and time operations to avoid blocking.
    
The last part is that in the previous version, all mount operations will share a same OSSFS_PASSWD_FILE and GOCRYPTFS_PASSWD_FILE path. This means that the operation cannot be idempotent. This is fixed by create a temp directory under /tmp for every mount operation to store those metadata files.

cc @LindaYu17 
related to https://github.com/kata-containers/kata-containers/pull/7965

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -273,6 +273,7 @@ dependencies = [
  "anyhow",
  "async-trait",
  "az-snp-vtpm",
+ "az-tdx-vtpm",
  "base64 0.21.5",
  "codicon",
  "csv-rs",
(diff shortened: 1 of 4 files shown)


fitzthum (approved):
Nice

LindaYu17:
lgtm, thanks!

Issue #352, opened 2023-09-11
github workflow improvements


mythi:
- [ ] from #350, I can see lots of duplicates. Move to something like `uses: "./.github/workflows/lib-common.yaml"` and `needs: xyz` 
- [ ] from #380, [pin DCAP versions](https://download.01.org/intel-sgx/sgx_repo/ubuntu/apt_preference_files/) to match packages with crate versions

Xynnn007:
cc @jialez0

PR #434, opened 2024-01-16
attester: add TSM REPORT module and move TDX to use it

mythi:
TODO:
- [ ] tests?
- [ ] better commit messages


mythi:
The CI error does not seem to be related to this PR (I can see it locally with `main` as well)

attestation-agent/attester/Cargo.toml
```diff
@@ -26,6 +26,8 @@ codicon = { version = "3.0", optional = true }
 hyper = { version = "0.14", features = ["full"], optional = true }
 hyper-tls = { version = "0.5", optional = true }
 tokio = { version = "1", features = ["full"], optional = true }
+tempfile = { version = "3.9.0", optional = true }
```
Xynnn007, line 29: We can update `tempfile` in the top `Cargo.toml` and refer it here
```suggestion
tempfile = { workspace = true, optional = true }
```

attestation-agent/attester/Cargo.toml
```diff
@@ -38,12 +40,14 @@ required-features = [ "bin" ]
 default = ["all-attesters"]
 all-attesters = ["tdx-attester", "sgx-attester", "az-snp-vtpm-attester", "az-tdx-vtpm-attester", "snp-attester", "csv-attester", "cca-attester"]
 
-tdx-attester = ["tdx-attest-rs"]
+tdx-attester = ["tsm-report", "tdx-deprecated-getquote-ioctl"]
+tdx-deprecated-getquote-ioctl = ["tdx-attest-rs"]
```
Xynnn007, line 44: Is getquote by ioctl already deprecated?

attestation-agent/attester/src/tsm_report/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+pub fn provider_is(wanted: &str) -> bool {
+    if !Path::new(TSM_REPORT_PATH).exists() {
+        return false;
+    }
+
+    match TsmReportPath::open() {
+        Ok(report_path) => {
+            let provider = report_path.provider.clone();
+            report_path.close();
+            provider.cmp(&wanted.to_string()).is_eq()
```
Xynnn007, line 122: Could we directly use
```suggestion
            provider == wanted
```

attestation-agent/attester/src/tsm_report/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+        }
+        Err(_) => false,
```
Xynnn007, line 124: Do we need some log when err of open TsmReport happens?

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -3,17 +3,23 @@
...
+#[cfg(feature = "tdx-deprecated-getquote-ioctl")]
 use tdx_attest_rs;
 
 const CCEL_PATH: &str = "/sys/firmware/acpi/tables/data/CCEL";
 
 pub fn detect_platform() -> bool {
-    Path::new("/dev/tdx-attest").exists() || Path::new("/dev/tdx-guest").exists()
+    provider_is(TSM_REPORT_PROVIDER_TDX)
+        || (cfg!(feature = "tdx-legacy-getquote-ioctl")
+            && (Path::new("/dev/tdx-attest").exists() || Path::new("/dev/tdx-guest").exists()))
```
Xynnn007, line 22: could you use `cfg-if!` to make the conditions more clearer?

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -37,20 +43,38 @@ impl Attester for TdxAttester {
...
+                }
+                Err(e) => {
+                    tsm.close();
+                    return Err(anyhow!("TDX Attester: {}", e));
+                }
+            },
+            #[cfg(feature = "tdx-deprecated-getquote-ioctl")]
+            Err(_) => {
+                let tdx_report_data = tdx_attest_rs::tdx_report_data_t {
+                    d: report_data.as_slice().try_into().unwrap(),
```
Xynnn007, line 60:
```suggestion
                    d: report_data.as_slice().try_into().ok_or_else(|| anyhow!(..))?,
```

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -37,20 +43,38 @@ impl Attester for TdxAttester {
...
+                };
+
+                match tdx_attest_rs::tdx_att_get_quote(Some(&tdx_report_data), None, None, 0) {
+                    (tdx_attest_rs::tdx_attest_error_t::TDX_ATTEST_SUCCESS, Some(q)) => q,
+                    (error_code, _) => {
+                        return Err(anyhow!(
+                            "TDX Attester: Failed to get TD quote using ioctl. Error code: {:?}",
+                            error_code
+                        ));
```
Xynnn007, line 69:
```suggestion
                        bail!(
                            "TDX Attester: Failed to get TD quote using ioctl. Error code: {:?}",
                            error_code
                        );
```

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -37,20 +43,38 @@ impl Attester for TdxAttester {
...
+                    }
+                }
             }
+            #[cfg(not(feature = "tdx-deprecated-getquote-ioctl"))]
+            Err(e) => return Err(anyhow!("TDX Attester: {}", e)),
```
Xynnn007, line 74:
```suggestion
            Err(e) => bail!("TDX Attester: {}", e),
```

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -37,20 +43,38 @@ impl Attester for TdxAttester {
...
-                ));
+        let quote_bytes = match TsmReportPath::open() {
+            Ok(tsm) => match tsm.attestation_report(TsmReportProvider::Tdx(report_data)) {
+                Ok(bytes) => {
+                    tsm.close();
+                    bytes
+                }
+                Err(e) => {
+                    tsm.close();
+                    return Err(anyhow!("TDX Attester: {}", e));
```
Xynnn007, line 54:
```suggestion
                    bail!("TDX Attester: {}", e);
```

Also, could we make the error information more descriptive?

Xynnn007 (commented):
Is TSM report module already supported in kernel `6.7-rc1`? I suggest that we add some comments inside the code. Also, if we want to leverage the ability, the payload builder side of kata-containers would also make some change. Do you have a plan for this?

attestation-agent/attester/src/tdx/mod.rs
mythi: The error covers `TsmReportError`. Would that be enough?

attestation-agent/attester/src/tsm_report/mod.rs
mythi: I though `provider_is()` just silently checks if the capability is available or not. The error can be logged by the user if the open fails when it's really needed.

mythi:
> s TSM report module already supported in kernel `6.7-rc1`? I suggest that we add some comments inside the code. Also, if we want to leverage the ability, the payload builder side of kata-containers would also make some change. Do you have a plan for this?

Linux 6.7 is available so TSM reports exist. I updated the commit message with the link to the ABI spec. There's no immediate additional changes needed on the Kata side other than just to get  6.7 used. I'm collaborating with @fidencio on that.

attestation-agent/attester/src/tdx/mod.rs
mythi: (moved to `bail!()`)

attestation-agent/attester/src/tdx/mod.rs
mythi: is that a good with here as there's no if-else needed?

attestation-agent/attester/Cargo.toml
Xynnn007: Ok. I prefer a name without `deprecated`. But really it does not matter

mythi: I'm OK to drop `deprecated`.

mythi: dropped

fitzthum:
Upstream CI issue is fixed btw. Please rebase at some point.

mythi:
> Upstream CI issue is fixed btw. Please rebase at some point.

done.

attestation-agent/attester/src/tsm_report/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+    }
+    pub fn attestation_report(
+        &self,
+        provider: TsmReportProvider,
+    ) -> Result<Vec<u8>, TsmReportError> {
+        let report_path = self.path.as_path();
+
+        let report_data = match provider {
+            TsmReportProvider::Tdx(inblob) => inblob,
+            TsmReportProvider::Sev(privlevel, inblob) => {
```
Xynnn007, line 66: I found that there is a logic here
- if `attester-tdx` is enabled, check either TSM Report or ioctl dev is supported
- if succeeds, try TSM Report first and ioctl if fails

However, TSM supports both TDX and SNP, so would it be
- if `tsm-report` is detected, go into `tsm-report` submodule and detect which provider is
- use match branch to provider in blob and then get out blob
- else, go legacy routine, s.t. detect ioctl TDX, SNP, ...

Xynnn007 (commented):
Let's disscuss more

attestation-agent/attester/src/tdx/mod.rs
mythi: I'll rework on it a bit

attestation-agent/attester/Cargo.toml
```diff
@@ -38,12 +40,14 @@ required-features = [ "bin" ]
 default = ["all-attesters"]
 all-attesters = ["tdx-attester", "sgx-attester", "az-snp-vtpm-attester", "az-tdx-vtpm-attester", "snp-attester", "csv-attester", "cca-attester"]
 
-tdx-attester = ["tdx-attest-rs"]
+tdx-attester = ["tsm-report", "tdx-getquote-ioctl"]
+tdx-getquote-ioctl = ["tdx-attest-rs"]
```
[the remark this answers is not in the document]
mythi: OK, I'll reorganize them. However, `tsm-report` is not TDX specific and I'm expecting the `sev_guest` based attesters to move it at some stage...

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -37,20 +43,38 @@ impl Attester for TdxAttester {
...
-
-        let engine = base64::engine::general_purpose::STANDARD;
-        let quote = match tdx_attest_rs::tdx_att_get_quote(Some(&tdx_report_data), None, None, 0) {
-            (tdx_attest_rs::tdx_attest_error_t::TDX_ATTEST_SUCCESS, Some(q)) => engine.encode(q),
-            (error_code, _) => {
-                return Err(anyhow!(
-                    "TDX Attester: Failed to get TD quote. Error code: {:?}",
-                    error_code
-                ));
+        let quote_bytes = match TsmReportPath::open() {
```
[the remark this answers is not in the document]
mythi: My idea was to keep `tsm-report` a "must-have" feature dependency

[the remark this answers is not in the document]
Xynnn007: If `tsm-report` is defined as a feature, then developers could choose to enable `tdx` while disable `tsm`. Thus this section will not pass compiling.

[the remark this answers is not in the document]
mythi: Isn't that true for any code that is built with mandatory features disabled?

attestation-agent/attester/src/tsm_report/mod.rs
mythi: > However, TSM supports both TDX and SNP, so would it be

the logic for other implementations would follow the TDX example: sev attester uses `provider_is(SEV`)  to check TSM reports for SEV are available. sev attesters `get_evidence` implements it's own logic for getting the quote using TSM reports (`Sev()` enum + attestation_report + supplemental_data)

mythi:
(rebased to trigger CI but no code changes at this time)

attestation-agent/attester/src/tdx/mod.rs
mythi: I don't have a way to test the deprecated flows so we could leave the change to another PR since it's not related to this PR

mythi: sorry, I took a closer look and looks like there's a minor change in it. I'm changing it back.

attestation-agent/attester/src/tdx/mod.rs
mythi: updated

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -37,20 +68,20 @@ impl Attester for TdxAttester {
...
-                return Err(anyhow!(
-                    "TDX Attester: Failed to get TD quote. Error code: {:?}",
-                    error_code
-                ));
+                match report {
+                    Ok(bytes) => bytes,
+                    Err(e) => bail!("TDX Attester: {}", e),
+                }
             }
+            Err(e) => try_getquote_ioctl_or_fail(e, report_data)?,
```
fitzthum, line 81: I don't like passing `e` into this function. Instead you could probably do something like `Err(e1) => try_getquote().map_err(|e2| anyhow!("{} {}",e1,e2)?`

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -37,20 +68,20 @@ impl Attester for TdxAttester {
...
         };
+        let engine = base64::engine::general_purpose::STANDARD;
+        let quote = engine.encode(quote_bytes);
```
fitzthum, line 84: Can probably just do `BASE64_STANDARD.encode(`

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -66,6 +97,7 @@ impl Attester for TdxAttester {
             .map_err(|e| anyhow!("Serialize TDX evidence failed: {:?}", e))
     }
 
+    #[cfg(feature = "tdx-getquote-ioctl")]
```
fitzthum, line 100: So TSM does not support extending the RTMRs? What happens if someone calls this function without the ioctl feature enabled?

attestation-agent/attester/src/tsm_report/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+                let _ = std::fs::remove_dir(path.as_path())
+                    .map_err(|r| log::error!("Failed to remove TSM Report directory: {}", r));
+                return Err(TsmReportError::Access("provider", e));
+            }
+        };
+
+        Ok(Self { path, provider })
+    }
+    pub fn close(&self) {
+        let _ = std::fs::remove_dir(self.path.as_path())
```
fitzthum, line 55: Wasn't this a temporary directory?

attestation-agent/attester/src/tsm_report/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+        let provider = match std::fs::read_to_string(path.as_path().join("provider")) {
+            Ok(tsm) => tsm.trim_matches('\n').to_string(),
+            Err(e) => {
+                let _ = std::fs::remove_dir(path.as_path())
+                    .map_err(|r| log::error!("Failed to remove TSM Report directory: {}", r));
+                return Err(TsmReportError::Access("provider", e));
+            }
+        };
+
+        Ok(Self { path, provider })
```
fitzthum, line 52: This reads more like a constructor. Maybe this function should be called `new`?

attestation-agent/attester/src/tsm_report/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+impl TsmReportPath {
+    pub fn open() -> Result<Self, TsmReportError> {
+        let p = tempdir_in(TSM_REPORT_PATH).map_err(TsmReportError::Open)?;
+
+        let path = p.into_path();
+
+        let provider = match std::fs::read_to_string(path.as_path().join("provider")) {
+            Ok(tsm) => tsm.trim_matches('\n').to_string(),
+            Err(e) => {
+                let _ = std::fs::remove_dir(path.as_path())
```
fitzthum, line 46: I think tempdir will be dropped when it goes out of scope. No need to manually delete this.

attestation-agent/attester/src/tsm_report/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+}
+
+impl TsmReportPath {
+    pub fn open() -> Result<Self, TsmReportError> {
+        let p = tempdir_in(TSM_REPORT_PATH).map_err(TsmReportError::Open)?;
+
+        let path = p.into_path();
+
+        let provider = match std::fs::read_to_string(path.as_path().join("provider")) {
+            Ok(tsm) => tsm.trim_matches('\n').to_string(),
```
fitzthum, line 44: nit: `tsm` -> `p`

attestation-agent/attester/src/tsm_report/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+    let generation = g.trim_matches('\n').to_string().parse::<u32>().unwrap();
+
+    if generation > 1 {
+        return Err(TsmReportError::InblobConflict(generation));
+    }
+
+    Ok(())
+}
+
+pub fn provider_is(wanted: &str) -> bool {
```
fitzthum, line 113: This function name might be a bit too generic, especially if we're using it outside of this file. Can we be clear that it is checking whether a provider is being reported by TSM interface.

attestation-agent/attester/src/tsm_report/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+//
+
+use log::error;
+use std::path::{Path, PathBuf};
+use tempfile::tempdir_in;
+use thiserror::Error;
+
+const TSM_REPORT_PATH: &str = "/sys/kernel/config/tsm/report";
+pub const TSM_REPORT_PROVIDER_TDX: &str = "tdx_guest";
+pub const TSM_REPORT_PROVIDER_SEV: &str = "sev_guest";
```
fitzthum, line 13: Can we use an enum for these?

attestation-agent/attester/src/tsm_report/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+        let aux = std::fs::read(report_path.join("auxblob"))
+            .map_err(|e| TsmReportError::Access("auxblob", e))?;
+
+        check_inblob_write_race(report_path)?;
+
+        Ok(aux)
+    }
+}
+
+fn check_inblob_write_race(report_path: &std::path::Path) -> Result<(), TsmReportError> {
```
fitzthum, line 100: I think this function isn't quite right. I think we should be treating this generation field like a sequence number and making sure that it is always the value that we expect (otherwise someone could have written to it which could potentially result in us getting a report with the wrong data). Here we're just checking if it is greater than one.

The attestation-agent can actually end up requesting multiple reports (mainly if the connection to the KBS expires) which I think would break in this case.

fitzthum (commented):
Thanks. I'm hoping we can use TSM for SNP as well.

I left a few comments. I think you can probably simplify a lot how the features are used.

attestation-agent/attester/src/tsm_report/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+pub struct TsmReportPath {
+    path: PathBuf,
+    pub provider: String,
+}
+
+impl TsmReportPath {
+    pub fn open() -> Result<Self, TsmReportError> {
+        let p = tempdir_in(TSM_REPORT_PATH).map_err(TsmReportError::Open)?;
+
+        let path = p.into_path();
```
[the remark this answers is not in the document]
mythi: This essentially removes the default `Drop` implemented by `tempdir`. We cannot use it because it does not work but leaks those directory resources.

attestation-agent/attester/Cargo.toml
```diff
@@ -44,7 +45,8 @@ all-attesters = ["tdx-attester", "sgx-attester", "az-snp-vtpm-attester", "az-tdx
 # quotes. It's an unconditional dependency for tdx-attester since that is the only way to
 # generate TDX quotes with upstream kernels.
 tsm-report = ["tempfile"]
-tdx-attester = ["tdx-attest-rs"]
+tdx-attester = ["tsm-report", "tdx-getquote-ioctl"]
```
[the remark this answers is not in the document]
mythi: Some of this is explained in the comment: with this PR we make the TSM as the primary mechanism without the option to disable it. `tdx-getquote-ioctl` is used during the transition to until we get 6.7 updated but them I'm looking to disable it in our default builds. However, there might still be users who need the out-of-tree patches so they can enabled it in their builds.

The reason for the separate feature is that when disabled, we also get to drop `tdx-attest-rs` from the builds which is a maintenance nightmare: see #352

attestation-agent/attester/src/tsm_report/mod.rs
mythi: Yes but the automatic removal is broken (details: with configfs, only mkdir/rmdir works but `tempdir` uses `remove_dir_all()` which first tries to remove files in the tmpdir and that returns an error. so, `tempdir` just silently ignores the error and leaks the tempdir)

attestation-agent/attester/src/tsm_report/mod.rs
mythi: I paired the name with open/close but I'm able to get rid of close() so will change this to `new`

attestation-agent/attester/src/tsm_report/mod.rs
mythi: answered in other places, resolving.

attestation-agent/attester/src/tsm_report/mod.rs
mythi: I was wondering the same, will update

attestation-agent/attester/src/tsm_report/mod.rs
mythi: `generation` is a read-only attribute and the kernel increments it when a write to `inblob` happens. With the `> 1` comparison the logic is very simple: only one quote per `TsmReportPath` instance can be generated. Each new instance gets a new `tempdir` and it is removed when the instance goes out of scope.

attestation-agent/attester/src/tdx/mod.rs
mythi: true. I did not change lines that I could just move from the original implementation but I can change this.

attestation-agent/attester/src/tsm_report/mod.rs
mythi: changed to an enum.

attestation-agent/attester/src/tdx/mod.rs
mythi: currently no, I believe the fall back is the trait default?

attestation-agent/attester/src/tdx/mod.rs
mythi: looks like `cc_eventlog` uses it too so I'm keeping it as is.

attestation-agent/attester/src/tsm_report/mod.rs
fitzthum: Hm I guess that checks out. It would be good to add a comment explaining this function.

attestation-agent/attester/src/tsm_report/mod.rs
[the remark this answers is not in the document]
fitzthum: What do you mean when you say that this "removes" the default drop? Are we keeping path in scope by putting it into the struct? Won't it eventually get dropped anyway?

attestation-agent/attester/src/tsm_report/mod.rs
mythi: added

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -3,17 +3,48 @@
...
+    if cfg!(feature = "tdx-getquote-ioctl") {
+        Path::new("/dev/tdx-attest").exists() || Path::new("/dev/tdx-guest").exists()
+    } else {
+        false
+    }
+}
+
+cfg_if::cfg_if! {
+    if #[cfg(feature = "tdx-getquote-ioctl")] {
+        fn try_getquote_ioctl_or_fail(_e: TsmReportError, report_data: Vec<u8>) -> Result<Vec<u8>> {
```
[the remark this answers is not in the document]
mythi: The downside of just one function is that it needs to handle unused variables but I've now changed to one.

portersrc (approved):
lgtm

attestation-agent/attester/src/tsm_report/mod.rs
```diff
@@ -0,0 +1,148 @@
...
+        // and leaks the created path. We implement our own Drop that removes the
+        // path (rmdir way) when TsmReportPath instance goes out of scope.
+        let path = p.into_path();
+
+        let provider = match std::fs::read_to_string(path.as_path().join("provider")) {
+            Ok(p) => match p.trim_matches('\n') {
+                TSM_REPORT_PROVIDER_TDX => TsmReportProvider::Tdx,
+                TSM_REPORT_PROVIDER_SEV => TsmReportProvider::Sev,
+                &_ => TsmReportProvider::Unknown,
+            },
```
fitzthum, line 76: Might be more idiomatic to implement this as a to_string method on the TsmReportProvider struct. Also you can probably use the crate `strum` to avoid having these constants.

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -5,22 +5,61 @@
...
+                d: report_data.as_slice().try_into().unwrap(),
+            };
+
+            match tdx_attest_rs::tdx_att_get_quote(Some(&tdx_report_data), None, None, 0) {
+                (tdx_attest_rs::tdx_attest_error_t::TDX_ATTEST_SUCCESS, Some(q)) => Ok(q),
+                (error_code, _) => Err(anyhow!("TDX getquote ioctl also failed (error code: {:?}) after:", error_code)),
+            }
+        } else {
+            // return empty error
+            Err(anyhow!(""))
```
fitzthum, line 56: Should this be something like "TDX Ioctl not supported?" Also, can you move this to the top of the function inside one small cfg_if?

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -44,22 +83,18 @@ impl Attester for TdxAttester {
...
         report_data.resize(TDX_REPORT_DATA_SIZE, 0);
 
-        let tdx_report_data = tdx_attest_rs::tdx_report_data_t {
-            // report_data.resize() ensures copying report_data to
-            // tdx_attest_rs::tdx_report_data_t cannot panic.
-            d: report_data.as_slice().try_into().unwrap(),
-        };
+        let quote_bytes = TsmReportPath::new(TsmReportProvider::Tdx).map_or_else(
+            |notsm| {
+                get_quote_ioctl(&report_data).map_err(|e| anyhow!("TDX Attester: {}{}", e, notsm))
```
fitzthum, line 88: I think you want `.context` here rather than `map_err` since the error you're already getting an `Anyhow::result`

attestation-agent/attester/Cargo.toml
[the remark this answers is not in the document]
fitzthum: I am all in favor of getting rid of the dcap dependency, but I would have just made those changes directly in the source when they were feasible rather than introducing a feature for it.

fitzthum:
This is looking good @mythi. I left a couple more comments.

attestation-agent/attester/src/tdx/mod.rs
mythi: fixed

attestation-agent/attester/src/tsm_report/mod.rs
mythi: fixed

attestation-agent/attester/src/tdx/mod.rs
mythi: > Also, can you move this to the top of the function inside one small cfg_if?

I don't understand what to do here. you mean have two separate cfg_ifs?

mythi:
> This is looking good @mythi. I left a couple more comments.

@fitzthum great comments, thanks!

attestation-agent/attester/src/tdx/mod.rs
fitzthum: I was thinking you could do it with just one, but actually I guess you need to avoid using the ioctl deps when they are not enabled.

fitzthum (approved):
LGTM. Let's make sure @Xynnn007 has another look at it

attestation-agent/attester/Cargo.toml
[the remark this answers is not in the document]
mythi: I thought about it too but this gives the flexibility for CoCo upstream to move faster to configfs-tsm based setups while some downstreams that are using the TDX "out-of-tree" work can keep their deprecated ioctl getquote working.

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -2,25 +2,51 @@
...
 
+#[cfg(feature = "tdx-dcap-ioctl")]
+use tdx_attest_rs::{self, tdx_report_t};
+
 const TDX_REPORT_DATA_SIZE: usize = 64;
 const CCEL_PATH: &str = "/sys/firmware/acpi/tables/data/CCEL";
 
+fn tdx_getquote_ioctl_is_available() -> bool {
+    if cfg!(feature = "tdx-dcap-ioctl") {
+        Path::new("/dev/tdx-attest").exists() || Path::new("/dev/tdx-guest").exists()
```
Lu-Biao, line 22: I have tried this PR with kernel 5.19, and it works well. but 
for kernel 6.x, the device path changed to `/dev/tdx_guest`, 
do you consider add this check for kernel 6.x?

attestation-agent/attester/Cargo.toml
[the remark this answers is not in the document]
mythi: But for the record: with the addition of `check_init_data` things changed a bit since GET_REPORT *is* supported by the upstream ioctl. I'm also OK to drop that additional feature for now and add it later once the upstream story for the ioctl future is clear.

attestation-agent/attester/src/tsm_report/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+// check_inblob_write_race checks that the returned outblob/auxblob
+// matches the quote generation request originally triggered when
+// inblob was written by the TsmReportPath instance. It prevents
+// the race condition that someone else could use the same temporary
+// directory to generate a quote.
+fn check_inblob_write_race(report_path: &std::path::Path) -> Result<(), TsmReportError> {
+    let g = std::fs::read_to_string(report_path.join("generation"))
+        .map_err(|e| TsmReportError::Access("generation", e))?;
+
+    let generation = g.trim_matches('\n').to_string().parse::<u32>().unwrap();
```
Lu-Biao, line 131: do you consider using 'map_err' instead of 'unwrap'?

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -2,25 +2,51 @@
...
+}
+
+#[allow(unused_variables)]
+fn get_quote_ioctl(report_data: &Vec<u8>) -> Result<Vec<u8>> {
+    cfg_if::cfg_if! {
+        if #[cfg(feature = "tdx-dcap-ioctl")] {
+            let tdx_report_data = tdx_attest_rs::tdx_report_data_t {
+                // report_data.resize() ensures copying report_data to
+                // tdx_attest_rs::tdx_report_data_t cannot panic.
+                d: report_data.as_slice().try_into().unwrap(),
```
Lu-Biao, line 35: do you consider using 'map_err' instead of 'unwrap'?

attestation-agent/attester/src/tdx/mod.rs
mythi: adding support for `/dev/tdx_guest` belongs to a separate PR. Let me prepare that

attestation-agent/attester/src/tdx/mod.rs
mythi: It's like this on purpose. See the comment and discussion in #452

attestation-agent/attester/Cargo.toml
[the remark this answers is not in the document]
mythi: dropped all the `cfg-if` for now

attestation-agent/attester/src/tsm_report/mod.rs
mythi: fixed

mythi:
> LGTM. Let's make sure @Xynnn007 has another look at it

@Xynnn007 there was one minor lint error after my updates yesterday but this should be good to go now.

arronwy (approved):
Thanks @mythi LGTM!

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -98,10 +111,12 @@ impl Attester for TdxAttester {
     }
 
     async fn check_init_data(&self, init_data: &[u8]) -> Result<()> {
+        use crate::utils::pad;
+        use scroll::Pread;
```
Xynnn007, line 115: Is this change related to the PR?

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -2,14 +2,10 @@
...
 // SPDX-License-Identifier: Apache-2.0
 //
-
-use crate::utils::pad;
-
+use super::tsm_report::*;
 use super::Attester;
 use anyhow::*;
 use base64::Engine;
-use log::debug;
```
Xynnn007, line 11: Is this change related to the PR?

Xynnn007 (approved):
Overall, lgtm. Some coding things to think about

attestation-agent/attester/src/tdx/mod.rs
mythi: it was needed when I had the `tdx-attest-rs` disabled but I can restore.

attestation-agent/attester/src/tdx/mod.rs
mythi: it was needed when I had the tdx-attest-rs disabled but I can restore. OTOH, it makes the code consistent with other places where `log::` is used in this file.

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -44,22 +60,19 @@ impl Attester for TdxAttester {
...
-            d: report_data.as_slice().try_into().unwrap(),
-        };
+        let quote_bytes = TsmReportPath::new(TsmReportProvider::Tdx).map_or_else(
+            |notsm| {
+                get_quote_ioctl(&report_data)
+                    .with_context(|| format!("TDX Attester: get_quote_ioctl() fallback failed after a TSM report error ({notsm})"))
+            },
+            |tsm| {
+                tsm.attestation_report(TsmReportData::Tdx(report_data.clone()))
+                    .map_err(|e| anyhow!("TDX Attester: {}", e))
```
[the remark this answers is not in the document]
mythi: good suggestion. as it is right now would be consistent with how it's done in all other places but I fine changing it here and we can fix other places later

attestation-agent/attester/src/tsm_report/mod.rs
```diff
@@ -0,0 +1,163 @@
...
+
+        Ok(aux)
+    }
+}
+
+// check_inblob_write_race checks that the returned outblob/auxblob
+// matches the quote generation request originally triggered when
+// inblob was written by the TsmReportPath instance. It prevents
+// the race condition that someone else could use the same temporary
+// directory to generate a quote.
```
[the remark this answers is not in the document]
mythi: is this useful or non `pub` functions?

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -44,22 +60,19 @@ impl Attester for TdxAttester {
...
-            // tdx_attest_rs::tdx_report_data_t cannot panic.
-            d: report_data.as_slice().try_into().unwrap(),
-        };
+        let quote_bytes = TsmReportPath::new(TsmReportProvider::Tdx).map_or_else(
+            |notsm| {
+                get_quote_ioctl(&report_data)
+                    .with_context(|| format!("TDX Attester: get_quote_ioctl() fallback failed after a TSM report error ({notsm})"))
+            },
+            |tsm| {
+                tsm.attestation_report(TsmReportData::Tdx(report_data.clone()))
```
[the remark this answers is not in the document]
mythi: the short answer is 'yes' but I'm open to brainstorm how this gets passed to `attestation_report()`

attestation-agent/attester/src/tdx/mod.rs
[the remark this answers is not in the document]
mythi: fixed

attestation-agent/attester/src/tsm_report/mod.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    if generation > 1 {
+        return Err(TsmReportError::InblobConflict(generation));
+    }
+
+    Ok(())
+}
+
+fn check_tsm_report_provider(
+    report_path: &std::path::Path,
```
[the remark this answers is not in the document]
mythi: fixed

attestation-agent/attester/src/tsm_report/mod.rs
```diff
@@ -0,0 +1,163 @@
...
+        Ok(aux)
+    }
+}
+
+// check_inblob_write_race checks that the returned outblob/auxblob
+// matches the quote generation request originally triggered when
+// inblob was written by the TsmReportPath instance. It prevents
+// the race condition that someone else could use the same temporary
+// directory to generate a quote.
+fn check_inblob_write_race(report_path: &std::path::Path) -> Result<(), TsmReportError> {
```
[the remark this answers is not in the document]
mythi: fixed

attestation-agent/attester/src/tdx/mod.rs
mythi: fixed

attestation-agent/attester/src/tdx/mod.rs
mythi: fixed

attestation-agent/attester/src/tsm_report/mod.rs
```diff
@@ -0,0 +1,163 @@
...
+
+pub enum TsmReportData {
+    Tdx(Vec<u8>),
+    Sev(u8, Vec<u8>),
+}
+
+// TsmReportPath instance represents a unique path on ConfigFS
+// provided by the TSM_REPORT attestation ABI. Currently, each
+// instance is a one-shot attestation request and the path is
+// automatically removed when the instance goes out of scope.
```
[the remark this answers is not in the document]
mythi: fixed

mythi:
Added some tests too.

attestation-agent/attester/src/tsm_report/mod.rs
```diff
@@ -0,0 +1,252 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::fs::write;
+    use tempfile::tempdir;
+
+    #[test]
+    fn test_tsm_report_provider() {
+        struct TestData<'a> {
```
Xynnn007, line 173: can use `rstest` to make the test cases more dry.

mythi: I can look into it but as it is right now is consistent with other test implementations under attesters.

attestation-agent/attester/src/tsm_report/mod.rs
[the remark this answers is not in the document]
Xynnn007: This is similar to a coding convention as we usually make the `//` comments to `///`. Like the rust official crate https://doc.rust-lang.org/src/alloc/vec/mod.rs.html#2394-2397 also does this upon non-pub funcs.

attestation-agent/attester/src/tsm_report/mod.rs
Xynnn007: Where did you see that? I found no test impls under attesters using this. btw, we are trying to using rstest in multiple places like https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/deps/crypto/src/native/aes256ctr.rs#L31-L47

But it is ok to leave it as-is. It's not a critical issue.

mythi: Let me try with `rstest`. It looks promising, IMO. I only checked that attester `Cargo.toml` does not add it which suggests attester tests do not use it. I also had some existing "template" for this so that is why I ended up doing it like this

attestation-agent/attester/src/tsm_report/mod.rs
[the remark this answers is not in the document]
mythi: makes sense, will fix

attestation-agent/attester/src/tsm_report/mod.rs
mythi: should be a bit better now, thanks!

mythi:
@fitzthum FYI, this has all review comments addressed and approvals from @arronwy and @Xynnn007

mythi:
rebased to get CI green

fitzthum:
Ok, sorry for the delays on this. Took a while since it touches some important things. We may need some follow-up work (for other platforms, for instance), but this seems like a good start.

mythi:
> We may need some follow-up work (for other platforms, for instance), but this seems like a good start.

Agree that follow-up work is needed. Do you think it would be good to have an issue about it? I'm counting at least 4 future work items depending on how they land upstream: SEV, SEV SVSM, RTMR writes, maybe TDX get-report and ioctl deprecation.

Issue #435, opened 2024-01-16
How to build CDH with only RESOURCE_PROVIDER and no KMS


bpradipt:
Currently CDH will build all KMSes if PROVIDER is empty - https://github.com/confidential-containers/guest-components/tree/main/confidential-data-hub#supported-features

Is there a way to build CDH without any KMSes and use only KBS. 
For example something like `make RESOURCE_PROVIDER=kbs PROVIDER=none`

cc @Xynnn007 @fitzthum @arronwy

fitzthum:
This seems like a good idea. 

I also think we might want to rename the `PROVIDER` config to make things a bit more clear.

bpradipt:
> This seems like a good idea.
> 
> I also think we might want to rename the `PROVIDER` config to make things a bit more clear.

Probably rename PROVIDER to KMS to align with the purpose

PR #437, opened 2024-01-17
build: Rename the feature flag and set default

bpradipt:
Rename the feature flag to align with the feature and provide for build time customisations of the options

Fixes: #435

### DIFF
--- a/confidential-data-hub/Makefile
+++ b/confidential-data-hub/Makefile
@@ -19,8 +19,8 @@ SOURCE_ARCH := $(shell uname -m)
 ARCH ?= $(shell uname -m)
 DEBUG ?=
 LIBC ?= gnu
-RESOURCE_PROVIDER ?=
-PROVIDER ?=
+RESOURCE_PROVIDER ?= kbs,sev
+KMS_PROVIDER ?= aliyun,ehsm
 DESTDIR ?= $(PREFIX)/bin
 RUSTFLAGS_ARGS ?=
 features ?=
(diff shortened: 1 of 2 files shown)


bpradipt:
@fitzthum @Xynnn007 I made an attempt to modify the Makefile to allow build time customisations. PTAL

confidential-data-hub/Makefile
```diff
@@ -19,8 +19,8 @@ SOURCE_ARCH := $(shell uname -m)
 ARCH ?= $(shell uname -m)
 DEBUG ?=
 LIBC ?= gnu
-RESOURCE_PROVIDER ?=
-PROVIDER ?=
+RESOURCE_PROVIDER ?= kbs,sev
```
Xynnn007, line 22: This is awaward for Makefile. If we does not want any `RESOURCE_PROVIDER`
```shell
make RESOURCE_PROVIDER=
```
will still set `RESOURCE_PROVIDER` to `kbs,sev`, because in Makefile an empty-value variable is equal to no-this-variable


The case is reasonable, because `kbs` feature only enables the ability of connecting to CoCoKBS. If we only want use legacy `offline-fs-kbc`, we does not need any features for `RESOURCE_PROVIDER`

confidential-data-hub/README.md
```diff
@@ -31,14 +39,16 @@ Confidential resource providers (flag `RESOURCE_PROVIDER`)
 | kbs                 | For TDX/SNP/Azure-SNP-vTPM based on KBS Attestation Protocol       |
 | sev                 | For SEV based on efi secret pre-attestation                        |
 
-Note: `offline-fs` is built-in, we do not need to manually enable. If no `RESOURCE_PROVIDER`
-is given, all features will be enabled.
+Note:
+- `offline-fs` is built-in, we do not need to manually enable. 
+- If no `RESOURCE_PROVIDER` flag is given, then all the resource providers will be enabled by default
```
Xynnn007, line 44: As mentioned, we should also delete this sentence (my fault, I did not delete it before).

Xynnn007 (commented):
Thanks @bpradipt for the fix. It now looks much more clearer! Only a comment FYI

confidential-data-hub/Makefile
bpradipt: Are you ok with the default RESOURCE_PROVIDER set to "kbs,sev" ?
Or you prefer a different default ?

Xynnn007: It sounds good

fitzthum (approved):
LGTM

PR #438, opened 2024-01-18
AA/kbs_protocol: fix the RCAR handshake unit test

Xynnn007:
There are some changes upon KBS side.

https://github.com/confidential-containers/kbs/pull/265/files#diff-88f80dee1e5f367cb319573b2d524906c5631100a0a3ce6cc62bf2ebf62fb251L95 replaces token type to a token config, and we do not have a default token config builder thus we need to explicitly add the attestation_token_config.

Also, the commit
https://github.com/confidential-containers/kbs/commit/43d56f3a4a92a1cc691f63a8e1311bcc0d2b3fc8 will block the sample_attester by the default policy of KBS when verifying the CoCoAS token. This is due to security. But in this test, we leverage sample-attester, so we need a policy that allows this.

The commit
https://github.com/confidential-containers/guest-components/pull/426/commits/006e1ffe69f2216c30eadb3094ee49fa209e6b2a enables sample_attester without setting environment variable.

This PR will resolve the blocker in #434 and #436

cc @fitzthum

### DIFF
--- a/attestation-agent/kbs_protocol/src/client/rcar_client.rs
+++ b/attestation-agent/kbs_protocol/src/client/rcar_client.rs
@@ -302,8 +302,10 @@ mod test {
         // a start script
         let mut start_kbs_script = env::current_dir().expect("get cwd");
         let mut kbs_config = start_kbs_script.clone();
+        let mut policy = start_kbs_script.clone();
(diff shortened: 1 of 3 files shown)


fitzthum (approved):
Thanks. I guess we missed this at first because it's the other repo. That's an interesting dependency.

PR #439, opened 2024-01-18
AA: Support CoCo-AS Attestation Token

jialez0:


attestation-agent/lib/src/lib.rs
```diff
@@ -12,12 +12,12 @@ use async_trait::async_trait;
...
+use std::{collections::HashMap, path::Path};
 
-#[cfg(feature = "cc_kbc")]
-mod token;
-#[cfg(feature = "cc_kbc")]
-use token::get_kbs_token;
+mod config;
+
+#[cfg(any(feature = "coco_as", feature = "kbs_as"))]
+use token::GetToken;
```
Xynnn007, line 20: Maybe we could have an intermediate feature named `_token` in `Cargo.toml`.
```toml
# Whether token provisioning is supported by AA
_token = []

coco_as_token = ["token/coco_as", "_token"]
kbs_token = ["token/kbs", "_token"]
```

And then we can here
```rust
#[cfg(feature = "_token")]
use token::GetToken;
```

attestation-agent/lib/src/lib.rs
```diff
@@ -174,23 +178,36 @@ impl AttestationAPIs for AttestationAgent {
...
-
-        // TODO: remove the feature flags after refactoring AA. Currently, kbs_host_url
-        // is only set by user in aa_kbc_params when cc_kbc is enabled.
-        #[cfg(not(feature = "cc_kbc"))]
-        {
-            bail!("unimplemented!");
-        }
+        let token = match _token_type {
+            #[cfg(any(feature = "cc_kbc", feature = "kbs_as"))]
+            "kbs" => {
```
Xynnn007, line 186: Could we define an enum for token types?

attestation-agent/lib/src/config.rs
```diff
@@ -4,60 +4,50 @@
...
-use serde::{Deserialize, Serialize};
+use serde::Deserialize;
 use std::env;
+use std::fs::File;
 use std::path::Path;
 use std::sync::OnceLock;
 use tokio::fs;
 
 const PEER_POD_CONFIG_PATH: &str = "/run/peerpod/daemon.json";
+pub const DEFAULT_AA_CONFIG_PATH: &str = "/etc/attestation.toml";
```
Xynnn007, line 16: Maybe `attestation-agent.toml`? 

Also there was once a discussion with @danmihai1 , he suggested to use JSONs for readibility in initdata. I prefer TOML as we can add comments but JSONs cannot.

attestation-agent/lib/src/lib.rs
```diff
@@ -88,22 +88,26 @@ pub trait AttestationAPIs {
     ) -> Result<()>;
 }
 
+#[allow(dead_code)]
 /// Attestation agent to provide attestation service.
-pub struct AttestationAgent {
+pub struct AttestationAgent<'a> {
+    config_file_path: &'a Path,
```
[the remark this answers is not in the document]
jialez0: `Path` will cause compile error here because it is an type which can not know its size when compiling. I think here we use `'a` is OK.

attestation-agent/token/src/lib.rs
```diff
@@ -0,0 +1,17 @@
...
+
+#[cfg(feature = "kbs")]
+pub mod kbs;
+
+#[cfg(feature = "coco_as")]
+pub mod coco_as;
+
+#[async_trait]
+pub trait GetToken {
+    async fn get_token(&self, service_url: String) -> Result<Vec<u8>>;
```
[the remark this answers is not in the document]
jialez0: I know this.

attestation-agent/lib/src/config.rs
bpradipt: Since the primary PR is about supporting attestation token, I think the changes to add config file mechanism should be in a separate PR.

This will help with the discussions and review especially w.r.to format, name and extensibility

cc @Xynnn007 @fitzthum

PR #444, opened 2024-01-19
CDH | Add configuration file when launching

Xynnn007:
~~Let's keep this PR as draft until initdata is finished, as after merging this PR kernel cmdline will no longer be used to parse kbs and kbc parameters.~~

This PR adds ability of setting configurations like `aa_kbc_params`, `kbs pub key` via a file when launching CDH. At the same time, parsing them from kernel cmdline is supported as a legacy way.

In future, if initdata mechanism is well implemented, we could deprecate the kernel cmdline way.


Xynnn007:
cc @fitzthum @danmihai1 @bpradipt

Xynnn007:
Fixes the logic and notes to ensure the initialization resources from KBS must be put under `/run/confidential-containers/cdh`. Now ready for review.

confidential-data-hub/hub/src/bin/confidential-data-hub/main.rs
```diff
@@ -21,29 +21,30 @@ use ttrpc::r#async::Server as TtrpcServer;
...
 mod api_ttrpc;
+mod config;
 mod keyprovider;
 mod keyprovider_ttrpc;
 mod server;
 
-const DEFAULT_CDH_SOCKET_ADDR: &str = "unix:///run/confidential-containers/cdh.sock";
+use config::*;
+
+const DEFAULT_CINFIG_PATH: &str = "/etc/confidential-data-hub.toml";
```
portersrc, line 31: Nitpick: "CINFIG" should be "CONFIG" (here and at its use below).

confidential-data-hub/hub/src/auth/kbs.rs
```diff
@@ -4,84 +4,93 @@
...
                 .await
                 .map_err(|e| {
                     Error::InitializationFailed(format!("kbs client get resource failed: {e}"))
                 })?;
 
-            fs::write(target_path, contents).await.map_err(|e| {
-                Error::InitializationFailed(format!(
-                    "kbs client get resource failed when writing to file: {e}"
-                ))
+            if !is_path_valid(k) {
```
portersrc, line 43: This check seems OK, but its placement seems a bit buried (in the middle of a key-value loop after a secret's already been fetched). Maybe `self.credentials` should be sanitized before this point.

portersrc (commented):
Just two minor comments. Nice work

confidential-data-hub/hub/src/bin/confidential-data-hub/config.rs
```diff
@@ -0,0 +1,138 @@
...
+                path: it.1.to_owned(),
+            });
+
+        self.credentials.extend(kbs_resources);
+        Ok(())
+    }
+}
+
+impl CdhConfig {
+    pub fn apply(&self) {
```
fitzthum, line 128: This is kind of a vague name. Is this function setting up the vars for the AA? Might be better to rename.

fitzthum (commented):
A few notes

confidential-data-hub/example.config.toml
```diff
@@ -0,0 +1,64 @@
...
+# credentials are items that will be retrieved from KBS when CDH
+# is launched. `resource_uri` refers to the KBS resource uri and
+# `path` is where to place the file.
+# `path` must be with prefix `/run/confidential-containers/cdh`,
+# or it will be blocked by CDH.
+[[credentials]]
+path = "/run/confidential-containers/cdh/kms-credential/aliyun/ecsRamRole.json"
+resource_uri = "kbs:///default/aliyun/ecs_ram_role"
+
+[[credentials]]
```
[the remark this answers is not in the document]
Xynnn007: You are right. The toml represents an array in this format. We also can use JSON instead of toml, but JSON does not support inline notes and well-printed multiple line strings. A JSON looks like
```json
{
  "kbc": {
    "name": "cc_kbc",
    "url": "http://example.io:8080",
    "kbs_cert": "-----BEGIN CERTIFICATE-----\nMIIFTDCCAvugAwIBAgIBADBGBgkqhkiG9w0BAQowOaAPMA0GCWCGSAFlAwQCAgUA\noRwwGgYJKoZIhvcNAQEIMA0GCWCGSAFlAwQCAgUAogMCATCjAwIBATB7MRQwEgYD\nVQQLDAtFbmdpbmVlcmluZzELMAkGA1UEBhMCVVMxFDASBgNVBAcMC1NhbnRhIENs\nYXJhMQswCQYDVQQIDAJDQTEfMB0GA1UECgwWQWR2YW5jZWQgTWljcm8gRGV2aWNl\nczESMBAGA1UEAwwJU0VWLU1pbGFuMB4XDTIzMDEyNDE3NTgyNloXDTMwMDEyNDE3\nNTgyNlowejEUMBIGA1UECwwLRW5naW5lZXJpbmcxCzAJBgNVBAYTAlVTMRQwEgYD\nVQQHDAtTYW50YSBDbGFyYTELMAkGA1UECAwCQ0ExHzAdBgNVBAoMFkFkdmFuY2Vk\nIE1pY3JvIERldmljZXMxETAPBgNVBAMMCFNFVi1WQ0VLMHYwEAYHKoZIzj0CAQYF\nK4EEACIDYgAExmG1ZbuoAQK93USRyZQcsyobfbaAEoKEELf/jK39cOVJt1t4s83W\nXM3rqIbS7qHUHQw/FGyOvdaEUs5+wwxpCWfDnmJMAQ+ctgZqgDEKh1NqlOuuKcKq\n2YAWE5cTH7sHo4IBFjCCARIwEAYJKwYBBAGceAEBBAMCAQAwFwYJKwYBBAGceAEC\nBAoWCE1pbGFuLUIwMBEGCisGAQQBnHgBAwEEAwIBAzARBgorBgEEAZx4AQMCBAMC\nAQAwEQYKKwYBBAGceAEDBAQDAgEAMBEGCisGAQQBnHgBAwUEAwIBADARBgorBgEE\nAZx4AQMGBAMCAQAwEQYKKwYBBAGceAEDBwQDAgEAMBEGCisGAQQBnHgBAwMEAwIB\nCDARBgorBgEEAZx4AQMIBAMCAXMwTQYJKwYBBAGceAEEBEDDhCejDUx6+dlvehW5\ncmmCWmTLdqI1L/1dGBFdia1HP46MC82aXZKGYSutSq37RCYgWjueT+qCMBE1oXDk\nd1JOMEYGCSqGSIb3DQEBCjA5oA8wDQYJYIZIAWUDBAICBQChHDAaBgkqhkiG9w0B\nAQgwDQYJYIZIAWUDBAICBQCiAwIBMKMDAgEBA4ICAQACgCai9x8DAWzX/2IelNWm\nituEBSiq9C9eDnBEckQYikAhPasfagnoWFAtKu/ZWTKHi+BMbhKwswBS8W0G1ywi\ncUWGlzigI4tdxxf1YBJyCoTSNssSbKmIh5jemBfrvIBo1yEd+e56ZJMdhN8e+xWU\nbvovUC2/7Dl76fzAaACLSorZUv5XPJwKXwEOHo7FIcREjoZn+fKjJTnmdXce0LD6\n9RHr+r+ceyE79gmK31bI9DYiJoL4LeGdXZ3gMOVDR1OnDos5lOBcV+quJ6JujpgH\nd9g3Sa7Du7pusD9Fdap98ocZslRfFjFi//2YdVM4MKbq6IwpYNB+2PCEKNC7SfbO\nNgZYJuPZnM/wViES/cP7MZNJ1KUKBI9yh6TmlSsZZOclGJvrOsBZimTXpATjdNMt\ncluKwqAUUzYQmU7bf2TMdOXyA9iH5wIpj1kWGE1VuFADTKILkTc6LzLzOWCofLxf\nonhTtSDtzIv/uel547GZqq+rVRvmIieEuEvDETwuookfV6qu3D/9KuSr9xiznmEg\nxynud/f525jppJMcD/ofbQxUZuGKvb3f3zy+aLxqidoX7gca2Xd9jyUy5Y/83+ZN\nbz4PZx81UJzXVI9ABEh8/xilATh1ZxOePTBJjN7lgr0lXtKYjV/43yyxgUYrXNZS\noLSG2dLCK9mjjraPjau34Q==\n-----END CERTIFICATE-----"
  },
  "credentials": [
    {
      "path": "/run/confidential-containers/cdh/kms-credential/aliyun/ecsRamRole.json",
      "resource_uri": "kbs:///default/aliyun/ecs_ram_role"
    },
    {
      "path": "/run/confidential-containers/cdh/test/file",
      "resource_uri": "kbs:///default/test/file"
    }
  ]
}
```

confidential-data-hub/hub/src/auth/kbs.rs
Xynnn007: Sounds good. Let me fix this

Xynnn007: I move the whole check logic before creating KBS client.

confidential-data-hub/hub/src/bin/confidential-data-hub/config.rs
Xynnn007: I once want to specify more than one envs, including `KBS_URL`, `KBC`, `KBS_PUB_KEY` etc. `set_configuration_envs()` might be a better name?

attestation-agent/lib/src/aa_kbc_params.rs
```diff
@@ -57,10 +57,17 @@ impl TryFrom<String> for AaKbcParams {
...
 
 async fn get_value() -> Result<String, ParamError> {
-    // first check whether we are in a peer pod
+    // first check env
+    if let Ok(params) = env::var("AA_KBC_PARAMS") {
+        debug!("get aa_kbc_params from env.");
+        return Ok(params);
+    }
+
+    // second check whether we are in a peer pod
```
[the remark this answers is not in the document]
Xynnn007: Ideally, we should remove `AA_KBC_PARAMS` in future and use `KBC type / KBS url` instead. However, before we have full initdata mechanism, the two parameters are still delivered via kernel cmdline claim `aa_kbc_params`. 

So this PR still temporarily use this way. Once the initdata is prepared, there will be two work
1. Abondon `aa_kbc_params` fully. Instead, use config vars to set `KBC` and `KBS URL`.
2. Use env to deliver `KBC` and `KBS URL` to underlying KMS crate without explicitly passing function calls layer by layer from the stack

confidential-data-hub/hub/src/bin/confidential-data-hub/config.rs
Xynnn007: Let me temporarily use the name `set_configuration_envs()` in the fixup. Please give any suggestions if it is not perfect.

Xynnn007:
Hi @portersrc I have fixed the things following your suggestions. PTAL

Hi @fitzthum I have fixed things you suggested. There are still some discussions left. PTAL

confidential-data-hub/hub/src/bin/confidential-data-hub/config.rs
```diff
@@ -0,0 +1,138 @@
...
+            std::result::Result::Ok(mut config) => {
+                info!("Read config file from {config_path}");
+                config.update_from_kernel_cmdline().await?;
+                return Ok(config);
+            }
+            Err(e) => warn!("read config file {config_path} failed {e:?}"),
+        }
+
+        warn!("Use default config. aa_kbc_params = offline_fs_kbc::null");
+        let mut config = CdhConfig::default();
```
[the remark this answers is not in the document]
fitzthum: You could move the logging to the default method itself if you want to be even more concise.

confidential-data-hub/hub/src/bin/confidential-data-hub/config.rs
```diff
@@ -0,0 +1,132 @@
...
+    pub fn set_configuration_envs(&self) {
+        // KBS configurations
+        env::set_var(
+            "AA_KBC_PARAMS",
+            format!("{}::{}", self.kbc.name, self.kbc.url),
+        );
+        if let Some(kbs_cert) = &self.kbc.kbs_cert {
+            env::set_var("KBS_PUBLICKEY_CERT", kbs_cert);
+        }
+    }
```
mkulke, line 131: Since we are using AA as a library can we initialize the AttestationAgent instance with configuration data instead of setting and reading an env within the same binary?



Nit: KBS_PUBLICKEY_CERT can be shortened to KBS_CERT probably

Xynnn007: The attestation agent binary/lib is not used in CDH. I just use `aa_kbc_params` to get kbs and kbc.

Why I set env and read in a same binary, is
1. function call depth. `kbs` and `kbc` parameter are needed in a `kms`, a separate crate. If we want to deliver the as the function argument, a chain of functions would be needed to modify to deliver from config to the underneath, which would be messy. 
2. easy to use. From the point of a separate crate, an env would be good as kbc and kbs are simple strings that could
3. different crates. `kms` is a crate and `hub` another. With env, `kms` could both work itself and with hub.

fitzthum (approved):
I can imagine some improvements down the road, but this seems like a good start.

confidential-data-hub/hub/src/bin/confidential-data-hub/config.rs
Xynnn007: > Nit: KBS_PUBLICKEY_CERT can be shortened to KBS_CERT probably

Thanks for this. I fixed

Xynnn007:
Just rebased the upstream code to fix CI.

fitzthum:
Btw, we might want to think about providing a default config file as well (in addition to the example). This would ship with all platforms that don't support InitData (or maybe with all platforms in general and then it would be overwritten by the init data). This would help with package stuff in Kata but we can do it in a separate PR.

Xynnn007:
Hey @mkulke does this pr look good now?

Xynnn007:
@mkulke let me think more about this before merging.

Xynnn007:
@mkulke Thanks for the link.

I quickly glanced at the [issue](https://github.com/rust-lang/rust/issues/27970). The unsafe key point is for multithread application due to the synchoronization and race of the envs. That means if the setting time and reading time are not easy to ensure, frequent modification would make the reader get a wrong env. However, in CDH this PR, the setting time is always the initialization of CDH when no code will call to read. After the initialization, the envs are fixed, and then the submodules could read the will-never-changed env.

BTW in kata-agent, we have already used this for a long time. An env is set in kata-agent https://github.com/kata-containers/kata-containers/blob/CCv0/src/agent/src/main.rs#L433

and then this env will be read in ocicrypt-rs https://github.com/confidential-containers/guest-components/blob/main/ocicrypt-rs/src/lib.rs#L40

This is isomorphism to this PR now.

TBO env might not be the best choice, and probably I should do some refactoring to deliver the parameter in function call. However for now, I think it's ok. wdyt?

mkulke:
> BTW in kata-agent, we have already used this for a long time. An env is set in kata-agent https://github.com/kata-containers/kata-containers/blob/CCv0/src/agent/src/main.rs#L433
> 
> and then this env will be read in ocicrypt-rs https://github.com/confidential-containers/guest-components/blob/main/ocicrypt-rs/src/lib.rs#L40

yes, I am aware, and I find this equally not-pretty. arguably ocicrypt-rs is a slightly different case (while our own code, it's supposed to be a re-implementation an original lib and it's called by kata, which is a different project).

So, if we have to resort to a global variable to pass down parameters to consumers, would we able to use a "proper" global variable?

https://docs.rs/once_cell/latest/once_cell/#safe-initialization-of-global-data

> TBO env might not be the best choice, and probably I should do some refactoring to deliver the parameter in function call. However for now, I think it's ok. wdyt?

I might be neurotic about those things, so maybe let's get another opinion on this.

Xynnn007:
> So, if we have to resort to a global variable to pass down parameters to consumers, would we able to use a "proper" global variable?

IMO `kms` could have ability to work as a separate crate, which provides the toolset to interact with different kinds of KMSes (including KBS). Thus I choose to use env.

Certainly, if `kms` would only be used in CDH, it will only be a module and we can use `OnceCell`. FWIW now, it seems still reasonable that `kms` would not aim to serve only CDH but more scenarios.

I prefer to keep it as-is. Let's collect some more ideas, kindly ping @fitzthum @1570005763 @portersrc

mkulke (approved):

fitzthum:
Yeah, I am also a bit suspicious of this ENV flow, but I think it isn't quite an anti-pattern. I think we can see this as setting up a one-time configuration for a child component. As the CDH evolves I think we should try to come up with something more robust. I have mentioned in the past that I don't like our reliance on the filesystem for other configuration parameters, but for now I guess I can live with this.

Xynnn007:
Let me merge this now. Any problems could be raised by an issue.

PR #449, opened 2024-01-24
AA: Support get CoCo-AS Attestation Token

jialez0:
@Xynnn007


Xynnn007 (approved):
The PR looks pretty good to me. Thanks. Only some nits.

PR #450, opened 2024-01-24
CDH | Add a client tool

Xynnn007:
This client tool helps to connect to the CDH and call the APIs. This is very useful in scenarios like Confidential VM where the CDH is used as a separate component.

Also, Fixes a bug for CDH/storage/OSS.


confidential-data-hub/hub/src/bin/client-tool/main.rs
```diff
@@ -0,0 +1,156 @@
...
+    match args.operation {
+        Operation::UnsealSecret(arg) => {
+            let client = SealedSecretServiceClient::new(inner);
+            let secret = tokio::fs::read(arg.secret_path).await.expect("read file");
+            let req = UnsealSecretInput {
+                secret,
+                ..Default::default()
+            };
+            let res = client
+                .unseal_secret(context::with_timeout(50 * 1000 * 1000 * 1000), &req)
```
fitzthum, line 100: Maybe some option to configure this timeout would be nice. At least maybe we should have it be a constant rather than doing the math in each of these cases.

fitzthum (commented):
Nice. This seems really good for testing and for people who want to use CDH without CoCo.

I made a couple small notes. It might be handy to add some kind of doc. One thing that would be handy would be an example of the JSON storage object that you provide to secure mount.

Also might be good to note that people will need to have the CDH running and configured correctly or none of this will work.

I also wonder if we could provide similar CLI interface in the CDH itself or using the CDH as a crate without having to add network interaction. Either way seems fine, though, and this might actually be better for testing.

Xynnn007:
@fitzthum Let me temporarily make this PR as draft before the whole test and documents are finished.

bpradipt:
@Xynnn007 this is very useful. And I agree with Tobin, having a doc will be very helpful.

Xynnn007:
The client tool is base on the CDH. If we want to use CDH in a VM, it needs some customized parameters. So let's finish https://github.com/confidential-containers/guest-components/pull/444 before continue this.

confidential-data-hub/hub/Cargo.toml
```diff
@@ -10,6 +10,10 @@ edition = "2021"
 name = "confidential-data-hub"
 required-features = ["bin"]
 
+[[bin]]
+name = "client-tool"
```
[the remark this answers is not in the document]
Xynnn007: Good idea.

confidential-data-hub/hub/build.rs
```diff
@@ -43,5 +43,25 @@ fn main() {
             "client: client",
             "client",
         );
+
+        ttrpc_codegen::Codegen::new()
```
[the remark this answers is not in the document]
Xynnn007: Right, fixed this. Thanks for digging into the detail

confidential-data-hub/hub/build.rs
```diff
@@ -43,5 +43,25 @@ fn main() {
...
+            .out_dir("./src/bin/client-tool")
```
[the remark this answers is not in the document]
Xynnn007: Nice suggestion. There is.

confidential-data-hub/hub/src/bin/client-tool/main.rs
```diff
@@ -0,0 +1,156 @@
...
+    let args = Cli::parse();
+    let inner = ttrpc::asynchronous::Client::connect(&args.socket).expect("connect ttrpc socket");
+
+    match args.operation {
+        Operation::UnsealSecret(arg) => {
+            let client = SealedSecretServiceClient::new(inner);
+            let secret = tokio::fs::read(arg.secret_path).await.expect("read file");
+            let req = UnsealSecretInput {
+                secret,
+                ..Default::default()
```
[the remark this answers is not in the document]
Xynnn007: Yes. `ttrpc` will have extra fields based on the original protobuf definition.

confidential-data-hub/hub/src/bin/client-tool/main.rs
Xynnn007: Sounds good. Let me add a parameter

Xynnn007:
The code is updatd. Let me add document before make it public.

Xynnn007:
Kindly ping @LindaYu17 @1570005763 for this.

Xynnn007:
Ok I think this PR is now ready for review @fitzthum @bpradipt

confidential-data-hub/docs/use-cases/secure-mount-with-aliyun-oss.md
```diff
@@ -0,0 +1,125 @@
...
+```json
+{
+    "driver": "",
+    "driver_options": [
+        "alibaba-cloud-oss={\"akId\":\"XXX\",\"akSecret\":\"XXX\",\"annotations\":\"\",\"bucket\":\"<bucket-id>\",\"encrypted\":\"gocryptfs\",\"encPasswd\":\"<PASSWORD>\",\"kmsKeyId\":\"\",\"otherOpts\":\"-o max_stat_cache_size=0 -o allow_other\",\"path\":\"<bucket-path>\",\"readonly\":\"\",\"targetPath\":\"/mnt/aliyun-oss\",\"url\":\"https://oss-cn-beijing.aliyuncs.com\",\"volumeId\":\"\"}"
+    ],
+    "source": "",
+    "fstype": "",
+    "options": [],
+    "mount_point": "/root/xynnn/mnt"
```
fitzthum, line 79: Make this path generic?

confidential-data-hub/hub/src/bin/cdh-tool.rs
```diff
@@ -0,0 +1,170 @@
...
+        Operation::UnsealSecret(arg) => {
+            let client = SealedSecretServiceClient::new(inner);
+            let secret = tokio::fs::read(arg.secret_path).await.expect("read file");
+            let req = UnsealSecretInput {
+                secret,
+                ..Default::default()
+            };
+            let res = client
+                .unseal_secret(
+                    context::with_timeout(args.timeout * 1000 * 1000 * 1000),
```
fitzthum, line 103: Maybe this scale factor can be a constant? Also are you sure that we need to multiply by such a larger number? What is `with_timeout` expecting?

fitzthum (commented):
Looks pretty good. Two small comments.

confidential-data-hub/docs/use-cases/secure-mount-with-aliyun-oss.md
Xynnn007: Ok. Fixed.

confidential-data-hub/hub/src/bin/cdh-tool.rs
Xynnn007: the large number means nanoseconds / second. Let me define a const here.

fitzthum (approved):
LGTM. Maybe we should add an e2e test workflow with this tool at some point.

PR #451, opened 2024-01-24
keyprovider: extend docker image and documentation

mkulke:
The keyprovider docker image has been extended to bundle a keyprovider-capable skopeo and include a convenience script that simplifies the creation of encrypted images for usage in CoCo.

Documentation has been added to use the image.

### DIFF
--- a/attestation-agent/coco_keyprovider/README.md
+++ b/attestation-agent/coco_keyprovider/README.md
@@ -13,6 +13,47 @@ The following guide will help make an encrypted image using [skopeo](https://git
 
 ## Encryption
 
+### Docker
+
+A docker image provides prebuilt CoCo keyprovider and skopeo to simplify image encryption:
+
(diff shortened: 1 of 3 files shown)


mkulke:
cc @bpradipt

Xynnn007 (approved):
Wow. It is a very coool thing and will be of great use. There are only one topic to talk about
1. I'm not sure if it could be better to keep `encrypt.sh` contents outside the Dockerfile, which would do help to maintainance.

mkulke:
> Wow. It is a very coool thing and will be of great use. There are only one topic to talk about

Indeed, that makes sense. I inlined the script to make it more more portable (i.e. it doesn't matter from where you invoke docker build) but since we build the code, we have to build from project root anyway.

bpradipt (approved):
/lgtm

PR #452, opened 2024-01-24
RFC: attester: tdx: try not to error on broken report_data

mythi:
The TDX attester code already tries very hard to ensure the report_data passed has the correct size. Because of that, initializing tdx_attest_rs::tdx_report_data_t should never fail. However, .try_into() returns a Result we must process.

Currently, the code just raises an error without a good reason what happened. Alternatively, we could just unwrap() but that could panic. The third approach is to just initialize tdx report data to zero and catch the attestation result later.

This picks the third option and changes the code to use a const to make sure Vec size conflicts are less likely to happen.

This was initially discussed in https://github.com/confidential-containers/guest-components/pull/434#discussion_r1456983754 and I wanted to have the conversation separately so that I could just follow the approach in #434

### DIFF
--- a/attestation-agent/attester/src/tdx/mod.rs
+++ b/attestation-agent/attester/src/tdx/mod.rs
@@ -10,6 +10,7 @@ use serde::{Deserialize, Serialize};
 use std::path::Path;
 use tdx_attest_rs;
 
+const TDX_REPORT_DATA_SIZE: usize = 64;
 const CCEL_PATH: &str = "/sys/firmware/acpi/tables/data/CCEL";
 
 pub fn detect_platform() -> bool {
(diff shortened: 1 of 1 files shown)


attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -31,14 +32,17 @@ pub struct TdxAttester {}
...
 
-        report_data.resize(64, 0);
+        report_data.resize(TDX_REPORT_DATA_SIZE, 0);
 
         let tdx_report_data = tdx_attest_rs::tdx_report_data_t {
-            d: report_data.as_slice().try_into()?,
+            d: report_data
+                .as_slice()
+                .try_into()
+                .unwrap_or([0; TDX_REPORT_DATA_SIZE]),
```
Xynnn007, line 45: If we need a default value, it's better to have a `warn!` or `info!` to tell this as it is often unexpected to the caller.

BTW, the error might not occur

As the try into from `&[u8]` to [u8; 64] will be performed by `TryFrom<&[T]>` defined in https://doc.rust-lang.org/std/primitive.slice.html#impl-TryFrom%3C%26%5BT%5D%3E-for-%5BT;+N%5D, which will return true if `slice.len() == N.` that is asserted by `.resize()`.

mythi: > which will return true if `slice.len() == N.` that is asserted by `.resize()`.

I was reading about the same that the error cannot happen and that was one of the reasons for this PR: just to make sure the size is always correctly set. What I also wanted to add but forgot is a comment about it.

What would be your preference? Simply `.unwrap()` after all?

Xynnn007 (approved):
lgtm

PR #453, opened 2024-01-25
Makefile: support to build components for all platforms and amd

Xynnn007:
`TEE_PLATFORM=all` to support a easy mode for users to build all guest components with all features enabled (including the attester plugins, etc)

`TEE_PLATFORM=amd` supports to build components of sev & snp.

Also, fixed a bug in CDH when both kbs and sev features are enabled. We not will initialize sev when `aa_kbc_param` is set to `sev`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -945,6 +945,7 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "async-trait",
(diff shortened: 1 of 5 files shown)


fidencio (approved):
lgtm, thanks @Xynnn007!

Issue #446, opened 2024-01-22
AA | Design ideas to verify the initdata


Xynnn007:
### Background

Currently, we are promoting the [initdata mechanism](https://github.com/confidential-containers/confidential-containers/issues/171) in the community. The core idea is to leverage initdata field (HOSTDATA for SNP, MRCONFIGID for TDX, etc.) to bind the hash value of any data injected into the guest. One of the most critical design is how AA exposes an interface for kata-agent to verify the binding between the hash of initdata plaintext and the corresponding initdata field inside TEE evidence. 

![image](https://github.com/confidential-containers/guest-components/assets/22250460/612ad2e6-eae8-4d85-ba20-c57e781e1ae6)

Overall, we should take a way to do the following things:
1. launch AA by being forked in kata-agent's main function or by systemd.
2. receive the initdata plaintext/hash into AA from kata-agent
3. AA gets the TEE attestation report. This process must be without network, as AA is a service launched at a very early stage in guest where network is not prepared. The codes for TDX and SNP
are finished by @danmihai1  perfectly in https://github.com/kata-containers/kata-containers/pull/8469/files#diff-df928933d70fb4e5616a6ecb3d8a1340adbf328dd3d6973b3a00ce5c75aae23aR1 which we can reuse in AA.
4. AA compares the hash of the initdata plaintext and the one inside evidence.

The step 2 is now to be designed. This issue aims to collect ideas upon this.

@fitzthum @danmihai1 and I have been discussing this and let's continue publicly.

### Design Concerns and Different Ways

There are some design concerns
1. Although TDX, SNP, CCA, etc have initdata field like `HOSTDATA`, `MRCONFIGID`, s390x and AMD SEV do not.
2. The `KBS url`, `KBS public key` together with some other parameters should be part of AA's configuration file. Those are all things provided via initdata.

Thus, we need a mechanism
1. Allow s390x and SEV to not verify initdata without blocking AA from launching. While on other platforms AA should panic/raise error if initdata is not matched.
2. The configurations should be correctly configured and used if initdata binding check passes.

#### A. Initdata as a launch parameter

Initdata check will be part of AA's launch parameter.
```shell
attestation-agent --config-file conf.json --initdata MDAwMDAwMAo=
```

The parameter `--initdata` item is optional. If it is given, it will check the binding against an evidence report as soon as AA is launched. 

[...]

PR #454, opened 2024-01-25
AA: Add Config file mechanism

jialez0:
cc @Xynnn007


confidential-data-hub/kms/src/error.rs
```diff
@@ -3,7 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-use attestation_agent::aa_kbc_params;
+use attestation_agent::config::aa_kbc_params;
```
Xynnn007, line 6: This PR also moves `aa_kbc_params` to another place. Could we bring this part out as a separate commit?

attestation-agent/lib/src/lib.rs
```diff
@@ -94,20 +95,26 @@ pub trait AttestationAPIs {
...
+            kbc_instance_map: HashMap::new(),
+            config_file_path: config::DEFAULT_AA_CONFIG_PATH.to_string(),
+        }
     }
 }
 
 impl AttestationAgent {
     /// Create a new instance of [AttestationAgent].
-    pub fn new() -> Self {
+    pub fn new(config_path: &str) -> Self {
```
Xynnn007, line 113: This function is never called due to the code. Is this intentded?

Xynnn007 (commented):
A few comments

1. Currently the config is read lazily when a get token request is received. What if we have the configuration file set at the launch of AA? e.g. firstly read the cmdline parameter to check if config-path is given. If not, try to read default path. If no default config is given, use a built-in default config.
2. Due to some discussion. Now we might be interested in add another API of AA name `UpdateConfiguration` which could be only called once. See https://github.com/confidential-containers/guest-components/issues/446#issuecomment-1906362841

bpradipt:
cc @mkulke if he gets some time to look into this

attestation-agent/lib/src/config/mod.rs
```diff
@@ -0,0 +1,31 @@
...
+#[derive(Clone, Debug, Deserialize)]
+#[allow(dead_code)]
+pub struct Config {
+    /// URL Address of Attestation Service
+    pub as_url: String,
+    // TODO: Add more fields that accessing AS needs.
+}
+
+impl TryFrom<&str> for Config {
+    type Error = anyhow::Error;
```
mkulke, line 23: In (new) library code, we should consider using concrete error types (`thiserror` crate makes this easy) in library code. this would probably translate to something like this: 

```rust
use thiserror::Error;

#[derive(Error, Debug)]
pub enum ConfigFileError {
    #[error("failed to open")]
    Io(#[from] io::Error), 
    #[error("failed to parse")]
    Parse(#[from] serde_json::Error),     
}

impl TryFrom<&str> for Confg {
    type Error = ConfigFileError;
    fn try_from(config_path: &str) -> Result<Self, Self::Error> {
        let file = File::open(config_path)?;
        let cfg: Config = serde_json::from_reader(file)?;
        Ok(cfg)
    }
}
```

attestation-agent/lib/src/config/mod.rs
```diff
@@ -0,0 +1,31 @@
...
+
+pub mod aa_kbc_params;
+
+pub const DEFAULT_AA_CONFIG_PATH: &str = "/etc/attestation-agent.toml";
+
+#[derive(Clone, Debug, Deserialize)]
+#[allow(dead_code)]
+pub struct Config {
+    /// URL Address of Attestation Service
+    pub as_url: String,
```
mkulke, line 18: `as_uri` maybe, as not all kbc connection strings are urls?

attestation-agent/lib/src/lib.rs
```diff
@@ -178,21 +185,28 @@ impl AttestationAPIs for AttestationAgent {
 
     #[allow(unreachable_code)]
     async fn get_token(&mut self, _token_type: &str) -> Result<Vec<u8>> {
-        let _params = aa_kbc_params::get_params().await?;
+        let _url = match aa_kbc_params::get_params().await {
```
mkulke, line 188: `_url` => `uri`

attestation-agent/lib/src/lib.rs
```diff
@@ -178,21 +185,28 @@ impl AttestationAPIs for AttestationAgent {
...
+            Ok(params) => params.uri().to_string(),
+            Err(_) => {
+                let config = config::Config::try_from(self.config_file_path.as_str())
```
mkulke, line 191: we probably don't want to re-read the configuration file on every get_token() call, maybe we can do this lazily or better yet, have the configuration option on the attestation-agent binary as a cli param and use it as initialization param for the  AA struct:

```rust
AttestationAgent {
    kbc_module_list: KbcModuleList::new(),
    kbc_instance_map: HashMap::new(),
    config: Option<config::Config>,
}
```

mkulke (approved):

mkulke (commented):
n/a

jialez0:
> Currently the config is read lazily when a get token request is received. What if we have the configuration file set at the launch of AA? e.g. firstly read the cmdline parameter to check if config-path is given. If not, try to read default path. If no default config is given, use a built-in default config.

> we probably don't want to re-read the configuration file on every get_token() call, maybe we can do this lazily or better yet, have the configuration option on the attestation-agent binary as a cli param and use it as initialization param for the AA struct:

@Xynnn007 @mkulke The reason why the config here is read lazily is that we want to retain the ability to dynamically modify the configuration file when AA is running (for example, other components in TEE want to modify some configuration parameters of AA). I'm not sure whether this is necessary, but it is flexible enough, or it may bring convenience to other scheme designs in the future.

jialez0:
> Due to some discussion. Now we might be interested in add another API of AA name `UpdateConfiguration` which could be only called once.

Instead of designing such an interface, I think it's better to modify the configuration file of AA directly.

confidential-data-hub/kms/src/error.rs
jialez0: I think these two parts can be classified as a commitment, because aa_ Kbc_ Parameters themselves are part of the AA configuration parameters, so I think it is OK to put them in a commit.

attestation-agent/lib/src/lib.rs
jialez0: This API is indeed not used in AA's gRPC binary, but AA library, as an independent Rust Crate, will provide support for a wider range of scenarios, so it is necessary to provide the API here.

mkulke:
> @Xynnn007 Instead of designing such an interface, I think it's better to modify the configuration file of AA directly.

It might be a bit unintuitive, the semantics of a configuration file is usually "read once, at launch". For a library the intended behaviour is a bit fuzzy, that's why I would suggest to:

- Use a configuration file for the AA application, read at startup, subsequent changes to the file are ignored
- Have an API in the AA library to be able to adjust the configuration

> for example, other components in TEE want to modify some configuration parameters of AA). I'm not sure whether this is necessary, but it is flexible enough, and it may bring convenience to other scheme designs in the future.

If we do this, we would have to account for race conditions. If there is API in the AA lib that would rewrite the config file, we could put it behind a mutex guard, I guess. (still, going from AA -> filesystem -> AA is to propagate config changes feels unnecessary) 

If we modify the config file out-of-process this wouldn't work.

jialez0:
> It might be a bit unintuitive, the semantics of a configuration file is usually "read once, at launch". For a library the intended behaviour is a bit fuzzy, that's why I would suggest to:
> 
> * Use a configuration file for the AA application, read at startup, subsequent changes to the file are ignored
> * Have an API in the AA library to be able to adjust the configuration
> 
> > for example, other components in TEE want to modify some configuration parameters of AA). I'm not sure whether this is necessary, but it is flexible enough, and it may bring convenience to other scheme designs in the future.
> 
> If we do this, we would have to account for race conditions. If there is API in the AA lib that would rewrite the config file, we could put it behind a mutex guard, I guess. (still, going from AA -> filesystem -> AA is to propagate config changes feels unnecessary)
> 
> If we modify the config file out-of-process this wouldn't work.

This makes sense. I will modify this PR to support configuration reading at startup.

jialez0:
@Xynnn007 @mkulke For how to maintain the dynamic change ability of AA configuration files, perhaps we can consider such a design (similar to systemd):

1) When the AA starts, read its own configuration file as the initial configuration; 2) The user modifies the configuration file at runtime, but it will not take effect; 3) The user needs to execute a special command to trigger AA to re-read the configuration.

What do you think?

mkulke:
> @Xynnn007 @mkulke For how to maintain the dynamic change ability of AA configuration files, perhaps we can consider such a design (similar to systemd and nginx):
> 
> 1. When the AA starts, read its own configuration file as the initial configuration; 2) The user modifies the configuration file at runtime, but it will not take effect; 3) The user needs to execute a special command (or send a linux process signal) to trigger AA to re-read the configuration.
> 
> What do you think?

sounds reasonable to me, maybe a bit cumbersome to implement. but we could implement such a reload-trigger logic once we need it and go with a read-once solution for now.

Xynnn007 (approved):
lgtm

PR #459, opened 2024-01-26
AA: Add `coco_as` feature to `cc_kbc` to default support CoCo-AS

jialez0:


attestation-agent/lib/Cargo.toml
```diff
@@ -33,7 +33,7 @@ default = ["sample_kbc", "rust-crypto", "coco_as"]
 kbs = ["kbs_protocol/background_check"]
 coco_as = ["reqwest"]
 
-cc_kbc = ["kbc/cc_kbc", "kbs_protocol/background_check", "kbs"]
+cc_kbc = ["kbc/cc_kbc", "kbs_protocol/background_check", "kbs", "coco_as"]
```
Xynnn007, line 36: This is interesting that `cc_kbc` is an old name to specify the kbc. But now kbc implementations are moved to CDH, thus only attester/platform should be configured for AA.

In future, we might also need to delete all `*-kbc` features and instead only leave `*-attester` and different kinds of token features.

But currently, it is ok to me.

Xynnn007 (approved):

PR #461, opened 2024-01-29
Replace unsafe NonNull::new_unchecked with NonNull:new

pingzhaozz:
rats_tls_handle ptr should be correct initialized by extern C func rats_tls_set_verification_callback

### DIFF
--- a/attestation-agent/kbc/src/eaa_kbc/rats_tls/mod.rs
+++ b/attestation-agent/kbc/src/eaa_kbc/rats_tls/mod.rs
@@ -31,7 +31,7 @@ unsafe impl ForeignType for RatsTls {
     type Ref = RatsTlsRef;
 
     unsafe fn from_ptr(ptr: *mut rats_tls_handle) -> RatsTls {
-        RatsTls(NonNull::new_unchecked(ptr))
+        RatsTls(NonNull::new(ptr).expect("rats_tls_handle ptr is null!"))
     }
 
     fn as_ptr(&self) -> *mut rats_tls_handle {


portersrc:
Looks OK to me.
I guess the comment might deserve a tweak, though? It looks like `rats_tls_init` (not `rats_tls_set_verification_callback`) is responsible for correctly initializing `rats_tls_handle` (assuming it's using [this C library](https://github.com/inclavare-containers/rats-tls/blob/3870d011a2061c8e6e14993fe2937f291017fed5/src/api/rats_tls_init.c#L132)).
`rats_tls_set_verification_callback` seems to just be setting the callback you give it (which in fact already assumes handle is non-null) (I'm looking [here](https://github.com/inclavare-containers/rats-tls/blob/3870d011a2061c8e6e14993fe2937f291017fed5/src/api/rats_tls_callback.c#L14)).

pingzhaozz:
> Looks OK to me. I guess the comment might deserve a tweak, though? It looks like `rats_tls_init` (not `rats_tls_set_verification_callback`) is responsible for correctly initializing `rats_tls_handle` (assuming it's using [this C library](https://github.com/inclavare-containers/rats-tls/blob/3870d011a2061c8e6e14993fe2937f291017fed5/src/api/rats_tls_init.c#L132)). `rats_tls_set_verification_callback` seems to just be setting the callback you give it (which in fact already assumes handle is non-null) (I'm looking [here](https://github.com/inclavare-containers/rats-tls/blob/3870d011a2061c8e6e14993fe2937f291017fed5/src/api/rats_tls_callback.c#L14)).

Yes, I updated the comment with ```rats_tls_init```. Thanks for pointing it out.

fitzthum:
> Yes, I updated the comment with rats_tls_init. Thanks for pointing it out.

Did you push this change?

pingzhaozz:
> 
> Did you push this change?

Yes, it was updated.

fitzthum (approved):
I'm not sure what the best way to handle errors in this type of binding is, but I guess this is fine. 

I think that all this `eaa_kbc` code is deprecated anyway.

Xynnn007:
Yes. `eaa_kbc` code is not used anymore in both kata-cc and enclave-cc

PR #462, opened 2024-01-29
AA: Add API of CheckInitData

### LINKED ISSUE #446: AA | Design ideas to verify the initdata

Xynnn007:
This new API is for kata-agent to check the data's integrity binding relationship with the hardware evidence.

Close #446 

The main work for TDX and SNP platform has been done by  @danmihai1 , this PR just refers most of the code.

cc @fitzthum

### DIFF
--- a/attestation-agent/attester/Cargo.toml
+++ b/attestation-agent/attester/Cargo.toml
@@ -19,6 +19,7 @@ kbs-types.workspace = true
 log.workspace = true
 nix = { version = "0.26.2", optional = true }
 occlum_dcap = { git = "https://github.com/occlum/occlum", tag = "v0.29.7", optional = true }
+scroll = { version = "0.11.0", default-features = false, features = ["derive"], optional = true }
 serde.workspace = true
 serde_json.workspace = true
 sev = { version = "1.2.0", default-features = false, features = [
@@ -26,6 +27,7 @@ sev = { version = "1.2.0", default-features = false, features = [
 ], optional = true }
 strum.workspace = true
 tdx-attest-rs = { git = "https://github.com/intel/SGXDataCenterAttestationPrimitives", tag = "DCAP_1.16", optional = true }
+thiserror.workspace = true
 # TODO: change it to "0.1", once released.
 csv-rs = { git = "https://github.com/openanolis/csv-rs", rev = "b74aa8c", optional = true }
 codicon = { version = "3.0", optional = true }
@@ -52,7 +54,7 @@ all-attesters = [
     "cca-attester",
 ]
 
-tdx-attester = ["tdx-attest-rs"]
+tdx-attester = ["scroll", "tdx-attest-rs"]
 sgx-attester = ["occlum_dcap"]
 az-snp-vtpm-attester = ["az-snp-vtpm"]
 az-tdx-vtpm-attester = ["az-tdx-vtpm"]
--- a/attestation-agent/attester/src/lib.rs
+++ b/attestation-agent/attester/src/lib.rs
@@ -74,6 +74,10 @@ pub trait Attester {
     ) -> Result<()> {
         bail!("Unimplemented")
     }
+
+    async fn check_init_data(&self, _init_data: &[u8]) -> Result<()> {
+        bail!("Unimplemented");
+    }
 }
 
 // Detect which TEE platform the KBC running environment is.
--- a/attestation-agent/attester/src/lib.rs
+++ b/attestation-agent/attester/src/lib.rs
@@ -7,6 +7,7 @@ use anyhow::*;
 use kbs_types::Tee;
 
 pub mod sample;
+pub mod utils;
 
 #[cfg(feature = "az-snp-vtpm-attester")]
 pub mod az_snp_vtpm;
--- a/attestation-agent/attester/src/snp/hostdata.rs
+++ b/attestation-agent/attester/src/snp/hostdata.rs
@@ -1,0 +1,24 @@
+// Copyright (c) 2024 Microsoft Corporation
+// Copyright (c) 2024 Alibaba Cloud
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum GetHostDataError {
+    #[error("Open Sev guest firmware failed: {0}")]
+    OpenSevGuestFirmware(#[from] std::io::Error),
+
+    #[error("Get report failed: {0}")]
+    GetReportError(#[from] sev::error::UserApiError),
+}
+
+pub fn get_snp_host_data() -> Result<[u8; 32], GetHostDataError> {
+    let mut firmware = sev::firmware::guest::Firmware::open()?;
+    let report_data: [u8; 64] = [0; 64];
+    let report = firmware.get_report(None, Some(report_data), Some(0))?;
+
+    Ok(report.host_data)
+}
--- a/attestation-agent/attester/src/snp/mod.rs
+++ b/attestation-agent/attester/src/snp/mod.rs
@@ -3,6 +3,8 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+use crate::utils::pad;
+
 use super::Attester;
 use anyhow::*;
 use serde::{Deserialize, Serialize};
@@ -10,6 +12,8 @@ use sev::firmware::guest::AttestationReport;
 use sev::firmware::guest::Firmware;
 use sev::firmware::host::CertTableEntry;
 use std::path::Path;
+
+mod hostdata;
 
 pub fn detect_platform() -> bool {
     Path::new("/sys/devices/platform/sev-guest").exists()
@@ -46,5 +50,15 @@ impl Attester for SnpAttester {
         };
 
         serde_json::to_string(&evidence).context("Serialize SNP evidence failed")
+    }
+
+    async fn check_init_data(&self, init_data: &[u8]) -> Result<()> {
+        let hostdata = hostdata::get_snp_host_data().context("Get HOSTDATA failed")?;
+        let init_data: [u8; 32] = pad(init_data);
+        if init_data != hostdata {
+            bail!("HOSTDATA does not match.");
+        }
+
+        Ok(())
     }
 }
--- a/attestation-agent/lib/src/lib.rs
+++ b/attestation-agent/lib/src/lib.rs
@@ -12,8 +12,7 @@ use async_trait::async_trait;
 use attester::{detect_tee_type, BoxedAttester};
 use kbc::{AnnotationPacket, KbcCheckInfo, KbcInstance, KbcModuleList};
 use resource_uri::ResourceUri;
-use std::collections::HashMap;
-use std::str::FromStr;
+use std::{collections::HashMap, str::FromStr};
 
 pub mod config;
 mod token;
@@ -90,6 +89,9 @@ pub trait AttestationAPIs {
         events: Vec<Vec<u8>>,
         register_index: Option<u64>,
     ) -> Result<()>;
+
+    /// Check the initdata binding
+    async fn check_init_data(&mut self, init_data: &[u8]) -> Result<()>;
 }
 
 /// Attestation agent to provide attestation service.
@@ -238,6 +240,15 @@ impl AttestationAPIs for AttestationAgent {
         attester
             .extend_runtime_measurement(events, register_index)
             .await?;
+        Ok(())
+    }
+
+    /// Check the initdata binding. If current platform does not support initdata
+    /// injection, return a success and raise a warning log.
+    async fn check_init_data(&mut self, init_data: &[u8]) -> Result<()> {
+        let tee_type = detect_tee_type();
+        let attester = TryInto::<BoxedAttester>::try_into(tee_type)?;
+        attester.check_init_data(init_data).await?;
         Ok(())
     }
 }
--- a/attestation-agent/protos/attestation-agent.proto
+++ b/attestation-agent/protos/attestation-agent.proto
@@ -25,8 +25,20 @@ message ExtendRuntimeMeasurementRequest {
 
 message ExtendRuntimeMeasurementResponse {}
 
+message InitDataPlaintext {
+    bytes Content = 1;
+    string Algorithm = 2; 
+}
+
+message CheckInitDataRequest {
+    bytes Digest = 1;
+}
+
+message CheckInitDataResponse {}
+
 service AttestationAgentService {
     rpc GetEvidence(GetEvidenceRequest) returns (GetEvidenceResponse) {};
     rpc GetToken(GetTokenRequest) returns (GetTokenResponse) {};
     rpc ExtendRuntimeMeasurement(ExtendRuntimeMeasurementRequest) returns (ExtendRuntimeMeasurementResponse) {};
+    rpc CheckInitData(CheckInitDataRequest) returns (CheckInitDataResponse) {};
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -287,11 +287,13 @@ dependencies = [
  "log",
  "nix 0.26.4",
  "occlum_dcap",
+ "scroll",
  "serde",
  "serde_json",
  "sev 1.2.1",
  "strum",
  "tdx-attest-rs",
+ "thiserror",
  "tokio",
 ]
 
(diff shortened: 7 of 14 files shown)


attestation-agent/attester/src/lib.rs
```diff
@@ -74,6 +74,10 @@ pub trait Attester {
     ) -> Result<()> {
         bail!("Unimplemented")
     }
+
+    async fn check_init_data(&mut self, _init_data: Vec<u8>) -> Result<()> {
```
mkulke, line 78: Does this have to be `&mut self`? 

Can we pass the _init_data as reference `&[u8]`?

attestation-agent/attester/src/snp/hostdata.rs
```diff
@@ -0,0 +1,19 @@
...
+        Ok(mut firmware) => {
+            let report_data: [u8; 64] = [0; 64];
+            match firmware.get_report(None, Some(report_data), Some(0)) {
+                Ok(report) => Ok(report.host_data.to_vec()),
+                Err(e) => Err(e.into()),
+            }
+        }
+        Err(e) => Err(e.into()),
+    }
+}
```
mkulke, line 19: Can we use concrete error types? 

It would also be good to unnest this into:

```
{
  let firmware = sev::firmware::guest::Firmware::open()?;
  ...
  let report = firmware.get_report(...)?;
  report.host_data()
}
```

attestation-agent/attester/src/snp/mod.rs
```diff
@@ -47,4 +49,14 @@ impl Attester for SnpAttester {
 
         serde_json::to_string(&evidence).context("Serialize SNP evidence failed")
     }
+
+    async fn check_init_data(&mut self, mut init_data: Vec<u8>) -> Result<()> {
+        let hostdata = hostdata::get_snp_host_data().context("Get HOSTDATA failed")?;
+        init_data.resize(core::mem::size_of::<[u8; 32]>(), b'0');
```
mkulke, line 55: i'd just create a slice `&init_data[..32]`

attestation-agent/attester/src/tdx/mrconfigid.rs
```diff
@@ -0,0 +1,185 @@
...
+
+// Get TDREPORT0 (a.k.a. TDREPORT subtype 0) using TDCALL[TDG.MR.REPORT].
+ioctl_iowr_nr!(
+    TDX_CMD_GET_REPORT0,
+    'T' as ::std::os::raw::c_uint,
+    1,
+    TdxReportReq
+);
+
+pub fn get_tdx_mrconfigid() -> Result<Vec<u8>> {
```
mkulke, line 149: are we able to use the tdx-attest-rs code that we use in the rest of the TDX attester impl to retrieve a report? It feels a bit odd that we have to have unsafe low-level code in AA directly.

attestation-agent/lib/src/lib.rs
```diff
@@ -88,6 +88,13 @@ pub trait AttestationAPIs {
         events: Vec<Vec<u8>>,
         register_index: Option<u64>,
     ) -> Result<()>;
+
+    /// Check the initdata binding
+    async fn check_init_data(
+        &mut self,
+        init_data: Vec<u8>,
+        _init_data_plaintext: Option<Vec<u8>>,
```
mkulke, line 96: Why is this optional?

fitzthum (commented):
Nice.

One thing I think we might have forgotten about when designing: ideally the KBS/AS will end up with the plaintext of the InitData so that it can be validated with a policy. We don't seem to have a mechanism for this. Perhaps if the plaintext is provided via this API it should be saved by the AA so that it can be sent as part of remote attestation. We could probably implement this in a follow-on.

attestation-agent/attester/Cargo.toml
```diff
@@ -52,7 +53,7 @@ all-attesters = [
     "cca-attester",
 ]
 
-tdx-attester = ["tdx-attest-rs"]
+tdx-attester = ["nix", "tdx-attest-rs", "vmm-sys-util"]
```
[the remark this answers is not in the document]
Xynnn007: This seems awkward. The initdata check might not be binding to cckbc. As currently `attester` is a sparate crate, which aims to apply to scenarios more than kbcs or even CoCo. Any developer could import this crate to get report or check initdata binding. But surely, in semantics, `attester` might not cover init data check things.

attestation-agent/attester/src/lib.rs
Xynnn007: You are right. We do not have to mark this as `mut` as we do not change any inside state. Also, let me add another commit to change other trait functions who does need to be `mut`.

`&[u8]` is a good choice. Let me fix. Also this conversion reminds me of old times XD.

attestation-agent/attester/src/snp/mod.rs
Xynnn007: The input `init_data` might be shorter than 32 bytes, so I use `resize` here

attestation-agent/attester/src/tdx/mrconfigid.rs
Xynnn007: The code here is a little trade-off.

In TDX, there are two different "hardware evidence"s.
- TDX_REPORT: an evidence directly got from `ioctl`. This is what this PR includes. Cons: more code. Pros: performance.
- TDX_QUOTE: a signed version of TDX_REPORT, together with some extra metadata. The signing process involves a vsock call from guest to the host. Also some function calls among host side services, like QGS, QE, etc. This is what `tdx-attest-rs` code does. Cons: performance. Pros: no more code.

I have no inclination, I just quote this part of the code out of respect for @danmihai1 's [existing work](https://github.com/kata-containers/kata-containers/pull/8469/files#diff-0ef258b1d68f4e2145febaede94507bbe0d168cdb5a75c2df1c231254e6ae859).

attestation-agent/lib/src/lib.rs
Xynnn007: In the future, we might want to allow carrying the initdata plaintext for more fine-grained check & matches. But current designs we do not need this and only use the initdata digest. I left this API parameter as `Option` for future use.

This once mentioned [the proposal](https://github.com/confidential-containers/confidential-containers/issues/171#issue-1985098907), Step 7 in Technical details

attestation-agent/protos/attestation-agent.proto
```diff
@@ -25,8 +25,16 @@ message ExtendRuntimeMeasurementRequest {
 
 message ExtendRuntimeMeasurementResponse {}
 
+message CheckInitDataRequest {
+    bytes InitDataDigest = 1;
+    optional bytes InitDataPlaintext = 2;
```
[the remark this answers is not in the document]
Xynnn007: Yes. Let me together change the API of AA lib.

attestation-agent/lib/src/lib.rs
Xynnn007: Inspired by https://github.com/confidential-containers/guest-components/pull/462/files#r1469814161, I'd make this an enum. Please help to check again if it is good.

attestation-agent/attester/src/snp/hostdata.rs
Xynnn007: Sounds good. Let me add the concrete error type for SNP first. Let's wait for how TDX should be implemented and then I will add for that.

Xynnn007:
Thanks @mkulke for the suggestions. I left the review conversions which might need to be ensured by you. PTAL

Hi @fitzthum 

> One thing I think we might have forgotten about when designing: ideally the KBS/AS will end up with the plaintext of the InitData so that it can be validated with a policy. We don't seem to have a mechanism for this. Perhaps if the plaintext is provided via this API it should be saved by the AA so that it can be sent as part of remote attestation. We could probably implement this in a follow-on.

Yes, agreed. Added the enum of `initdata` for plaintext, which could be relied on. Also some conversions to be determined. PTAL

attestation-agent/attester/src/snp/mod.rs
mkulke: I see. maybe it's a good idea to introduce a generic `pad()` helper fn, since we have to use that often:

```rust
fn pad<const T: usize>(input: &[u8]) -> [u8; T] {
    let mut output = [0; T];
    let len = input.len();
    if len > T {
         output.copy_from_slice(&input[..T]);
    } else {
         output[..len].copy_from_slice(input);
    }
    output
}

fn main() {
    let x: [u8; 5] = pad(&vec![1,2,3,4]);
    println!("padded: {x:?}");
}
```

attestation-agent/attester/src/tdx/mrconfigid.rs
mkulke: For consistency we could use `tdx_attest_rs::tdx_att_get_report()`, no? we can still use the TdReport/TdInfo types (or the parts that we need), but interacting with ioctl devices (or any devices) the feels out-of-scope for this crate.

Maybe like Dan suggested, the ioctl could be put in a shared crate, but I guess that's pretty much what the tdx_attest_rs crate from Intel does (minus the types, apparently).

Xynnn007: Oh! Right. I nearly forgot the function `tdx_att_get_report()`. Let me fix this.

attestation-agent/lib/src/lib.rs
mkulke: The link doesn't work for me, but I assume the parameter would be something like this? (pseudocode):

```
enum InitData {
  Hash(Vec<u8>),
  PlainText(Vec<u8>),
  HashAndPlainText({ hash: Vec<u8>, plain_text: Vec<u8>}),
}
```
It might be cumbersome to use, if we really anticipate to pass an init_data body to the function and it doesn't end up being a perpetual dead parameter, I think the Option is fine.

Alternatively, we can also go with the a limited `check_init_data_digest()` fn for now and introduce a more comprehensive `check_init_data()` fn later.

attestation-agent/attester/Cargo.toml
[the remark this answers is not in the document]
fitzthum: You're right we can use the attester crates without using cc-kbc, but it is a bit weird semantically. For now I think it is fine.

attestation-agent/protos/attestation-agent.proto
```diff
@@ -25,8 +25,28 @@ message ExtendRuntimeMeasurementRequest {
 
 message ExtendRuntimeMeasurementResponse {}
 
+message InitDataPlaintext {
+    bytes Content = 1;
+    string Algorithm = 2; 
+}
```
fitzthum, line 31: So it's worth noting that this is InitData specific code. If we still haven't reached an agreement there, maybe we should leave support for the Plaintext to a later PR.

fwiw, I think this shows a clear benefit of the InitData approach. The policy does not include any information about how it is hashed and even if it did, we would have to do parsing of OPA to get it. We could add an additional field to pass the hash type from the shim (where the data is first hashed), to the kata agent, to the AA, to the KBS, but it seems much nicer to have it simply included in the InitData. cc @mkulke

attestation-agent/attester/src/snp/mod.rs
```diff
@@ -47,4 +49,18 @@ impl Attester for SnpAttester {
 
         serde_json::to_string(&evidence).context("Serialize SNP evidence failed")
     }
+
+    async fn check_init_data(&self, init_data: &[u8]) -> Result<()> {
```
jialez0, line 53: I think of an issue: I'm not sure whether it's a good idea to pass init data as an API parameter, because this will lead to any user mode program can pass in an init data for AA to check, which may lead to violent collision cracking of initdata.

Perhaps a better way is to let AA read the init data to be verified from a file with a fixed path. This can ensure that only users with specific permissions (which can modify that file) can use this function of AA by setting file access permissions.

Xynnn007: I might get your point. The core idea of initdata is to protect the integrity, while the confidentiality is not.

Because the init data plaintext is set by untrusted host when the VM is created, we might not need to worry about the confidentiality.

Furthermore, the API of AA is now provided by forwarding from ASR(API Server Restful) via restful api, which could be configured which api of AA could be exposed to app containers. This could help to prevent unexpected access.

fitzthum: Interesting point. I think as @Xynnn007 says it is probably not an issue. The InitData is not secret and this API won't be available to the workload containers. I can't think of any attack where a userspace component could leverage this API to do something dangerous.

attestation-agent/protos/attestation-agent.proto
fitzthum: I don't mind having one endpoint for both even if only the digest is supported at first. I think we will need a plaintext option for the attestation flow to really add up, but this will require some agreement on the format.

Btw, @Xynnn007 this `InitDataPlaintext` message doesn't quite match up with your RFC spec which contains a version and digest field as well.

Xynnn007: > Btw, @Xynnn007 this InitDataPlaintext message doesn't quite match up with your RFC spec which contains a version and digest field as well.

Yes. I once hope that the format in RFC is handled inside kata-agent. That means
- version check
- data field into files
- use the digest field & hash alg to match the data field and call AA's `CheckInitData()` semantically /  DIrectly use data field and hash alg to call AA's `CheckInitDataPlaintext()` semantically, where the `InitDataPlaintext`

@mkulke I once thought that one API is enough, and to decrease the potential adaptation work when API changes, I used one API to receive different kinds of parameters as they both are "checking the initdata field binding relationship". 

But with your opinions I realize that the steps are probably relatively big. Let me add one simple parameter `initDataDigest` and left the plaintext extension in future together with the format issue.

attestation-agent/attester/src/snp/mod.rs
Xynnn007: Added this, thanks!

Xynnn007:
Thanks @fitzthum @mkulke @jialez0 , I fixed the suggestions you've left.
1. Use `pad()` 
2. Only check digest for now and API
3. use `tdx_attest::get_evidence` rather than impl manually

PTAL

mkulke (approved):
lgtm

fitzthum (approved):
LGTM

jialez0 (approved):
LGTM

jiazhang0:
@fitzthum @Xynnn007 @jialez0 Let me clarify something for the future decision and this time just let it go.

I think it is an not easy decision to add a new API which may introduces extra attack interface. In other word, there is still the possibility to use an approach to check the validation of init data without adding new API. In addition, CoCo is not the only case where AA will be deployed. In some cases, the proxy is not used.

Furthermore, considering certain security principles such as defense in depth and the principle of least privilege, adding a new API is not the only solution to address the problem and requirement at hand. Even if someone argues that the newly added API has successfully passed penetration testing without any vulnerabilities found, the results of penetration testing can only provide temporary confidence. Past performance is not indicative of future results. Therefore, we cannot exclude the potential for unknown vulnerabilities and security risks with new API.

Security often focuses on the future possibilities that arise when the timeline is extended, and aims to proactively address potential risks and threats that may arise in the future.

PR #463, opened 2024-01-29
Fix: Use strum string to parse AA token type string

jialez0:


Cargo.toml
```diff
@@ -56,6 +56,7 @@ serde_json = "1.0"
 serial_test = "1"
 sha2 = "0.10.7"
 strum = { version = "0.25", features = ["derive"] }
+strum_macros = "0.25"
```
Xynnn007, line 59: We do not need this dep when `derive` feature is enabled for `strum`

Xynnn007 (approved):
lgtm. Only a comment

PR #464, opened 2024-01-31
cdh/secret: add unseal, ehsm support and integration test

1570005763:
1. eHSM support for secret management is added. We can build Sealed Secret with eHSM now.
2. Unseal is added to secret_cli tool. We can unseal Sealed Secret with the tool now.
3. Integration test is added for secret_cli tool. Thought it can not run on CI, since either Aliyun KMS server or eHSM KMS Server is required. I test them locally, and here are outputs.
  - test with Aliyun KMS server
    image
  - test with eHSM KMS server
    image


confidential-data-hub/secret/src/bin/secret_cli.rs
```diff
@@ -58,9 +79,18 @@ struct EnvelopeCommand {
...
     /// Alibaba KMS driver to seal the envelope
+    #[cfg(feature = "aliyun")]
     Ali(AliProviderArgs),
+
+    /// Intel eHSM driver to seal the envelope
+    #[cfg(feature = "ehsm")]
+    Ehsm(EhsmProviderArgs),
+
+    /// NoFeatureOpened is added in case no feature is opened, which is not prefered
+    NoFeatureOpened,
```
fitzthum, line 90: It's a bit weird that this enum will still have `NoFeatureOpened` even if other features are enabled. I wonder if we want to create some kind of SampleKMS that can always be enabled as a fallback.

fitzthum (commented):
A few initial comments.

Xynnn007 (approved):
Nice work. Only some minor comments. Please also wait for a double check from @fitzthum

confidential-data-hub/secret/src/bin/secret_cli.rs
1570005763: KMS plugins may not be secure enough to save encrypt keys. If no KMS is supported, I think KBS is better to be the fallback to save secrets. And `NoFeatureOpened` is weird indeed. So I revised the implementation of this part. Hope it looks better.

confidential-data-hub/secret/src/bin/secret_cli.rs
```diff
@@ -80,12 +106,57 @@ struct AliProviderArgs {
...
-            // TODO
+        Cli::Unseal(para) => {
+            let secret_json = fs::read(&para.file_path)
+                .await
+                .expect("failed to read sealed secret");
+            let secret: Secret = serde_json::from_slice(&secret_json)
+                .expect("illegal input sealed secret format (json deseralization failed)");
+
+            match secret.r#type {
+                SecretContent::Envelope(ref envlope) => match envlope.provider.as_str() {
```
fitzthum, line 133: nit: `envlope` -> `envelope`

confidential-data-hub/secret/src/bin/secret_cli.rs
```diff
@@ -80,12 +106,57 @@ struct AliProviderArgs {
     client_key_file_path: String,
 }
 
+#[cfg(feature = "ehsm")]
+#[derive(Args)]
+struct EhsmProviderArgs {
+    /// path of the crendential file
+    #[arg(short, long)]
+    crendential_file_path: String,
```
fitzthum, line 114: nit: `crendential_file_path` -> `credential_file_path`.

confidential-data-hub/secret/src/bin/secret_cli.rs
```diff
@@ -80,12 +106,57 @@ struct AliProviderArgs {
...
+    endpoint: String,
+}
+
 #[tokio::main]
 async fn main() {
     let args = Cli::parse();
     match args {
-        Cli::Unseal => {
-            // TODO
+        Cli::Unseal(para) => {
```
fitzthum, line 125: It might be nice to make a separate function for unseal and seal rather than have main be so large. also, nit: `para` -> `unseal_args`

confidential-data-hub/secret/tests/secret_cli.rs
```diff
@@ -0,0 +1,187 @@
...
+        .read_to_end(&mut unsealed_secret)
+        .expect("read unsealed_secret_file fail");
+
+    assert_eq!(original_secret, unsealed_secret);
+}
+
+#[test]
+#[ignore]
+#[cfg(feature = "ehsm")]
+fn test_cli_envlope_secret_lifetime_ehsm() {
```
fitzthum, line 102: There is a lot of repetition in this function (with the one above. Is there a way to make some helper functions that can be shared?

fitzthum (commented):
A few more comments but I think this is pretty close

fitzthum (approved):
LGTM. Thnks @1570005763

PR #467, opened 2024-02-04
AA/attester: improve tdx runtime measurement

Lu-Biao:
improve implementation of tdx runtime measurement API, 'extend_runtime_measurement'.

tdx attester should call intel DCAP lib api with the specified parameter type to avoid the high level needing to know the details of the hardware platform when calling the runtime measurement interface.

### DIFF
--- a/attestation-agent/attester/src/tdx/mod.rs
+++ b/attestation-agent/attester/src/tdx/mod.rs
@@ -11,6 +11,8 @@ use base64::Engine;
 use log::debug;
 use scroll::Pread;
 use serde::{Deserialize, Serialize};
+use sha2::{Digest, Sha384};
+use std::mem;
 use std::path::Path;
 use tdx_attest_rs::{self, tdx_report_t};
 
@@ -18,6 +20,7 @@ mod report;
 
 const TDX_REPORT_DATA_SIZE: usize = 64;
 const CCEL_PATH: &str = "/sys/firmware/acpi/tables/data/CCEL";
+const RUNTIME_MEASUREMENT_RTMR_INDEX: u64 = 2;
 
 pub fn detect_platform() -> bool {
     Path::new("/dev/tdx-attest").exists() || Path::new("/dev/tdx-guest").exists()
@@ -81,7 +84,17 @@ impl Attester for TdxAttester {
         _register_index: Option<u64>,
     ) -> Result<()> {
         for event in events {
-            match tdx_attest_rs::tdx_att_extend(&event) {
+            let mut event_buffer = [0u8; mem::size_of::<tdx_attest_rs::tdx_rtmr_event_t>()];
+            let mut hasher = Sha384::new();
+            hasher.update(&event);
+            let hash = hasher.finalize().to_vec();
+            let rtmr_event = unsafe {
+                &mut *(event_buffer.as_mut_ptr() as *mut tdx_attest_rs::tdx_rtmr_event_t)
+            };
+            rtmr_event.version = 1;
+            rtmr_event.rtmr_index = RUNTIME_MEASUREMENT_RTMR_INDEX;
+            rtmr_event.extend_data.copy_from_slice(&hash);
+            match tdx_attest_rs::tdx_att_extend(&event_buffer) {
                 tdx_attest_rs::tdx_attest_error_t::TDX_ATTEST_SUCCESS => {
                     log::debug!("TDX extend runtime measurement succeeded.")
                 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -291,6 +291,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sev 1.2.1",
+ "sha2 0.10.8",
  "strum",
  "tdx-attest-rs",
  "thiserror",
--- a/attestation-agent/attester/Cargo.toml
+++ b/attestation-agent/attester/Cargo.toml
@@ -25,6 +25,7 @@ serde_json.workspace = true
 sev = { version = "1.2.0", default-features = false, features = [
     "snp",
 ], optional = true }
+sha2.workspace = true
 strum.workspace = true
 tdx-attest-rs = { git = "https://github.com/intel/SGXDataCenterAttestationPrimitives", tag = "DCAP_1.16", optional = true }
 thiserror.workspace = true


Xynnn007:
Hi @Lu-Biao , overall I think this PR looks good. A question from my side is do you think it is good to have a standard eventlog format / API definition in AA to record the events like image pulling, and TEE-platform specific way to binding the integrity of the eventlog? like using RTMR for TDX, and PCR for vTPM platforms.

I draft two commits aiming for this
https://github.com/Xynnn007/guest-components/commit/bf9d79f7923cf8e5b7375c51ca42e64cd56a3070 for eventlog definition and https://github.com/Xynnn007/guest-components/commit/fd07a8433da9a09c2b877696b1d6629b587e550e for TDX implementation

I don't think the things above will not block this PR. They might be future work.

arronwy:
@binxing

fitzthum:
@Xynnn007 

> A question from my side is do you think it is good to have a standard eventlog format / API definition in AA to record the events like image pulling, and TEE-platform specific way to binding the integrity of the eventlog? like using RTMR for TDX, and PCR for vTPM platforms.

There is a [kernel thread](https://lore.kernel.org/linux-integrity/20240128212532.2754325-1-sameo@rivosinc.com/) sort of about this

Xynnn007:
> @Xynnn007
> 
> > A question from my side is do you think it is good to have a standard eventlog format / API definition in AA to record the events like image pulling, and TEE-platform specific way to binding the integrity of the eventlog? like using RTMR for TDX, and PCR for vTPM platforms.
> 
> There is a [kernel thread](https://lore.kernel.org/linux-integrity/20240128212532.2754325-1-sameo@rivosinc.com/) sort of about this

I took a check upon the thread. AFAIK, the work focuses on kernel API to extend RTMR registers. The influence upon AA would not be the concrete implementations of `extend_runtime_measurement()` functions for TDX/CCA/RISC-V. However, the eventlog definition and API of AA would not be influenced.

image

binxing:
Hi @Xynnn007, the kernel ABI discussion was started by @sameo with a TSM patch without logging support, then questions arose on whether or not the event log  should be kept by the kernel as part of the ABI and what the log format should be. AA wouldn't have to keep its own log should the kernel community agree to keep/maintain the log in the kernel.

There are proposals on using the existing TCG2 or the CEL log formats but IMHO neither can satisfy CoCo's requirements, so I proposed a [text/line based format](https://lore.kernel.org/linux-integrity/226df539-b3f4-4099-b6a6-293fa200c536@intel.com/) and tried to articulate the [reasoning](https://lore.kernel.org/linux-integrity/fe77c42c-7a61-49ee-b729-abc87ae6e352@intel.com/). I guess we (the CoCo community) should discuss the same topic too and voice our preference/concerns in the kernel community,

Xynnn007:
@binxing Nice. If kernel will keep the eventlog, AA does not need to. And IMO it's better for kernel to keep the eventlog because if kernel cannot, basically any process with root privilege could keep its own, which would make conflicts.



> There are proposals on using the existing TCG2 or the CEL log formats but IMHO neither can satisfy CoCo's requirements, so I proposed a [text/line based format](https://lore.kernel.org/linux-integrity/226df539-b3f4-4099-b6a6-293fa200c536@intel.com/) and tried to articulate the [reasoning](https://lore.kernel.org/linux-integrity/fe77c42c-7a61-49ee-b729-abc87ae6e352@intel.com/). I guess we (the CoCo community) should discuss the same topic too and voice our preference/concerns in the kernel community,

I have roughly read [the spec](https://trustedcomputinggroup.org/wp-content/uploads/TCG_PCClient_PFP_r1p05_05_3feb20.pdf) and find that there is an event type named `Tagged Event Log` in section `9.4.2`. It says
> This event type was originally specified in the TCG PC Client Implementation Specification for Conventional BIOS.
It was deprecated in the original version of the PC Client Specific Platform Firmware Profile for TPM 2.0 systems.
As a result of more interest in platform attestation services, which are not just confined to the results of the Platform
Firmware measurements, but also the host platform OS and software environment, this event type has been
redefined in this specification for use by OS and software on a PC Client platform. Other than the Event Type and
structure, it is not standardized.

And the `taggedEventID`
> This is a unique identifier defined by the measuring OS or application. Note: This is Host OS or software defined data and not defined by this specification.

`taggedEventData`
> The data of the event. Note: This is Host OS or software defined data and not defined by this specification.

This means we could define arbitrary event types based on `Tagged Event`, which follows TCG standard. Surely, a text/human readable format is also ok to embed as `taggedEventData`. This is what I did in the commits.

I am not sure if this makes sense?

binxing:
@Xynnn007, thanks for digging up the details in the TCG PC Client spec! Yes, `EV_EVENT_TAG` will work for CoCo as we are allowed to choose arbitrary values for `taggedEventID` for whatever events we define. With that said, are we saying we prefer TCG2 format?

The discussion on event log format is out of scope for, hence shouldn't block, this PR. Shall we merge this PR and continue event log discussion in a new issue?

Xynnn007:
> @Xynnn007, thanks for digging up the details in the TCG PC Client spec! Yes, `EV_EVENT_TAG` will work for CoCo as we are allowed to choose arbitrary values for `taggedEventID` for whatever events we define. With that said, are we saying we prefer TCG2 format?
> 
> The discussion on event log format is out of scope for, hence shouldn't block, this PR. Shall we merge this PR and continue event log discussion in a new issue?

Yes. To me this is another thread and will never block this PR.

Xynnn007 (approved):

Xynnn007:
cc @jialez0 might be interested too

mythi:
> To sum up, for CoCo usage, we can define a lot of events like `ImagePulling`, `ContainerDeployment`, etc. The deserialization and serialization code could be a separate crate.

we can but the question is what is useful/really needed. With kubernetes, we have no control over the ordering of the container events inside the pod so that easily leads to many sets of reference values we'd have to replay to do the verification

mythi (approved):
LGTM also but one generic coding style question: when would it make sense to add `tdx_attest_rs::*` to the scope?

Xynnn007:
> we can but the question is what is useful/really needed. With kubernetes, we have no control over the ordering of the container events inside the pod so that easily leads to many sets of reference values we'd have to replay to do the verification

My first impression is this depends how to define the events.

If the events are for image pulling, then the order does not matter. The replay only checks the integrity of data content but not order. Then the separate image pull events could be verified. As the order change does not matter. If there are true scenarios that event log order matters, the origin of the events would also need to carefully implement. At that time we can handle from the source.

mythi:
> Then the separate image pull events could be verified. As the order change does not matter.

Ah, right, so it's just an additional step in the verification.

arronwy (approved):
Thanks @Lu-Biao LGTM!

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -81,7 +84,17 @@ impl Attester for TdxAttester {
         _register_index: Option<u64>,
     ) -> Result<()> {
         for event in events {
-            match tdx_attest_rs::tdx_att_extend(&event) {
+            let mut event_buffer = [0u8; mem::size_of::<tdx_attest_rs::tdx_rtmr_event_t>()];
+            let mut hasher = Sha384::new();
+            hasher.update(&event);
+            let hash = hasher.finalize().to_vec();
+            let rtmr_event = unsafe {
```
fitzthum, line 91: Any kind of validation we can do here before the unsafe cast? This API could be called by anyone in user space, right?

fitzthum (commented):
One comment.

attestation-agent/attester/src/tdx/mod.rs
Xynnn007: The definition of the underlying `tdx_rtmr_event_t` is [here](https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/master/QuoteGeneration/quote_wrapper/tdx_attest/tdx_attest.h#L89-L96)

The lib inside will [cast it too](https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/master/QuoteGeneration/quote_wrapper/tdx-attest-rs/src/lib.rs#L208). 
If we have an same size of memory buffer `event_buffer` as `tdx_rtmr_event_t`, we can cast the struct. What we need to do is to fill some fields as what `DCAP` expects, which is implemented in the following sentences.

I think the unavoidable `unsafe` here is because the underlying crate `tdx-attest-sys` uses [bindgen](https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/master/QuoteGeneration/quote_wrapper/tdx-attest-sys/build.rs#L59-L78) which will only convert a C struct into a Rust struct without making it `pub` or including any (de)serialization implementations. The result is we can only use memory and pointers to do the dirty (de)serialization work.

Xynnn007: btw, I've get in touch with DCAP team to update the code. Hopefully after some updates we can avoid unsafe calling in upper stack, s.t. in AA.

arronwy: @fitzthum can we merge this PR and update `unsafe` part when dcap team upgrade their API?

fitzthum: Yeah that's fine. Sorry for the delay.

Issue #470, opened 2024-02-06
Image-rs build fails on rust nightly-2024-02-05


ChengyuZhu6:
Rustc 1.78.0-nightly (f067fd608 2024-02-05) was the latest nightly release, but it has a compatibility issue with curve25519-dalek. You can see the details of the issue https://github.com/dalek-cryptography/curve25519-dalek/issues/618. This issue also affects guest-components, as shown by this https://github.com/confidential-containers/guest-components/actions/runs/7795476730/job/21258488007?pr=469. Therefore, any PRs related to image-rs might fail the nightly test job under this version.

PR #471, opened 2024-02-06
Cargo.lock: Update dep of curve25519-dalek and x25519-dalek

ChengyuZhu6:
Rustc 1.78.0-nightly (f067fd608 2024-02-05) was the latest nightly release, but it has a compatibility issue with
curve25519-dalek https://github.com/dalek-cryptography/curve25519-dalek/issues/618.

Now curve25519-dalek and x25519-dalek have a new release to fix the
issue. So we can update the dependencies about curve25519-dalek and
x25519-dalek.

Fixes: #470

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1246,9 +1246,9 @@ dependencies = [
 
 [[package]]
 name = "curve25519-dalek"
-version = "4.1.1"
+version = "4.1.2"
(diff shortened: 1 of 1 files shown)


Xynnn007 (approved):
Cool. Thanks for the fix

PR #476, opened 2024-02-14
CI: fix rust-nightly static checks

portersrc:
This PR attempts to fix these two rust nightly errors we started seeing:
```
    Checking image-rs v0.1.0 (/home/porter/wo/confidential-containers/guest-components/image-rs)
error: trait `Protocol` is never used
  --> image-rs/src/resource/mod.rs:33:7
   |
33 | trait Protocol: Send + Sync {
   |       ^^^^^^^^
   |
   = note: `-D dead-code` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(dead_code)]`

error: direct implementation of `ToString`
  --> image-rs/src/signature/image/digest.rs:79:1
   |
79 | / impl ToString for Digest {
80 | |     fn to_string(&self) -> String {
81 | |         format!("{}:{}", self.algorithm, self.value)
82 | |     }
83 | | }
   | |_^
   |
   = help: prefer implementing `Display` instead
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#to_str
ing_trait_impl
   = note: `-D clippy::to-string-trait-impl` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(clippy::to_string_trait_impl)]`
```

### DIFF
--- a/image-rs/src/pull.rs
+++ b/image-rs/src/pull.rs
@@ -470,7 +470,7 @@ mod tests {
 
         for image_url in nydus_images.iter() {
             let tempdir = tempfile::tempdir().unwrap();
-            let image = Reference::try_from((*image_url).clone()).expect("create reference failed");
+            let image = Reference::try_from(*image_url).expect("create reference failed");
             let mut client = PullClient::new(
                 image,
                 tempdir.path(),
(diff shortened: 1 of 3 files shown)


fitzthum (approved):
Nice. The dead code warning is slightly interesting. I don't think that is dead code, but it's being consumed by another crate. I have a vague memory of this being a problem in the past. I'm not against the fix here.

Issue #485, opened 2024-02-27
AA | Add IBM Secure Execution driver for attester


huoqifeng:
We'd like to add s390x attestation support and add the driver in attester.
s390x support remote attestation via [pvattest](https://github.com/ibm-s390-linux/s390-tools/tree/master/pvattest), we'll introduce this binaries and help facilitate the attestation flow. Which include:
- [ ] Add the pvattest bin for s390x -- might be rust crate.
- [ ] Add s390x driver in attester
- [ ] Add corresponding code in guest-components to support the new attester.  

KBS flow will be addressed in a separate issue in [kbs](https://github.com/confidential-containers/trustee/issues/342)

fitzthum:
This sounds really cool. I am curious what the whole flow will look like.

PR #492, opened 2024-02-29
aa/attester: add IBM Secure Execution driver framework

huoqifeng:
This is the initial draft to fix: https://github.com/confidential-containers/guest-components/issues/485 and the draft PR provides the API framework without a real pvattest crate yet.

Try it after adding the new Tee type `Tee::Se` in https://github.com/virtee/kbs-types/blob/main/src/lib.rs#L24, run:
```
cd guest-components
TEE_PLATFORM=se KBC=cc_kbc make
```
Depends on: 
- [x] https://github.com/virtee/kbs-types/issues/26


Xynnn007:
Thanks @huoqifeng for the contribution! I'd love to review this if it is ok : D

huoqifeng:
@Xynnn007 this is not fully ready for review, but I think we can start discuss the potential changes both on guest-components side and on trustee side, which is tracked by https://github.com/confidential-containers/trustee/pull/345.

We want to move the `challenge` generation from session to attestation-service and introduce a new fn `generate_challenge`. like in https://github.com/huoqifeng/trustee/blob/s390x-verifier/attestation-service/attestation-service/src/lib.rs#L243, we want use this `challenge.extra_params` in guest like https://github.com/huoqifeng/guest-components/blob/s390x-attester/attestation-agent/kbs_protocol/src/client/rcar_client.rs#L140

The reason is: for IBM Secure Execution (SE), `attester` need an `attestation-request` to generate the `evidence`, while the `attestation-request` should always be created on `verifier`. CC @BbolroC  @steffen-eiden

huoqifeng:
@Xynnn007 @fitzthum I'd like start the review process of the API change to support IBM SE. May you please help on that?

attestation-agent/attester/src/se/seattest.rs
fitzthum, line 1: Might want to rename this file. It looks like it says "seat test"

attestation-agent/attester/src/se/seattest.rs
```diff
@@ -0,0 +1,27 @@
...
+//
+
+use anyhow::*;
+use async_trait;
+
+#[derive(Default)]
+pub struct FakeSeAttest {}
+
+#[async_trait::async_trait]
+pub trait SeImplAttester {
```
fitzthum, line 13: This trait doesn't seem necessary to me. Why not just put the logic in the methods above (and call into the real SE crate from there once it is ready)? Also, `SeImplAttester` is a slightly strange name for a trait.

attestation-agent/kbs_protocol/test/policy.rego
```diff
@@ -5,3 +5,7 @@ default allow = false
 allow {
 	input["tee"] == "sample"
 }
+
+allow {
+        input["tee"] == "se"
+}
```
fitzthum, line 11: I think this policy file is intended to be used with the nontee (sample tests) so I wouldn't add the SE clause here. Isn't the fake SE attester always returning false for platform detection anyway?

fitzthum (commented):
A few high-level comments. I am still thinking about the bigger changes to the protocol.

attestation-agent/attester/src/se/seattest.rs
huoqifeng: Thanks @fitzthum , we'll put the crate in https://crates.io/crates/s390_pv, it's still in-progress.

attestation-agent/attester/src/se/seattest.rs
huoqifeng: Thanks @fitzthum , we'll put the real crate in https://crates.io/crates/s390_pv, it's still in-progress. So, `FakeSeAttest` and `SeImplAttester` will be replaced the crate eventually.

attestation-agent/kbs_protocol/test/policy.rego
huoqifeng: Yeah, this is what I added when testing se driver, we need drop it when merge.

attestation-agent/kbs_protocol/src/client/rcar_client.rs
```diff
@@ -179,11 +181,19 @@ impl KbsClient<Box<dyn EvidenceProvider>> {
...
+    ) -> Result<String> {
         let mut hasher = Sha384::new();
         hasher.update(runtime_data);
 
-        let ehd = hasher.finalize().to_vec();
+        let mut ehd = hasher.finalize().to_vec();
+        // IBM SE uses challenge_extra_params as runtime_data to pass attestation_request
+        if challenge_extra_params.chars().count() > 0 {
+            ehd = challenge_extra_params.into_bytes();
+        }
```
mkulke, line 196: do we want to have this as default behaviour? other TEE  impls might have `extra_params` set and still want to use the hashed runtime_data.

we have the `tee` available on the calling function site, so we can do a special case handling.

huoqifeng: Yes, do some special for IBM SE and other TEE based on `tee` is alternative and I think it's hood also.

huoqifeng: Fixed

attestation-agent/kbs_protocol/test/policy.rego
huoqifeng: Fixed.

attestation-agent/attester/src/se/seattest.rs
huoqifeng: Fixed

attestation-agent/attester/src/se/seattest.rs
huoqifeng: Fixed.

fitzthum:
@huoqifeng Are we waiting for the underlying SE crate before we merge this? Any idea when it will be ready?

huoqifeng:
@fitzthum because the code `tools/client` in trustee PR depends on the `kbs_protocol` in guest-components, I'm think we merge this PR after review completed without the underlying IBM SE crate, and then add it after the PR in trustee merged. So the steps will be:
- Review and merge  https://github.com/confidential-containers/guest-components/pull/492 without IBM SE crate
- Review and merge https://github.com/confidential-containers/trustee/pull/345/ with IBM SE crate (@Xynnn007 suggest we should add the real crate in trustee)
- Add IBM SE crate in guest-components in a new PR
@fitzthum @Xynnn007 wdyt?

Xynnn007 (commented):
There are some nits. Let's make it better

attestation-agent/attester/src/se/ibmse.rs
```diff
@@ -0,0 +1,148 @@
...
+            encr_measurement_key: encr_measurement_key.to_vec(),
+            encr_request_nonce: encr_request_nonce.to_vec(),
+            image_hdr_tags: *image_hdr_tags,
+        })
+    }
+}
+
+pub fn calc_userdata() -> Result<UserData> {
+    // TODO, calculate userdata based on the boot partition
+    let image_btph = "ddddddd";
```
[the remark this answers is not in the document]
huoqifeng: UserData is optional check point, it's not checked on trustee side yet. We're still in discussion what value should be put in.

attestation-agent/attester/src/se/ibmse.rs
```diff
@@ -0,0 +1,148 @@
...
+    request::BootHdrTags,
+    uv::{AttestationCmd, ConfigUid, UvDevice},
+};
+use serde::{Deserialize, Serialize};
+use serde_json;
+use serde_with::{base64::Base64, serde_as};
+
+#[allow(unreachable_code)]
+pub fn is_se_guest() -> bool {
+    #[cfg(not(target_arch = "s390x"))]
```
mkulke, line 22: Can we move this to attester/src/lib, since the whole module doesn't make sense if the target arch is not s390x?

attestation-agent/attester/Cargo.toml
```diff
@@ -15,9 +15,11 @@ kbs-types.workspace = true
 log.workspace = true
 nix = { version = "0.28", optional = true, features = ["ioctl", "fs"] }
 occlum_dcap = { git = "https://github.com/occlum/occlum", tag = "v0.29.7", optional = true }
+pv = { version = "0.10.0", package = "s390_pv" }
```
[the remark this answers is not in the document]
huoqifeng: Fixed, thanks!

attestation-agent/attester/Cargo.toml
```diff
@@ -64,5 +67,6 @@ az-tdx-vtpm-attester = ["az-tdx-vtpm"]
 snp-attester = ["sev"]
 csv-attester = ["csv-rs", "codicon", "hyper", "hyper-tls", "tokio"]
 cca-attester = ["nix"]
+se-attester  = []
```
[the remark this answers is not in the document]
huoqifeng: Fixed, thanks!

attestation-agent/attester/src/se/ibmse.rs
```diff
@@ -0,0 +1,148 @@
+// Copyright (C) Copyright IBM Corp. 2024
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+// Copyright (C) Copyright IBM Corp. 2024
+//
+// SPDX-License-Identifier: Apache-2.0
+//
```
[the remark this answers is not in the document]
huoqifeng: Fixed, thanks!

attestation-agent/attester/src/se/ibmse.rs
```diff
@@ -0,0 +1,148 @@
...
+}
+
+impl SeAttestationRequest {
+    pub fn from_slice(request: &[u8]) -> Result<Self> {
+        Ok(serde_json::from_slice(request).unwrap())
+    }
+
+    pub fn from_string(request: &str) -> Result<Self> {
+        Ok(serde_json::from_str(request).unwrap())
+    }
```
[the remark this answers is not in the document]
huoqifeng: Fixed, thanks!

attestation-agent/attester/src/se/ibmse.rs
```diff
@@ -0,0 +1,148 @@
...
+    let image_btph = "ddddddd";
+    Ok(UserData {
+        image_btph: image_btph.into(),
+    })
+}
+
+pub fn perform(req: &[u8], userdata: &UserData) -> Result<SeAttestationResponse> {
+    // req is serialized SeAttestationRequest String bytes
+    let req_str = std::str::from_utf8(req)?;
+    let request = SeAttestationRequest::from_string(req_str)?;
```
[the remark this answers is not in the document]
huoqifeng: fixed, thanks!

attestation-agent/attester/src/se/ibmse.rs
```diff
@@ -0,0 +1,148 @@
...
+            measurement: measurement.to_vec(),
+            additional_data: additional_data.to_vec(),
+            user_data: user_data.to_vec(),
+            cuid: *cuid,
+            encr_measurement_key: encr_measurement_key.to_vec(),
+            encr_request_nonce: encr_request_nonce.to_vec(),
+            image_hdr_tags: *image_hdr_tags,
+        })
+    }
+}
```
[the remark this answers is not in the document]
huoqifeng: fixed, thanks!

attestation-agent/attester/src/se/ibmse.rs
```diff
@@ -0,0 +1,148 @@
...
+
+    SeAttestationResponse::create(
+        &measurement,
+        &additional_data,
+        &user_data,
+        cuid,
+        encr_measurement_key,
+        encr_request_nonce,
+        image_hdr_tags,
+    )
```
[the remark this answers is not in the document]
huoqifeng: Fixed, thanks!

attestation-agent/attester/src/se/ibmse.rs
```diff
@@ -0,0 +1,148 @@
...
+
+    pub fn to_uvc(&self, user_data: &[u8]) -> Result<AttestationCmd> {
+        let cmd = AttestationCmd::new_request(
+            self.request_blob.clone().into(),
+            Some(user_data.to_vec()),
+            self.measurement_size,
+            self.additional_size,
+        )?;
+        Ok(cmd)
+    }
```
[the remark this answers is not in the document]
huoqifeng: Fixed, thanks!

attestation-agent/attester/src/se/ibmse.rs
```diff
@@ -0,0 +1,148 @@
...
+    #[serde_as(as = "Base64")]
+    encr_request_nonce: Vec<u8>,
+    #[serde_as(as = "Base64")]
+    image_hdr_tags: BootHdrTags,
+}
+
+impl SeAttestationRequest {
+    pub fn from_slice(request: &[u8]) -> Result<Self> {
+        Ok(serde_json::from_slice(request).unwrap())
+    }
```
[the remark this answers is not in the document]
huoqifeng: Fixed, thanks!

Cargo.toml
```diff
@@ -51,7 +51,9 @@ resource_uri = { path = "attestation-agent/deps/resource_uri" }
 ring = "0.17"
 rsa = "0.9.2"
 rstest = "0.17"
+s390_pv = "0.10.0"
```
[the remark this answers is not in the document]
huoqifeng: Fixed, thanks!

attestation-agent/kbs_protocol/src/client/rcar_client.rs
```diff
@@ -179,11 +181,21 @@ impl KbsClient<Box<dyn EvidenceProvider>> {
...
         let mut hasher = Sha384::new();
         hasher.update(runtime_data);
 
-        let ehd = hasher.finalize().to_vec();
+        let mut ehd = hasher.finalize().to_vec();
+        // IBM SE uses challenge_extra_params as runtime_data to pass attestation_request
+        if tee == Tee::Se {
+            // challenge_extra_params is serialized SeAttestationRequest String
+            ehd = challenge_extra_params.into_bytes();
+        }
```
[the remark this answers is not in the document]
huoqifeng: Fixed, thanks!

attestation-agent/attester/src/se/mod.rs
```diff
@@ -0,0 +1,44 @@
...
+
+#[async_trait::async_trait]
+impl Attester for SeAttester {
+    async fn get_evidence(&self, attestation_request: Vec<u8>) -> Result<String> {
+        let userdata = ibmse::calc_userdata()?;
+        debug!("userdata json: {:#?}", &userdata.clone());
+        // attestation_request is serialized SeAttestationRequest String bytes
+        let evidence = ibmse::perform(&attestation_request, &userdata)?;
+        debug!("response json: {:#?}", evidence.clone());
+        serde_json::to_string(&evidence).context("Serialize SE evidence failed")
```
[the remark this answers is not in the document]
huoqifeng: Fixed, thanks!

attestation-agent/attester/src/se/ibmse.rs
```diff
@@ -0,0 +1,148 @@
...
+pub fn perform(req: &[u8], userdata: &UserData) -> Result<SeAttestationResponse> {
+    // req is serialized SeAttestationRequest String bytes
+    let req_str = std::str::from_utf8(req)?;
+    let request = SeAttestationRequest::from_string(req_str)?;
+    let user_data = serde_json::to_vec(userdata)?;
+    let mut uvc: AttestationCmd = request.to_uvc(&user_data)?;
+    let uv = UvDevice::open()?;
+    uv.send_cmd(&mut uvc)?;
+
+    let measurement = uvc.measurement().to_vec().clone();
```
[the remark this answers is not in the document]
huoqifeng: Fixed, thanks!

attestation-agent/attester/src/se/ibmse.rs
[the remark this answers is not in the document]
huoqifeng: Changed it to optional for now.

attestation-agent/attester/src/se/ibmse.rs
huoqifeng: @mkulke there is a `#[cfg(feature = "se-attester")]` to build this module selectively, is that OK for you?

mkulke: to understand the reasoning behind this additional flag: the motivation is to compile the se-attester on all archs into the bin but disable the code at runtime on the wrong archs?

huoqifeng: Yes, intention is to use `#[cfg(feature = "se-attester")]` to include `se` or not, `#[cfg(not(target_arch = "s390x"))]` to make sure it works well also on x86 if `se-attester` selected.

mkulke: But is this what we want? As I understand se-attester would never work on x86_64, why package the code into the bins?

attestation-agent/attester/Cargo.toml
```diff
@@ -15,9 +15,11 @@ kbs-types.workspace = true
 log.workspace = true
 nix = { version = "0.28", optional = true, features = ["ioctl", "fs"] }
 occlum_dcap = { git = "https://github.com/occlum/occlum", tag = "v0.29.7", optional = true }
+pv = { version = "0.10.0", package = "s390_pv", optional = true }
```
mkulke, line 18: is pv_core sufficient maybe, to avoid linking to openssl?

attestation-agent/attester/src/se/ibmse.rs
huoqifeng: Not sure whether its reasonable, I want it also be checked by CI before we add a s390x runner in CI and GHA.

mkulke: ok, to be fair we have the same for the arm attester, we'll include it in the x86_64 bins even if it's dead code. we can leave it like this, I suppose, but maybe check whether we can use pv_core instead to get rid of the openssl dep (it might cause issues in initrd-based guest-components)

huoqifeng: oh, I got your concern!

attestation-agent/attester/Cargo.toml
huoqifeng: Got your real concern.

attestation-agent/attester/src/se/ibmse.rs
huoqifeng: Fixed, thanks!

huoqifeng:
@mkulke @Xynnn007 comments addressed, thanks for all of your review! I'm curious why the CI is not health with no failure task?

mkulke:
> @mkulke @Xynnn007 comments addressed, thanks for all of your review! I'm curious why the CI is not health with no failure task?

some workflows have been cancelled, seemingly for no reason. I've noticed this problem elsewhere this week, I'll re-schedule the cancelled ones, that worked so far.

mkulke (approved):
lgtm

attestation-agent/attester/Cargo.toml
```diff
@@ -51,6 +53,7 @@ all-attesters = [
     "snp-attester",
     "csv-attester",
     "cca-attester",
+    "se-attester",
```
Xynnn007, line 56: We'd better add a comment here.
```
`se-attester` feature can work only on `s390x`  target arch.
```

attestation-agent/attester/src/se/ibmse.rs
```diff
@@ -0,0 +1,105 @@
...
+    uv::{AttestationCmd, ConfigUid, UvDevice},
+};
+use serde::{Deserialize, Serialize};
+use serde_json;
+use serde_with::{base64::Base64, serde_as};
+
+#[allow(unreachable_code)]
+pub fn is_se_guest() -> bool {
+    #[cfg(not(target_arch = "s390x"))]
+    return false;
```
Xynnn007, line 19: As we already mark the code path here would only be executed when `target_arch` is `s390x` [here](https://github.com/confidential-containers/guest-components/pull/492/files#diff-8fd997b18e87172dbdcec4c6ab0268c8d5b8e4dcd7b3a4d493ab1f9519d6c10aR137-R138), we might not need this condition.

attestation-agent/attester/src/se/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+mod tests {
+    use super::*;
+
+    #[ignore]
+    #[tokio::test]
+    async fn test_se_get_evidence() {
+        let attester = SeAttester::default();
+        let report_data: Vec<u8> = vec![0; 64];
+
+        let evidence = attester.get_evidence(report_data).await;
```
Xynnn007, line 36: I am afraid that this unit test cannot be run. As the input `report_data` will be treated as a `SeAttestationRequest` and the test here defines an empty one. Thus I suggest either
1. add a real worked `report_data` in under `https://github.com/confidential-containers/guest-components/tree/main/attestation-agent/attester/test_data` (requires a new dir creation)
2. delete this unit test

attestation-agent/attester/src/se/ibmse.rs
```diff
@@ -0,0 +1,105 @@
...
+        user_data,
+        cuid: *cuid,
+        encr_measurement_key: request.encr_measurement_key,
+        encr_request_nonce: request.encr_request_nonce,
+        image_hdr_tags: request.image_hdr_tags,
+    };
+
+    debug!("response json: {:#?}", response.clone());
+    Ok(serde_json::to_string(&response)?)
+}
```
Xynnn007, line 105: I noticed that the logic under `perform` is direct and compresihensive. Could we move all the logic directly to `get_evidence` in `se/mod.rs`. Also, we might not need extra definition of `calc_userdata()`, could we just 
```rust
let userdata = UserData {
        image_btph: image_btph.into(),
    };
```

Xynnn007 (commented):
Some more nits..

huoqifeng:
@Xynnn007 I guess the failure in https://github.com/confidential-containers/guest-components/actions/runs/9358373555/job/25760080271?pr=492 is caused by network connection?

huoqifeng:
@mkulke @Xynnn007 thank you very much for reviewing the PR and gave so many useful suggestions! thanks for your approval.

attestation-agent/attester/src/se/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+            measurement: uvc.measurement().to_vec(),
+            additional_data,
+            user_data,
+            cuid: *cuid,
+            encr_measurement_key: request.encr_measurement_key,
+            encr_request_nonce: request.encr_request_nonce,
+            image_hdr_tags: request.image_hdr_tags,
+        };
+
+        debug!("response json: {:#?}", response.clone());
```
Xynnn007, line 104: I think we might not need `.clone()` here. Every `.clone()` would alloc a slice of memory.

Xynnn007 (approved):
Looks good. Thanks @huoqifeng .

attestation-agent/attester/src/se/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+impl Attester for SeAttester {
+    async fn get_evidence(&self, req: Vec<u8>) -> Result<String> {
+        // req is serialized SeAttestationRequest String bytes
+        // TODO, calculate optional userdata based on the boot partition etc.
+        let image_btph = "optional check";
+        let userdata = UserData {
+            image_btph: image_btph.into(),
+        };
+
+        debug!("userdata json: {:#?}", &userdata.clone());
```
Xynnn007, line 78: Also this `.clone()`
```suggestion
        debug!("userdata json: {userdata:#?}");
```

attestation-agent/attester/src/se/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+        // req is serialized SeAttestationRequest String bytes
+        let request: SeAttestationRequest = serde_json::from_slice(req)?;
+        let user_data = serde_json::to_vec(&userdata)?;
+        let mut uvc: AttestationCmd = AttestationCmd::new_request(
+            request.request_blob.clone().into(),
+            Some(user_data.to_vec()),
```
Xynnn007, line 84: I noticed that the `user_data` was to be validated [on the AS side](https://github.com/confidential-containers/trustee/pull/345/files#diff-915457db8ca21ef25fda79ebeadd19a81a62971717017b3d7ffc06b9bc00ebceR197) . Also the [document of attester](https://docs.rs/s390_pv/0.10.0/s390_pv/uv/struct.AttestationCmd.html#method.new_request) and [verifier side](https://docs.rs/s390_pv/0.10.0/s390_pv/attest/struct.AttestationItems.html#method.new) show this nonce is generated before measurement. I wonder if this value should be generated on the AS side first, and then do the check -- like a challenge value to prevent replay attack, also just like `report_data`/`runtime_data` for other platforms.

huoqifeng: user_data is optional data which will be injected on attester side and checked on AS side saved beforehand -- yes, which is optional today. We are going to add more info such as initdata hash in user_data in future. But it's not used today. We check OS Image and Firmware info only today. @Xynnn007

attestation-agent/attester/src/se/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+    request::BootHdrTags,
+    uv::{AttestationCmd, ConfigUid, UvDevice},
+};
+use serde::{Deserialize, Serialize};
+use serde_json;
+use serde_with::{base64::Base64, serde_as};
+
+pub fn detect_platform() -> bool {
+    // run always on s390x machine
+    let v = std::fs::read("/sys/firmware/uv/prot_virt_guest").unwrap_or_else(|_| vec![0]);
```
fitzthum, line 19: Why not just return false in the else case here?

And for the next step, can you compare the string directly rather than converting it to an int.

Also if we need the variable v, can we rename it to something more descriptive?

attestation-agent/attester/src/se/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+
+#[derive(Debug, Default)]
+pub struct SeAttester {}
+
+#[async_trait::async_trait]
+impl Attester for SeAttester {
+    async fn get_evidence(&self, req: Vec<u8>) -> Result<String> {
+        // req is serialized SeAttestationRequest String bytes
+        // TODO, calculate optional userdata based on the boot partition etc.
+        let image_btph = "optional check";
```
fitzthum, line 73: Can you add a comment about what `image_btph` is? There are a lot of terms here that I am not familiar with. Is there something we could add to the docs to describe the flow in general?

attestation-agent/kbs_protocol/src/client/rcar_client.rs
```diff
@@ -179,11 +182,23 @@ impl KbsClient<Box<dyn EvidenceProvider>> {
...
     }
 
-    async fn generate_evidence(&self, runtime_data: String) -> Result<String> {
+    async fn generate_evidence(
+        &self,
+        tee: Tee,
+        runtime_data: String,
+        nonce: String,
+    ) -> Result<String> {
+        debug!("Challenge nonce: {nonce}");
```
fitzthum, line 191: not sure we need this

fitzthum (commented):
Looks pretty good. A couple minor comments.

attestation-agent/attester/src/se/mod.rs
steffen-eiden: Using an int completely makes sense.
He took the code from [pv_guest_bit_set](https://docs.rs/s390_pv/0.10.0/s390_pv/misc/fn.pv_guest_bit_set.html) in the `s390_pv` crate.
-> just use the exported function from that crate (you use anyways for the attestation flow).
 
 This also describes better what you are doing:
 that sysfs entry (if it exists) just says if its a SE-guest (not backed by attestation) and can be used as an early exit for e.g. an attestation flow.

attestation-agent/kbs_protocol/src/client/rcar_client.rs
huoqifeng: If possible, we can have this remain for trouble shooting purpose?

attestation-agent/attester/src/se/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+        let userdata = UserData {
+            image_btph: image_btph.into(),
+        };
+
+        debug!("userdata json: {:#?}", &userdata.clone());
+        // req is serialized SeAttestationRequest String bytes
+        let request: SeAttestationRequest = serde_json::from_slice(req)?;
+        let user_data = serde_json::to_vec(&userdata)?;
+        let mut uvc: AttestationCmd = AttestationCmd::new_request(
+            request.request_blob.clone().into(),
```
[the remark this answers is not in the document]
huoqifeng: @steffen-eiden what do you mean? could you please give an example?

attestation-agent/attester/src/se/mod.rs
huoqifeng: Hopefully, @fitzthum you agree with what @steffen-eiden explained, thank you Steffen!

attestation-agent/attester/src/se/mod.rs
huoqifeng: Comments refined to explain it.

attestation-agent/attester/src/se/mod.rs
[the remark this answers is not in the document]
steffen-eiden: have a look on [Rust by Example](https://doc.rust-lang.org/rust-by-example/flow_control/match/destructuring/destructure_structures.html):
```
    let faa = Foo { x: (1, 2), y: 3 };

    // You do not need a match block to destructure structs:
    let Foo { x : x0, y: y0 } = faa;
    println!("Outside: x0 = {x0:?}, y0 = {y0}");

```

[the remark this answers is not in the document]
huoqifeng: I never used it, easy and powerful, cool!

attestation-agent/attester/src/se/mod.rs
fitzthum: You can use the function that @steffen-eiden mentions. Otherwise, fixup the style of this function by using an `if let`. 

> that sysfs entry (if it exists) just says if its a SE-guest (not backed by attestation) and can be used as an early exit for e.g. an attestation flow.

This detection function should only be triggered if attestation is supported. Is that what is happening here?

huoqifeng:
> @huoqifeng I noticed the CI will not alert about code that doesn't compile in the se-attester module. I suppose due to the target-toggle. Do you plan to add CI for s390? If not, maybe we should revisit that toggle, otherwise the code will break unnoticed w/ future refactorings.

I know @BbolroC is working on this, maybe, we bring back the change in lib.rs to enable CI for se?

attestation-agent/attester/src/se/mod.rs
huoqifeng: I just noticed @steffen-eiden made it public. Yeah, I'll just use it directly.

attestation-agent/attester/src/se/mod.rs
huoqifeng: Let me remove the unused user_data for now.

fitzthum (approved):
LGTM

Some kind of test coverage would be good. Maybe that is a follow-on?

BbolroC:
For reviewers, you can find CI enablement for s390x at #576 . Thanks! 😉

PR #494, opened 2024-03-01
cdh: make the config path configurable by env

### LINKED ISSUE #489: CDH: way to define alternative location of the config file?
At the moment (CCv0) kata-agent launches CDH with no arguments. 

```rust
if let Err(e) = launch_process(
    logger,
    CDH_PATH,
    &vec![],
    CDH_SOCKET,
    DEFAULT_LAUNCH_PROCESS_TIMEOUT,
) ...
```

This means it will attempt to read a default config file at `/etc/confidential-data-hub.toml`. The config file is supposed to be dynamic, i.e. it will be populated at guest launch with values retrieved from init-data.

In a measured rootfs for peerpods `/etc` is on a read-only, dmverity-checked volume. all the dynamic files go into a (memory backed) `/run` folder. (I think this would also be a good pattern independent of how peerpod does it).

I see different ways of addressing this:

- default config path to `/run/confidential-containers/confidential-data-hub.toml` (unintuitive, possibly)
- make the default path for the config file configurable via env `CDH_CONFIG_PATH` (fine IMO, feels a bit weird that we have to do such gymnastics, there is precedent with `KATA_AGENT_CONFIG_PATH`, though)
- do not launch CDH programmatically from agent, but use an init system instead (a comfortable option, a systemd directive can be easily adjusted. I think there were plans for this in the merge-to-main, however since those dynamic values are supposed to be set via agent RPC endpoint I wouldn't know how this works. can someone confirm?)
- instead of (or prior to) defaulting to `offline-fs-kbc` if the config-file is not found, we could fall back to the AA_KBC_PARAMS env (not sure, we probably want to retire that variable).

mkulke:
fixes #489 

In the CoCo scenario the CDH configuration is dynamic, since it is specific to the container workload, so using /etc, which might reside on a read-only volume is not a good default in those cases. We need a way to override this (to reference a config in /run). --config is not an option, since the process is spawned statically w/o cli arguments by kata-agent.

Hence the configuration path lookup has been adjusted in this way:
- --config is optional
- if --config is unset, consider env.CDH_CONFIG_PATH
- defaults to DEFAULT_CONFIG_PATH if both are unset

### DIFF
--- a/confidential-data-hub/hub/src/bin/confidential-data-hub/main.rs
+++ b/confidential-data-hub/hub/src/bin/confidential-data-hub/main.rs
@@ -3,7 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-use std::{path::Path, sync::Arc};
+use std::{env, path::Path, sync::Arc};
 
 use anyhow::{anyhow, Context, Result};
 use api_ttrpc::{
@@ -38,8 +38,8 @@ struct Cli {
     /// Path to the config  file
     ///
     /// `--config /etc/confidential-data-hub.toml`
-    #[arg(default_value_t = DEFAULT_CONFIG_PATH.to_string(), short)]
-    config: String,
+    #[arg(short)]
+    config: Option<String>,
 }
 
 macro_rules! ttrpc_service {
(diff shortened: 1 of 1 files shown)


Xynnn007 (approved):
Thanks. lgtm

fitzthum (approved):
LGTM

Issue #422, opened 2023-12-22
image-rs re-introduce ring version that is incompatible with s390x?


stevenhorsman:
I'm not completely sure what the situation is, so apologies if this isn't clear.

During testing of https://github.com/kata-containers/kata-containers/pull/8670, we ran the kata-agent build with a new commit from image-rs and they failed with:
```
13:13:34 error: failed to run custom build command for `ring v0.16.20`
13:13:34 
13:13:34 Caused by:
13:13:34   process didn't exit successfully: `/kata-containers/src/agent/target/release/build/ring-884cb6961366a545/build-script-build` (exit status: 101)
13:13:34   --- stderr
13:13:34   thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /root/.cargo/registry/src/index.crates.io-d11c229612889eed/ring-0.16.20/build.rs:358:10
```
I know ring didn't use to support s390x. That got added in October in https://github.com/briansmith/ring/pull/1297, but 0.16.20 is ~3 years old, so it won't be in there. I know that Ding did some work to ensure that ring wasn't needed to unblock s390x builds of image-rs, has that been undone recently?

It seems to be commit https://github.com/confidential-containers/guest-components/commit/4ddac3882c1aed1ab562a4fa10d587ff464fcc1e that introduced this issue as a151bca worked, so we just used that one in the end, but once we re-integrate image-rs with kata-containers we'll hit this issue.

For some of the other guest-components projects we have [Makefiles](https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/Makefile) to make it easier to build for multiple architectures and added workflows: https://github.com/confidential-containers/guest-components/blob/5d4bb95578b611d15d6907b3419f5cb129572e51/.github/workflows/cdh_basic.yml#L59-L61 that test them, so maybe we need to consider this for image-rs so we can find out about issues before them get integrated with kata-containers?

mythi:
> Ding did some work to ensure that ring wasn't needed to unblock s390x builds of image-rs, has that been undone recently?

Looks like the latest sigstore pulls "tough" which depends on this old ring... :/

mythi:
sigstore-rs has a dependabot PR open that brings in updated ring. We should just get them to upgrade and cut a new release?

Xynnn007:
Let me fix the dep in upstream sigstore-rs.

Xynnn007:
pending pr https://github.com/sigstore/sigstore-rs/pull/320

[...]

PR #501, opened 2024-03-13
image-rs: replace rand from openssl to ring

Xynnn007:
openssl does not support cross compilation while ring supports, this makes sense when we do cross compilation tests.


Xynnn007:
btw, this makes me more think about use `ring` instead of `openssl` on the whole stack due to https://github.com/confidential-containers/guest-components/issues/422#issuecomment-1990996409.

arronwy (approved):
Thanks @Xynnn007 LGTM!

PR #503, opened 2024-03-13
attester: tdx: Improve extend runtime measurement handling

jodh-intel:
Some systems do not yet provide the ability to extend runtime measurement so add a check to allow a more meaningful error to be returned.

### DIFF
--- a/attestation-agent/attester/src/tdx/mod.rs
+++ b/attestation-agent/attester/src/tdx/mod.rs
@@ -47,6 +47,16 @@ fn get_quote_ioctl(report_data: &Vec<u8>) -> Result<Vec<u8>> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


Xynnn007:
cc @jialez0

mythi (approved):
LGTM. I had chatted about this with @jodh-intel offline. 

`detect_platform()` cannot cover all the different variations between "in-tree" and "out-of-tree" TDX support so this is the simplest approach to let the user know what is expected to work.

Xynnn007:
Let's merge https://github.com/confidential-containers/guest-components/pull/467 before this PR, because potential conflict might not be resolved as @Lu-Biao would not be here for some time.

Xynnn007 (approved):
lgtm

PR #506, opened 2024-03-15
Fix makefile

Xynnn007:
This makes the LIBC env could be overwritten for `all`, `tdx` and `amd` platforms.

Also update the ci workflow.

Related to https://github.com/kata-containers/kata-containers/pull/8870

### DIFF
--- a/.github/workflows/aa_basic.yml
+++ b/.github/workflows/aa_basic.yml
@@ -65,13 +65,13 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y libdevmapper-dev
 
-      - name: Build and install with default features
+      - name: Gnu build and install with ttrpc
         run: |
-          make && make install
+          make ttrpc=true && make install
 
-      - name: Musl build with none platform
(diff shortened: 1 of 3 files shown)


.github/workflows/aa_basic.yml
```diff
@@ -65,13 +65,13 @@ jobs:
...
+      - name: Gnu build and install with ttrpc
         run: |
-          make && make install
+          make ttrpc=true && make install
 
-      - name: Musl build with none platform
+      - name: Musl build with all platform
         run: |
-          make LIBC=musl ATTESTER=none
+          make LIBC=musl ttrpc=true ATTESTER=none
```
mkulke, line 74: to speed up builds, do we maybe want to use a matrix? LIBC would be set as `env: { "LIBC": ${{ strategy }}`

fitzthum: Good suggestion but let's take this on a followup so we can keep moving forward with https://github.com/kata-containers/kata-containers/pull/8870

PR #508, opened 2024-03-15
Makefile: Fix error

stevenhorsman:
echo command outside of recipe not valid, and results in:
```
Makefile:55: *** missing separator.  Stop.
```
, so replace with $(info)


fitzthum (approved):
LGTM

PR #510, opened 2024-03-18
image-rs: use ring in unit test

Xynnn007:
Get rid of openssl in `dev-dependencies`

cc @ChengyuZhu6


ChengyuZhu6 (approved):
Thanks @Xynnn007 ! LGTM

arronwy (approved):

PR #511, opened 2024-03-19
aa_kbc_params: parse from kernel cmdlind first

Xynnn007:
In CoCo, the aa_kbc_params is still passed from the kernel cmdline and will be read in Confidential Data Hub. If we do not provide a default configuration file for CDH, a default one will set the aa_kbc_params env to offline-fs-kbc. This would make the kernel cmdline not work.

As a workaround, this commit bring the kernel cmdline parsing first. The disadvantage is that we will not support to use any other aa_kbc_params if we already set the kernel cmdline. This disadvantage will be covered once initdata is implemented.

following up to https://github.com/kata-containers/kata-containers/pull/8870#issuecomment-2005588652

cc @fitzthum @mkulke @wainersm


mkulke (commented):
as an alternative approach, could we not set the ENV at CDH's launch if `get_value()` returns something?

mkulke (approved):
lgtm

fitzthum (approved):
LGTM, but there are some test failures. One minor check thing and one `await` that needs to be removed.

confidential-data-hub/hub/src/bin/confidential-data-hub/config.rs
```diff
@@ -120,11 +120,15 @@ impl CdhConfig {
 
 impl CdhConfig {
     pub fn set_configuration_envs(&self) {
-        // KBS configurations
-        env::set_var(
-            "AA_KBC_PARAMS",
-            format!("{}::{}", self.kbc.name, self.kbc.url),
-        );
+        if let Err(_) = attestation_agent::config::aa_kbc_params::get_value() {
```
wainersm, line 123: Hi @Xynnn007 !

`attestation_agent::config::aa_kbc_params::get_value()` is a very simple function. I wonder if it wouldn't make sense to simply duplicate it to CDH to avoid creating this dependency CDH -> AA as the build of AA can get overly complicated.

wainersm: Never mind. I just noticed cdh -> AA dependency is already established.

mkulke: indeed, but they communicate via API and a clear interface, so a direct dep like this is unfortunate. I think we want to establish what configuration options we want to provide and then build configuration accordingly. At the moment this is more or less spec' ed in the implementation (e.g. "if aa_kbc_params kernel param is unset and there is no CDH config file, set aa_kbc_params as environment parameter to configure the CDH KMS - but without KBS cert").

So, there are a few assumptions, and it would be a good idea to simplify, even if that means duplication, e.g. the introduction of a `cdh_kbs_params` kernel param if we want to support passport on platforms without dynamic initdata.

PR #519, opened 2024-03-26
AA: add envlogger to enable log

Xynnn007:
Before this commit, AA will not enable log by default, so that the process will not print any output even if `RUST_LOG` env is set.


arronwy (approved):
Thanks @Xynnn007 LGTM!

PR #524, opened 2024-03-29
image-rs: refactor secure channel initialization logic

Xynnn007:
Currently image signature verification keys, auth files are got from confidential datahub who has a fixed ttrpc socket. So we do not require image-rs to have to configure the aa_kbc_params.

In Kata-CC this is important because in this way we can avoid parsing kernel cmdline in kata-agent.

Only when in enclave-cc, aa_kbc_params is still used by image-rs.


fitzthum (approved):
LGTM

PR #526, opened 2024-04-02
AA | Bug Fix: correct API endpoint when get AS token

jialez0:


Xynnn007 (approved):
btw we'd better add some tests upon this to make the code robust everytime we change it.

PR #531, opened 2024-04-07
AA | Enable the launch configuration

Xynnn007:
This PR leverages the configuration file.

The configuration file defines the context of the plugins, and currently only attestation tokens are supported. It is in JSON format.

If no `--config` is provided, a default one will be used. It means that the aa_kbc_params will be tried to find. If not aa_kbc_params is found in kernel cmdline, peer-pod.json, kata-agent config or ENV the launch will fail.

Also an example is given.

btw @jialez0 as the config is still not be used until the PR get merged. Do you think it is good to use JSON or we use toml? When use JSON, a cert would be hard to set as we need escape chars. Unlike TOML, example is https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/example.config.toml#L19-L51


attestation-agent/attestation-agent/src/config/kbs.rs
```diff
@@ -0,0 +1,28 @@
...
+    pub url: String,
+
+    /// Cert of KBS
+    pub cert: Option<String>,
+}
+
+impl Default for KbsConfig {
+    fn default() -> Self {
+        let aa_kbc_params =
+            aa_kbc_params::get_params().expect("No aa_kbc_params specified in kernel cmdline");
```
mkulke, line 22: get_params will use multiple methods
```suggestion
            aa_kbc_params::get_params().expect("failed to set aa_kbc_params");
```

attestation-agent/attestation-agent/src/lib.rs
```diff
@@ -68,44 +69,55 @@ pub trait AttestationAPIs {
...
-    pub fn new(config_path: &str) -> Self {
-        let config = Config::try_from(config_path).ok();
+    pub fn new(config_path: Option<&str>) -> Result<Self> {
+        let _config = match config_path {
+            Some(config_path) => {
+                info!("Use AA config file: {config_path}");
+                Config::try_from(config_path)?
+            }
+            None => {
+                warn!("No AA config file specified. Use a default one.");
```
mkulke, line 96:
```suggestion
                warn!("No AA config file specified. Using a default configuration.");
```

attestation-agent/attestation-agent/src/lib.rs
```diff
@@ -68,44 +69,55 @@ pub trait AttestationAPIs {
...
 }
 
 impl AttestationAgent {
     /// Create a new instance of [AttestationAgent].
-    pub fn new(config_path: &str) -> Self {
-        let config = Config::try_from(config_path).ok();
+    pub fn new(config_path: Option<&str>) -> Result<Self> {
+        let _config = match config_path {
+            Some(config_path) => {
+                info!("Use AA config file: {config_path}");
```
mkulke, line 92:
```suggestion
                info!("Using config file: {config_path}");
```

attestation-agent/attestation-agent/src/lib.rs
```diff
@@ -68,44 +69,55 @@ pub trait AttestationAPIs {
...
+        }
     }
 }
 
 impl AttestationAgent {
     /// Create a new instance of [AttestationAgent].
-    pub fn new(config_path: &str) -> Self {
-        let config = Config::try_from(config_path).ok();
+    pub fn new(config_path: Option<&str>) -> Result<Self> {
+        let _config = match config_path {
```
mkulke, line 90: we don't need an underscore prefix, do we?

mkulke (approved):
couple of nits, otherwise lgtm

attestation-agent/attestation-agent/src/config/kbs.rs
Xynnn007: Thanks. I assume you mean `get`

attestation-agent/attestation-agent/src/lib.rs
Xynnn007: We do need this. As if no features in `kbs` and `coco_as` is enabled, the config is not used thus the lint checker will raise a no-use warning.

jialez0:
> btw @jialez0 as the config is still not be used until the PR get merged. Do you think it is good to use JSON or we use toml? When use JSON, a cert would be hard to set as we need escape chars. Unlike TOML, example is https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/example.config.toml#L19-L51

For me, I think TOML is better.

Xynnn007:
> > btw @jialez0 as the config is still not be used until the PR get merged. Do you think it is good to use JSON or we use toml? When use JSON, a cert would be hard to set as we need escape chars. Unlike TOML, example is https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/example.config.toml#L19-L51
> 
> For me, I think TOML is better. But if we use TOML, will it cause problems when we pass AA configuration as Initdata?

It's all right. AA config could be treated as a raw string with `'''` quoted, same as CDH.

attestation-agent/attestation-agent/src/bin/grpc-aa/main.rs
```diff
@@ -25,6 +26,13 @@ struct Cli {
     /// `--attestation_sock 127.0.0.1:11223`
     #[arg(default_value_t = DEFAULT_ATTESTATION_AGENT_ADDR.to_string(), short, long = "attestation_sock")]
     attestation_sock: String,
+
+    /// Configration file for Attestation Agent
```
fitzthum, line 30: nit: `Configration` -> `Configuration`

attestation-agent/attestation-agent/src/bin/ttrpc-aa/main.rs
```diff
@@ -33,6 +34,13 @@ struct Cli {
     /// `--attestation_sock unix:///tmp/attestation`
     #[arg(default_value_t = DEFAULT_ATTESTATION_SOCKET_ADDR.to_string(), short, long = "attestation_sock")]
     attestation_sock: String,
+
+    /// Configration file for Attestation Agent
```
fitzthum, line 38: Same

fitzthum (approved):
LGTM. One nit

attestation-agent/attestation-agent/src/bin/grpc-aa/main.rs
Xynnn007: Thanks. Let me fix.

Xynnn007:
Hi @mkulke @jialez0 @fitzthum , I realize that the dep `config` could help to parse both `toml` and `json` config. Thus I
1. added both JSON and TOML example config for AA & CDH
2. added a unit test to parse the configs for AA
3. change the default config path to `/etc/attestation-agent.conf` and `/etc/confidential-data-hub.conf` without a suffix specifying format

Please ensure the change 3 will not influence current cases, or I can abondon 3
cc @wainersm

jialez0 (approved):
LGTM.

Issue #487, opened 2024-02-27
-v should print the git sha for the guest bins


mkulke:
at the moment the `-v` doesn't tell us much, it would be nice to get a git revision.

fitzthum:
On a related note, should we be synchronizing our Cargo versions with the actual release version? It's more work, but it's a bit goofy for everything to be v0.1.0 permanently.

mkulke:
> On a related note, should we be synchronizing our Cargo versions with the actual release version? It's more work, but it's a bit goofy for everything to be v0.1.0 permanently.

yeah, that too. once we release crates we have to do this anyway.

PR #532, opened 2024-04-07
AA & CDH | Add feature information and git sha via `-V`

### LINKED ISSUE #530: CDH & AA | Add option to show built-in plugins

Xynnn007:
Fixes #530 

After this PR, the AA would show its built-in plugins by `-V` like the following
```
$ ./attestation-agent -V

attestation-agent 

Commit Hash: 083cb8750c325c7d90147f3bce1f498664e376fb (dirty)
Supported Attesters: TDX_ATTESTER, SGX_ATTESTER, AZ_SNP_VTPM_ATTESTER, AZ_TDX_VTPM_ATTESTER, SNP_ATTESTER
Token plugins: KBS, COCO_AS
```

and CDH
```
$ ./confidential-data-hub -V

confidential-data-hub 

Commit Hash: 083cb8750c325c7d90147f3bce1f498664e376fb (dirty)
Resource Providers: KBS, SEV
KMS plugins: ALIYUN, EHSM
```

Related to also #487 

cc @fitzthum @surajssd @mkulke

### DIFF
--- a/attestation-agent/attestation-agent/Cargo.toml
+++ b/attestation-agent/attestation-agent/Cargo.toml
@@ -55,12 +55,12 @@ kbs = ["kbs_protocol/background_check", "token"]
 # CoCoAS Attestation Token
 coco_as = ["reqwest", "token"]
 
-all-attesters = ["kbs_protocol?/all-attesters", "attester/all-attesters"]
-tdx-attester = ["kbs_protocol/tdx-attester", "attester/tdx-attester"]
-sgx-attester = ["kbs_protocol/sgx-attester", "attester/sgx-attester"]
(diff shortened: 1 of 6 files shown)


mkulke (commented):
there's this:

image

maybe we can use it for the attester features?

fitzthum:
> maybe we can use it for the attester features?

We are using that. The list is just all the different features to check for.

Xynnn007:
> LGTM.
> 
> Slightly annoying that we'll need to update the feature lists here, but this seems like a pretty standard way of doing things.

yes. We need to break the all-attesters features into small features or the information would be only one All-attester rather than separate different attesters.

Xynnn007:
Ok @mkulke Please feel free to merge this if it looks good to you, or share any suggestions.

mkulke (approved):
thanks, this will be very helpful for debugging! lgtm

PR #537, opened 2024-04-09
CDH: move sev init into KMS plugins

### LINKED ISSUE #534: SEV: Client Broken
We missed something when we refactored the KBC into the SEV Client.

[Here](https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/hub/src/auth/sev.rs#L18) we load the kernel module, but we don't use the kernel module struct as intended. 

The struct has drop implemented on it to unload the kernel module when the struct goes out of scope. This is a precaution to avoid other components accessing the secret.

Since we drop the struct during the auth step, the securityfs will be empty by the time we setup the client.

Xynnn007:
The initailization of sev should including setting a kernel module, mount a security fs and read the file content from the fs. In the old version before reading the fs, the kernel module has already been removed. This commit fixes this by move the logic into the initialization of online-sev-kbc.

Also, the `credentials` items inside CDH config would also make sense to online-sev-kbc and offline-fs-kbc except for cc-kbc.

I cannot test this PR if it works locally as no machine is avaliable, so might need help from @fitzthum 

Fixes #534


Xynnn007:
One thing to note that the online-sev-kbc's aa_kbc_param looks like `online_sev_kbc::127.0.0.1:5432` rather than cc-kbc's `cc_kbc::http://127.0.0.1:8000` with an extra protocol indicator `http`. This follows the legacy online-sev-kbc code. I am not sure if it's ok to assume all simple-kbs serves as HTTP but not HTTPS.

fitzthum (approved):
This looks good. I don't have the opportunity to test, but I will return to this when @ryansavino gets his shim PR upstream in Kata and we can make sure that it is all good and optimize a few things.

Xynnn007:
Let me merge to avoid potential rebases.

PR #539, opened 2024-04-11
CDH | Refactor secure mount module

Xynnn007:
The old secure mount API aligns with kata-agent's Storage object, thus brings more parameters of no use, like `source`, `fstype`, etc.

This new definition organizes parameters into options(k-v pairs) and flags (purely strings), also a `volume_type` field to specify the concrete secure mount type, and last a `mount_point` to specify the target mount point.

The change makes it much easier and straightforward to write requests when we directly use CDH inside a non-CoCo scenario.

This PR would break PR https://github.com/kata-containers/kata-containers/pull/7965/files. But as it is not merged now, we can make a new PR based on that when doing secure mount feature merge-to-main.

cc @ChengyuZhu6


confidential-data-hub/docs/use-cases/secure-mount-with-aliyun-oss.md
```diff
@@ -66,39 +66,39 @@ Run CDH
...
+        "otherOpts": "-o max_stat_cache_size=0 -o allow_other",
+        "path": "<bucket-path>",
+        "readonly": "",
+        "targetPath": "/mnt/aliyun-oss",
+        "url": "https://oss-cn-beijing.aliyuncs.com",
+        "volumeId": ""
+    },
+    "flags": [],
     "mount_point": "/mnt/target-path"
 }
```
Xynnn007, line 90: I hope that this new request format makes much more sense to you. : ) @bpradipt @fitzthum

bpradipt: @Xynnn007 indeed :-) 
A related question, can I use the secure mount functionality today to mount S3 or Azure blob volumes inside the container today ? Or code changes will be needed to support different object storages ?

Xynnn007: Take aliyun oss for example, we might need
1. CDH to impl the plugin. If 1 is achieved, a confidential VM support is ok. The following points are for confidential containers.
2. kata-runtime & kata-agent deliver `secure-mount` metadata to CDH. This is what https://github.com/kata-containers/kata-containers/pull/7965/files did.
3. The CSI driver of S3/Azure blob/other kinds of storages should support deliver mount info to kata as [aliyun-oss does](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/master/pkg/oss/coco.go#L94-L100)

bpradipt: Thanks @Xynnn007. This makes sense

fitzthum (commented):
Looks pretty good overall.

I guess you're removing the source parameter (at least from the top level). I think that's fine, but I wonder if it could be useful in cases where the encrypted dir is already inside the guest i.e. mounted from the host by the kata agent.

Xynnn007:
> I guess you're removing the source parameter (at least from the top level). I think that's fine, but I wonder if it could be useful in cases where the encrypted dir is already inside the guest i.e. mounted from the host by the kata agent.

Oh. I once thought from the CDH point of view, such a parameter could be included inside `options` map, as it is used to indicate how to mount. Thus setting an key named `kata.source` in the `options` when kata-agent gets a non-empty `source` `Storage` object might also work? I think the key name `kata.source` is rarely to collide

confidential-data-hub/storage/src/volume_type/alibaba_cloud_oss/error.rs
```diff
@@ -0,0 +1,29 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use thiserror::Error;
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+#[derive(Error, Debug)]
+pub enum Error {
+    #[error("Error when getting plaintext of OSS parameters")]
```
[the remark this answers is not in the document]
Xynnn007: Let me add the logic of printing the error sources/stack in the CDH high level API.

fitzthum:
> Oh. I once thought from the CDH point of view, such a parameter could be included inside options map, as it is used to indicate how to mount. Thus setting an key named source in the options when kata-agent gets a non-empty source Storage object might also work? This hints that kata-agent should do some parameter organization work.

yeah, should be fine to us an option for it, but that might mean we do a bit more juggling in the kata agent.

ChengyuZhu6 (approved):
LGTM thanks @Xynnn007

Xynnn007:
> yeah, should be fine to us an option for it, but that might mean we do a bit more juggling in the kata agent. i guess it might actually be better to have an option for it since the source probably won't be used in most cases

Well. I am still thinking about the scenarios outside CoCo, like in confidential VM. 
1. The requests to CDH should be easy and clear. Extra `source` would make it confusing for users beyond kata-containers.
2. Even in kata-contianers, as you've mentioned, some storage does not require a host-to-guest storage volume (what `source` points to) like mounts directly from the internet, so we might not need such a fixed field, even it is optional.
3. Try to think about this: what's the scenarios where `source` is set? Obviously when some storage is shared from the host to the guest, we need `source`. We can almost ensure that `directVolume` is such case in kata. We can enhence https://github.com/kata-containers/kata-containers/tree/main/src/tools/csi-kata-directvolume to support encrypted mode, where `source` will be delivered inside `options`. This would meet the barrier of no-`source` field and we need `source` semantics.

fitzthum:
> So, maybe we can leave this as-is?

That's fine. We can always change it in the future once we start implementing more storage backends.

PR #542, opened 2024-04-15
dependency updates

mythi:
My 'carry-water-chop-wood' contribution to what I think is important: keep our dependencies update-to-date. This PR updates some dependencies that are know to require manual updates (i.e., `nix`) AND changes dependabot to focus on direct dependencies for now so that the "backlog" gets emptied.


attestation-agent/coco_keyprovider/Cargo.toml
```diff
@@ -12,7 +12,7 @@ base64.workspace = true
 clap = { workspace = true, features = ["derive"] }
 ctr.workspace = true
 daemonize = "0.5.0"
-env_logger = "0.10.0"
+env_logger = { workspace = true }
```
Xynnn007, line 15:
```suggestion
env_logger.workspace = true
```

Xynnn007 (commented):
Thanks for the contribution.

attestation-agent/coco_keyprovider/Cargo.toml
mythi: fixed

mythi:
/cc @fitzthum 

it looks to me the codeowners setting does not give sufficient coverage.

PR #543, opened 2024-04-17
Optimize error for guest-components

### LINKED ISSUE #529: api-server-rest: Responded with 200 OK although providing  error in the body
api-server-rest responds with "**200 OK**" with a body indicating a timeout error.

The body included:
`rpc status: Status { code: INTERNAL, message: "[CDH] [ERROR]: Get Resource failed: get resource failed: get rersource failed: Kbs client error: get resource failed: http request failed: get failed: error sending request for url (http://127.0.0.1:8080/kbs/v0/resource/default/sshclient/publicKey): connection closed before message completed", details: [], special_fields: SpecialFields { unknown_fields: UnknownFields { fields: None }, cached_size: CachedSize { size: 0 } } }`

This is probably an example of a broader set of potential errors that lead to a "**200 OK**" with an error body.

When a. REST service reply with "**200 OK**" the REST client assumes that the body contains the requested information. 
In this case, the REST client will assume that the body contains key material. 

The REST client has no way to identify that an issue occurred and will only discover this later when trying to use the information as a key material and fail. 

The api-server-rest should respond with a 4XX or 5XX to the client if the key material is not delivered. 

The issue was observed when the connection between the CDH and KBS was abruptly disconnected during a key request.

### LINKED ISSUE #541: Promote the error print of guest-components
Currently, the error information printed by guest-components is not well-formatted. An expected one from my point of view should looks like
```plaintext
some error
Caused by: level 1 reason
Caused by: some more detailed reason
...
```

This follows the discussion of https://github.com/confidential-containers/trustee/pull/327#discussion_r1503563955

BTW, we should think about whether the concrete error information should be reported to higher level. A most simple way is to report a rough error type to the caller of the guest components, and the concrete guest component itself should log the error information in the format mentioned above.

wdyt? @mkulke and probably @kartikjoshi21

Xynnn007:
This PR does the following
1. Reorganize error handling and information in logs in CDH/AA. Only a brief error information will be responsed to caller, and detailed information will be printed in the log of CDH/AA. Close #541
2. ASR now will response an internal error (HTTP 500) when failed to execute API call. Close #529
3. Based on the code change, we bring out the duplicated code for `unseal_secret` in storage plugin

### DIFF
--- a/attestation-agent/attestation-agent/src/bin/grpc-aa/server.rs
+++ b/attestation-agent/attestation-agent/src/bin/grpc-aa/server.rs
@@ -39,17 +39,17 @@ impl AttestationAgentService for AA {
 
         let mut attestation_agent = self.inner.lock().await;
 
-        debug!("Call AA to get token ...");
+        debug!("AA (grpc): get token ...");
 
         let token = attestation_agent
             .get_token(&request.token_type)
             .await
             .map_err(|e| {
-                error!("Call AA to get token failed: {}", e);
-                Status::internal(format!("[ERROR:{}] AA get token failed: {}", AGENT_NAME, e))
+                error!("AA (grpc): get token failed:\n{e:?}");
+                Status::internal(format!("[ERROR:{AGENT_NAME}] AA get token failed"))
             })?;
 
-        debug!("Get token successfully!");
+        debug!("AA (grpc): Get token successfully!");
 
         let reply = GetTokenResponse { token };
 
@@ -64,20 +64,17 @@ impl AttestationAgentService for AA {
 
         let mut attestation_agent = self.inner.lock().await;
 
-        debug!("Call AA to get evidence ...");
+        debug!("AA (grpc): get evidence ...");
 
         let evidence = attestation_agent
             .get_evidence(&request.runtime_data)
             .await
             .map_err(|e| {
-                error!("Call AA to get evidence failed: {}", e);
-                Status::internal(format!(
-                    "[ERROR:{}] AA get evidence failed: {}",
-                    AGENT_NAME, e
-                ))
+                error!("AA (grpc): get evidence failed:\n{e:?}");
+                Status::internal(format!("[ERROR:{AGENT_NAME}] AA get evidence failed"))
             })?;
 
-        debug!("Get evidence successfully!");
+        debug!("AA (grpc): Get evidence successfully!");
 
         let reply = GetEvidenceResponse { evidence };
 
@@ -92,20 +89,19 @@ impl AttestationAgentService for AA {
 
         let mut attestation_agent = self.inner.lock().await;
 
-        debug!("Call AA to extend runtime measurement ...");
+        debug!("AA (grpc): extend runtime measurement ...");
 
         attestation_agent
             .extend_runtime_measurement(request.events, request.register_index)
             .await
             .map_err(|e| {
-                error!("Call AA to extend runtime measurement failed: {}", e);
+                error!("AA (grpc): extend runtime measurement failed:\n{e:?}");
                 Status::internal(format!(
-                    "[ERROR:{}] AA extend runtime measurement failed: {}",
-                    AGENT_NAME, e
+                    "[ERROR:{AGENT_NAME}] AA extend runtime measurement failed"
                 ))
             })?;
 
-        debug!("Extend runtime measurement successfully!");
+        debug!("AA (grpc): extend runtime measurement succeeded.");
 
         let reply = ExtendRuntimeMeasurementResponse {};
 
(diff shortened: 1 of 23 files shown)


fitzthum (commented):
Great stuff. Made a few comments

attestation-agent/attestation-agent/src/bin/grpc-aa/server.rs
```diff
@@ -45,8 +45,8 @@ impl AttestationAgentService for AA {
             .get_token(&request.token_type)
             .await
             .map_err(|e| {
-                error!("Call AA to get token failed: {}", e);
-                Status::internal(format!("[ERROR:{}] AA get token failed: {}", AGENT_NAME, e))
+                error!("Call AA to get token failed: {e:?}");
```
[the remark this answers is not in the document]
Xynnn007: good idea.

Xynnn007:
@fitzthum Fixed as suggestions. Thanks. PTAL

Xynnn007:
> Do you want to fixup [this trace](https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/hub/src/bin/config/mod.rs#L64) as well? I saw something about this in the slack.
> 
> There are a couple of places where you use the same format string in this PR that you might want to double check.

I think it is enough now. `{e:?}` (concrete error information with backtrace) only shows in CDH/AA's own stdout/err and will not be delivered to kata-agent. This would help to debug.

confidential-data-hub/secret/src/secret/mod.rs
```diff
@@ -31,15 +31,12 @@ pub const VERSION: &str = "0.1.0";
...
-                "Unsupported secret version {}. Only support {VERSION} now.",
-                self.version
-            )));
+            return Err(SecretError::VersionError);
         }
 
         match &self.r#type {
-            SecretContent::Envelope(env) => env.unseal().await,
-            SecretContent::Vault(v) => v.unseal().await,
+            SecretContent::Envelope(env) => env.unseal().await.map_err(Into::into),
```
mkulke, line 38: i'm suprised we need this mapping here, since the fn returns the same error type?

confidential-data-hub/secret/src/error.rs
```diff
@@ -5,13 +5,24 @@
...
+    #[error("unseal envelope secret failed")]
+    UnsealEnvelopeFailed(#[from] EnvelopeError),
+
+    #[error("unseal vault secret failed")]
+    UnsealVaultFailed(#[from] VaultError),
 
-    #[error("unseal vault secret failed: {0}")]
-    UnsealVaultFailed(String),
+    #[error("parse SealedSecret failed: {0}")]
+    Parse(&'static str),
```
mkulke, line 27: maybe `ParseFailed` to be consistent with the other errors

confidential-data-hub/image/src/error.rs
```diff
@@ -5,16 +5,44 @@
...
+
+    #[error("kms interface when {context}")]
+    KmsError {
+        #[source]
+        source: kms::Error,
+        context: &'static str,
+    },
+
+    #[error("base64 decoding failed when {context}")]
+    Base64Decoding {
```
mkulke, line 25: `Base64DecodingFailed`

mkulke, line 25: `Base64DecodeFailed`

confidential-data-hub/image/src/error.rs
```diff
@@ -5,16 +5,44 @@
...
+        #[source]
+        source: base64::DecodeError,
+        context: &'static str,
+    },
+
+    #[error("decrypt LEK using KEK failed")]
+    Decrypt {
```
mkulke, line 32: `DecryptFailed`

mkulke, line 32: `DecryptFailed`

confidential-data-hub/image/src/error.rs
mkulke, line 25: `Base64DecodeFailed`

confidential-data-hub/image/src/annotation_packet/v2.rs
```diff
@@ -61,36 +62,27 @@ fn default_provider() -> String {
...
 impl TryInto<super::v1::AnnotationPacket> for AnnotationPacketV2 {
-    type Error = Error;
+    type Error = anyhow::Error;
 
     fn try_into(self) -> std::result::Result<super::v1::AnnotationPacket, Self::Error> {
         if self.version != DEFAULT_VERSION {
-            return Err(Error::ConvertAnnotationPacketFailed(format!(
-                "`version` must be {DEFAULT_VERSION}."
-            )));
+            bail!("`version` must be {DEFAULT_VERSION}.");
```
mkulke, line 69: should we resort to anyhow in library code? If we want to reduce verbosity, we could move the strings to a thiserror enum:

```
return Err(Error::PackageVersionMismatch(self.version));
...
return Err(Error::UnknownProvider(self.provider));
...
```

confidential-data-hub/secret/src/secret/mod.rs
Xynnn007: actually not. the inyo here will map the inner error type into outer one

Xynnn007: actually not. the inyo here will map the inner error type into outer one

confidential-data-hub/image/src/annotation_packet/v2.rs
Xynnn007: Oh. This is all about transform `PacketV1` to `PacketV2`, let me change the error information of `ConvertAnnotationPacketFailed`.

arronwy (approved):
Thanks @Xynnn007 LGTM!

mkulke (approved):

PR #548, opened 2024-04-23
AA | Add Eventlog Recording for Attestation Agent

### LINKED ISSUE #495: AA: Measurement Event Log Format
PR #467 introduces code to measure container images into a TDX RTMR during loading. However, the absence of  an event log poses challenges to verifiers when images are loaded out of order. An event log is crucial for attesting to runtime events like additional process execution in existing containers and security policy changes.

To institute an event log, a consensus on the log format is imperative. We have the option to adopt an established standard log format, like the TPM2 log format outlined in the [TCG PC Client Profile](https://trustedcomputinggroup.org/resource/pc-client-specific-platform-firmware-profile-specification/) or the [Canonical Event Log Format (*CEL*)](https://trustedcomputinggroup.org/resource/canonical-event-log-format/) defined by the TCG Infrastructure Working Group. Alternatively, we can formulate a bespoke format. Generally, opting for a standardized format is preferred to facilitate wider acceptance. It's worth noting that the TPM2 log format predominantly caters to platform firmware-oriented events, making *CEL* a more reasonable choice for our purpose.

*CEL* introduces the concepts of *NELR* (Native Event Log Record) and *CELR* (Canonical Event Log Record), and allows applications to keep event log in *NELR*s, and translate/convert them to *CELR*s at attestation time. In essence, we are free to define the *NELR* format according to our requirements while remaining compliant with *CEL* standards.

## *NELR* (Native Event Log Record) Format

An event record consists of an event type and associated parameters. One encoding approach is TLV (Type-Length-Value), akin to the `union` construct in C, using a tag, usually an integer, to specify the record's structure. Alternatively, the event record can be encoded using JSON serialization. However, formats relying on integer event types are generally suboptimal, requiring a centralized entity for event type assignment. While currently not problematic, this may hinder user containers from generating customized measurement events in the future. Despite its preference, JSON introduces the challenge of "structural equivalence", where the same JSON object can have various valid encodings due to differences in field order and insignificant space characters. This inherent variability poses a challenge for verifiers uncertain about which encoding to hash. Furthermore, it's important to note that JSON lacks support for comments, complicating the conveyance of Event Log Informative Data (*ELID*) in *CEL* terminology.

Hence, a proposed text-based *NELR* format is presented below. Text is preferred over binary in this design due to its human readability, ease of encoding in JSON strings for transmission via RESTful APIs, and its capacity to accommodate descriptive event types/tags rather than numerical ones.

The format is succinctly described as follows:

- Each *NELR* is a non-empty line of text terminated by a newline character. The entire line (including leading/trailing spaces/tabs but excluding the trailing newline) is hashed/extended as is.
- A log file is a textual document composed of *NELR*s.
- A log file may include empty lines (solely a newline character); these lines are excluded from hashing/extension.
- Additionally, a log file may incorporate comments, identified as lines starting with `#` (no leading spaces/tabs allowed). Comments, serving purposes such as conveying *ELID* or enhancing readability, are treated in the same manner as empty lines.

**NOTE**: The choice of a plain text line over JSON is deliberate to eliminate variability in JSON encoding, ensuring verifiers have certainty regarding what to hash. Nevertheless, a JSON object can be encapsulated inside a *NELR* but all newline characters must be omitted.

The following exemplifies a log file containing *NELR*s (which can be locally maintained by the Attestation Agent), formatted in accordance with the previously described specifications.

```sh
#
# Lines starting from '#' (no leading space allowed) are comments and are
# treated in the same manner as empty lines.
#
# All empty lines are ignored (not hashed/extended).
#

#
# We have the option of either having one log file per Measurement Register
# (MR) or a global log for all MRs. The format definition currently assumes one
# log per MR. In the event of an agreement on an aggregated log for all MRs, we
# can prefix each line with an MR index.
#

#
# A NELR is an entire line of text without the trailing newline. All
# characters, including leading and trailing spaces and tabs, are significant
# and hashed as is.
#
# In establishing a convention, the initial NELR must explicitly serve as an
# INIT record, delineating the hash algorithm and MR value before the log
# initiation. This is imperative, as the virtual Firmware may have previously
# extended this MR and maintained a log in an alternative format before
# initiating the Kernel boot.
#
# The following INIT record indicates that the MR has not been extended by the
# virtual Firmware, and the hash algorithm employed is SHA384.
#
INIT sha384/000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000

#
# The following represents an ImageLoad event, and the SHA256 digest of the
# loaded image manifest is
# 0123456789abcdef1123456789abcdef2123456789abcdef3123456789abcdef.
#
# As another convention, the first column should be a "domain identifier",
# offering insight into the entity generating the event. The use of uppercase
# "domain identifiers", such as INIT, is reserved exclusively for log keepers
# and verifiers.
#
github.com/confidential-containers ImageLoad sha256/0123456789abcdef1123456789abcdef2123456789abcdef3123456789abcdef

#
# The content of a record is indeed free-form. Therefore, if needed, JSON
# objects can be inline within the structure.
#
github.com/confidential-containers EventWithJSONParams { "key1":"value1\tmore values and 'quotes'\n", "key2": [ "value2", 2, true, null ] }
```

## Converting *NELR* to *CELR* (Canonical Event Log Record)

The provided table outlines essential fields within a *CELR* per *CEL* information model, complete with default values and instructions on determining their specific values. The conversion of *NELR* to *CELR* involves filling in these fields, encoding *NELR* into a JSON string, and then assigning the resulting string to the field `content`.

|Field|Value|Description
|-|-|-
|`recnum`|Assigned sequentially by conversion tool|This field is an integer starting from `0` for the initial *CELR* and incremented for each subsequent *CELR*.
|`pcr`|`2`|On TDX, this designates the index of the RTMR to which the event is extended, default to `2`.
|`digests`|SHA-384 digest of the *NELR*|This comprises an array of hash algorithm and digest pairs corresponding to the `content` field. TDX (and most other HW TEEs), supports only one algorithm, resulting in a single entry within this array.
|`content_type`|`"CC-CEL"`|`content_type` dictates the type or structure of the `content` field. Currently, "CC-CEL" is utilized, with suggestions for alternative options welcomed. It's important to note, as per the *CEL* spec, `content_type` must also possess an integer value to enable binary encoding, such as in TLV or CBOR formats. The absence of an integer value restricts *CELR* to JSON encoding in our scenario.
|`content`|*NELR*|This represents the *NELR* line encoded as a JSON string.

Below is a Bash script for converting a log file containing *NELR* entries into a JSON array of *CELR*s.

```sh
#!/usr/bin/env bash

NELR_to_CELR_json() {
    local _RECNO=0
    local _HALG _MR _J _D
    while read -r; do
        test -z "$REPLY" -o "${REPLY:0:1}" = "#" && continue

        case $(cut -f1 -d' ' <<< "$REPLY") in
            INIT)
                if test -z "$_HALG"; then
                    _MR=$(cut -f2 -d' ' <<< "$REPLY")
                    _HALG=${_MR%%/*}
                    _MR=${_MR#$_HALG/}
                else
                    echo error: Multiple INIT found >&2
                    return 1
                fi
                ;;
            *)
                if test -z "$_HALG"; then
                    echo error: First record is not INIT >&2
                    return 2
                fi
                ;;
        esac

        _D=$(echo -n "$REPLY" | ${_HALG}sum | cut -f1 -d' ')
        _J+="$(printf \
            '{"recnum":%d,"pcr":%d,"digests":[{"hashAlg":"%s","digest":"%s"}],"content_type":"CC-CEL","content":"%s"},' \
            $_RECNO $1 $_HALG $_D "$(sed 's/[\\"]/\\&/g' <<< "$REPLY")")"
        echo -e "mr$_RECNO: $_MR" >&2
        _MR=$(echo $_MR$_D | xxd -r -p | ${_HALG}sum | cut -f1 -d' ')
        _RECNO=$((_RECNO+1))
    done

    echo "{\"mr\":\"$_MR\",\"log\":[${_J%,}]}"
}

# AA stores measurements in RTMR[2]
NELR_to_CELR_json 2
```

If the previous log file is saved as `mr.log`, and the script file is named `n2celr.sh`, then executing the following command

```sh
./n2celr.sh < mr.log | jq .
```

will yield the following result.

```json
{
  "mr": "4693cf54a9c8b386c67290abfa981696fe07d985d06e7f06c6929370db90bbb647faff2bc0375d270aa2f0dd42635568",
  "log": [
    {
      "recnum": 0,
      "pcr": 2,
      "digests": [
        {
          "hashAlg": "sha384",
          "digest": "5d97ae953f7b03a1e93f58f9d621f84d1bdf053d1ec5ad2461262a938ecd4ed4d42d3c5c42084828da1c8051a6c08e6a"
        }
      ],
      "content_type": "CC-CEL",
      "content": "INIT sha384/000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
    },
    {
      "recnum": 1,
      "pcr": 2,
      "digests": [
        {
          "hashAlg": "sha384",
          "digest": "98cfd27eaba5b57ee1b9bbb2da9886ec9ebc40ef428ffd5eda1a4024039beed7d0f4b46b2e63bb419780f619add308ef"
        }
      ],
      "content_type": "CC-CEL",
      "content": "github.com/confidential-containers ImageLoad sha256/0123456789abcdef1123456789abcdef2123456789abcdef3123456789abcdef"
    },
    {
      "recnum": 2,
      "pcr": 2,
      "digests": [
        {
          "hashAlg": "sha384",
          "digest": "dfcf2f7ac9a8e19cbe4233837a1cd34a4c6517a8d6852aa8a566e1f43b39dc1f32bcb51fa4e6c7a139c844ab5c766361"
        }
      ],
      "content_type": "CC-CEL",
      "content": "github.com/confidential-containers EventWithJSONParams { \"key1\":\"value1\\tmore values and 'quotes'\\n\", \"key2\": [ \"value2\", 2, true, null ] }"
    }
  ]
}
```

Please note in the JSON output provided above, the key `mr` contains the ultimate value of the MR, while the key `log` represents the JSON array of *CELR*s. Therefore,

```sh
./n2celr.sh < mr.log | jq -r .mr
```

will yield the final value of the MR; and

```sh
./n2celr.sh < mr.log | jq .log
```

will yield the *CEL* log as an JSON array; while

```sh
n2celr.sh < mr.log | jq -r .log[].content
```

will convert *CELR*s back to *NELR*s (with comments stripped).

Xynnn007:
Related to/Close #495

This PR 
- refactors the API `extend_runtime_measurement` to a one aligned with NELR proposed in #495
- refactors TDX's RTMR extension logic to avoid unsafe code

This PR is **still a draft**. TODOs
- [ ] ~~A specification document to define CoCo event types~~
- [ ] ~~logic to convert NELR to CELR~~
- [ ] ~~logic to deliver eventlogs to the verifiers~~

Needs some comments to ensure we are on the same way.

cc @binxing @mythi

btw @mkulke I am not sure if vTPM crates are ready for extending PCRs now?


mkulke:
> btw @mkulke I am not sure if vTPM crates are ready for extending PCRs now?

Not yet, we'll have to add this to the API, since we also need it for Peerpod's initdata

zvonkok:
@Xynnn007 @fitzthum Are we looking to integrate coconut-svsm vTPM support was recently merged (SNP). There is some work going on to add TDX-partioning as well.

fitzthum:
.> Are we looking to integrate coconut-svsm? vTPM support was recently merged (SNP). There is some work going on to add TDX-partioning as well.

I am open to this. One of my close colleagues has been working on adding vTPM support there. I think the biggest challenge might be supporting all the forks that they currently rely on to get things running. We will also probably need a new vTPM attester/verifier or possibly one for each platform coconut runs on.

attestation-agent/attestation-agent/src/eventlog.rs
```diff
@@ -0,0 +1,36 @@
...
+        let parent_dir = eventlog_path
+            .parent()
+            .ok_or_else(|| anyhow!("illegal eventlog path: cannot locate the parent dir"))?;
+
+        std::fs::create_dir_all(parent_dir).context("create eventlog parent dir")?;
+        let file = File::create(path).context("create eventlog")?;
+        Ok(Self { file })
+    }
+
+    pub fn write_log(&mut self, log: &[u8]) -> Result<()> {
```
binxing, line 26: Just wonder, why using `u8` instead of `String` (UTF-8 encoding) as the log entry type?

attestation-agent/attestation-agent/src/eventlog.rs
```diff
@@ -0,0 +1,36 @@
...
+impl EventLog {
+    pub fn new(path: &str) -> Result<Self> {
+        let eventlog_path = PathBuf::from(path);
+        let parent_dir = eventlog_path
+            .parent()
+            .ok_or_else(|| anyhow!("illegal eventlog path: cannot locate the parent dir"))?;
+
+        std::fs::create_dir_all(parent_dir).context("create eventlog parent dir")?;
+        let file = File::create(path).context("create eventlog")?;
+        Ok(Self { file })
```
binxing, line 23: Besides creating the log file, the 1st entry (per #495) should be an `INIT` entry that records the hash algorithm and initial value of the MR. But I don't see that entry created anywhere. Am I missing anything?

Xynnn007: Yes, it has not been implemented. I will add this entry if others looks right.

attestation-agent/attestation-agent/src/eventlog.rs
Xynnn007: There is no any special reasons. If you prefer, I can turn it into `String`

attestation-agent/attestation-agent/src/eventlog.rs
```diff
@@ -0,0 +1,48 @@
...
+
+        std::fs::create_dir_all(parent_dir).context("create eventlog parent dir")?;
+        let file = File::create(path).context("create eventlog")?;
+        let mut eventlog = Self { file };
+
+        let init_entry = match hash_algorithm {
+            HashAlgorithm::Sha256 => "INIT sha256/0000000000000000000000000000000000000000000000000000000000000000",
+            HashAlgorithm::Sha384 => "INIT sha384/000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
+            HashAlgorithm::Sha512 => "INIT sha512/00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
+        };
```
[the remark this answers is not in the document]
Xynnn007: Do you mean that we should verify the existing eventlogs & extended RTMR here before extending it?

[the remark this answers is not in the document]
binxing: We only need to extend this `INIT` entry. And to create the `INIT` entry, we have to read the current RTMR value.

It's **unnecessary** to verify the existing log (left off by the BIOS).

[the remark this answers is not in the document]
Xynnn007: ok. this is as expected.

attestation-agent/attestation-agent/config.example.json
```diff
@@ -7,5 +7,10 @@
...
         }
+    },
+    "eventlog_config": {
+        "eventlog_algorithm": "sha384",
+        "eventlog_path": "/run/confidential-containers/eventlog",
```
arronwy, line 13: the eventlog path can located in `/run/attestation-agent/` since this daemon may used by projects beyond CoCo

attestation-agent/attestation-agent/config.example.toml
```diff
@@ -29,3 +29,9 @@ DYEbQvoZIkUTc1gBUWDcAUS5ztbJg9LCb9WVtvUTqTP2lGuNymOvdsuXq+sAZh9b
 M9QaC1mzQ/OStg==
 -----END CERTIFICATE-----
 '''
+
+[eventlog_config]
+
+eventlog_algorithm = "sha384"
+eventlog_path = "/run/confidential-containers/eventlog"
```
arronwy, line 36: same as above

attestation-agent/attestation-agent/src/config/mod.rs
```diff
@@ -16,11 +19,66 @@ pub mod kbs;
 
 pub const DEFAULT_AA_CONFIG_PATH: &str = "/etc/attestation-agent.conf";
 
+pub const DEFAULT_EVENTLOG_HASH: &str = "sha384";
+
+pub const DEFAULT_EVENTLOG_PATH: &str = "/run/confidential-containers/eventlog";
```
arronwy, line 24: same here

attestation-agent/attestation-agent/src/config/mod.rs
```diff
@@ -16,11 +19,66 @@ pub mod kbs;
...
+    pub eventlog_config: EventlogConfig,
+}
+
+#[derive(Clone, Debug, Deserialize, Default)]
+pub struct EventlogConfig {
+    /// Hash algorithm used to extend runtime measurement for eventlog.
+    pub eventlog_algorithm: HashAlgorithm,
+
+    /// Path to store eventlog.
+    pub eventlog_path: String,
```
arronwy, line 78: we may need implement self defined `Default` for `EventlogConfig` and assign the default eventlog path, else the path may empty.

attestation-agent/attestation-agent/src/lib.rs
```diff
@@ -70,21 +75,50 @@ pub trait AttestationAPIs {
...
+            .eventlog_algorithm
+            .digest(init_entry.as_bytes());
+        let tee_type = detect_tee_type();
+        let attester = TryInto::<BoxedAttester>::try_into(tee_type)?;
+
+        attester
+            .extend_runtime_measurement(event_digest, DEFAULT_PCR_INDEX)
+            .await
+            .context("write INIT entry")?;
+
```
arronwy, line 118: we need `write_log` for `init_entry` here to allow remote verifier to get the digest algorithm and reproduce the rtmr digest

arronwy (commented):
Thanks @Xynnn007 , verified with this PR, this feature works well.

attestation-agent/attestation-agent/config.example.json
Xynnn007: Yes. When AA is launched, the path `/run/confidential-containers` will be created also. I think maybe we could have another issue & PR to promote this altogether with both AA/CDH/ASR side?

Xynnn007: Let's start with eventlog in this PR.

attestation-agent/attestation-agent/src/lib.rs
Xynnn007: Yes, thanks!

arronwy (approved):
Thanks @Xynnn007 LGTM!

Xynnn007:
With great help from @arronwy , the whole stream from AA -> AS passed. I will put the AS side PR then.

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -54,13 +55,17 @@ fn runtime_measurement_extend_available() -> bool {
...
+
 #[derive(Serialize, Deserialize)]
 struct TdxEvidence {
     // Base64 encoded CC Eventlog ACPI table
     // refer to https://uefi.org/specs/ACPI/6.5/05_ACPI_Software_Programming_Model.html#cc-event-log-acpi-table.
     cc_eventlog: Option<String>,
     // Base64 encoded TD quote.
     quote: String,
+    // Eventlog of Attestation Agent
+    aa_eventlog: Option<String>,
```
mythi, line 68: the open here is do we want to keep `cc_eventlog` and `aa_eventlog` separate or they get sent out in a unified structure? The NELR format used here can work independent of `cc_eventlog` when `INIT`ed properly but it might still be beneficial for the verifier to use one structure only.

FWIW, while working with ITA I discovered they have a rather special format for the `event_log` that needs to be passed to it:

```
[{"rtmr":{"index":1,"bank":"SHA384"},"rtmr_events":[{"type_id":"0x8000000b","type_name":"EV_EFI_HANDOFF_TABLES2","tags":["TdxTable"],"measurement":"0b8772e5b0b41b83e6044a68397e02f49fb47066b4fbe4917ea2c45c64f323fdacbb37948f821ebaf8bc9c938ba8a749"},{"type_id":"0x8000000a","type_name":"EV_EFI_PLATFORM_FIRMWARE_BLOB2","tags":["Fv(XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX)"],"measurement":"344bc51c980ba621aaa00da3ed7436f7d6e549197dfe699515dfa2c6583d95e6412af21c097d473155875ffd561d6790"},
...
```

It's similar to what `eventlog-rs` builds but the TCG log parsing is done by the attester/relying party.

attestation-agent/attester/src/tdx/rtmr.rs
```diff
@@ -0,0 +1,76 @@
...
+
+/// The actual rtmr event data handled in DCAP
+#[repr(C, packed)]
+pub struct TdxRtmrEvent {
+    /// Always 1
+    version: u32,
+
+    /// The RTMR that will be extended.
+    /// As defined in https://github.com/confidential-containers/td-shim/blob/main/doc/tdshim_spec.md#td-measurement
+    /// we will use RTMR 2 for guest application code and configuration.
```
mythi, line 14: nit: with the `DEFAULT_PCR_INDEX` it maps to RTMR3.

mythi (commented):
I did not do a thorough review yet, just commenting on the `TdxEvidence` mainly.

attestation-agent/attester/src/tdx/mod.rs
Xynnn007: Yeah as how this NELR could be embedded inside a CELR is not finally decided, I think it is fair to use a separate `AAEL` to carry such NELR information. Once kernel community reached an agreement, we could merge `AAEL` into `CCEL` in a standard way without any extra assumptions.

attestation-agent/attester/src/tdx/rtmr.rs
Xynnn007: Thanks for this, will fix

attestation-agent/attestation-agent/src/lib.rs
```diff
@@ -152,17 +197,39 @@ impl AttestationAPIs for AttestationAgent {
...
+        content: &str,
         register_index: Option<u64>,
     ) -> Result<()> {
+        let register_index = register_index.unwrap_or_else(|| {
+            info!("No PCR index provided, use default {DEFAULT_PCR_INDEX}");
+            DEFAULT_PCR_INDEX
+        });
+
+        let log_entry = format!("{domain} {operation} {content}");
+        self.eventlog.write_log(&log_entry)?;
```
mythi, line 219: the flow here writes the log and then extends. perhaps a safer approach (following the systemd way) would be to take an flock, extend, write and release.

attestation-agent/attestation-agent/src/eventlog.rs
```diff
@@ -0,0 +1,33 @@
...
+//
+
+use std::{fs::File, io::Write};
+
+use anyhow::{Context, Result};
+
+/// AA's eventlog will be put into this path
+pub const EVENTLOG_PATH: &str = "/run/attestation-agent/eventlog";
+
+pub const EVENTLOG_PARENT_DIR_PATH: &str = "/run/attestation-agent";
```
mythi, line 13: this can also be determined runtime using `.parent()` from `Path` with `EVENTLOG_PATH`.

attestation-agent/attestation-agent/src/lib.rs
```diff
@@ -70,21 +75,54 @@ pub trait AttestationAPIs {
...
 pub struct AttestationAgent {
     _config: Config,
+    eventlog: EventLog,
 }
 
 impl Default for AttestationAgent {
     fn default() -> Self {
         if let Ok(_config) = Config::try_from(config::DEFAULT_AA_CONFIG_PATH) {
-            return AttestationAgent { _config };
+            let eventlog = EventLog::new().expect("eventlog initialize");
```
mythi, line 84: is the idea to make AA to always log independent of the underlying extension mechanism used by the TEE?

attestation-agent/attestation-agent/src/lib.rs
Xynnn007: Nice idea. Let me add a Mutex wrapper for `eventlog`

attestation-agent/attestation-agent/src/eventlog.rs
Xynnn007: Right. As they two are hardcoded a const would be better

attestation-agent/attestation-agent/src/lib.rs
Xynnn007: Yes. AA will do the conversion logic from events into underlying TEE. After kernel has got a converged eventlog and rtmr API. AA could call that API to extend, and also wrap the events into the target log format.

attestation-agent/attestation-agent/src/eventlog.rs
Xynnn007: sounds good. Let me add a fix

Xynnn007:
Rebased to resolve the conflicts. Does this PR now look good?

attestation-agent/attestation-agent/src/eventlog.rs
```diff
@@ -0,0 +1,35 @@
...
+
+impl EventLog {
+    pub fn new() -> Result<Self> {
+        std::fs::create_dir_all(EVENTLOG_PARENT_DIR_PATH).context("create eventlog parent dir")?;
+        let file = File::create(EVENTLOG_PATH).context("create eventlog")?;
+        Ok(Self { file })
+    }
+
+    pub fn write_log(&mut self, log: &str) -> Result<()> {
+        writeln!(&mut self.file, "{log}").context("failed to write log")?;
```
mkulke, line 29:
```suggestion
        writeln!(self.file, "{log}").context("failed to write log")?;
```

attestation-agent/attestation-agent/src/lib.rs
```diff
@@ -154,17 +200,41 @@ impl AttestationAPIs for AttestationAgent {
...
+            info!("No PCR index provided, use default {DEFAULT_PCR_INDEX}");
+            DEFAULT_PCR_INDEX
+        });
+
+        let log_entry = format!("{domain} {operation} {content}");
+        let event_digest = self
+            ._config
+            .eventlog_config
+            .eventlog_algorithm
+            .digest(log_entry.as_bytes());
```
mkulke, line 226: I think it would be good to move this transformation to a dedicated fn + unit_test, if just to specify it
```rust
struct Event<'a> {
  domain: &'a str,
  operation: &'a str,
  content: &'a str,
}

impl Event {
  fn log_digest(&self) -> Vec<u8> {
    let log_entry = format!("{domain} {operation} {content}");
    ...
  }
}
```

attestation-agent/protos/attestation-agent.proto
```diff
@@ -18,9 +18,26 @@ message GetTokenResponse {
     bytes Token = 1;
 }
 
+// - `events`: a event slice. Any single event will be calculated into a hash digest to extend the current
```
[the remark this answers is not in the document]
Xynnn007: Oh you are right. Let me fix

attestation-agent/attestation-agent/src/lib.rs
```diff
@@ -70,23 +76,56 @@ pub trait AttestationAPIs {
...
         AttestationAgent {
             _config: Config::new().expect("AA initialize"),
+            eventlog,
         }
     }
 }
 
 impl AttestationAgent {
+    pub async fn init(&mut self) -> Result<()> {
+        // We should get the current platform's evidence to see the RTMR value.
```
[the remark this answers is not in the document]
Xynnn007: Yes. I must mix the term here. I want to say the register that is used to work as runtime measurement register. For TDX it is RTMR (coincidently the same name), for (v)TPM platforms it is a specific PCR, for ARM CCA it is REM, for RISCV it is `vm-runtime-measurements` ...

I don't know a concrete term to cover all of them, so I use RTMR here. Hope it makes sense

attestation-agent/attestation-agent/src/lib.rs
Xynnn007: I added `Event` struct, `to_string()` and `digest_with()` function.

attestation-agent/attestation-agent/src/lib.rs
```diff
@@ -69,27 +76,43 @@ pub trait AttestationAPIs {
...
+            .config
+            .eventlog_config
+            .eventlog_algorithm
+            .digest(init_entry.as_bytes());
+
+        let tee_type = detect_tee_type();
+        let attester = TryInto::<BoxedAttester>::try_into(tee_type)?;
+        let mut eventlog = self.eventlog.lock().await;
+
+        attester
```
mkulke, line 104: we should be able to use `self.attester` here

mkulke (approved):
small nit, otherwise lgtm

attestation-agent/attestation-agent/src/lib.rs
Xynnn007: Yes, fixed

mkulke:
@Xynnn007 can we merge this one? I would start working on a PR for vTPM runtime measurements based on this

PR #549, opened 2024-04-24
AA: add UpdateConfiguration API

Xynnn007:
As discussed, this API is a workaround API for CoCo to apply initdata proposal. Once we agree on a better way (like systemd scheme), we can deprecate the API.

See https://github.com/kata-containers/kata-containers/issues/9468

cc @fitzthum

### DIFF
--- a/attestation-agent/attestation-agent/src/lib.rs
+++ b/attestation-agent/attestation-agent/src/lib.rs
@@ -3,7 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-use std::str::FromStr;
+use std::{io::Write, str::FromStr};
 
 use anyhow::{Context, Result};
 use async_trait::async_trait;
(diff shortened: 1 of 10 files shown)


attestation-agent/attestation-agent/src/lib.rs
```diff
@@ -100,6 +100,27 @@ impl AttestationAgent {
...
+    pub fn update_configuration(&mut self, conf: &str) -> Result<()> {
+        let mut tmpfile = tempfile::NamedTempFile::new()?;
+        let _ = tmpfile.write(conf.as_bytes())?;
+        tmpfile.flush()?;
+
+        let _config = Config::try_from(
+            tmpfile
+                .path()
+                .as_os_str()
+                .to_str()
```
fitzthum, line 116: If this function takes a str, can we just that directly in `try_from` rather than writing it to a tmp file?

fitzthum (commented):
Looks good. One question.

attestation-agent/attestation-agent/src/lib.rs
Xynnn007: The str here means the path. This is actually misleading.

fitzthum: oh right. I didn't read this closely enough.

I'm not really a fan of the tmpfile thing but I guess we can live with it temporarily. Another approach would be to have the kata agent write the file, like it will with other configs and just take the path here.

What is the purpose of the underscore here?

Xynnn007: config is actually used only in the features `coco-as` and `kbs`. Without a underscore the lint will cry when we enable neither of them.

PR #555, opened 2024-05-13
bumps sev crate to 3.1.1

wobito:
This PR bumps the SEV crate for support in https://github.com/confidential-containers/trustee/pull/385
and handles the new return structure from SEV 3.1.1

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -407,9 +407,9 @@ dependencies = [
 
 [[package]]
 name = "az-cvm-vtpm"
-version = "0.5.2"
+version = "0.5.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "53f5fa6fe32f6409dca8202ddd267027ac1f1de4cd91b9e21d92062c14d63cae"
+checksum = "1e2d89967f683d16dafdaacb578a2841daff9f43c856c6a6dc9939cc11272712"
 dependencies = [
  "bincode",
  "jsonwebkey",
@@ -419,7 +419,7 @@ dependencies = [
  "serde",
  "serde-big-array",
  "serde_json",
- "sev 1.2.1",
+ "sev 3.1.1",
  "sha2 0.10.8",
  "thiserror",
  "tss-esapi",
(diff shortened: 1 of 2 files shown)


attestation-agent/attester/src/snp/mod.rs
```diff
@@ -44,6 +44,11 @@ impl Attester for SnpAttester {
             .get_ext_report(None, Some(data), Some(0))
             .context("Failed to get attestation report")?;
 
+        let certs = match certs {
+            Some(c) => c,
+            None => bail!("Certificate chain was not provided"),
+        };
```
fitzthum, line 50: I think we are going to add support in Trustee for the verifier to pull the cert chain if needed, so I would prefer not to have a fatal error here.

fitzthum (commented):
Looks good. One note and it looks like you might have a conflict.

I am slightly worried about picking up some random changes with this jump, but let's see.

wobito:
Thanks for the comments will pick this up tomorrow and rectify the conflict. 

We have tested this in aws with the verifier changes in my other pr mentioned above and bare metal using the kbs-client.

attestation-agent/attester/src/snp/mod.rs
mkulke, line 50: apart from the @fitzthum comment, this pattern matching can be rewritten to an if let

```suggestion
        let Some(certs) = certs else {
             bail!("Certificate chain was not provided");
        }
```

wobito: I've removed the bail and just unwrapped the value.

fitzthum: Unwrapping it is still fatal. If there is no cert chain we probably want to pass an empty Vec. You could do `cert_chain: certs.unwrap_or(vec![])` or we could change the SnpEvidence struct to have `cert_chain: Option<Vec<CertTableEntry>>`

fitzthum: Looks like the SEV crate returns `Option<Vec<CertTableEntry>>` so you might just change the struct and use that value directly.

mkulke (approved):
lgtm.

I think we probably want to start versioning the evidence structs somewhere. Maybe not explicitly, but at the moment it's somewhat intransparent which pair of attester and verifier will work.

I assume sev's v3 `AttestationReport` struct will also be different from the one in v1.

the trustee repo references guest-components in the kbs/tools/client crate, it might be a good idea to bump it there together with the verifier PR, so we have a least a working pair documented.

fitzthum (approved):
LGTM. Let's wait to merge this until the verifier PR is a bit closer since we'll need to fixup the struct there. ofc this will still have to go first.

wobito:
Thanks everyone!! happy to contribute help!
Working on the verifier pr at the moment too.

Issue #556, opened 2024-05-17
Add cdh golang client interface


ChengyuZhu6:
This client interface streamlines the process of connecting to CDH and leveraging its APIs. The Go version is very useful for integrating the CDH with containerd plugins, including NRI and image verifiers, where CDH operates as an independent module. However, the absence of a Go-based client for the CDH presents a gap that needs to be addressed.

PR #557, opened 2024-05-17
Add cdh golang client interface

ChengyuZhu6:
This client interface streamlines the process of connecting to CDH and leveraging its APIs. The Go version is very useful for integrating the CDH with containerd plugins, including NRI and image verifiers, where CDH operates as an independent module.

Note that this PR includes only the interface for `UnsealSecret`. 
I think other interfaces can be implemented in the following PRs.

Fixes: #556


ChengyuZhu6:
cc @arronwy @Xynnn007

confidential-data-hub/golang/cmd/grpc-client/main.go
```diff
@@ -0,0 +1,45 @@
...
+	"os"
+
+	common "github.com/confidential-containers/guest-components/confidential-data-hub/golang/pkg/core"
+	cdhgrpcapi "github.com/confidential-containers/guest-components/confidential-data-hub/golang/pkg/grpc"
+)
+
+func main() {
+
+	vFlag := flag.String("v", "", "sealed secret value")
+	fFlag := flag.String("f", "", "sealed secret file")
```
Xynnn007, line 19: So the client now only supports unseal secret, right? We might need to add some TODO notes inside the code for future extension.

btw, we might also want to add two parameters:
1. GRPC socket address
2. to tell whether this is a gRPC or ttRPC client

confidential-data-hub/golang/pkg/api/cdhgrpc/cdhgrpc.proto
```diff
@@ -0,0 +1,46 @@
+syntax = "proto3";
```
Xynnn007, line 1: Can we reuse the proto already defined in https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/hub/protos/api.proto? This could help with so-called "single truth of source".

confidential-data-hub/golang/pkg/api/cdhttrpc/cdhttrpc.proto
```diff
@@ -0,0 +1,46 @@
+syntax = "proto3";
```
Xynnn007, line 1: same as mentioned. We use the same proto defined in https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/hub/protos/api.proto

Xynnn007 (commented):
Thanks @ChengyuZhu6 .

I think the `pkg` in golang for CDH would make great sense to users working in k8s ecosystem. Some comments

confidential-data-hub/golang/pkg/api/cdhttrpc/cdhttrpc.proto
ChengyuZhu6: Done.

confidential-data-hub/golang/pkg/api/cdhgrpc/cdhgrpc.proto
ChengyuZhu6: Done.

confidential-data-hub/golang/cmd/grpc-client/main.go
ChengyuZhu6: Done.

Xynnn007 (approved):
Nice to me. One last question is that we'd better have CI to cover this, but I think it is ok to be in a separate PR.

fitzthum (commented):
So this will be used inside of a guest? I associate Go more with k8s on the host. Can you elaborate on the integrations that you mentioned? I am slightly wary of supporting more languages although overall this PR looks fine (with my modest understanding of Go).

ChengyuZhu6:
> So this will be used inside of a guest? I associate Go more with k8s on the host. Can you elaborate on the integrations that you mentioned? I am slightly wary of supporting more languages although overall this PR looks fine (with my modest understanding of Go).

Ok. I have successfully integrated CDH with NRI plugin in containerd. This allows us to unseal secrets in NRI plugin using the CDH Go client, rather than having to execute CDH binary to get unsealed values.

ChengyuZhu6:
> So this will be used inside of a guest? I associate Go more with k8s on the host. Can you elaborate on the integrations that you mentioned? I am slightly wary of supporting more languages although overall this PR looks fine (with my modest understanding of Go).


I unseal secret with NRI plugin during CreateContainer stage in k8s.

result: 
image

pod yaml: 
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: registry
spec:
  containers:
    - name: registry
      image: quay.io/chengyu_zhu/registry:latest
      command: [ "/bin/sh", "-c", "env && sleep 3000" ]
      env:
        - name: SECRET_USERNAME
          valueFrom:
            secretKeyRef:
              name: secret-sealed
              key: sealed-secret
```

fitzthum (approved):
Ok, I'm a bit wary of supporting another language, but I think I understand this use case.

Xynnn007:
@fitzthum @mkulke Ok. Let's could move the golang things to a separate repo if needed. 

Over all, thanks @ChengyuZhu6 for the work!

PR #558, opened 2024-05-21
Improve image unpacking feature set

skaegi:
This PR contains three tiny fixes for gzip unpacking:
1) Multiple member support
2) Preservation of ownership and permissions
3) Correct Whiteout handling


image-rs/src/unpack.rs
```diff
@@ -33,6 +33,11 @@ pub fn unpack<R: io::Read>(input: R, destination: &Path) -> Result<()> {
     let mut dirs: HashMap<CString, [timeval; 2]> = HashMap::default();
     for file in archive.entries()? {
         let mut file = file?;
+
+        if ! convert_whiteout(&file.path().unwrap(), file.header(), destination)? {
```
Xynnn007, line 37: Let's try to avoid using `unwrap()` here as it will cause the program panic. Try this `ok_or(anyhow!(...))` to raise a proper error information.

image-rs/src/unpack.rs
```diff
@@ -83,6 +88,54 @@ pub fn unpack<R: io::Read>(input: R, destination: &Path) -> Result<()> {
     Ok(())
 }
 
+fn convert_whiteout(path: &Path, header: &Header, destination: &Path) -> Result<bool> {
+    // Handle whiteout conversion
+    let name = path.file_name().unwrap();
+    let parent = path.parent().unwrap();
+    let name = name.to_str().unwrap();
```
Xynnn007, line 95: Same as before. Please try to use error handling and avoid `unwrap()`s.

like
```rust
let name = path.file_name().ok_or(anyhow!("No `file name` is in the whiteout path"));
```

image-rs/src/unpack.rs
```diff
@@ -83,6 +88,54 @@ pub fn unpack<R: io::Read>(input: R, destination: &Path) -> Result<()> {
...
+        xattr::set(parent, "trusted.overlay.opaque", b"y")?;
+        return Ok(false)
+    }
+
+    if name.starts_with(".wh.") {
+        let original_name = name.strip_prefix(".wh.").unwrap();
+        let original_path = parent.join(original_name);
+        let path = CString::new(format!(
+            "{}/{}",
+            destination.display(),
```
Xynnn007, line 106: In rust edition >=2021 (guest components/image meets this) due to https://doc.rust-lang.org/std/fmt/trait.Display.html
```rust
format!(
    "{}/{}",
    destination.display(),
    original_path.display(),
}
```
can be shortten to
```rust
format!("{destination}/{original_path}")
```

Xynnn007 (commented):
Welcome @skaegi and thanks for the fix! Some nits and we need to make the CI happy

image-rs/src/unpack.rs
skaegi: Thanks -- total work in progress here :)

image-rs/src/unpack.rs
skaegi: nice!

skaegi: hmm... I ended up reverting this as I was getting...
```
error[E0277]: `Path` doesn't implement `std::fmt::Display`
   --> image-rs/src/unpack.rs:106:42
    |
106 |         let path = CString::new(format!("{destination}/{original_path}"))?;
    |                                          ^^^^^^^^^^^^^ `Path` cannot be formatted with the default formatter; call `.display()` on it
    |
    = help: the trait `std::fmt::Display` is not implemented for `Path`, which is required by `&Path: std::fmt::Display`
    = note: in format strings you may be able to use `{:?}` (or {:#?} for pretty-print) instead
    = note: call `.display()` or `.to_string_lossy()` to safely print paths, as they may contain non-Unicode data
    = note: this error originates in the macro `$crate::__export::format_args` which comes from the expansion of the macro `format` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0277]: `PathBuf` doesn't implement `std::fmt::Display`
   --> image-rs/src/unpack.rs:106:56
    |
106 |         let path = CString::new(format!("{destination}/{original_path}"))?;
    |                                                        ^^^^^^^^^^^^^^^ `PathBuf` cannot be formatted with the default formatter; call `.display()` on it
    |
    = help: the trait `std::fmt::Display` is not implemented for `PathBuf`
    = note: in format strings you may be able to use `{:?}` (or {:#?} for pretty-print) instead
    = note: call `.display()` or `.to_string_lossy()` to safely print paths, as they may contain non-Unicode data
    = note: this error originates in the macro `$crate::__export::format_args` which comes from the expansion of the macro `format` (in Nightly builds, run with -Z macro-backtrace for more info)
```

PR #561, opened 2024-05-24
Support gzip.multiple_members

skaegi:
Images like `cgr.dev/chainguard/busybox` built using gzip directly might legitimately be using gzip multiple members. This really has to be supported to have broad compatibility with docker images as this style is prevalent in the distroless community.


fitzthum (approved):
LGTM

PR #562, opened 2024-05-24
Gzip should preserve ownership/permissions

skaegi:
This is necessary to support images using non-root users and a mixture of runtime users.


fitzthum (approved):
LGTM

PR #563, opened 2024-05-24
Handle gzip whiteouts correctly

skaegi:
When looking at the resulting file contents of multilayer image we frequently are seeing `.wh.` files. We need to correctly handle whiteouts. See
1) https://github.com/opencontainers/image-spec/blob/main/layer.md#whiteouts
2) https://github.com/containerd/containerd/blob/v1.7.17/archive/tar_opts_linux.go#L34

### DIFF
--- a/image-rs/src/unpack.rs
+++ b/image-rs/src/unpack.rs
@@ -2,16 +2,23 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
+use anyhow::anyhow;
 use anyhow::Context;
 use anyhow::{bail, Result};
 use libc::timeval;
 use log::warn;
 use std::collections::HashMap;
+use std::convert::TryInto;
 use std::ffi::CString;
 use std::fs;
 use std::io;
 use std::path::Path;
 use tar::Archive;
+use tar::Header;
+
+// See https://github.com/opencontainers/image-spec/blob/main/layer.md#whiteouts
+const WHITEOUT_PREFIX: &str = ".wh.";
+const WHITEOUT_OPAQUE_DIR: &str = ".wh..wh..opq";
 
 /// Unpack the contents of tarball to the destination path
 pub fn unpack<R: io::Read>(input: R, destination: &Path) -> Result<()> {
@@ -33,6 +40,11 @@ pub fn unpack<R: io::Read>(input: R, destination: &Path) -> Result<()> {
     let mut dirs: HashMap<CString, [timeval; 2]> = HashMap::default();
     for file in archive.entries()? {
         let mut file = file?;
+
+        if !convert_whiteout(&file.path()?, file.header(), destination)? {
+            continue;
+        }
+
         file.unpack_in(destination)?;
 
         // tar-rs crate only preserve timestamps of files,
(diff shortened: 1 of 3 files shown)


fitzthum:
Might need to do `cargo fmt` on this

image-rs/src/unpack.rs
```diff
@@ -81,6 +93,63 @@ pub fn unpack<R: io::Read>(input: R, destination: &Path) -> Result<()> {
...
+            .strip_prefix(WHITEOUT_PREFIX)
+            .ok_or(anyhow!("Failed to strip whiteout prefix for: {}", name))?;
+        let original_path = parent.join(original_name);
+        let path = CString::new(format!(
+            "{}/{}",
+            destination.display(),
+            original_path.display()
+        ))?;
+
+        let ret = unsafe { libc::mknod(path.as_ptr(), libc::S_IFCHR, 0) };
```
mkulke, line 125: can we use the nix crate for mknod?

image-rs/src/unpack.rs
```diff
@@ -81,6 +93,63 @@ pub fn unpack<R: io::Read>(input: R, destination: &Path) -> Result<()> {
...
+        let uid: libc::uid_t = header
+            .uid()?
+            .try_into()
+            .map_err(|_| io::Error::new(io::ErrorKind::Other, "UID is too large!"))?;
+        let gid: libc::gid_t = header
+            .gid()?
+            .try_into()
+            .map_err(|_| io::Error::new(io::ErrorKind::Other, "GID is too large!"))?;
+
+        let ret = unsafe { libc::lchown(path.as_ptr(), uid, gid) };
```
mkulke, line 139: can we use the nix crate?

fidencio (approved):
lgtm, thanks @skaegi!

mkulke (commented):
Thank you for plugging this particular hole. It would be useful to have a test for this to ensure we don't break this during refactoring, we introduced subtle layer ordering bugs before.

image-rs/src/unpack.rs
skaegi: I think we can and happy to do it. There is existing use of `libc::lutimes` and `libc::utimes` in `unpack.rs` too. @mkulke Do you want me to do this work in this PR or should I open a separate issue/PR dealing with moving to nix?

skaegi:
> Thank you for plugging this particular hole. It would be useful to have a test for this to ensure we don't break this during refactoring, we introduced subtle layer ordering bugs before.

Sure. The existing pull image exercises this code so minimally we can add verification that the whiteouts are handled correctly.

image-rs/src/unpack.rs
mkulke: ah, wasn't aware, in this case then let's stick with the unsafe blocks here and open a dedicated issue. I think there's is good reason to use the nix crate's abstraction, but maybe there were considerations I'm not aware of

skaegi:
I added whiteout tests to unpack as that seemed best and rebased on the latest. The test failure looks like an unrelated networking flake. Ready to merge…

mkulke (approved):
lgtm

PR #571, opened 2024-06-04
tdx-attester updates

mythi:


attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -49,16 +43,6 @@ fn get_quote_ioctl(report_data: &Vec<u8>) -> Result<Vec<u8>> {
     }
 }
 
-// Return true if the TD environment can extend runtime measurement,
-// else false.
-fn runtime_measurement_extend_available() -> bool {
```
fitzthum, line 54: I would keep this function and move the check from below into it even though it's only one line. We'll probably update this check later.

fitzthum (approved):
One small comment

attestation-agent/attester/src/tdx/mod.rs
mythi: ok. itt saved me one extra `cfg()` check with https://github.com/confidential-containers/guest-components/pull/572/commits/1c8291f20d801a1242c0cf7c549fa6a04a29b333 but I'm fine adding it back.

Xynnn007 (approved):
lgtm

attestation-agent/attester/Cargo.toml
```diff
@@ -65,4 +65,4 @@ snp-attester = ["sev"]
 csv-attester = ["csv-rs", "codicon", "hyper", "hyper-tls", "tokio"]
 cca-attester = ["nix"]
 
-bin = ["tokio/rt", "tokio/macros", "all-attesters"]
+bin = ["tokio/rt", "tokio/macros", "codicon"]
```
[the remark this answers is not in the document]
mythi: I had a local patch for something like that some time ago. I'll try to find it and add it here.

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -9,7 +9,6 @@ use crate::utils::pad;
 use crate::InitdataResult;
 use anyhow::*;
 use base64::Engine;
-use log::debug;
```
Xynnn007, line 12: This might be a leftover of some commit

mythi: actually, it's on purpose to handle macros scope consistently because with  `use log:: {debug, warn}` and https://github.com/confidential-containers/guest-components/pull/572/commits/1c8291f20d801a1242c0cf7c549fa6a04a29b333 combined it triggers unused imports.

PR #572, opened 2024-06-04
tdx-attester: make DCAP libs optional

mythi:
Still WIP/RFC.

In #434 I had proposed to make use of `tdx-attest-rs` behind a new feature but dropped the idea since the initdata work moved to use it too and that is a valid case event with upstream kernels.

However, @fidencio mentioned a use case that building `kbs-client` without having to install `libtdx-attest*` (because of added complexity and the fact that the libs are not yet avail on all OS'es, e.g, 24.04) but still get the basic `get_evidence()` working can be useful. So I restored my original code and added here for feedback on how it's best handled.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -322,6 +322,7 @@ dependencies = [
  "az-snp-vtpm",
  "az-tdx-vtpm",
  "base64 0.21.7",
+ "cfg-if",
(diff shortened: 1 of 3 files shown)


mythi:
closing for now

PR #573, opened 2024-06-06
image-rs: bump dependencies

Xynnn007:
This PR updates some dependencies version.
- `jwt-simple` from 0.11 to 0.12. This helps the transive dependency `zeroize` to have a proper version when I rebase https://github.com/kata-containers/kata-containers/pull/9382
- `sigstore` to v0.9.0. Regular update
- `reqwest` to 0.12. Regular update

~~There will be a [lint error](https://github.com/confidential-containers/guest-components/actions/runs/9395257087/job/25876291283?pr=573) for rust toolchain v1.72.0 if we use `sigstore v0.9.0` upstream. At the meanwhile, the latest toolchain accepts this so [this test](https://github.com/confidential-containers/guest-components/actions/runs/9395257087/job/25876291505?pr=573) passes.~~

~~I once thought we could allow this in `guest-components` but soon found that there is `clippy` check using `v1.72.0` rust toolchain to check clippy in kata side.~~

~~I think it is hard and not reasonable for upstream `sigstore` for this to fix a historical version toolchain lint error. Thus I created a forked version of `sigstore v0.9.0` with [only one extra patch](https://github.com/Xynnn007/sigstore-rs/commit/232b33720067337618187fe3c0b58c56312475e6) to fix this. This is why the `sigstore`'s rev points to a personal repo. Hopefully this workaround could be acceptable.~~

~~For a long time, I wonder how we update rust toolchain to higher level in kata-side. The frequency?~~


image-rs/Cargo.toml
```diff
@@ -39,7 +39,7 @@ serde = { workspace = true, features = ["serde_derive", "rc"] }
 serde_json.workspace = true
 serde_yaml = { version = "0.9", optional = true }
 sha2.workspace = true
-sigstore = { git = "https://github.com/sigstore/sigstore-rs.git", rev = "d5ba303", default-features = false, optional = true }
+sigstore = { git = "https://github.com/Xynnn007/sigstore-rs.git", rev = "232b337", default-features = false, optional = true }
```
stevenhorsman, line 42: I understand that a fork is being used to a linting issue with the 1.72 rust toolchain (https://github.com/Xynnn007/sigstore-rs/commit/232b33720067337618187fe3c0b58c56312475e6). I think that a better approach is for the kata-agent build to use a newer version, and we are discussing this on the Kata AC meeting later today, so we should be able to switch to the 0.9.0 release if that goes well

.github/workflows/image_rs_build.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       fail-fast: false
       matrix:
         rust:
-          - 1.72.0
+          - 1.76.0
```
mythi, line 31: IIRC we agreed (https://github.com/confidential-containers/confidential-containers/issues/193) to use 'latest stable' with the explicit version pinned. Trustee uses 1.76.0 which was the 'latest stable' back when this topic was discussed but is also outdated.

mythi: > I guess you would lobby for us to bump image-rs to 1.78, trustee to 1.76 and unpinned for the other guest components?

Given Kata uses a single version and all of these are consumed by Kata (or its CI), perhaps it's good to have a unified version specified. Initially my thinking was that following the latest stable is enough but that may not be a good choice after all. Thinking about more, would it make sense to have a clear 'MSRV policy' defined for CoCo which set guidelines for ourselves and downstream users?

stevenhorsman: Sorry to mislead - Kata containers has individual versions for trustee, guest-components (AA, CDH, ASR) and the kata-agent (image-rs & ocicrypt-rs) as we had to split them out to get things to build, however divergent versions does lead to some confusion and complications with CVEs. We'll see what the discussion today brings and I'll report back here with a summary.

An alternative approach is that we could consider publishing binaries (via oras packages) for the guest-components and trustee in confidential-containers and then just consume them directly in kata-containers rather than building them there, which might help reduce some of the dependencies, but will probably introduce other issues!

stevenhorsman: Yes - I also was referring to the rust version too, so my clarification didn't help either! 😿

stevenhorsman:
I think the reqwest bump has broken the signature verification test (I backed down to `0.11.27` and it passed):
```
---- signature_verification stdout ----
cdh_path: /home/runner/work/guest-components/guest-components/image-rs/scripts/ttrpc/confidential-data-hub
script_dir: /home/runner/work/guest-components/guest-components/image-rs/scripts
thread 'signature_verification' panicked at image-rs/tests/signature_verification.rs:214:13:
assertion `left == right` failed: Test: Allow pulling a unencrypted signed image from a protected registry., Signing scheme: Simple Signing, Err(Security validate failed: Validate image failed: Failed to query extensions supported by registry v2 endpoint

Caused by:
    0: error sending request for url (https://quay.io/v2/)
    1: client error (Connect)
    2: invalid URL, scheme is not http)
  left: false
 right: true
```
The code is failing in https://github.com/confidential-containers/guest-components/blob/0b28991babc7d5c064b904bef1d318882fe6e976/image-rs/src/signature/mechanism/simple/xrss.rs#L102-L107

Interesting I think this is the negative test as quay.io doesn't support xrss, so we expect the headers to be missing, but it obviously shouldn't fail like this, so I'll try and debug it more.

.github/workflows/image_rs_build.yml
stevenhorsman: So we didn't get a great response from the Kata AC - Zvonko needs to bump the version to at least 1.74, but there is some resistance from RH folks to stay latest stable at the moment as they have an old version of rust in their base image. We've asked them to go and investigate and see what version they can bump to and then get back to us.

stevenhorsman:
> I think the reqwest bump has broken the signature verification test (I backed down to `0.11.27` and it passed)

After too much local debugging and trial and error I found the fix for this. Adding the `"default-tls"` feature to reqwests.

Cargo.toml
```diff
@@ -38,15 +38,15 @@ ctr = "0.9.2"
...
+jwt-simple = { version = "0.12", default-features = false, features = ["pure-rust"] }
 kbs-types = "0.5.3"
 lazy_static = "1.4.0"
 log = "0.4.14"
 openssl = "0.10"
 prost = "0.11"
 protobuf = "3.4.0"
 rand = "0.8.5"
-reqwest = { version = "0.11.27", default-features = false }
+reqwest = { version = "0.12", default-features = false }
```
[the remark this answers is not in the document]
Xynnn007: Thanks for the effort and help! Let me fix this

Xynnn007:
Let me also update `ehsm-client` and `oci-distribution` dep in this PR.

stevenhorsman (approved):
LGTM. Thanks for the updates

fitzthum (approved):
LGTM, but we probably want some clarity on kata msrv before merging.

Xynnn007:
Rebased the code. 

> LGTM, but we probably want some clarity on kata msrv before merging.

What would be the trigger of this?

stevenhorsman:
> > LGTM, but we probably want some clarity on kata msrv before merging.
> 
> What would be the trigger of this?

The conversation in the Kata AC meeting I guess.

I ran `cargo msrv` on this code and it looks like the required version is 1.72.1, so we are super close with 1.72.0 in kata-containers, so I don't anticipate any issues bumping to that at the minimum as Zvonko is planning to move to move us to 1.74 as part of the CDI work, so my personal view is that we don't need to wait to merge this.

mkulke (approved):

Issue #569, opened 2024-06-03
tdx-attester: strip CCEL before adding it to the evidence


mythi:
`tdx-attester` encodes and sends the full 64K CCEL blob as part of the evidence but only a small part of it contains relevant eventlog data. The evidence creation/processing can be optimized by stripping any unset values

Xynnn007:
It might be a trade-off. CCEL maps to a region of memory and most part of them would be `0x00`. If we want to decrease the size of transferred data/network loads, we could use some compression before transfer, but this would bring some extra work for calculation/cpu loads.

mythi:
True, I wish the kernel offered better interface for this, similar to TPM `bios_binary_measurements`. The added value for sanitizing/stripping it properly would be to get it compatible with existing tooling:

```
# ./tcglog-dump -v /sys/firmware/acpi/tables/data/CCEL 
cannot read log: log entry has an out-of-range PCR index (4294967295)
# xxd -c2  /sys/firmware/acpi/tables/data/CCEL | grep -v ffff | xxd -r > /tmp/ccel.bin 
# ./tcglog-dump -v /tmp/ccel.bin 
PCR  DIGEST                                                                                            TYPE                              DETAILS
1                                                                                                      EV_NO_ACTION                      
1    0b8772e5b0b41b83e6044a68397e02f49fb47066b4fbe4917ea2c45c64f323fdacbb37948f821ebaf8bc9c938ba8a749  EV_EFI_HANDOFF_TABLES2            
1    f87302177b059d54a2cf0c5f13340dbabf5c9dd60dc3f996c68b776fbe4de959769443a3d8ef6538b97d7e151c8298e8  EV_EFI_PLATFORM_FIRMWARE_BLOB2    
1    cfa4e2c606f572627bf06d5669cc2ab1128358d27b45bc63ee9ea56ec109cfafb7194006f847a6a74b5eaed6b73332ec  EV_EFI_VARIABLE_DRIVER_CONFIG     SecureBoot: 0
...
```

Compression would not work because some of the network hops would have to send it uncompressed anyways (e.g., KBS->AS).

PR #575, opened 2024-06-10
attester: tdx: strip CCEL

mythi:
Fixes: #569 

The CCEL log is made available through an ACPI sysfs entry and is of size "log_area_minimum_length". OVMF sets it to 64k.

The current tdx-attester code reads the whole blob and it's used as is in encoding and when sent over the wire.

Test runs suggests that it could be beneficial to strip the log before processing it further:

Squeezed from 65536 to 5064 bytes

The stripping follows the same pattern as what eventlog-rs does on the receiving end (we keep the same "stop flag" in the blob to keep things compatible).

### DIFF
--- a/attestation-agent/attester/src/tdx/mod.rs
+++ b/attestation-agent/attester/src/tdx/mod.rs
@@ -106,6 +106,22 @@ impl TdxAttester {
     }
 }
 
+fn find_stop_pos(buffer: &[u8]) -> usize {
+    let mut stop_pos: usize = 0;
+
+    const CHUNK_SIZE: usize = std::mem::size_of::<u64>();
+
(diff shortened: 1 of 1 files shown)


mythi:
I had forgotten this. It's tested and ready now:

```
[2024-10-31T07:34:34Z INFO  verifier::tdx] CCEL integrity check succeeded. 
```

Xynnn007 (approved):
Thanks for the patch and this would make great sense to efficiency. This is the producer side code. I wonder if any unit tests with ccels that produced by this patch could be added to the [verifier side](https://github.com/confidential-containers/trustee/blob/main/deps/verifier/src/tdx/eventlog.rs#L32)?

mythi:
It's possible to update `CCEL_data` blob pretty easily if needed.

PR #583, opened 2024-06-13
ci: Fix occlum CI

### LINKED ISSUE #581: ci: Fix occlum attester CI
Now occlum ci does not work correctly. Because the runner does not run normally, s.t. https://github.com/confidential-containers/guest-components/actions/workflows/aa_occlum_sgx.yml. Some solutions
1. delete this test as [enclave-cc HW e2e test](https://github.com/confidential-containers/enclave-cc/blob/main/.github/workflows/e2e.yaml#L23) will cover this
2. re-register the runner of ecc to org. Re-use the runner and fix the ci pipeline

Xynnn007:
Related to #581

~~As SGX ci runner is not prepared now, also the occlum yaml marks `runs-on` as `self-hosted`, which will cause all self-hosted runner to run this CI, including IBM SE. This is not expected. So This pr temporily close the CI of occlum, until we set up a good runner (after close #581) we can go back to open it.~~

Fixes occlum ci


Xynnn007:
Yea I think it is ready for review now. cc @mythi

Xynnn007:
Seems that there are two errors upon s390x runner
1. pulling `mcr.microsoft.com/hello-world` will fail ([image-rs test](https://github.com/confidential-containers/guest-components/actions/runs/9510506441/job/26215112571?pr=583))
2. No `docker` command is found on the runner, which is used in [rcar unit test](https://github.com/confidential-containers/guest-components/actions/runs/9510506448/job/26215112540?pr=583#step:10:391).

cc @BbolroC

mythi (approved):
Looks good and nice to see the use of `evidence_getter` here.

arronwy (approved):
Thanks @Xynnn007 LGTM!

PR #585, opened 2024-06-14
GHA: Place pre-action for s390x self-hosted runner

BbolroC:
It is observed that some packages (e.g. docker) are required to run workflows for s390x. Like the post-action, we need to place pre-action for the platform.

### DIFF
--- a/.github/workflows/aa_basic.yml
+++ b/.github/workflows/aa_basic.yml
@@ -47,6 +47,12 @@ jobs:
             cargo_lint_opts: "--no-default-features --features openssl,se-attester,kbs,coco_as -p attestation-agent -p attester -p coco_keyprovider -p kbc -p kbs_protocol -p crypto -p resource_uri"
     runs-on: ${{ matrix.instance }}
     steps:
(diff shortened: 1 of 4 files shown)


Xynnn007 (approved):

BbolroC:
For reviewers, the script will be track via `set -x`:

```bash
function pre_action_cc_guest_components() {
    set -x
    echo "remove files under $GITHUB_WORKSPACE ..."
    sudo chown -R $USER:$USER $GITHUB_WORKSPACE
    sudo rm -rf $GITHUB_WORKSPACE/*

    echo "Check if docker is installed"
    if ! which docker > /dev/null; then
        echo "install docker ..."
        sudo DEBIAN_FRONTEND=noninteractive apt-get install -y docker.io
        sleep 5
        sudo systemctl enable --now docker.service
    fi
    echo "Adjust an ownership for .docker directory"
    if [ -d ${HOME}/.docker ]; then
        sudo chown -R $USER:$USER ${HOME}/.docker
    fi
    set +x
}
```

PR #586, opened 2024-06-14
GHA: Fix indentation error and skip image.rs tests for s390x

BbolroC:
This commit fixes the wrongly indented configuration in #585 and place the pre action step after the code check-out.

There is an issue on pulling an image `mcr.microsoft.com/hello-world` on s390x.
It looks a load balancer for the microsoft registry is unstable so that the runner was able to pull the image with 10& success ratio (see https://github.com/microsoft/containerregistry/issues/144)
It is not reasonable to let the test run under the unstable environment.

This commit skips the tests at image.rs for the platform.


Xynnn007 (approved):

fitzthum (approved):
LGTM

Issue #564, opened 2024-05-25
Report Cause of Handshake Failure More Clearly


fitzthum:
Currently we offer a very [general message](https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/kbs_protocol/src/client/rcar_client.rs#L59) when we are unable to connect to the KBS. This isn't ideal. We should propagate at least the error from the last failure upwards.

Xynnn007:
We have a log [here](https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/kbs_protocol/src/client/rcar_client.rs#L61) for every try. So do you mean we record the error information every time fails, and return the latest one if the retries run out?

PR #588, opened 2024-06-14
aa: surface error from RCAR handshake

fitzthum:
I've noticed a few people running into an error like `AA-KBC get token failed: RCAR handshake failed: Get token failed because of RCAR handshake retried 5 times.` which is not very useful. This error can be caused by the KBS URI being incorrect or by the IKBS rejecting the evidence or a number of other situations.

Fixes https://github.com/confidential-containers/guest-components/issues/564

### DIFF
--- a/attestation-agent/kbs_protocol/src/client/rcar_client.rs
+++ b/attestation-agent/kbs_protocol/src/client/rcar_client.rs
@@ -41,54 +41,16 @@ struct AttestationResponseData {
 
 impl KbsClient<Box<dyn EvidenceProvider>> {
     /// Get a [`TeeKeyPair`] and a [`Token`] that certifies the [`TeeKeyPair`].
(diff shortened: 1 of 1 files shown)


Xynnn007 (approved):
lgtm

mythi:
> Also, refactor the kbs client slightly.

The refactoring made the actual change rather invisible. Would have been good to split the changes in two commits at least

related to this PR: I think the error handling here would need to understand a bit more about when the error is not recoverable and that a re-attempt won't help..

fitzthum:
> related to this PR: I think the error handling here would need to understand a bit more about when the error is not recoverable and that a re-attempt won't help..

Let me think about this for a follow-up. In theory we should only be retrying when it is connection-related, but it might be good to test that somehow.

PR #590, opened 2024-06-17
ibmse: add se attester to all-attesters

huoqifeng:


huoqifeng:
@fitzthum @Xynnn007 may you help have a look?

arronwy (approved):

Xynnn007:
Not related to this PR. But can you help to change line 67 to
```
se-attester = ["kbs_protocol?/se-attester", "attester/se-attester"]
```

s.t. add an extra `?` after `kbs_protocol`, this would enable `kbs_protocol/se-attester` when `kbs_protocol` is already enabled. Or, when we enable `se-attester`, the optional dependency `kbs_protocol` will be enabled.

attestation-agent/attestation-agent/Cargo.toml
```diff
@@ -58,7 +58,7 @@ kbs = ["kbs_protocol/background_check", "token"]
 # CoCoAS Attestation Token
 coco_as = ["reqwest", "token"]
 
-all-attesters = ["tdx-attester", "sgx-attester", "az-snp-vtpm-attester", "az-tdx-vtpm-attester", "snp-attester"]
+all-attesters = ["tdx-attester", "sgx-attester", "az-snp-vtpm-attester", "az-tdx-vtpm-attester", "snp-attester", "se-attester"]
```
Xynnn007, line 61: I wonder why the CI passes to build the binary as `attester/se-attester` seems to be built only on s390x? https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/attester/Cargo.toml#L57-L58

https://github.com/confidential-containers/guest-components/actions/runs/9546379968/job/26309121829?pr=590#step:9:241

Xynnn007 (approved):

attestation-agent/attestation-agent/Cargo.toml
huoqifeng: se-attester can also be built and run on x86 machine but can only been run on a s390x machine correctly. @Xynnn007

huoqifeng: It can actually be run on x86 correctly because https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/attester/src/se/mod.rs#L19 returns false on x86 instance.

fitzthum (approved):
LGTM

PR #591, opened 2024-06-18
CDH | Add Aliyun STS Token support for KMS

Xynnn007:
cc @1570005763


1570005763 (commented):
Nice work. Just curious, given that STS token is often a short-lived access credential, in what scenarios will 'sts_token_client' be used?

Xynnn007:
Yes. This is used in some scenarios where aliyun KMS will only be used at the launch time. In this scene a short-term token is enough.

1570005763 (changes requested):
LGTM. Once all checks have been passed, it can be merged.

Xynnn007:
@1570005763 Yea. Also a commit to make aliyun kms suites statically built-able.

1570005763 (approved):
LGTM

jialez0 (approved):

PR #596, opened 2024-06-24
CDH: fix aa_kbc_params and config reading

### LINKED ISSUE #593: CDH: kbs kms plugin should get aa_kbc_params from cdh config file if it's provided
Flow like below causes the aa_kbc_params still comes from kata-agent configure file (`/etc/agent-config.toml`) even a cdh configure file provided via like `-c cdh.toml`

> https://github.com/confidential-containers/guest-components/blob/2f1380fe67b5dfe188052f85f5c11ff9aecb7351/confidential-data-hub/kms/src/plugins/kbs/mod.rs#L36 
> https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/attestation-agent/src/config/aa_kbc_params.rs#L73 
> https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/attestation-agent/src/config/aa_kbc_params.rs#L65
> https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/attestation-agent/src/config/aa_kbc_params.rs#L89

```
fn from_config_file() -> Result<String, ParamError> {
    debug!("get aa_kbc_params from file");

    // check env for KATA_AGENT_CONFIG_PATH, fall back to default path
    let path: &String = KATA_AGENT_CONFIG_PATH.get_or_init(|| {
        env::var("KATA_AGENT_CONFIG_PATH").unwrap_or_else(|_| "/etc/agent-config.toml".into())
    });

    debug!("reading agent config from {}", path);
    let agent_config_str = std::fs::read_to_string(path)?;

    let agent_config: AgentConfig = toml::from_str(&agent_config_str)?;

    Ok(agent_config.aa_kbc_params)
}
```

We need figure out an algorithm that cdh.toml will always take top priority as parameters when it's provided.

Xynnn007:
CDH should firstly read config specified with commandline, and then env.

aa_kbc_param should be firstly read directly from CDH's config file. If there is no config file, try to read from env and then kernel cmdline. If still no values found, use a default one with offline_fs_kbc.

Close #593

cc @huoqifeng @mkulke

### DIFF
--- a/confidential-data-hub/hub/src/bin/config/mod.rs
+++ b/confidential-data-hub/hub/src/bin/config/mod.rs
@@ -3,13 +3,12 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-use std::env;
+use std::{env, fs, path::Path};
 
 use anyhow::*;
 use config::{Config, File};
-use log::{debug, warn};
+use log::{debug, info};
 use serde::Deserialize;
-use tokio::fs;
 
 cfg_if::cfg_if! {
     if #[cfg(feature = "ttrpc")] {
@@ -19,7 +18,7 @@ cfg_if::cfg_if! {
     }
 }
 
-#[derive(Deserialize, Debug)]
+#[derive(Deserialize, Debug, PartialEq)]
 pub struct KbsConfig {
     pub name: String,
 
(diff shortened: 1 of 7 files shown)


confidential-data-hub/hub/src/bin/config/mod.rs
```diff
@@ -65,13 +75,31 @@ impl Default for CdhConfig {
 }
 
 impl CdhConfig {
-    pub async fn init(config_path: &str) -> Result<Self> {
-        let mut config = Self::from_file(config_path).unwrap_or_else(|e| {
-            warn!("read config file {config_path} failed {e:?}, use a default config where `aa_kbc_params` = offline_fs_kbc::null.");
-            Self::default()
+    pub fn new(config_path: Option<String>) -> Result<Self> {
+        let config_path = config_path.or_else(|| {
+            if let std::result::Result::Ok(env_path) = env::var("CDH_CONFIG_PATH") {
```
mkulke, line 80: nit: can we move this one layer up to the main() fn invoking `CdhConfig::new()`? It might unclutter the logic here. We already log which config file we are using below ("Use configuration file...").

```rust
let cfg_path_env = env::var("CDH_CONFIG_PATH").ok();
let cfg_path = cli.config.or(cfg_path_env);
let config = CdhConfig::new(cfg_path)?;
```

confidential-data-hub/hub/src/bin/config/mod.rs
```diff
@@ -65,13 +75,31 @@ impl Default for CdhConfig {
...
+            Some(path) => {
+                info!("Use configuration file {path}");
+                if !Path::new(&path).exists() {
+                    bail!("Config file {path} not found.")
+                }
+
+                Self::from_file(&path)?
+            }
+            None => {
+                info!("No config path specified, use a default one.");
```
mkulke, line 97: it's a bit ambiguous whether we use a default config path or default config.

```suggestion
                info!("No config path specified, using a default configuration.");
```

confidential-data-hub/hub/src/bin/config/mod.rs
```diff
@@ -65,13 +75,31 @@ impl Default for CdhConfig {
...
+                Self::default()
+            }
+        };
+
+        config.update_from_kernel_cmdline()?;
```
mkulke, line 102: should we rename that to `extend_credentials_from_kernel_cmdline()` and/or add a comment what this fn does? I was initially confused why the kernel cmdline can override the config again, but it seems to only extend the credentials vector.

Xynnn007: Nice idea. Let me rename the fn

mkulke (approved):

huoqifeng (approved):
LGTM
thank you @Xynnn007 to fix it!

PR #599, opened 2024-06-24
AA: fix launch configuration

Xynnn007:
Before this commit, if no configuration and proper aa_kbc_param given to AA, the launch of AA will panic.

This commit will use a default value for aa_kbc_param for AA/CDH. If no env or kernel cmdline, default value offline_fs_kbc and null will be used. If a aa_kbc_param is specified, it will be parsed as kbc:url. If the aa_kbc_param is malwared, the launch process will still fail.

cc @mkulke


mkulke (approved):
small nit

PR #600, opened 2024-06-24
CDH/AA: make offline_fs_kbc::null default in get_params()

mkulke:
At the moment we don't support a case in which attestation-agent can start without either an AA config provided or aa_kbc_params specified. In CDH we default to `aa_kbc_params=offline_fs_kbc::null` for those cases. This change will make it the default for both AA and CDH.


mkulke:
closed in favor of #599

PR #602, opened 2024-06-30
image-rs: make tar reader async

Xynnn007:
before this commit, tar reader is sync, thus we need a separate os thread for every image layer pulling green thread. This would cost more memory and CPU resource because on linux a thread actually means another process.

This commit replace the tar crate to async version, thus we do not need another thread to handle this to decrease memory occupation.

Also, the original thread way would make the error of `unpack()` function missing because it would not be captured by original thread. Thus always a inclear `Send Error {...}` error message will be returned. The new way will not face such situation. cc @stevenhorsman 

Some abilities like permission maintain, ownership maintain are manually implemented as the original crate does not support well.

cc @arronwy

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1097,9 +1097,9 @@ checksum = "c2459377285ad874054d797f3ccebf984978aa39129f6eafde5cdc8315b612f8"
 
 [[package]]
 name = "const_fn"
-version = "0.4.9"
+version = "0.4.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fbdcdcb6d86f71c5e97409ad45898af11cbc995b4ee8112d59095a28d376c935"
+checksum = "373e9fafaa20882876db20562275ff58d50e0caa2590077fe7ce7bef90211d0d"
 
 [[package]]
 name = "const_format"
@@ -2013,7 +2013,7 @@ checksum = "1ee447700ac8aa0b2f2bd7bc4462ad686ba06baa6727ac149a2d6277f0d240fd"
 dependencies = [
  "cfg-if",
  "libc",
- "redox_syscall",
+ "redox_syscall 0.4.1",
  "windows-sys 0.52.0",
 ]
 
@@ -2767,6 +2767,7 @@ dependencies = [
  "futures-util",
  "hex",
  "kbc",
+ "krata-tokio-tar",
  "lazy_static",
  "libc",
  "log",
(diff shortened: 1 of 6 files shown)


Xynnn007:
Rebased and added another commit to replace `libc` with `nix`. This commit should be test on kata pipeline before merge.

stevenhorsman:
> Rebased and added another commit to replace `libc` with `nix`. This commit should be test on kata pipeline before merge.

Sure - as I don't have any green ticks for the current image-rs bump in kata, I'll temporarily re-purpose it to test your branch

stevenhorsman:
> > Rebased and added another commit to replace `libc` with `nix`. This commit should be test on kata pipeline before merge.
> 
> Sure - as I don't have any green ticks for the current image-rs bump in kata, I'll temporarily re-purpose it to test your branch

In https://github.com/kata-containers/kata-containers/actions/runs/9806138253/job/27078629312?pr=9828 the non-TEE, SEV and SNP tests passed and the TDX is failing with normal results, so I think this is okay from the kata integration tests. I will drop the change for now, but if this gets merged before the image-rs bump does then I'll update to pick this up too. Thanks!

fitzthum (approved):
LGTM although I am super familiar with the image-rs code.

fitzthum:
@arronwy do these changes look good to you?

arronwy (approved):
Thanks @Xynnn007 LGTM!

arronwy:
> @arronwy do these changes look good to you?

Yes, the primary concern for this pr is the possibility that crate `krata-tokio-tar` may not be actively maintained in future.

The init async support for tar is `async-tar` project https://github.com/dignifiedquire/async-tar  which is forked from `tar-rs`, and this project is not update for three years.

Next is `tokio-tar` https://github.com/vorot93/tokio-tar which is forked from `async-tar` and not update for one year and may PR is not review: https://github.com/vorot93/tokio-tar/pulls

Finally is the crate we use `krata-tokio-tar` https://github.com/edera-dev/tokio-tar which is forked from `tokio-tar` with several bug fixings.

After sync with @Xynnn007 if future have issues in this crate, we can fork and fix it.

@fitzthum any suggestions?

fitzthum:
@arronwy @Xynnn007 

the state of those repos is far from ideal, but I guess we can make it work. Maybe we can get in touch with some of those maintainers and figure out what the status is. We can fork if we have to.

PR #605, opened 2024-07-05
cdh:golang support to dynamic generate go code with proto file

ChengyuZhu6:
Since there is already "api.proto" defining the CDH api interface in `confidential-data-hub/hub/protos`, we can reuse the api.proto to minimize redundant code and reduce the workload of updating the API interface.


Xynnn007 (approved):
Cool. This would help to keep one origin

fitzthum (approved):
LGTM

PR #607, opened 2024-07-08
Add aa option for runtime register

mkulke:
## Why? 

If we use extend_runtime_measurement to measure policy or init-data we need to specify in which register it is measuring into, since the default might not work in each situation.

There is an option to specify it at the caller site, which would be kata-agent. it would be an attestation-related parameter that we would have to pipe through via kata to the attester, which is impractical.

## How

Currently the attester is being initialized in each RPC call that is interacting with the TEE, this change moves the TEE detection into the initialization of the `AttestationAgent` instance.

A `default_register_index` field is now added to the AA configuration, if this per-guest default is not set it will fall back to a hardcoded default.

### DIFF
--- a/attestation-agent/attestation-agent/config.example.toml
+++ b/attestation-agent/attestation-agent/config.example.toml
@@ -29,3 +29,6 @@ DYEbQvoZIkUTc1gBUWDcAUS5ztbJg9LCb9WVtvUTqTP2lGuNymOvdsuXq+sAZh9b
 M9QaC1mzQ/OStg==
 -----END CERTIFICATE-----
 '''
+
+[attester]
+default_register_index = 42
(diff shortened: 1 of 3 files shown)


Xynnn007:
Oh. I think there would be some overlap thing for pcr index [here ](https://github.com/confidential-containers/guest-components/pull/548/files#diff-bfc44b3434ebb686b602e48861e5d3b01d6ecb69cb2591d48d6f3ad2dd4cf89dR172). Besides this, I am ok with make attester a struct member of AA

mkulke:
@Xynnn007 there are some overlaps with 

> Oh. I think there would be some overlap thing for pcr index [here ](https://github.com/confidential-containers/guest-components/pull/548/files#diff-bfc44b3434ebb686b602e48861e5d3b01d6ecb69cb2591d48d6f3ad2dd4cf89dR172). Besides this, I am ok with make attester a struct member of AA

yes, I just noticed the redundancy. let's get the other PR merged first, or maybe you can cherry-pick ad78ee2c80ba9444265c40cddc785a9422faa79e?

Xynnn007:
@mkulke Yea, I cherry picked the commit in that PR

mkulke:
superseded by #548

PR #608, opened 2024-07-09
CDH | add image pull API

Xynnn007:
This commit adds a new API `pull_image()` for Confidential DataHub. This new api will converge all operations related to image pull to one API, hinting that image signature validation, auth file retrievement, security policy enforcement will all be done inside CDH.

Related configurations is included in CDH's config.image item.

Although the basic workflow passes due to the unit test, let's keep this PR as draft because some details still needs to promote.
- [ ] Configuration file for image pull of CDH
- [ ] More test cases
- [ ] Determine details of interacting with kata-agent

Related to https://github.com/kata-containers/kata-containers/issues/9266


fitzthum:
Do you think we should do this before doing encryped/signed images in Kata? Also, are we going to run into problems here if we don't have init-data supported? I saw you mentioned some problems with provisioning the config file on your encrypted image PR.

Xynnn007:
@fitzthum Before initdata is implemented, the config for CDH is hard to inject so we might need to wait for that.

So logically we might still merge to main with what we are doing, and then initdata, and then this PR and move things to CDH in kata.

confidential-data-hub/hub/Cargo.toml
```diff
@@ -69,6 +70,9 @@ sev = ["image/sev", "kms/sev", "secret/sev"]
 # support eHSM stacks (KMS, ...)
 ehsm = ["image/ehsm", "secret/ehsm"]
 
+# suppot for image-pull API
+image-pull = ["image-rs/kata-cc-rustls-tls"]
```
Xynnn007, line 74: With this feature, CDH is ~ 17.53MB (x86-64 & gnu). Without the feature, is ~ 16.16MB (x86-64 & gnu)

Thus I think we might not need this extra flag. We can always use `image-rs` as a dep as CDH.

Xynnn007:
Ok, for image decryption, it does not work because some dead lock happens inside ocicrypt-rs. let me try to refactor some code inside ocicrypt-rs to make it work

confidential-data-hub/hub/Cargo.toml
Xynnn007: Deleted the feature

confidential-data-hub/hub/src/config.rs
```diff
@@ -41,19 +41,82 @@ impl KbsConfig {
...
+    pub auth_uri: String,
+
+    /// Proxy that will be used to pull image
+    pub https_proxy: String,
+
+    /// No proxy env that will be used to pull image
+    pub no_proxy: String,
+
+    /// The path to store the pulled data
+    pub work_dir: String,
```
Xynnn007, line 89: Hi @ChengyuZhu6 , do you think we need to include this field in CDH's config? I have set a default value if it is not set, but there is still risk. If an user set an unexpected dir, would this influence the image pull in CoCo?

ocicrypt-rs/build.rs
```diff
@@ -37,7 +37,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
             .include("src/utils")
             .rust_protobuf()
             .customize(ttrpc_codegen::Customize {
-                async_all: true,
+                async_all: false,
```
Xynnn007, line 40: Hi @arronwy , I changed the ttrpc client to sync here. Please take a look if it is good to you

Xynnn007:
Hi @BbolroC , the CI for SE cannot pass https://github.com/confidential-containers/guest-components/actions/runs/10398806362/job/28796702259?pr=608#step:6:479, but the [x86-64](https://github.com/confidential-containers/guest-components/actions/runs/10398806362/job/28796702005?pr=608) succeeded. Could you please help to take a look? Thanks!

Hi @fitzthum @stevenhorsman I think this is now ready for review.

confidential-data-hub/hub/src/config.rs
```diff
@@ -41,19 +41,82 @@ impl KbsConfig {
...
+    pub max_concurrent_download: usize,
+
+    /// sigstore config file for simple signing
+    ///
+    /// This defaults to [`image_rs::config::SIG_STORE_CONFIG_DEFAULT_FILE`].
+    pub sigstore_config_uri: String,
+
+    /// Path to `Policy.json`
+    ///
+    /// This defaults to [`image_rs::config::POLICY_FILE_PATH`].
```
stevenhorsman, line 74: Minor idea:
Is it worth a reference to https://github.com/containers/image/blob/main/docs/containers-policy.json.5.md, to help clarify what it's referring to, and distinguish  it from the other types of policy we have involved?

confidential-data-hub/hub/src/config.rs
```diff
@@ -41,19 +41,82 @@ impl KbsConfig {
...
+    /// The path to store the pulled data
+    pub work_dir: String,
+}
+
+pub const IMAGE_WORK_DIR: &str = "/tmp/image-rs";
+
+impl Default for ImageConfiguration {
+    fn default() -> Self {
+        Self {
+            security_validate: false,
```
stevenhorsman, line 97: I'm mixed as to whether we should consider have security validation on by default for "Secure by Default" purposes, or if that would cause issues?

confidential-data-hub/hub/src/config.rs
```diff
@@ -95,6 +159,21 @@ impl CdhConfig {
     fn from_file(config_path: &str) -> Result<Self> {
         let c = Config::builder()
             .set_default("socket", DEFAULT_CDH_SOCKET_ADDR)?
+            .set_default("image.security_validate", "true")?
+            .set_default("image.auth", "true")?
```
stevenhorsman, line 163: Should auth default to true? If it is set to true and `kbs:///default/credential/test` hasn't been set will it error, or just ignore the "empty" creds?

confidential-data-hub/hub/src/config.rs
```diff
@@ -227,6 +326,7 @@ path = "/run/confidential-containers/cdh/kms-credential/aliyun/config.toml"
             },
             credentials: Vec::new(),
             socket: DEFAULT_CDH_SOCKET_ADDR.into(),
+            image: crate::ImageConfiguration::default(),
```
stevenhorsman, line 329: Sorry if I've misunderstood something, but is it worth the unit test config file (or another test) having non-default values, so we can check the file reading is working and not just failing, but falling back on Default?

confidential-data-hub/hub/src/config.rs
```diff
@@ -41,19 +41,82 @@ impl KbsConfig {
...
+pub struct ImageConfiguration {
+    /// Whether to enable image security validation
+    ///
+    /// This defaults to `false`.
+    pub security_validate: bool,
+
+    /// Use `auth.json` control
+    /// ///
+    /// This defaults to `false`.
+    pub auth: bool,
```
stevenhorsman, line 60: Optionally idea: Maybe this is an opportunity to make this field clearer, which something like `use_credentials`, or `authenticated_registry_support` rather then just `auth`?

confidential-data-hub/hub/tests/image-pull.rs
```diff
@@ -0,0 +1,71 @@
...
+    "cosign signed unencrypted image"
+)]
+#[case(
+    "ghcr.io/confidential-containers/test-container:encrypted",
+    "unsigned encrypted image"
+)]
+#[case("quay.io/liudalibj/private-busy-box", "private registry")]
+#[tokio::test]
+#[serial_test::serial]
+async fn test_pull_image(#[case] image_ref: &str, #[case] item: &str) {
```
stevenhorsman, line 27: Would it be cleaner to use [case descriptions](https://docs.rs/rstest/latest/rstest/attr.rstest.html#optional-case-description) here, rather than passing in `item` as that e.g.
```
#[case("busybox:latest", "unsigned unencrypted image")]
```
becomes
```
#[case::unsigned_unencrypted_image("busybox:latest")]
```
?

.github/workflows/cdh_basic.yml
```diff
@@ -45,12 +45,11 @@ jobs:
           fetch-depth: 1
 
       - name: Install Rust toolchain (${{ matrix.rust }})
-        uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
-          toolchain: ${{ matrix.rust }}
-          override: true
-          components: rustfmt, clippy
+        run: |
```
stevenhorsman, line 48: Did you consider https://github.com/dtolnay/rust-toolchain for this as it seems widely used? This isn't nescessarily a suggestion, I've just realised that we use actions-rs all over the place and as you point out it has been unmaintained for a while and it using older node versions, so deciding on a single thing to replace it with (which could be just using cargo directly as you have), probably makes sense.

stevenhorsman (commented):
A few comments/questions from me, but overall it's looking good.

confidential-data-hub/Makefile
```diff
@@ -24,6 +24,7 @@ KMS_PROVIDER ?= aliyun,ehsm
 DESTDIR ?= $(PREFIX)/bin
 RUSTFLAGS_ARGS ?=
 features ?=
+IMAGE_PULL ?= false
```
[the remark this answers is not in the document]
Xynnn007: Oh it should be something left. Let me delete this.

confidential-data-hub/hub/src/config.rs
Xynnn007: One thing to note that the original config inside image-rs is marked as false. But we can re-write this in high level callers like what we do in this PR in CDH.

confidential-data-hub/hub/src/config.rs
Xynnn007: Oh. In kata we mark it as true by default [here](https://github.com/kata-containers/kata-containers/blob/main/src/agent/src/image.rs#L64). And with [this fix](https://github.com/confidential-containers/guest-components/pull/620/files), we'd better mark it by default false as you suggested. Thanks.

confidential-data-hub/hub/src/config.rs
Xynnn007: Nice suggestion.

Xynnn007: Is `use_registry_credential` a good name?

confidential-data-hub/hub/src/config.rs
stevenhorsman: Yeah, I think in `CCv0` we just said if we used the KBS at all we made the credentials check, so it was true a lot of the time and we'd handle missing creds gracefully, but now it fails (which is good for error checking), then we want to deliberately enable it when expected now I guess.

confidential-data-hub/hub/src/config.rs
stevenhorsman: Maybe `use_registry_credentials` as IIRC you can specify multiple credentials in a single auth.json file?

confidential-data-hub/hub/tests/image-pull.rs
Xynnn007: Oh! THis is a really cool feature! Let me do this!

stevenhorsman: Yeah, I need to remember to use it more myself too!

.github/workflows/cdh_basic.yml
Xynnn007: I think it is a good choice. Prpbably we can replace all yamls with this GHA in another PR. But sure let me make an issue to record this

confidential-data-hub/hub/src/config.rs
Xynnn007: Nice suggestion. Let me add a simple unit test w/o any field and see if all will be filled with default values. Also another test with an undefined field, and see if the config parser would ignore that field and fill other fields w/ default values.

.github/workflows/cdh_basic.yml
Xynnn007: finished in https://github.com/confidential-containers/guest-components/issues/675

image-rs/src/pull.rs
```diff
@@ -142,10 +142,16 @@ impl<'a> PullClient<'a> {
         };
 
         let decryptor = Decryptor::from_media_type(&layer.media_type);
+        let decryptor_i = decryptor.clone();
```
fitzthum, line 145: not sure about this naming convention.

I've seen people create a new scope around their closure and then shadow the variables like this.
```
tokio::spawn({
    let foo = foo.clone();
    async move {
        blah(foo)
    }
});
```
not sure if that would work here. If not maybe just `decryptor_clone`.

.github/workflows/cdh_basic.yml
mythi: independent of this conversation, I had reported the same in https://github.com/confidential-containers/confidential-containers/issues/230. would be good to track the work for the whole project, not just guest-components?

fitzthum (commented):
Nice. A couple initial comments.

confidential-data-hub/hub/src/hub.rs
```diff
@@ -6,19 +6,64 @@
...
+            let https_proxy = config.image.https_proxy;
+            if !https_proxy.is_empty() {
+                env::set_var("HTTPS_PROXY", https_proxy);
+            }
+        }
+
+        match env::var("HTTPS_PROXY") {
+            Ok(val) => info!("https_proxy is set to: {val}"),
+            Err(e) => log::warn!("failed to set https_proxy: {e}"),
+        };
```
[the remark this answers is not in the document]
Xynnn007: Yes I think you are right that we might probably only need only match statement. Btw, these messages would be printed by the CDH's stderr. In CoCo currently we do not have a way to expose, but in future we might need to redirect the logs of CDH and AA to somewhere for users.

In scenarios outside CoCo, it might be good to have these logs.

confidential-data-hub/hub/tests/image-pull.rs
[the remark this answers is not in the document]
Xynnn007: Oh. The tests named `tests/*` under the same dir of `src` are [integration tests](https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html) who can call the bianries of the crate. While the tests in the same file inside `#[cfg(test)]` are [unit tests](https://doc.rust-lang.org/rust-by-example/testing/unit_testing.html) and they can only call the functions of that mod.

In this case, to test the image pull we need to create a CDH process as the image decryption would access itself via ttrpc to get decryption keys, while a unit test cannot provide the ttrpc sock file.

confidential-data-hub/hub/tests/image-pull.rs
```diff
@@ -0,0 +1,64 @@
...
+        .unwrap()
+        .arg("pull-image")
+        .arg("--image-url")
+        .arg(image_ref)
+        .arg("--bundle-path")
+        .arg(tempdir.path())
+        .assert()
+        .success();
+
+    tokio::time::sleep(Duration::from_secs(4)).await;
```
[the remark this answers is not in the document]
Xynnn007: Good point, let me add this

image-rs/src/pull.rs
Xynnn007: Oh I test locally like 
```rust
// decrypt_config is `Option<&str>`, so we need to clone the &str to string outside the closure
let decrypt_config_clone = decrypt_config.as_ref().map(|inner| inner.to_string());
let decrypt_key = tokio::task::spawn_blocking(move || {
                let decryptor_clone = decryptor.clone();
                let layer_clone = layer.clone();
                decryptor_clone
                    .get_decrypt_key(&layer_clone, &decrypt_config_clone.as_deref())
                    .map_err(|e| anyhow!("failed to get decrypt key {}", e.to_string()))
            })
```

The compiler still cries because the moved value, hinting that we must do `clone` operation outside the closure.

Xynnn007: Let me just rename the variables

fitzthum: The  new scope has to go before the closure not inside of it.

confidential-data-hub/hub/tests/image-pull.rs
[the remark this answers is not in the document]
fitzthum: I guess the integration tests are a bit scattered, but we can try to organize later.

Xynnn007:
When I seem to test locally, found some little bugs in `image-rs` about image pulling idempotence when in same work dir. See the commit messages for details.

image-rs/src/pull.rs
Xynnn007: Oh! I get your point now. Let me rewrite as
```rust
let decrypt_key = tokio::task::spawn_blocking({
                let decryptor_clone = decryptor.clone();
                let layer_clone = layer.clone();
                let decrypt_config_clone = decrypt_config.as_ref().map(|inner| inner.to_string());
                move || {
                    decryptor_clone
                        .get_decrypt_key(&layer_clone, &decrypt_config_clone.as_deref())
                        .context("failed to get decrypt key")
                }
            })
```

confidential-data-hub/hub/tests/image-pull.rs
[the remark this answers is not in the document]
Xynnn007: Yes. Now both image-rs and confidential-data-hub have integration tests, overall their functions overlap. Maybe we could consider delete part of image-rs'.

confidential-data-hub/hub/src/config.rs
Xynnn007: Good, I deletes the ImageClient's default work dir para in CDH to leverage both functions without re-define consts.

ChengyuZhu6:
> When I seem to test locally, found some little bugs in `image-rs` about image pulling idempotence when in same work dir. See the commit messages for details.

Cool! I think it’s useful in some scenarios to prepare image layers and meta store files in the rootfs manually.

ChengyuZhu6:
cc @bpradipt

BbolroC:
I think an error below for s390x looks valid:

```
error: single-character string constant used as pattern
  --> confidential-data-hub/secret/src/secret/mod.rs:46:45
   |
46 |         let sections: Vec<_> = secret.split(".").collect();
   |                                             ^^^ help: try using a `char` instead: `'.'`
   |
```

Thanks!

Xynnn007:
> I think an error below for s390x looks legit:
> 
> ```
> error: single-character string constant used as pattern
>   --> confidential-data-hub/secret/src/secret/mod.rs:46:45
>    |
> 46 |         let sections: Vec<_> = secret.split(".").collect();
>    |                                             ^^^ help: try using a `char` instead: `'.'`
>    |
> ```
> 
> Thanks!

this error is not related to this PR. Probably we could fix it in another pr then

image-rs/src/pull.rs
fitzthum: If you have this new scope you can shadow the variables (i.e. `let decryptor = decryptor.clone()`), but it might be clearer to have new names. Up to you.

Xynnn007: Good, changed the name.

Xynnn007:
@BbolroC Fixed in https://github.com/confidential-containers/guest-components/pull/680 and get the pr rebased. and now it works

confidential-data-hub/hub/src/config.rs
```diff
@@ -0,0 +1,404 @@
...
+    pub resource_uri: String,
+    pub path: String,
+}
+
+#[derive(Clone, Deserialize, Debug, PartialEq)]
+pub struct ImageConfiguration {
+    /// Whether to enable image security validation
+    ///
+    /// This defaults to `false`.
+    pub security_validate: bool,
```
fitzthum, line 55: I've never really liked this parameter name. This refers to signature validation, right? Why not just call it that?

We should probably add slightly more extensive comments to this struct or the sample config.

confidential-data-hub/hub/src/config.rs
```diff
@@ -0,0 +1,404 @@
...
+    /// The flag to control whether we use registry credential file to
+    /// access the image registries.
+    ///
+    /// This defaults to `false`.
+    pub use_registry_credentials: bool,
+
+    /// Maximum number of concurrent downloads to perform during image pull.
+    ///
+    /// This defaults to [`image_rs::config::DEFAULT_MAX_CONCURRENT_DOWNLOAD`].
+    pub max_concurrent_download: usize,
```
fitzthum, line 66: Let's clarify whether this refers to pulling multiple images at once, multiple layers at once, or something else.

Also we might want to say what the actual default value is and when people might want to set it higher or lower.

image-rs/src/image.rs
```diff
@@ -356,11 +349,22 @@ impl ImageClient {
         let image_id = create_bundle(&image_data, bundle_dir, snapshot)?;
 
         self.meta_store
-            .lock()
+            .write()
             .await
             .image_db
             .insert(image_data.id.clone(), image_data.clone());
 
+        let metafile = self
```
fitzthum, line 357: nit: `metafile` -> `meta_file`

confidential-data-hub/hub/src/config.rs
```diff
@@ -0,0 +1,404 @@
...
+    /// <https://github.com/containers/image/blob/main/docs/containers-policy.json.5.md>.
+    ///
+    /// At the same time, some enhencements based on CoCo is used, that is the
+    /// keyPath field can be filled with a KBS URI like `kbs:///default/key/1`
+    pub policy_uri: String,
+
+    /// Path to the auth file
+    ///
+    /// This defaults to [`image_rs::config::AUTH_FILE_PATH`].
+    pub auth_uri: String,
```
fitzthum, line 86: I think we could group these args a little better. Put all the signature stuff together; all the authenticated registry stuff together, etc.

confidential-data-hub/hub/src/config.rs
```diff
@@ -0,0 +1,404 @@
...
+
+    /// Path to `Policy.json`
+    ///
+    /// This defaults to [`image_rs::config::POLICY_FILE_PATH`].
+    /// The policy follows the format of
+    /// <https://github.com/containers/image/blob/main/docs/containers-policy.json.5.md>.
+    ///
+    /// At the same time, some enhencements based on CoCo is used, that is the
+    /// keyPath field can be filled with a KBS URI like `kbs:///default/key/1`
+    pub policy_uri: String,
```
fitzthum, line 81: It's not obvious from the name or the comment that this has anything to do with signatures.

confidential-data-hub/hub/src/config.rs
```diff
@@ -0,0 +1,404 @@
...
+    /// Path to the auth file
+    ///
+    /// This defaults to [`image_rs::config::AUTH_FILE_PATH`].
+    pub auth_uri: String,
+
+    /// Proxy that will be used to pull image
+    pub https_proxy: String,
+
+    /// No proxy env that will be used to pull image
+    pub no_proxy: String,
```
fitzthum, line 92: This could also be a lot clearer. The `no_proxy` name is not great. I think this allows you to specify addresses that you don't need to use the proxy to connect to. We should give an example and maybe consider changing the name to `override` or `skip`.

confidential-data-hub/hub/src/config.rs
```diff
@@ -0,0 +1,404 @@
...
+
+    /// The path to store the pulled data
+    pub work_dir: Option<String>,
```
fitzthum, line 95: Will this ever be used by someone using CoCo?

fitzthum (commented):
A few more comments. These are mainly about the config file. I think we can take this opportunity to clarify a lot of our parameter names.

We should also make sure we have really clear documentation of all the parameters either in the source or in the sample config.

confidential-data-hub/hub/src/config.rs
Xynnn007: The image policy not only enables signature verification, but also can directly `reject` or `allAll` images. How about `enable_image_policy`?

confidential-data-hub/hub/src/config.rs
Xynnn007: What about this? We only leave one parameter `extra_registry_auth_uri` to cover both `use_registry_credentials` and `auth_uri`. That means if it is set, hinting `use_registry_credentials` is `true`. The same is for `image_policy_uri`.

confidential-data-hub/hub/src/config.rs
Xynnn007: No. But CDH can work for cases like directly inside IaaS platform rather than CoCo. This would help users to set different workdir which is to store the downloaded layers.

confidential-data-hub/hub/src/config.rs
Xynnn007: Ignore this comment and plz see https://github.com/confidential-containers/guest-components/pull/608#discussion_r1729722755

confidential-data-hub/hub/src/config.rs
bpradipt: @Xynnn007 this is for the case where we can embed image layers in the rootfs for very large images right ?

confidential-data-hub/hub/src/config.rs
bpradipt: May be we can use `insecureAcceptAnything` like used in containers policy.json ?

confidential-data-hub/hub/src/config.rs
Xynnn007: This `no_proxy` name is widely used. Maybe we can add some comments for this?

confidential-data-hub/hub/src/config.rs
Xynnn007: @bpradipt What do you mean by `insecureAcceptAnything`? This is a policy item that can be used in the policy.json

confidential-data-hub/hub/src/config.rs
Xynnn007: Yes, very large images is typical use case. Also beyond CoCo, CDH could has this configured.

confidential-data-hub/hub/src/config.rs
bpradipt: I meant naming the bool as "insecure_accept_anything" or "insecureacceptanything" to indicate whether image image security validation will be done or not.

Xynnn007:
@fitzthum I changed the config names, ptal

confidential-data-hub/hub/src/config.rs
Xynnn007: @bpradipt If needed, I think you can raise an issue to record this. After these technique preparations are made, we can edit an best practice document for this

confidential-data-hub/hub/src/config.rs
Xynnn007: Oh, got it. I changed the parameters and combined `security_validate` and `policy_uri` into one. Plaese help to check if it is now more straightforward

confidential-data-hub/hub/src/hub.rs
```diff
@@ -6,19 +6,79 @@
...
+        }
+
+        if let Some(number) = config.image.max_number_layer_download_currently {
+            image_client.config.max_concurrent_download = number;
+        }
+
+        // TODO: move the proxy envs to image-rs' PullClient once it supports
+        // Current the whole process of CDH would be influenced by the HTTPS_PROXY env
+        if let Some(https_proxy) = config.image.https_proxy {
+            match env::var("HTTPS_PROXY") {
```
fitzthum, line 57: Are we trying to keep supporting the configs set by kata? I guess in the short-term that is a good idea, but in the long term I think the only kata configs that we should follow is `aa_kbc_params`.

fitzthum (commented):
Looking good. I appreciate the changes to the config. I think combining the flags to enable the features with the URIs was a good idea.

I have a few more comments. Let's just try to settle on the names for the parameters. We can tweak the descriptions in a follow-up.

confidential-data-hub/hub/src/hub.rs
Xynnn007: For now the only way to set the proxy of image pull is to set the env of the whole process. This would have some side effects where other parts also wants to set the proxy, because the overall proxy env would also influence the KBS client. The check here is to ensure no other env is set ahead of time.

There is a thread https://github.com/oras-project/rust-oci-client/pull/160 to make `oci-client` that image-rs is using to pull image to support the two proxy parameters only for the client struct. Once that gets merged, we can fix this.

ChengyuZhu6 (approved):
Overall code looks good to me. LGTM ! thanks @Xynnn007

Xynnn007:
Just did a rebase to resolve conflicts.

confidential-data-hub/hub/src/hub.rs
```diff
@@ -6,19 +6,85 @@
...
+
+        if let Some(uri) = config.image.sigstore_config_uri {
+            if !uri.is_empty() {
+                image_client.config.file_paths.sigstore_config = uri;
+            }
+        }
+
+        if let Some(number) = config.image.max_concurrent_layer_downloads_per_image {
+            image_client.config.max_concurrent_download = number;
+        }
```
fitzthum, line 58: Do you think it would make sense to move this logic into the initializer for the `image_client` (and pass `config.image` into that) rather than putting all this image-specific stuff in this function?

fitzthum (approved):
I left one possibly suggestion but generally I think this is good. 

I can live with `no_proxy` for now I guess and we can clean up the docs a little bit later.

confidential-data-hub/hub/src/hub.rs
Xynnn007: There are some barriers because the parameters is different from `image-rs`'s. If you are worrying about the exposion of the args of `iamge_client`, we can add a `Builder` logic to `ImageClient`. If you are worrying about the duplication of settings, we could do a refactoring upon image-rs to have a `ClientBuilder` logic to make it better.

wdyt?

PR #610, opened 2024-07-10
attester: implement runtime measurement for az vtpm TEEs

mkulke:
the trait fn's extend_runtime_measurement() have been implemented for both az-snp-vtpm and az-tdx-vtpm. There is a bit of duplicate code, but it's better than introducing dependencies across attesters.


Xynnn007 (approved):
LGTM. only need to make the CI happy

fitzthum (approved):
Nice

PR #613, opened 2024-07-15
AA: add GetTeeType API

Xynnn007:
This new API is used for a caller to get current platform name.

---New things added

In some scenarios, callers want to get the name of current platform. Thus we need this basic semantics for AA.


imlk0:
> Rebased and ready for review. cc also @imlk0

Thanks @Xynnn007. That solves my problem.

mythi:
> In some scenarios, callers want to get the name of current platform.

I think it would be good to clarify in the commit message what are the scenarios and who are the callers.

attestation-agent/attestation-agent/src/bin/grpc-aa/server.rs
```diff
@@ -162,6 +163,31 @@ impl AttestationAgentService for AA {
...
+        let tee = attestation_agent.get_tee_type();
+
+        let tee = serde_json::to_string(&tee)
+            .map_err(|e| {
+                error!("AA (ttrpc): get tee type failed:\n {e:?}");
+                Status::internal(format!("[ERROR:{AGENT_NAME}] AA get tee type failed"))
+            })?
+            .trim_end_matches('"')
+            .trim_start_matches('"')
+            .to_string();
```
fitzthum, line 184: This json hack is a bit weird. Is there no way to deserialize as a string?

fitzthum (approved):
LGTM but see one nit

Xynnn007:
> I think it would be good to clarify in the commit message what are the scenarios and who are the callers.

Yes. Added the details. AA can run outside CoCo scenarios. For example, AA would run on an IaaS platform, and a user of such IaaS platform could get the tee type. Another example is some PaaS services on top of such IaaS platform, they also want to know the platform name.

attestation-agent/attestation-agent/src/bin/grpc-aa/server.rs
Xynnn007: The serialization comes from the enum `Tee` from `kbs-types`. In rust a `enum` will be serialized into forms like `"xxx"` rather than `xxx`. Thus we need to manually trim the `"` at both starting and end.

mythi:
> AA would run on an IaaS platform, and a user of such IaaS platform could get the tee type.

I was suspecting it might be something like this. It makes me think if it overlaps with what node-feature-discovery already gives.

Xynnn007:
> > AA would run on an IaaS platform, and a user of such IaaS platform could get the tee type.
> 
> I was suspecting it might be something like this. It makes me think if it overlaps with what node-feature-discovery already gives.

Yes. As a confidential runtime component, AA should have such ability itself.

fitzthum (approved):
LGTM

PR #615, opened 2024-07-16
AA: handle multiline content in log events

mkulke:
The NELR spec doesn't prescribe a format for the `content` payload of an event. However, since a line break is semantic in the log file, the content cannot be a multi-line string:

```
content = "one two\nthree four five"
               |
               v
my-domain my-operation one two
three four five
```

The result would be ambiguous. This change enforces this assertion. Also added tests for the event log and moved the init log line generation to the eventlog module (as the digest calc is already there).

### DIFF
--- a/attestation-agent/attestation-agent/src/lib.rs
+++ b/attestation-agent/attestation-agent/src/lib.rs
@@ -3,11 +3,10 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-use std::{io::Write, str::FromStr};
-
 use anyhow::{Context, Result};
 use async_trait::async_trait;
 use attester::{detect_tee_type, BoxedAttester};
+use std::{io::Write, str::FromStr};
 use tokio::sync::Mutex;
 
 pub use attester::InitdataResult;
@@ -16,8 +15,7 @@ pub mod config;
 mod eventlog;
 pub mod token;
 
-use config::HashAlgorithm;
-use eventlog::{EventEntry, EventLog};
+use eventlog::{Content, EventLog, LogEntry};
 use log::{debug, info, warn};
 use token::*;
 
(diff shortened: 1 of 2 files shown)


attestation-agent/attestation-agent/src/lib.rs
```diff
@@ -200,16 +191,20 @@ impl AttestationAPIs for AttestationAgent {
...
+            content,
+        };
+        let event_digest = log_entry.digest_with(self.config.eventlog_config.eventlog_algorithm);
 
         self.attester
             .extend_runtime_measurement(event_digest, register_index)
             .await?;
 
-        eventlog.write_log(&log_entry.to_string())?;
+        self.eventlog.lock().await.write_log(&log_entry)?;
```
Xynnn007, line 207: There's a very subtle synchronization issue here. 

Line 207 makes the lock only owned by the execution stream this line, while the original the whole `async fn extend_runtime_measurement()` function body owns the lock.

We'd better keep the lock being owned before and after `self.attester.extend_runtime_measurement()` calling to make `self.attester.extend_runtime_measurement()` and `write_log()` an atomic operation. without atomic semantics, when multiple `extend_runtime_measurement` requests come together, `write_log()` might not happen right after its `self.attester.extend_runtime_measurement()` but in out of order.

Xynnn007 (commented):
Nice catch. Besides line separators, spaces `' '` are also need to be carefully handled. In the original PR on AS side https://github.com/confidential-containers/trustee/pull/408/files#diff-aa0f7de1cbbcb4efaafaaa30a4b065987cf9d16f1fa01cfdb4e05c1323a1c59eR27 I didnot care about this. Let me fix this issue

mkulke:
> Nice catch. Besides line separators, spaces `' '` are also need to be carefully handled. In the original PR on AS side https://github.com/confidential-containers/trustee/pull/408/files#diff-aa0f7de1cbbcb4efaafaaa30a4b065987cf9d16f1fa01cfdb4e05c1323a1c59eR27 I didnot care about this. Let me fix this issue

Indeed. the first two fields have similar problems, maybe there should be a formal spec document for the log file in the repo, defines which characters are legal, are multiple INIT statements legal, etc...

attestation-agent/attestation-agent/src/lib.rs
mkulke: ah, good observation. I thought the lock was mostly about encapsulating the sync io from the async wrapper. I put it into a block to make that more obvious

Xynnn007 (approved):
lgtm

Issue #540, opened 2024-04-11
CDH | EncryptedDirectMount plugin for SecureMount


Xynnn007:
After https://github.com/confidential-containers/guest-components/pull/539 is merged. I'd like to implement a simple secure mount plugin independent on any enterprise storage services. The principle behind is like @arronwy 's excellent work https://github.com/kata-containers/kata-containers/pull/4885/files#diff-261600ce2c691940f7e65ecf8f5768c47d9efabe42000272e408029449dc22fb

The architecture diagram looks like
![image](https://github.com/confidential-containers/guest-components/assets/22250460/c59f5b9c-404e-41e8-b48b-24bbb99aea34)

The straightforward result is that
- TEE could use outer storage media but outside the TEE only ciphertext could be seen

The rationale behind this is to provide
- A quick start up for users to use secure mount
- Make it easier e2e ci for secure mount

#### Discussions

1. Where to find the decryption key?

There should provide two ways:
- The key is generated inside TEE guest. This hints that the storage will be lost if the pod exits or crashes.
- The key is retrieved from KBS/KMS

The key source could be specified by parameter.

fitzthum:
This is a great proposal. I think it will also help us develop tests for the secure storage feature (which will be required when we try to add it to kata). It would be cumbersome if the tests depended on some specific cloud offering.

This can also help us convince people to include gocryptfs in the generic confidential image.

I guess the question about the decryption key is about whether it is persistent storage or ephemeral storage. Ideally we could support both.

ChengyuZhu6:
Currently, we'll move the logic that encrypts the block device from host to guest in order to achieve trusted storage for pulling images.

ChengyuZhu6:
> Currently, we'll move the logic that encrypts the block device from host to guest in order to achieve trusted storage for pulling images.

Discussion: https://github.com/kata-containers/kata-containers/pull/9999#discussion_r1676356366

ChengyuZhu6:
So I think the task list is like as: 

- [ ] Secure mount api in cdh.
- [ ] Encrypt a block device and mount it to a specified path (moving the logic of trusted storage from Kata to CDH) .
- [ ] Decrypt a block device and mount it to a specified path (the feature of `secure mount`).

bpradipt:
@ChengyuZhu6 few questions for my understanding
[...]

PR #617, opened 2024-07-17
cdh: support to encrypt block device

ChengyuZhu6:
Support to encrypt block device in cdh.

Fixed: #540 -- part II

### DIFF
--- a/confidential-data-hub/docs/SECURE_STORAGE.md
+++ b/confidential-data-hub/docs/SECURE_STORAGE.md
@@ -62,3 +62,28 @@ flowchart LR
 ```
 
 For more details, please refer to [the guide](use-cases/secure-mount-with-aliyun-oss.md).
+
+### Block Device
+
+The [plugin](../storage/src/volume_type/blockdevice) provides ways to encrypt a block device and mount it to a specific mount point. Currently only support LUKS in [cryptsetup](https://gitlab.com/cryptsetup/cryptsetup/) for block device encryption.
+
+#### LUKS Encryption
+
+In this mode, the device would be encrypted as LUKS device first, and then mount it to a target path to store the data to protect the confidentiality and integrity of the data.
+
+The architecture diagram is
+
+```mermaid
+flowchart LR
+    A[Local/Network] -- mount --> B[Block Device]
+    subgraph TEE Guest
+        B -- Check if encrypted --> F{Is Encrypted?}
+        F -- No --> G[Encrypt by cryptsetup]
+        G -- encrypt --> C[LUKS Encrypted Block Device]
+        F -- Yes --> C[LUKS Encrypted Block Device]
+        C -- open and mapping --> D[Mapped Device]
+        D -- mount --> E[Target Path]
+    end
+```
+
+For more details, please refer to [the guide](use-cases/secure-mount-with-block-device.md).
--- a/confidential-data-hub/docs/use-cases/secure-mount-with-block-device.md
+++ b/confidential-data-hub/docs/use-cases/secure-mount-with-block-device.md
@@ -1,0 +1,77 @@
+# Secure mount with Block Device
+
+This guide helps an user to use confidential data hub to secure mount inside TEE environment.
+
+## Preliminaries
+
+- Ensure that [cryptsetup](https://gitlab.com/cryptsetup/cryptsetup/) is installed.
+
+## Example
+
+### Create a loop device
+
+```shell
+$ loop_file="/tmp/test.img"
+$ sudo dd if=/dev/zero of=$loop_file bs=1M count=1000
+$ sudo losetup -fP $loop_file
+$ device=$(sudo losetup -j $loop_file | awk -F'[: ]' '{print $1}')
+$ echo $device
+# Output should be something like /dev/loop0
+$ device_num=$(sudo lsblk -no MAJ:MIN $device)
+$ echo $device_num
+# Output should be something like 7:0
+```
+
+### Secure mount inside a TEE environment
+
+1. Build the CDH and its client tool
+
+Follow the instructions in the [CDH README](../../README.md#confidential-data-hub) and [Client Tool README](../../README.md#client-tool) to build the CDH and its client tool.
+
+2. Install `luks-encrypt-storage`
+
+Install [luks-encrypt-storage](../../storage/scripts/luks-encrypt-storage) into `/usr/local/bin`
+
+3. Run CDH
+```shell
+$ confidential-data-hub
+```
+
+4. Prepare a request JSON `storage.json`
+```json
+{
+    {
+        "volume_type": "BlockDevice",
+        "options": {
+            "deviceId": "7:0",
+            "encryptType": "LUKS",
+            "dataIntegrity": "true"
+        },
+        "flags": [],
+        "mount_point": "/mnt/test-path"
+    }
+}
+
+```
+- Fields:
+    - `volume_type`: The secure mount plugin type name. It determines how the rest of the fields are used.
+    - `options`: A key-value map specifying the settings for the mount operation. Different plugins can define different keys in the options. In this example, all keys are for block devices.
+    - `flags`: A string list specifying settings for the mount operation. Different plugins can define different uses for this field.
+    - `mount_point`: The target mount path for the operation.
+
+- Options Fields:
+    - `deviceId`: The device number, formatted as "MAJ:MIN".
+    - `encryptType`: The encryption type. Currently, only LUKS is supported.
+    - `encryptKey`: Encryption key. It can be a sealed secret or a resource uri. If not set, it means that the device is unencrypted and a random 4096-byte key will be generated to encrypt the device.
+    - `dataIntegrity`: Enables dm-integrity to protect data integrity. Note that enabling data integrity will reduce IO performance by more than 30%.
+
+5. Make a request to CDH
+```shell
+$ client-tool secure-mount --storage-path storage.json
+
+# Check the target path to see if the mount succeeded
+$ lsblk |grep "encrypted_disk"
+# Expected output:
+└─encrypted_disk_OEyEj_dif 253:1    0 968.6M  0 crypt
+  └─encrypted_disk_OEyEj   253:2    0 968.6M  0 crypt /mnt/test-path
+```
--- a/confidential-data-hub/storage/Cargo.toml
+++ b/confidential-data-hub/storage/Cargo.toml
@@ -10,6 +10,7 @@ anyhow.workspace = true
 async-trait.workspace = true
 base64.workspace = true
 log.workspace = true
+kms = { path = "../kms", features = ["kbs"] }
 rand = { workspace = true, optional = true }
 secret = { path = "../secret" }
 serde.workspace = true
--- a/confidential-data-hub/storage/src/volume_type/blockdevice/luks/mod.rs
+++ b/confidential-data-hub/storage/src/volume_type/blockdevice/luks/mod.rs
@@ -1,0 +1,91 @@
+// Copyright (c) 2024 Intel
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+use super::Interpreter;
+use super::{get_plaintext_key, BlockDeviceError, BlockDeviceParameters, Result};
+use async_trait::async_trait;
+use log::error;
+use rand::{distributions::Alphanumeric, Rng};
+use tokio::{
+    fs,
+    io::{AsyncReadExt, AsyncWriteExt},
+    process::Command,
+};
+
+/// LUKS encrypt storage binary
+const LUKS_ENCRYPT_STORAGE_BIN: &str = "/usr/local/bin/luks-encrypt-storage";
+
+async fn random_encrypt_key() -> Vec<u8> {
+    let mut buffer = vec![0u8; 4096];
+    rand::thread_rng().fill(&mut buffer[..]);
+    buffer
+}
+
+async fn create_storage_key_file(
+    storage_key_path: &str,
+    encrypt_key: Option<String>,
+) -> anyhow::Result<()> {
+    let mut storage_key_file = fs::File::create(storage_key_path).await?;
+
+    let plain_key = match encrypt_key {
+        Some(encrypt_key) => get_plaintext_key(&encrypt_key).await?,
+        None => random_encrypt_key().await,
+    };
+
+    storage_key_file.write_all(&plain_key).await?;
+    storage_key_file.flush().await?;
+    Ok(())
+}
+
+pub(crate) struct LuksInterpreter;
+
+#[async_trait]
+impl Interpreter for LuksInterpreter {
+    async fn secure_device_mount(
+        &self,
+        parameters: BlockDeviceParameters,
+        mount_point: &str,
+    ) -> Result<()> {
+        let random_string: String = rand::thread_rng()
+            .sample_iter(&Alphanumeric)
+            .take(5)
+            .map(char::from)
+            .collect();
+        let storage_key_path = format!("/tmp/encrypted_storage_key_{}", random_string);
+        create_storage_key_file(&storage_key_path, parameters.encryption_key.clone()).await?;
+
+        let parameters = vec![
+            parameters.device_id,
+            match parameters.encryption_key {
+                None => "false".to_string(),
+                Some(_) => "true".to_string(),
+            },
+            mount_point.to_string(),
+            storage_key_path,
+            parameters.data_integrity.clone(),
+        ];
+
+        let mut encrypt_device = Command::new(LUKS_ENCRYPT_STORAGE_BIN)
+            .args(parameters)
+            .spawn()
+            .map_err(|e| {
+                error!("luks-encrypt-storage cmd fork failed: {:?}", e);
+                BlockDeviceError::BlockDeviceMountFailed
+            })?;
+
+        let bd_res = encrypt_device.wait().await?;
+        if !bd_res.success() {
+            let mut stderr = String::new();
+            if let Some(mut err) = encrypt_device.stderr.take() {
+                err.read_to_string(&mut stderr).await?;
+                error!("BlockDevice mount failed with stderr: {:?}", stderr);
+            } else {
+                error!("BlockDevice mount failed");
+            }
+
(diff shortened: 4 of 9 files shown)


confidential-data-hub/storage/src/volume_type/rbd/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+pub(crate) struct Rbd;
+
+async fn random_encrypt_key() -> anyhow::Result<String> {
+    let storage_key_path = "/tmp/random_encrypt_rbd.key";
+    let mut file = fs::File::create(storage_key_path).await?;
+
+    let mut buffer = vec![0u8; 4096];
+    rand::thread_rng().fill(&mut buffer[..]);
+
+    file.write_all(&buffer).await?;
```
Xynnn007, line 56: we also need `file.flush().await` because operation system might delay the real write back, which may cause some unexpected synchronization issues.

confidential-data-hub/storage/src/volume_type/rbd/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+
+    /// Indicates whether to enable dm-integrity.
+    #[serde(rename = "dataIntegrity")]
+    pub data_integrity: String,
+}
+
+pub(crate) struct Rbd;
+
+async fn random_encrypt_key() -> anyhow::Result<String> {
+    let storage_key_path = "/tmp/random_encrypt_rbd.key";
```
Xynnn007, line 50: Do we support multiple multiple encrypted rbd mounts? If so, do we need different keys?

confidential-data-hub/storage/src/volume_type/rbd/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+
+#[derive(Serialize, Deserialize, PartialEq, Debug)]
+struct RbdParameters {
+    /// The device number, formatted as "MAJ:MIN".
+    #[serde(rename = "deviceId")]
+    pub device_id: String,
+
+    /// The encryption type. Currently, only LUKSfs is supported.
+    #[serde(rename = "encryptType")]
+    pub encrypt_type: String,
```
Xynnn007, line 32: Could we use an `enum` for this?

confidential-data-hub/storage/src/volume_type/rbd/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+
+    /// Indicates whether the encryption key is generated by CDH or set by users.
+    #[serde(rename = "randomEncrypt")]
+    pub random_encrypt: String,
+
+    /// Encryption key if `random_encrypt` is "false".
+    #[serde(rename = "encryptKey")]
+    pub encrypt_key: Option<String>,
```
Xynnn007, line 40: Maybe we could combine the two parameters into one `encrypt_key`? s.t.
1. if `encrypt_key` is given, the key will be got by unseal the secret a.k.a the value of `encrypt_key`
2. if not, use a random key generated inside CDH

confidential-data-hub/storage/src/volume_type/rbd/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+        _flags: &[String],
+        mount_point: &str,
+    ) -> Result<()> {
+        // construct RbdParameters
+        let parameters = serde_json::to_string(options)?;
+        let rbd_parameter: RbdParameters = serde_json::from_str(&parameters)?;
+
+        if rbd_parameter.encrypt_type == "luksFS" {
+            let key_path = if rbd_parameter.random_encrypt == "false" {
+                // TODO get key from secret
```
Xynnn007, line 73: it is relatively simple
```rust
let key = secret::unseal_secret(rbd_parameter.encrypt_key).await?;
// some logic to write `key` into a file
```

Xynnn007 (commented):
Nice PR! We could image that storage of CoCo could be handled in near future

zvonkok:
Can we make sure we have proper documentation about this feature? Issues, PRs and discussions are easily consumed by developers, end users need a concise way of consuming such cool features without digging through github.

ChengyuZhu6:
> Can we make sure we have proper documentation about this feature? Issues, PRs and discussions are easily consumed by developers, end users need a concise way of consuming such cool features without digging through github.

Of course we should have a good document about it!

confidential-data-hub/storage/src/volume_type/rbd/mod.rs
ChengyuZhu6: Done.

confidential-data-hub/storage/src/volume_type/rbd/mod.rs
ChengyuZhu6: Done.

confidential-data-hub/storage/src/volume_type/rbd/mod.rs
ChengyuZhu6: Fixed.

confidential-data-hub/storage/src/volume_type/rbd/mod.rs
ChengyuZhu6: Done

confidential-data-hub/storage/src/volume_type/rbd/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+
+pub(crate) struct Rbd;
+
+async fn random_encrypt_key() -> anyhow::Result<String> {
+    let mut buffer = vec![0u8; 4096];
+    rand::thread_rng().fill(&mut buffer[..]);
+    Ok(base64::engine::general_purpose::STANDARD.encode(&buffer))
+}
+
+async fn get_plaintext_secret(secret: &str) -> anyhow::Result<String> {
```
fitzthum, line 59: I think I was expecting the encryption key to be a resource uri. Maybe having it be a sealed secret is better because that contains info about the KBS/HSM, integrity protection, etc., but so far we haven't used a sealed secret this way in any other parts of the CDH. Are we planning to adopt sealed secret format as a general wrapper around a resource uri?

cc: @Xynnn007

confidential-data-hub/storage/src/volume_type/rbd/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+        options: &HashMap<String, String>,
+        _flags: &[String],
+        mount_point: &str,
+    ) -> Result<()> {
+        // construct RbdParameters
+        let parameters = serde_json::to_string(options)?;
+        let rbd_parameter: RbdParameters = serde_json::from_str(&parameters)?;
+
+        if rbd_parameter.encrypt_type == RbdEncryptType::LUKS {
+            let storage_key_path = create_storage_key_file(&rbd_parameter).await?;
```
fitzthum, line 101: should we delete this file at some point?

confidential-data-hub/storage/src/volume_type/mod.rs
```diff
@@ -61,6 +62,12 @@ impl Storage {
                     .await?;
                 Ok(self.mount_point.clone())
             }
+            Volume::Rbd => {
```
fitzthum, line 65: Rbd = rados block device? I don't see anything in this PR that is specific to RBD. Maybe this should just be called `Volume::Luks` or `Volume::Encrypted` in case we support gocryptfs in the future.

It's also interesting to note that a number of storage implementations might share the same encryption scheme but have different ways to mount the volume. I wonder if at some point we should refactor the storage design to have two phases (setup device and decrypt volume) so that some of the code could be shared. In this case we aren't doing anything to setup the device so we can probably worry about it later.

fitzthum (commented):
Looks good. A few comments.

Might be nice to come up with some kind of test.

confidential-data-hub/storage/src/volume_type/mod.rs
ChengyuZhu6: Ah, I mean RBD = raw block device.  To avoid any ambiguity, we may consider coming up with a better name.

confidential-data-hub/storage/src/volume_type/rbd/mod.rs
ChengyuZhu6: It would be removed in the binary `luks-encrypt-storage`

confidential-data-hub/storage/src/volume_type/rbd/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+    pub encrypt_key: Option<String>,
+
+    /// Indicates whether to enable dm-integrity.
+    #[serde(rename = "dataIntegrity")]
+    pub data_integrity: String,
+}
+
+pub(crate) struct Rbd;
+
+async fn random_encrypt_key() -> anyhow::Result<String> {
```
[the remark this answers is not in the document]
ChengyuZhu6: Fixed.

confidential-data-hub/storage/src/volume_type/rbd/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+
+#[derive(Serialize, Deserialize, PartialEq, Debug)]
+struct RbdParameters {
+    /// The device number, formatted as "MAJ:MIN".
+    #[serde(rename = "deviceId")]
+    pub device_id: String,
+
+    /// The encryption type. Currently, only LUKS is supported.
+    #[serde(rename = "encryptType")]
+    pub encrypt_type: RbdEncryptType,
```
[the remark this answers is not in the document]
ChengyuZhu6: Fixed.

confidential-data-hub/storage/src/volume_type/rbd/mod.rs
Xynnn007: I think they are for different scenarios. Given that we now support both KMS as a KBS backend (use KBS URI to index a key) and KMS as a CDH plugin (use Sealed Secret to index a key), we might need to support both. We can add a condition judgement here, if with `kbs://` prefix, use 
```rust
new_getter("kbs", ProviderSettings::default()).await?.get_secret(&secret, &Annotations::default())
```
if with `sealed.` prefix, use
```rust
secret::unseal_secret(secret.as_bytes()).await?
```
else, raise an error saying that illegal keyid format.

Thus we could support both inside CDH code. The consequence is both deploy model could be used and the user does need to worry about underlying logic.

confidential-data-hub/storage/src/volume_type/mod.rs
Xynnn007: Maybe we should add some document for this. I think the `RBD` plugin here would mean "given a raw block device, mount it to the target path and decrypt it with given/ephemeral key".


The initial code does not care about the device setup. This also raises some open question. If CDH cares about the setup of the device, how to ensure the availability? I am not sure how kata now handles volumes if volumes are broken during runtime. Maybe we can learn something from kata.

confidential-data-hub/storage/Cargo.toml
```diff
@@ -10,6 +10,7 @@ anyhow.workspace = true
 async-trait.workspace = true
 base64.workspace = true
 log.workspace = true
+kms = { path = "../kms", default-features = false }
```
Xynnn007, line 13: Do we need `features = ["kbs"]`?

confidential-data-hub/storage/src/volume_type/blockdevice/luks/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+    process::Command,
+};
+
+/// LUKS encrypt storage binary
+const LUKS_ENCRYPT_STORAGE_BIN: &str = "/usr/local/bin/luks-encrypt-storage";
+
+async fn random_encrypt_key() -> anyhow::Result<String> {
+    let mut buffer = vec![0u8; 4096];
+    rand::thread_rng().fill(&mut buffer[..]);
+    Ok(base64::engine::general_purpose::STANDARD.encode(&buffer))
```
Xynnn007, line 23: This function does not return `Err`, thus we could directly use `async fn random_encrypt_key() -> String`?

confidential-data-hub/storage/src/volume_type/blockdevice/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+#[async_trait]
+pub trait Encryptor {
+    async fn encrypt_and_mount(
+        &self,
+        parameters: BlockDeviceParameters,
+        mount_point: &str,
+    ) -> Result<()>;
+}
+
+async fn get_plaintext_key(resource: &str) -> anyhow::Result<String> {
```
Xynnn007, line 53: Must the key be tuf8 string, or could be byte slice like `Vec<u8>`?

Xynnn007 (commented):
Nice, some nits.

confidential-data-hub/storage/Cargo.toml
ChengyuZhu6: Done.

confidential-data-hub/storage/src/volume_type/blockdevice/luks/mod.rs
ChengyuZhu6: Done.

confidential-data-hub/storage/src/volume_type/mod.rs
fitzthum: Maybe spell it out as `RawBlockDevice` or just call it something else to avoid confusion.

Provisioning the device can involve operations by the kata agent (i.e. mounting something from the host) and/or operations by the CDH (i.e. pulling data from cloud storage). In the first case (this PR) I think we basically have to assume that the kata agent (or more generally the caller of the CDH) has set things up correctly.

When we start implementing more sophisticated storage backends maybe it will make sense to make things a little bit more modular.

confidential-data-hub/storage/src/volume_type/rbd/mod.rs
fitzthum: Ok, I think supporting both makes sense. Each scheme has some pros/cons. I wonder if we should do something similar for other features.

confidential-data-hub/docs/SECURE_STORAGE.md
```diff
@@ -62,3 +62,24 @@ flowchart LR
...
+In this mode, the device would be encrypted as LUKS device first, and then mount it to a target path to store the data to protect the confidentiality and integrity of the data.
+
+The architecture diagram is
+
+```mermaid
+flowchart LR
+    A[Block Device] -- mount --> B[Block Device]
+    subgraph TEE Guest
+        B -- encrypt by cryptsetup --> C[LUKS Block Device]
+        C -- mount --> D[Target Path]
```
Xynnn007, line 81: Could we mark which devices are encrypted, and which are plaintext?

confidential-data-hub/docs/use-cases/secure-mount-with-block-device.md
```diff
@@ -0,0 +1,77 @@
...
+
+### Secure mount inside a TEE environment
+
+1. Build the CDH and its client tool
+
+Follow the instructions in the [CDH README](../../README.md#confidential-data-hub) and [Client Tool README](../../README.md#client-tool) to build the CDH and its client tool.
+
+2. Install `luks-encrypt-storage`
+
+Install [luks-encrypt-storage](../../storage/scripts/luks-encrypt-storage) into `/usr/local/bin`
```
Xynnn007, line 33: As a separate service, CDH could require install extra things. I am thinking about whether we can integrate the installations of `luks-encrypt-storage`, `cachefs` and `ossfs` binaries/scripts in CDH's Makefile. In CCv0, we explicit install them [in kata scripts](https://github.com/kata-containers/kata-containers/pull/7965/files#diff-4a3807c6c399f5750341b6dd154009c87f1dc69fc0345b66a0af381ebdf1dab4R79-R80). This would bring extra maintaining cost for kata. cc @fitzthum @fidencio

Xynnn007 (commented):
Nice. Only one last question

confidential-data-hub/docs/use-cases/secure-mount-with-block-device.md
ChengyuZhu6: I agree that we can move the tasks of installing binaries or scripts to CDH, as they are only used in the CDH. CDH is responsible for installing them, and we can simplify the code in kata.

confidential-data-hub/docs/SECURE_STORAGE.md
ChengyuZhu6: Done.

confidential-data-hub/storage/src/volume_type/blockdevice/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+    pub encryption_key: Option<String>,
+
+    /// Indicates whether to enable dm-integrity.
+    #[serde(rename = "dataIntegrity")]
+    pub data_integrity: String,
+}
+pub(crate) struct BlockDevice;
+
+#[async_trait]
+pub trait Encryptor {
```
fitzthum, line 44: should this be encryptor or decryptor?

fitzthum (approved):
LGTM

I'm not sure exactly the best way to handle dependencies like the script. Maybe it would be ideal if we converted that to rust at some point. For now you can probably find some way to get it into the rootfs.

ChengyuZhu6:
@zvonkok Does the documentation make sense to you? Is there anything else that needs to be added or clarified?

ChengyuZhu6:
> I'm not sure exactly the best way to handle dependencies like the script. Maybe it would be ideal if we converted that to rust at some point. For now you can probably find some way to get it into the rootfs.

Sounds like a good idea for the future. I’ll keep that in mind and we can revisit it later.

confidential-data-hub/storage/src/volume_type/blockdevice/mod.rs
ChengyuZhu6: For trusted storage, this should be encryptor. The guest will need to follow three steps: encrypt the block device from the host as a LUKS-encrypted block device, open the encrypted device with the key, and mount it. However, for secure storage, the guest will only need to perform the latter two steps. Overall, I think `encryptor` or `decryptor` are not suitable names for it. Therefore, I have renamed it to `Interpreter`.

Xynnn007 (approved):

PR #618, opened 2024-07-17
image-rs: update cosign signed image test materials

Xynnn007:
Now, the cases

Case: Deny pulling an unencrypted unsigned image from a protected registry
Image: ghcr.io/confidential-containers/test-container-image-rs:unsigned

Case: Allow pulling an unencrypted signed image with cosign-signed signature
Image: ghcr.io/confidential-containers/test-container-image-rs:cosign-signed

Case: Deny pulling an unencrypted signed image by cosign using a wrong public key
Image: ghcr.io/confidential-containers/test-container-image-rs:cosign-signed-key2

At the same time, the images on the ghcr.io side is updated. The original tag `cosign-signed-key2` is actually the `unsigned` one, and we updated a new real `unsigned` one.

Related policy file updated.

cc @stevenhorsman


fitzthum (approved):
We should really automate the process of image generation at some point.

Xynnn007:
> We should really automate the process of image generation at some point.

Yes. We are now manually doing this now. Hopefully we could have a full e2e test that includes encryption and signing process

PR #620, opened 2024-07-19
image-rs: bail out if unable to get registry auth credentials

wainersm:
The current behavior of switching to anonymous authentication does hide the error to get the credentials. So changed the logic to bail out in case of error.

### DIFF
--- a/image-rs/src/image.rs
+++ b/image-rs/src/image.rs
@@ -3,7 +3,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use anyhow::{anyhow, bail, Result};
-use log::warn;
 use oci_distribution::manifest::{OciDescriptor, OciImageManifest};
 use oci_distribution::secrets::RegistryAuth;
 use oci_distribution::Reference;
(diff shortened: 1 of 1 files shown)


image-rs/src/image.rs
```diff
@@ -229,19 +228,15 @@ impl ImageClient {
...
+                        bail!("failed to get registry authentication credentials: {:?}", e)
                     }
                 }
             }
             (false, true) => RegistryAuth::Anonymous,
             _ => auth.expect("unexpected uninitialized auth"),
         };
 
-        let mut client = PullClient::new(
+        let mut client: PullClient = PullClient::new(
```
Xynnn007, line 239: do we need this?

Xynnn007 (approved):
thanks @wainersm! only a nit

image-rs/src/image.rs
wainersm: Hi @Xynnn007 !

Good catch. Probably it was automatically refactoried by vscode. I will revert that.

wainersm: done

fitzthum (approved):
Cool

PR #624, opened 2024-07-23
cdh/kms: modify get_secret() function

1570005763:
Change get_secret() function of AliyunKmsClient::ClientKey, AliyunKmsClient::EcsRamRole, and AliyunKmsClient::StsToken into immutable form. And also separate them with Getter trait.


Xynnn007 (approved):
good. let's see how ci says.

PR #625, opened 2024-07-23
CDH/KMS: mark Get trait immutable

Xynnn007:
Typically, a `get` operation should not change any inner state of the storage. Thus we mark this trait immutable.

This would also do help to outer callers.

cc @fitzthum


1570005763 (approved):
LGTM

jialez0 (approved):

PR #627, opened 2024-07-23
AA: add flag to enable eventlog

### LINKED ISSUE #626: Init Attestation agent failed on tdx
The initialization of attestation agent failed on TDX in CoCo. The AA is built from commit `0d8146321c3de023f0f7f40e47fc0f860133dfc7` , error log: 
```
"Error: init AA"
"Caused by:"
"    0: write INIT entry"
"    1: TDX Attester: Cannot extend runtime measurement on this system"
"
"Stack backtrace:"
"   0: <unknown>"
"   1: <unknown>"
"   2: <unknown>"
"   3: <unknown>"
"   4: <unknown>"
"   5: <unknown>"
"   6: <unknown>"
"   7: <unknown>"
"   8: __libc_start_main"
"   9: <unknown>"
```

Xynnn007:
On most platforms, runtime measurement is now not supported. Thus the init PCR extension will fail.

This commit adds a flag for AA launch config to control whether eventlog is enabled. By default it will be closed.

In future, with initdata, the deployer would specify whether to enable eventlog in AA's config.

fixes #626

cc @ChengyuZhu6 

cc also @arronwy

### DIFF
--- a/attestation-agent/attestation-agent/src/config/mod.rs
+++ b/attestation-agent/attestation-agent/src/config/mod.rs
@@ -69,6 +69,9 @@ pub struct EventlogConfig {
 
     /// PCR Register to extend INIT entry
     pub init_pcr: u64,
+
+    /// Flag whether enable eventlog recording
+    pub enable_eventlog: bool,
 }
 
 impl Default for EventlogConfig {
@@ -76,6 +79,7 @@ impl Default for EventlogConfig {
         Self {
             eventlog_algorithm: HashAlgorithm::Sha384,
             init_pcr: DEFAULT_PCR_INDEX,
+            enable_eventlog: false,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


arronwy:
we may have two options for enabling eventlog, one is current PR, user decide whether enable in initdata and disable by default, if user enabled but stack/platform not supported, the `attestation-agent` will failed to launch. another option is `attestation-agent` detect whether current stack support runtime measurement, if not, just skip the operation and ensure `attestation-agent` can still launch success: 
https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/attester/src/tdx/mod.rs#L127-L129
https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/attester/src/lib.rs#L88-L91

any suggestions @Xynnn007 @fitzthum @ChengyuZhu6

Xynnn007:
@arronwy This is a good point that will save the effort of users, while at the same time lose some flexibility. I am open to both ways -- they both have pros and cons.

ChengyuZhu6:
Tested and verified with TDX: https://github.com/kata-containers/kata-containers/actions/runs/10069229690/job/27836772221?pr=9999

fitzthum:
Either option is fine with me. I think explicitly enabling/disabling is a bit less error-prone.

ChengyuZhu6 (approved):
LGTM thanks @Xynnn007

mkulke (approved):

arronwy (approved):
Thanks @Xynnn007 LGTM!

PR #628, opened 2024-07-25
Improve KBS protocol version handling and bump the version to v0.1.1 due to kbs-types changes

mythi:
Pull in @Xynnn007's work for https://github.com/confidential-containers/trustee/issues/242#issuecomment-2146864188

### DIFF
--- a/attestation-agent/kbs_protocol/src/client/rcar_client.rs
+++ b/attestation-agent/kbs_protocol/src/client/rcar_client.rs
@@ -109,16 +109,34 @@ impl KbsClient<Box<dyn EvidenceProvider>> {
 
         debug!("send auth request to {auth_endpoint}");
 
-        let challenge = self
+        let resp = self
             .http_client
             .post(auth_endpoint)
             .header("Content-Type", "application/json")
             .json(&request)
             .send()
-            .await?
-            .json::<Challenge>()
             .await?;
 
+        match resp.status() {
+            reqwest::StatusCode::OK => {
+                debug!("KBS request OK");
+            }
+            reqwest::StatusCode::UNAUTHORIZED => {
+                let error_info = resp.json::<ErrorInformation>().await?;
+                bail!(
+                    "KBS request unauthorized, ErrorInformation: {:?}",
+                    error_info
+                );
+            }
+            _ => {
+                bail!(
+                    "KBS Server Internal Failed, Response: {:?}",
+                    resp.text().await?
+                );
+            }
+        }
+
+        let challenge = resp.json::<Challenge>().await?;
         debug!("get challenge: {challenge:#?}");
         let tee_pubkey = self.tee_key.export_pubkey()?;
         let runtime_data = json!({
--- a/attestation-agent/kbs_protocol/src/client/rcar_client.rs
+++ b/attestation-agent/kbs_protocol/src/client/rcar_client.rs
@@ -363,10 +363,19 @@ mod test {
             .try_into()
             .expect("resource uri");
 
-        let resource = client
-            .get_resource(resource_uri)
-            .await
-            .expect("get resource");
+        let resource = match client.get_resource(resource_uri).await {
+            Ok(resource) => resource,
+            Err(e) => {
+                // Skip the test if the kbs server returned ProtocolVersion error. Any other
+                // error is treated as a failure.
+                assert!(e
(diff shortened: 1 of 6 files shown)


fitzthum:
whoops, didn't mean to make this ready for review

mythi:
This PR now depends on getting https://github.com/confidential-containers/trustee/pull/449 merged and released with with `:latest` kbs image.

mythi:
> This PR now depends on getting [confidential-containers/trustee#449](https://github.com/confidential-containers/trustee/pull/449) merged and released with `:latest` kbs image.

`:latest` tag does not seem to get updated due to pending s390x jobs.

mythi:
rebased to include the latest `Cargo.toml` updates that were merged

@mkulke @fitzthum this PR now breaks the cyclic dependency in a way that we can continue to use `test_client` test.

mythi:
@Xynnn007 PTAL also

jodh-intel (approved):
Thanks, @mythi.

lgtm!

attestation-agent/kbs_protocol/src/client/rcar_client.rs
```diff
@@ -135,7 +147,7 @@ impl KbsClient<Box<dyn EvidenceProvider>> {
         let attest_endpoint = format!("{}/{KBS_PREFIX}/attest", self.kbs_host_url);
         let attest = Attestation {
             tee_pubkey,
-            tee_evidence: evidence,
+            tee_evidence: serde_json::from_str(&evidence)?, // TODO: change attesters to return Value?
```
mkulke, line 150: see comment above about using serde types directly.

Xynnn007 (approved):
lgtm

attestation-agent/kbs_protocol/src/client/rcar_client.rs
mythi: is the comment missing from the review?

mkulke: oh indeed. sorry

attestation-agent/kbs_protocol/src/client/rcar_client.rs
```diff
@@ -104,21 +104,33 @@ impl KbsClient<Box<dyn EvidenceProvider>> {
         let request = Request {
             version: String::from(KBS_PROTOCOL_VERSION),
             tee,
-            extra_params: String::new(),
+            extra_params: serde_json::Value::String(String::new()),
```
mkulke, line 107: It's uncommon to use serde types directly like this. Usually this would be solved by a serde annotation on the struct.

mythi: so the `kbs-types` for `Request` would assign an empty string by default so that this line could be dropped completely? That would probably makes sense. The change here is what the compiler suggested

mkulke: oh, I see. the kbs-types package exposes a serde_json type on its public types. that's even more unusual. I'll open an issue on virtee/kbs-types. but there's not much we can do here.

mkulke (approved):

attestation-agent/kbs_protocol/src/client/rcar_client.rs
```diff
@@ -104,21 +104,33 @@ impl KbsClient<Box<dyn EvidenceProvider>> {
...
+        match resp.status() {
+            reqwest::StatusCode::UNAUTHORIZED => {
+                let error_info = resp.json::<ErrorInformation>().await?;
+                bail!(
+                    "KBS request unauthorized, ErrorInformation: {:?}",
+                    error_info
+                );
+            }
+            _ => {
+                debug!("KBS request OK");
```
fitzthum, line 129: Maybe we should explicitly check for the success status code here and use the fall through case to catch other types of errors.

fitzthum (commented):
Sorry for the delay in reviewing. Seems good. One suggestion

attestation-agent/kbs_protocol/src/client/rcar_client.rs
mythi: fixed (the same way it's done for the `attest_response` below).

fitzthum (approved):
LGTM

PR #630, opened 2024-07-27
ocicrypt-rs: dont't swallow pre_unwrap_key() error

mkulke:
The error for encryption::pre_unwrap_key() is silently discarded, which means we get no meaningful error in consumers, like the kata-agent log on ttrpc connectivity problems. I don't think it makes sense to not abort on errors from pre_unwrap_key.

### DIFF
--- a/ocicrypt-rs/src/encryption.rs
+++ b/ocicrypt-rs/src/encryption.rs
@@ -188,10 +188,9 @@ pub fn decrypt_layer_key_opts_data(
                 priv_key_given = true;
             }
 
-            if let Ok(opts_data) = pre_unwrap_key(keywrapper, dc, &b64_annotation) {
-                if !opts_data.is_empty() {
-                    return Ok(opts_data);
-                }
+            let opts_data = pre_unwrap_key(keywrapper, dc, &b64_annotation)?;
+            if !opts_data.is_empty() {
+                return Ok(opts_data);
             }
             // try next keywrapper
         }


Xynnn007 (approved):
Thanks for the contribution. The error log for this surely made trouble when debugging. One thing to note is that The original logic will try all keywrapper plugins, but current by default assume that only one keywrapper is configured.

mkulke:
> Thanks for the contribution. The error log for this surely made trouble when debugging. One thing to note is that The original logic will try all keywrapper plugins, but current by default assume that only one keywrapper is configured.

ok, looking at the original golang code, this seems to be a 1:1 conversion. However, I wrote a test for that original preUnwrapKeys() fn and it doesn't seem to discard the error ("supposed to fail"):

```golang
type DummyKeyWrapper struct{}

func (dkw *DummyKeyWrapper) UnwrapKey(dc *config.DecryptConfig, annotation []byte) ([]byte, error) {
	if string(annotation) == "fail" {
		return nil, fmt.Errorf("supposed to fail")
	}
	return []byte("unwrapped_key"), nil
}

...

func TestPreUnwrapKey(t *testing.T) {
	kw := &DummyKeyWrapper{}
	// "fail,ok"
	// _, err := preUnwrapKey(kw, dc, "ZmFpbA==,b2s=")
	_, err := preUnwrapKey(kw, dc, "ZmFpbA==")
	if err != nil {
		t.Fatal(err)
	}
}
```

yields

```bash
$ go test ./encryption*.go
--- FAIL: TestPreUnwrapKey (0.00s)
    encryption_test.go:115: no suitable key found for decrypting layer key:
        - supposed to fail

FAIL
FAIL    command-line-arguments  0.005s
FAIL
```

mkulke:
the same on the caller. this will fail with "supposed to fail", since no keywapper was successful.

```golang
func TestDecryptLayerKeyOptsData(t *testing.T) {
	RegisterKeyWrapper("dummy", &DummyKeyWrapper{"dummy"})
	RegisterKeyWrapper("dummy2", &DummyKeyWrapper{"dummy2"})
	t.Log("Registered dummy key wrappers", keyWrapperAnnotations)
	annotations := map[string]string{
		// fail
		"org.opencontainers.image.enc.keys.dummy": "ZmFpbA==",
		// ok
		// "org.opencontainers.image.enc.keys.dummy2": "b2s=",
	}
	desc := ocispec.Descriptor{
		Annotations: annotations,
	}
	_, err := decryptLayerKeyOptsData(dc, desc)
	if err != nil {
		t.Fatal(err)
	}
}
```

mkulke:
yeah, so the error propagation is missing in the rust port, this is definitely a bug. So, if we want to keep true to the original error behaviour of the golang implementation (which I don't like, fwiw, I'd prefer to bail out early), we could just fix the error reporting. WDYT?

```patch
diff --git a/ocicrypt-rs/src/encryption.rs b/ocicrypt-rs/src/encryption.rs
index 3da2345..890831f 100644
--- a/ocicrypt-rs/src/encryption.rs
+++ b/ocicrypt-rs/src/encryption.rs
@@ -116,7 +116,7 @@ fn pre_unwrap_key(
 
         match keywrapper.unwrap_keys(dc, &annotation) {
             Err(e) => {
-                errs.push_str(&e.to_string());
+                errs.push_str(format!("- {}\n", e));
                 continue;
             }
             Ok(opts_data) => {
@@ -175,6 +175,7 @@ pub fn decrypt_layer_key_opts_data(
     annotations: Option<&HashMap<String, String>>,
 ) -> Result<Vec<u8>> {
     let mut priv_key_given = false;
+    let mut errs = vec![];
     let annotations = annotations.unwrap_or(&DEFAULT_ANNOTATION_MAP);
 
     for (annotations_id, scheme) in KEY_WRAPPERS_ANNOTATIONS.iter() {
@@ -188,16 +189,23 @@ pub fn decrypt_layer_key_opts_data(
                 priv_key_given = true;
             }
 
-            let opts_data = pre_unwrap_key(keywrapper, dc, &b64_annotation)?;
-            if !opts_data.is_empty() {
-                return Ok(opts_data);
+            match pre_unwrap_key(keywrapper, dc, &b64_annotation) {
+                Ok(opts_data) => {
+                    if !opts_data.is_empty() {
+                        return Ok(opts_data);
+                    }
+                }
+                Err(err) => errs.push(err),
             }
             // try next keywrapper
         }
     }
 
     if !priv_key_given {
-        return Err(anyhow!("missing private key needed for decryption"));
+        return Err(anyhow!(
+            "missing private key needed for decryption:\n {}",
+            errs.join("\n")
+        ));
     }
 
     Err(anyhow!(
```

Xynnn007:
@mkulke I think it is fine up to now as it is only used in CoCo for now and the promotion in the pr will work well.

If you like, we could use the temp var way as you did. Also raise an error when some error happens even if `Ok` returns.

arronwy (approved):
Thanks @mkulke LGTM!

PR #631, opened 2024-07-29
add vault support to secret-cli tool

fitzthum:
The secret-cli tool doesn't support vault secrets, which are currently the type that we are testing with Kata. There are a few reasons for this. For instance, creating a sealed vault secret is really just creating a simple json file and pushing the resource to the KBS/KMS (which may or may not be within scope of this tool). Even so we might as well add support for vault secrets.

I cleaned up a few things in the secret-cli tool, but there is still more to do.

Marking this PR as a draft until I add some documentation.


confidential-data-hub/docs/SEALED_SECRET.md
```diff
@@ -140,25 +173,30 @@ Then add a prefix `sealed.`
 sealed.ewoJInZlcnNpb24iOiAiMC4xLjAiLAoJInR5cGUiOiAiZW52ZWxvcGUiLAoJInByb3ZpZGVyIjogInh4eCIsCgkia2V5X2lkIjogInh4eCIsCgkiZW5jcnlwdGVkX2tleSI6ICJhYjI3ZGM9IiwgCgkiZW5jcnlwdGVkX2RhdGEiOiAieHh4IiwKCSJ3cmFwX3R5cGUiOiAiQTI1NkdDTSIsCgkiaXYiOiAieHh4IiwKCSJhbm5vdGF0aW9ucyI6IHsKCQkiY3J5cHRvX2NvbnRleHQiOiB7CgkJCSJhbGdvcml0aG0iOiAiQTI1NkdDTSIKCQl9LAoJCSJwcm92aWRlcl9zZXR0aW5nIjogewoJCQkia21zX2luc3RhbmNlX2lkIjogInh4eCIKCQl9Cgl9Cn0=
```
Xynnn007, line 173: I am  afraid that we need also to update this document part. Sealed secret format should be like `"sealed".<JWS header>.<JWS body, s.t. base64 payload>.<signature>`. But for now, `<JWS header>` and `<signature>` are ignored. See [example](https://github.com/kata-containers/kata-containers/pull/9719/files#diff-8bc9cea233848b1c184a4052fee2b24d8124539de2aa25c9efa3458b298ccb03R63)

confidential-data-hub/secret/src/bin/secret_cli.rs
```diff
@@ -37,7 +38,7 @@ enum Cli {
 struct SealArgs {
     /// path of the file which contains the content to be sealed
     #[arg(short, long)]
-    file_path: String,
+    file_path: Option<String>,
```
Xynnn007, line 41: I think the reason to mark this field to `Option`, is that we want `Vault` secrets are set separately rather than by this cli tool. If so, we can move this field to `EnvelopeCommand`, thus that only Envelope type would use this field.

btw, in the original design, we want even for `VaultSecret`s this tool will help to set secret to remote service, like what `Envelope` does. But surely, this seems making this tool code logic complex. Thus I am ok with this PR, s.t. leave the setting to separate vault separately.

Xynnn007 (approved):
Thanks Tobin, overall looks great to me.

ChengyuZhu6 (approved):
LGTM  thanks @fitzthum

confidential-data-hub/secret/src/bin/secret_cli.rs
fitzthum: Ah yeah I'll move that to the `EnvelopeCommand`.

I think we should enable provisioning secrets in the future, but that will take some work. I also think we might want to expose this functionality as part of a kubectl plugin.

confidential-data-hub/docs/SEALED_SECRET.md
fitzthum: Ah right. I updated the docs.

Btw do you think this secret-cli tool should output a json secret or the base64 format?

fitzthum:
I probably broke the `secret-cli.rs` test script by reshuffling some of the command line args. I will try to patch it up although I can't fully run those tests.

confidential-data-hub/docs/SEALED_SECRET.md
Xynnn007: The final aim is to output the format `"sealed".<JWS header>.<JWS body, s.t. base64 payload>.<signature>`. But currently, I am ok with both.

fitzthum:
This is probably enough for one PR. I will likely enable signatures soonish. Someone should revisit the HSM tests at some point.

Xynnn007 (approved):
lgtm  thanks!

PR #634, opened 2024-07-31
attestation-agent: Extend ResourceUri to support query string

cclaudio:
With this change, the kbs-client is able to send a get-resource request that includes the query string provided e.g.:

kbs-client --url http://127.0.0.1:8080 \
           get-resource \
           --path 'plugin/nebula/credential?ip[ip]=10.11.12.13&ip[netbits]=21&name=pod1'

This is required by the KBS repository plugin interface.

### DIFF
--- a/attestation-agent/deps/resource_uri/src/lib.rs
+++ b/attestation-agent/deps/resource_uri/src/lib.rs
@@ -22,6 +22,7 @@ pub struct ResourceUri {
     pub repository: String,
     pub r#type: String,
     pub tag: String,
+    pub query: Option<String>,
 }
 
 impl TryFrom<&str> for ResourceUri {
@@ -62,6 +63,7 @@ impl TryFrom<url::Url> for ResourceUri {
                 repository: values[0].into(),
                 r#type: values[1].into(),
                 tag: values[2].into(),
+                query: value.query().map(|s| s.to_string()),
             })
         } else {
             Err(RESOURCE_ID_ERROR_INFO)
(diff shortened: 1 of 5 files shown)


attestation-agent/deps/resource_uri/src/lib.rs
```diff
@@ -160,6 +167,23 @@ mod tests {
...
             tag: "213".into(),
+            query: None,
+        };
+        assert_eq!(expected, resource);
+    }
+
+    const TEST_QS_URL: &str = "kbs:///plugin/plugname/resourcename?param1=value1&param2=value2";
+
+    #[test]
+    fn deserialize_qs() {
```
Xynnn007, line 178: rstest could help to make the tests more dry. I am sure that you will love it : )

examples https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/attestation-agent/src/eventlog.rs#L158-L174

Xynnn007 (approved):
LGTM. Thanks @cclaudio for this!

fitzthum (approved):
LGTM

attestation-agent/deps/resource_uri/src/lib.rs
cclaudio: thanks, rstest definitely makes the code cleaner. Let me rework that test  case

cclaudio:
I will update this PR shortly with a edge case I forgot and also the test case clean-up

cclaudio:
Done.
- patched token_client.rs as well to support query string
- merged all ResourceUri tests into one rstest test with multiple cases. Thanks @Xynnn007  for the feedback.

PR #638, opened 2024-07-31
ci: increase open-pull-requests-limit from 1 to 3

arronwy:
the default value for open-pull-requests-limit is 5, to avoid block by some PR may need rebase source code, change the value from 1 to 3

Signed-off-by: Wang, Arron <arron.wang@intel.com>


ChengyuZhu6 (approved):
LGTM thanks @arronwy

Xynnn007 (approved):

PR #644, opened 2024-08-02
ci: Remove duplicate build when a PR is merged to main

arronwy:
Make CI only build when there is a PR, or a direct push to main.

### DIFF
--- a/.github/workflows/aa_basic.yml
+++ b/.github/workflows/aa_basic.yml
@@ -24,6 +24,7 @@ concurrency:
 
 jobs:
   basic_ci:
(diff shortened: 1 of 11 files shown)


Xynnn007 (approved):
LGTM. Thanks!

BbolroC:
Hmm.. it seems the redundant checks are still scheduled to run, and some of the checks where s390x is involved are gone in https://github.com/confidential-containers/guest-components/pull/647, which has rebased on main after this merge. I think something is going wrong here.

BbolroC:
The CI is behaving as this PR intended. Thanks.

PR #653, opened 2024-08-05
add reproducible build for CDH

zny666:
构建两次镜像，并生成两次容器，使用diffoscope比较后，二者无差异。


1570005763:
Two suggestions that are unrelated to the code.
- Use English so that everyone can understand.
- If you want to modify a particular PR, you can use `force push` to overwrite the commits that need to be changed, instead of opening a new PR.

PR #658, opened 2024-08-08
image-rs: update cosign signature verification unit test

Xynnn007:
This commit deletes a duplicated unit case for cosign signature scheme.

Also, the original cosign signed test image is based on busybox, thus a successfully pulling for that image (busybox) via nydus-snapshotter would cause another pulling with policy preventing the image successful. This is because of some logic inside nydus-snapshotter.

To handle this, we update the signed image to be a different one from busybox. Separately we will fix this on the snapshotter side.

cc @stevenhorsman


arronwy (approved):
Thanks @Xynnn007 LGTM!

PR #666, opened 2024-08-13
cdh: improves the luks-encrypt-storage script

wainersm:
While reviewing https://github.com/kata-containers/kata-containers/pull/9999 I wanted to understand confidential-data-hub/storage/scripts/luks-encrypt-storage better so I took a deep look at this file. I began to worry whether the logic to detect the blocks could fail or not, and in case of failure if it should continue the script as if everything went well (possible not cleaning up the block pages if any). Is it concern? So I decided to add a check for the block numbers, that if empty then the script bails out. Let me know if it is wrong and will introduce another issue that I could anticipate.

While in here, I delinted the script. 

Cc @Xynnn007 @ChengyuZhu6 @fitzthum

### DIFF
--- a/confidential-data-hub/storage/scripts/luks-encrypt-storage
+++ b/confidential-data-hub/storage/scripts/luks-encrypt-storage
@@ -99,15 +99,15 @@ if [[ -n "$device_name" && -b "$device_path" ]]; then
 		fi
 	fi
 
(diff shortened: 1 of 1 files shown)


ChengyuZhu6 (approved):
It makes sense to me. LGTM thanks for your improvement.

huoqifeng (approved):
LGTM
Thanks for harden the script! @wainersm

PR #667, opened 2024-08-14
AA: deletes `UpdateConfiguration` API and add `--initdata` launch parameter

Xynnn007:
This commit deletes `UpdateConfiguration` API for AA, also adds a new cmdline parameter `--initdata` for AA to check initdata when launching. The two changes are for initdata feature.

The deletion of `UpdateConfiguration` would help avoid a partial initialized AA.

### DIFF
--- a/attestation-agent/attestation-agent/src/bin/grpc-aa/main.rs
+++ b/attestation-agent/attestation-agent/src/bin/grpc-aa/main.rs
@@ -6,7 +6,8 @@
 mod server;
 
 use anyhow::*;
-use attestation_agent::AttestationAgent;
+use attestation_agent::{AttestationAPIs, AttestationAgent};
+use base64::Engine;
 use clap::Parser;
 use log::{debug, info};
 use tokio::signal::unix::{signal, SignalKind};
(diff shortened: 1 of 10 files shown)


Xynnn007:
Test passed.

attestation-agent/attestation-agent/src/bin/grpc-aa/main.rs
```diff
@@ -45,6 +54,23 @@ pub async fn main() -> Result<()> {
...
     let mut aa = AttestationAgent::new(cli.config_file.as_deref()).context("start AA")?;
+    if let Some(initdata) = cli.initdata {
+        info!("Initdata is given by parameter, try to check.");
+        let initdata = base64::engine::general_purpose::STANDARD
+            .decode(&initdata)
+            .context("base64 decode initdata")?;
+        let res = aa
+            .bind_init_data(&initdata)
+            .await
+            .context("check initdata")?;
```
fitzthum, line 65: This where the AA will crash if the initdata check does not match, right?

fitzthum (approved):
LGTM. I think there are a few more pieces needed for initdata in the longterm, like storing the initdata in the AA so we can give it to Trustee, but we can do this later.

huoqifeng (approved):
LGTM thanks! @Xynnn007

attestation-agent/attestation-agent/src/bin/grpc-aa/main.rs
Xynnn007: Thanks. Added the more expressive language.

PR #669, opened 2024-08-14
keyprovider: Pin a specific version of skopeo

fidencio:
The dependency on skopeo is quite fragile as there are several versions of the project that would generate an encrypted image with a gzip header that Confidential Containers simply cannot deal with.

For now, let's use f64a376, from Aug 14th 2024 as, empirically, we know that it works.

We must have tests implemented in order to cover the documentation we provide, and whether it works or not.

### DIFF
--- a/attestation-agent/docker/Dockerfile.keyprovider
+++ b/attestation-agent/docker/Dockerfile.keyprovider
@@ -26,7 +26,13 @@ RUN apt-get update && apt-get install -y \
 	pkg-config
 RUN git clone https://github.com/containers/skopeo $GOPATH/src/github.com/containers/skopeo 
 WORKDIR $GOPATH/src/github.com/containers/skopeo
-RUN git checkout v1.14.1
+# The dependency on skopeo is quite fragile as there are several versions of
+# the project that would generate an encrypted image with a gzip header that
+# Confidential Containers simply cannot deal with.
+#
+# For now, let's use f64a376, from Aug 14th 2024 as, empirically, we know that
+# it works.
+RUN git reset --hard f64a376
 ENV DISABLE_DOCS=1
 RUN make bin/skopeo
 RUN make install


attestation-agent/docker/Dockerfile.keyprovider
```diff
@@ -26,7 +26,13 @@ RUN apt-get update && apt-get install -y \
 	pkg-config
 RUN git clone https://github.com/containers/skopeo $GOPATH/src/github.com/containers/skopeo 
 WORKDIR $GOPATH/src/github.com/containers/skopeo
-RUN git checkout v1.14.1
+# The dependency on skopeo is quite fragile as there are several versions of
```
stevenhorsman, line 29: So was v1.14.1 working before and now isn't, or did we just not test it properly before, or are we not sure? I'm trying to work out if this pinned commit (which in theory is as stable as picking a released version?) might cause issues in future too?

fidencio: I simply don't know if it worked before, but it for sure contradicts with the documentation that says that skopeo version should be v1.14.2+.

I've retried several combinations of skopeo from v1.14.x, and also the latest v1.16.0 release, none of them worked for me.

May it cause issues in the future?  I am pretty sure it may and it will, unless we work on tests either on the skopeo side or on our side to ensure we'll keep everything working all the time.  But that's a rather big task that I'm not volunteering myself to work on.

wainersm: I had a hope that `skopeo inspect docker://ghcr.io/confidential-containers/test-container:multi-arch-encrypted` could give me information about the version of skopeo used, but nope.

Definitively we should be building the test image out of code (vs someone build and publish manually). There is the (https://github.com/kata-containers/kata-containers/issues/9360) tracking that work but unfortunately it didn't move forward.

Xynnn007 (approved):

mkulke (approved):

PR #678, opened 2024-08-16
AA: avoid creating AAEL if it is disabled

Xynnn007:
Before this commit, if we disabled eventlog recording in AA's config, there will still be a file `/run/attestation-agent/eventlog` but with no contents.

This file would be collected by tdx attester, treated as a valid AAEL. But on CoCo-AS side the AAEL parser will fail because nothing is in it.

This commit will avoid creating this file once we disabled in AA's config.

cc @ChengyuZhu6


ChengyuZhu6 (approved):
Thanks for your quick fix @Xynnn007 !

jialez0 (approved):

PR #681, opened 2024-08-19
AA: fix timeout when processing multiple incoming requests

imlk0:
The attestation-agent provided two binaries `aa-ttrpc` and `aa-grpc`. However, due to the existing of Mutex, we can't handle multiple `get_token()` and `get_evidence()` requests at the same time.

This commit removes Mutex from `attestation-agent/attestation-agent/src/bin/grpc-aa/server.rs` and `attestation-agent/attestation-agent/src/bin/ttrpc-aa/server.rs` to solve this problem.

### DIFF
--- a/attestation-agent/attestation-agent/src/lib.rs
+++ b/attestation-agent/attestation-agent/src/lib.rs
@@ -8,7 +8,7 @@ use async_trait::async_trait;
 use attester::{detect_tee_type, BoxedAttester};
 use kbs_types::Tee;
 use std::{io::Write, str::FromStr};
-use tokio::sync::Mutex;
+use tokio::sync::{Mutex, RwLock};
 
 pub use attester::InitdataResult;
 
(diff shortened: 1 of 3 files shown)


attestation-agent/attestation-agent/src/lib.rs
```diff
@@ -207,7 +214,9 @@ impl AttestationAPIs for AttestationAgent {
             operation,
             content,
         };
-        let alg = self.config.eventlog_config.eventlog_algorithm;
+        let alg = config.eventlog_config.eventlog_algorithm;
+        drop(config);
```
Xynnn007, line 218: ditto

Xynnn007 (commented):
Thanks for this fix and this would greatly promote the concurrent performance! some nits.

attestation-agent/attestation-agent/src/lib.rs
imlk0: optimized

attestation-agent/attestation-agent/src/lib.rs
```diff
@@ -106,7 +109,7 @@ impl AttestationAgent {
 
     /// Create a new instance of [AttestationAgent].
     pub fn new(config_path: Option<&str>) -> Result<Self> {
-        let config = match config_path {
+        let config = RwLock::new(match config_path {
```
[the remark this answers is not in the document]
imlk0: updated

Xynnn007 (approved):

Issue #604, opened 2024-07-02
image-rs: Whiteout code causing problems with image pull?


stevenhorsman:
In kata-containers we tried to bump image-rs to latest in (https://github.com/kata-containers/kata-containers/pull/9828#issuecomment-2197182995) and the pull-image failed on the agnhost:2.21 image. We had quite a bit of discussion in https://cloud-native.slack.com/archives/C039JSH0807/p1719475621870489 and found that the specific error was:
```
error="rpc error: code = Internal desc = failed to handle layer: hasher sha256: channel: send failed SendError { .. }; unpack error: failed to unpack convert whiteout\n\nCaused by:\n    Not supported (os error 95)" name=containerd-shim-v2 pid=40880 
```
We we tried the commit before https://github.com/confidential-containers/guest-components/pull/563 it seemed to work, so there are a few things we probable should do as a result of this:
- Revert the whiteout change until we can understand the problem
- I did try and re-produce the error with the image-rs integration tests and didn't manage it, so understanding that will be important too
- Improving the image-rs errors by adding extra `.context` to unwraps was crucial to debug this, so apply that sounds like a good approach

fitzthum:
cc @skaegi

squarti:
@stevenhorsman @skaegi  I believe the original code failed when running on platforms where xattr is not supported. PR above reintroduces the code on supported platforms.

squarti:
@stevenhorsman Do you know which OS these failures are happening? Curious to know whether it is an unsupported OS or the FS does not suppport xattr.

Xynnn007:
> @stevenhorsman Do you know which OS these failures were happening? Curious to know whether it is an unsupported OS or FS that does not support xattr.

I am afraid that this is because the previous whiteout PR will directly set `xattr` without detecting if current path supports xattr. For some cases, a tmpfs would not support xattr as it is compiled without enable specific feature. I am afraid that #687 also has this problem. We should do some change to align with https://github.com/confidential-containers/guest-components/pull/691 which fixes the `xattr` detecting.

squarti:
> I am afraid that this is because the previous whiteout PR will directly set `xattr` without detecting if current path supports xattr. For some cases, a tmpfs would not support xattr as it is compiled without enable specific feature. I am afraid that #687 also has this problem. We should do some change to align with #691 which fixes the `xattr` detecting.

Makes sense.  I will update #687 once #691 is merged.

Xynnn007:
#691 has been merged

PR #687, opened 2024-08-20
image-rs: Handle gzip whiteouts correctly

squarti:
This PR readds whiteout handling on platforms that support xattr

Fixes: #604

### DIFF
--- a/image-rs/src/unpack.rs
+++ b/image-rs/src/unpack.rs
@@ -7,6 +7,8 @@ use filetime::FileTime;
 use futures::StreamExt;
 use log::{debug, warn};
 use nix::libc::timeval;
+use nix::sys::stat::{mknod, Mode, SFlag};
+
 use std::{
     collections::HashMap,
     convert::TryInto,
@@ -80,10 +82,7 @@ async fn convert_whiteout(
         original_path.display()
     ))?;
 
-    let ret = unsafe { nix::libc::mknod(path.as_ptr(), nix::libc::S_IFCHR, 0) };
-    if ret != 0 {
-        bail!("mknod: {:?} error: {:?}", path, io::Error::last_os_error());
-    }
+    mknod(path.as_c_str(), SFlag::S_IFCHR, Mode::empty(), 0)?;
 
     set_perms_ownerships(&path, ChownType::LChown, uid, gid, mode).await
 }
--- a/image-rs/src/unpack.rs
+++ b/image-rs/src/unpack.rs
@@ -2,7 +2,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-use anyhow::{bail, Context, Result};
+use anyhow::{anyhow, bail, Context, Result};
 use filetime::FileTime;
 use futures::StreamExt;
 use log::{debug, warn};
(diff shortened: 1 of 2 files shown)


image-rs/src/unpack.rs
```diff
@@ -60,6 +130,10 @@ pub async fn unpack<R: AsyncRead + Unpin>(input: R, destination: &Path) -> Resul
             .try_into()
             .context("GID is too large!")?;
 
+        if !convert_whiteout(&file.path()?, file.header(), destination)? {
```
Xynnn007, line 133: I noticed that there are some duplicated logic between `convert_whiteout` and `unpack` after line 139. Could we do some refactoring upon this commit to avoid this?

bpradipt:
@squarti @Xynnn007 are you aware of any tools that helps to inspect a container image to figure out if it contains gzip whiteouts?

image-rs/src/unpack.rs
squarti: Yes, I will refactor it

squarti:
> @squarti @Xynnn007 are you aware of any tools that helps to inspect a container image to figure out if it contains gzip whiteouts? Or any sample image that can be used for testing this PR?

I am not aware of any tool.  I used this image `ghcr.io/actions/actions-runner:latest` to test it

image-rs/src/unpack.rs
squarti: @Xynnn007 I pushed some changes. Is that what you have in mind?

image-rs/src/unpack.rs
```diff
@@ -22,6 +22,60 @@ use std::{
...
+const WHITEOUT_PREFIX: &str = ".wh.";
+const WHITEOUT_OPAQUE_DIR: &str = ".wh..wh..opq";
+
+#[cfg(not(any(target_os = "linux", target_os = "freebsd", target_os = "macos")))]
+fn convert_whiteout(path: &Path, uid: u32, gid: u32, mode: Option<u32>, destination: &Path) -> Result<bool> {
+    Ok(true)
+}
+
+#[cfg(any(target_os = "linux", target_os = "freebsd", target_os = "macos"))]
+use anyhow::anyhow;
```
Xynnn007, line 34: Let's move this
```
use anyhow::anyhow;
```

directly inside `convert_whiteout`

image-rs/src/unpack.rs
```diff
@@ -22,6 +22,60 @@ use std::{
 use tokio::{fs, io::AsyncRead};
 use tokio_tar::ArchiveBuilder;
 
+const WHITEOUT_PREFIX: &str = ".wh.";
+const WHITEOUT_OPAQUE_DIR: &str = ".wh..wh..opq";
+
+#[cfg(not(any(target_os = "linux", target_os = "freebsd", target_os = "macos")))]
+fn convert_whiteout(path: &Path, uid: u32, gid: u32, mode: Option<u32>, destination: &Path) -> Result<bool> {
+    Ok(true)
```
Xynnn007, line 30: Probably we could use a shared `convert_whiteout` funciton. With `cfg-if` to firstly judge the platform and return `Ok(true)` if needed. Thus we do not need two `convert_whiteout` versions.

Xynnn007 (commented):
Thanks for the refactoring! Two last comments and other parts looks perfect to me

mythi:
> > @squarti @Xynnn007 are you aware of any tools that helps to inspect a container image to figure out if it contains gzip whiteouts? Or any sample image that can be used for testing this PR?
> 
> I am not aware of any tool. I used this image `ghcr.io/actions/actions-runner:latest` to test it

FWIW, I've built images with whiteouts using the distroless tooling in the past. In case it's useful..

Xynnn007 (commented):
One last thing. Sorry I did not make it clear

Xynnn007 (approved):
good. A rebase is needed

image-rs/src/unpack.rs
```diff
@@ -39,6 +39,57 @@ fn is_attr_available(path: &Path) -> Result<bool> {
...
+            destination.display(),
+            original_path.display()
+        ))?;
+
+        let ret = unsafe { nix::libc::mknod(path.as_ptr(), nix::libc::S_IFCHR, 0) };
+        if ret != 0 {
+            bail!("mknod: {:?} error: {:?}", path, io::Error::last_os_error());
+        }
+
+        set_perms_ownerships(&path, ChownType::LChown, uid, gid, mode).await?;
```
Xynnn007, line 85: Do we need also to `set_file_times()` here?

squarti: Good question.  Looking into the golang base implementation, `set_file_times` is not done. 

The purpose of the whiteout file is to hide the original file. Even if we update the file times, it is not going to be visible in the file system.

fitzthum (commented):
Nice. A few comments. I guess this is mainly being ported from the Go code, but I'd like to understand a little better what's happening.

fitzthum (approved):
This `mknod` thing is still a bit mysterious to me, but I guess we can live with it.

image-rs/src/unpack.rs
```diff
@@ -39,6 +39,55 @@ fn is_attr_available(path: &Path) -> Result<bool> {
...
+        .strip_prefix(WHITEOUT_PREFIX)
+        .ok_or(anyhow!("Failed to strip whiteout prefix for: {}", name))?;
+    let original_path = parent.join(original_name);
+    let path = CString::new(format!(
+        "{}/{}",
+        destination.display(),
+        original_path.display()
+    ))?;
+
+    let ret = unsafe { nix::libc::mknod(path.as_ptr(), nix::libc::S_IFCHR, 0) };
```
Xynnn007, line 83: I am not sure we use `nix` here, s.t. https://docs.rs/nix/0.29.0/nix/sys/stat/fn.mknod.html  If we can, we will get code w/o `unsafe` parts

squarti: I am not sure this is what you are after.

fitzthum:
Thanks @squarti 

Let's keep an eye on image unpacking when we next  bump guest-components just in case.

PR #691, opened 2024-08-23
image-rs: check xattrs for target dir when image unpacking

### LINKED ISSUE #690: image-rs | Failed to set xattr for tmpfs
Thanks to @bpradipt , when we pull an image to tmpfs that does not support xattr, e.g. [quay.io/bpradipt/hello-ocp:no-long-link](http://quay.io/bpradipt/hello-ocp:no-long-link)

Whether xattr is supported by a mount path can be checked by `mount | grep tmpfs`. If there is no `xattr` inside the `(...)`, it does not supports.

failures:

---- pull::tests::test_async_pull_client stdout ----
Got error on function call attempt 0. Will retry in 1s: failed to handle layer: hasher sha256: failed to unpack `/tmp/.tmpbPSr5B/sha256_fda7f57d59e2441f415e2f05ff824b63f0d64d521f0afe0c1f10e09f5c30f22c/var/lib/dnf/history.sqlite`

Caused by:
    0: failed to set extended attributes to /tmp/.tmpbPSr5B/sha256_fda7f57d59e2441f415e2f05ff824b63f0d64d521f0afe0c1f10e09f5c30f22c/var/lib/dnf/history.sqlite. Xattrs: key="user.overlay.origin", value="".
    1: Operation not supported (os error 95)
Got error on function call attempt 1. Will retry in 1s: failed to handle layer: hasher sha256: failed to unpack `/tmp/.tmpbPSr5B/sha256_fda7f57d59e2441f415e2f05ff824b63f0d64d521f0afe0c1f10e09f5c30f22c/var/lib/dnf/history.sqlite`

Xynnn007:
Some platforms/file-systems do not support xattrs, this would make the image pull fail because of failing to set xattr. This commit will check whether the target path supports xattr. If yes, the unpacking will maintain xattrs; if not, it will not set xattrs.

Fixes #690

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2783,6 +2783,7 @@ dependencies = [
  "ttrpc-codegen",
  "url",
  "walkdir",
+ "xattr",
  "zstd 0.12.4",
 ]
 
--- a/image-rs/Cargo.toml
+++ b/image-rs/Cargo.toml
@@ -61,6 +61,7 @@ tonic = { workspace = true, optional = true }
 ttrpc = { workspace = true, features = ["async"], optional = true }
 url = "2.5.2"
 walkdir = "2"
+xattr = "1"
 zstd = "0.12"
 
 nydus-api = { version = "0.3.0", optional = true }
--- a/image-rs/src/unpack.rs
+++ b/image-rs/src/unpack.rs
@@ -5,7 +5,7 @@
 use anyhow::{bail, Context, Result};
 use filetime::FileTime;
 use futures::StreamExt;
-use log::warn;
+use log::{debug, warn};
 use nix::libc::timeval;
 use std::{
     collections::HashMap,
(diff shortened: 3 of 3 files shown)


Xynnn007:
cc @bpradipt

Xynnn007:
oh, this fails to check xattr correctly. please do not review for now

bpradipt:
> oh, this fails to check xattr correctly. please do not review for now

@Xynnn007 I was trying out something on the following lines. Seems to work
```
diff --git a/image-rs/src/unpack.rs b/image-rs/src/unpack.rs
index 126369c..bc814b6 100644
--- a/image-rs/src/unpack.rs
+++ b/image-rs/src/unpack.rs
@@ -12,6 +12,7 @@ use std::{
     convert::TryInto,
     ffi::CString,
     fs::Permissions,
+    fs::File,
     io,
     os::{
         fd::{AsFd, AsRawFd},
@@ -21,12 +22,33 @@ use std::{
 };
 use tokio::{fs, io::AsyncRead};
 use tokio_tar::ArchiveBuilder;
+use xattr::FileExt;

 /// Unpack the contents of tarball to the destination path
 pub async fn unpack<R: AsyncRead + Unpin>(input: R, destination: &Path) -> Result<()> {
+    let is_xattrs_supported = destination.metadata().map(|meta| {
+        if meta.is_dir() {
+            // Try setting and removing a test xattr on the directory itself
+            let key = "user.test";
+            let value = b"test";
+            if let Ok(file) = File::open(destination) {
+                let set_result = file.set_xattr(key, value);
+                let remove_result = file.remove_xattr(key);
+
+                // Check if both setting and removing xattrs worked
+                set_result.is_ok() && remove_result.is_ok()
+            } else {
+                false
+            }
+        } else {
+            false
+        }
+
+    }).unwrap_or(false);
+
     let mut archive = ArchiveBuilder::new(input)
         .set_ignore_zeros(true)
-        .set_unpack_xattrs(true)
+        .set_unpack_xattrs(is_xattrs_supported)
         .set_preserve_permissions(true)
         .build();
     let mut entries = archive
```

Xynnn007:
@bpradipt I test locally but has false negative. Need to inspect more

Xynnn007:
In a dir that supports `xattr`, the `file.set_xattr(key, value)` would say `Operation not supported`. But the `set_xattr` inside `krata-tokio-tar` will work.

Xynnn007:
Now it works with `key` starting with `user.overlay.origin`

fitzthum (approved):
Two nits but looks fine assuming you've tested it out. Might be good to have a unit test, but I'm not sure how to do that without some messing around.

Xynnn007:
> Two nits but looks fine assuming you've tested it out. Might be good to have a unit test, but I'm not sure how to do that without some messing around.

I have test myself locally in both dir that supports xattrs and not. It is hard to do a unit test as `mount` on a platform could be compiled without `xattr` support, thus we might not to mount both xattr dir and non-xattr dir.

Xynnn007:
@bpradipt Thanks for the right direction. PTAL if this now works

bpradipt (approved):
/lgtm
thanks @Xynnn007

PR #695, opened 2024-08-23
docs: add coco_keyprovider to tools section

fitzthum:
It always takes me a while to find the CoCo keyprovider since it is in the attestation-agent dir. Let's add it to the tools section, which lists all the tools provided in this repo. We might also think about moving the keyprovider itself at some point.


Xynnn007 (approved):
lgtm, thanks!

PR #710, opened 2024-09-06
AA: use enhanced version of vtpm get report for ITA

pawelpros:
Changed evidence generation method which will include report_data in re-generated HCL report (azTdxVtpm)

Signed-off-by: Pawel Proskurnicki <pawel.proskurnicki@intel.com>

### DIFF
--- a/attestation-agent/attester/src/az_tdx_vtpm/mod.rs
+++ b/attestation-agent/attester/src/az_tdx_vtpm/mod.rs
@@ -34,7 +34,7 @@ struct Evidence {
 #[async_trait::async_trait]
 impl Attester for AzTdxVtpmAttester {
     async fn get_evidence(&self, report_data: Vec<u8>) -> Result<String> {
-        let hcl_report_bytes = vtpm::get_report()?;
+        let hcl_report_bytes = vtpm::get_report_with_report_data(&report_data)?;
         let hcl_report = hcl::HclReport::new(hcl_report_bytes.clone())?;
         let td_report = hcl_report.try_into()?;
         let td_quote_bytes = imds::get_td_quote(&td_report)?;


attestation-agent/kbs_protocol/src/client/rcar_client.rs
```diff
@@ -193,8 +193,12 @@ impl KbsClient<Box<dyn EvidenceProvider>> {
         nonce: String,
     ) -> Result<String> {
         debug!("Challenge nonce: {nonce}");
-        let mut hasher = Sha384::new();
-        hasher.update(runtime_data);
+        let mut hasher: Box<dyn DynDigest + Send> = match tee {
+            Tee::AzTdxVtpm => Box::new(Sha256::new()),
+            _ => Box::new(Sha384::new()),
+        };
```
mkulke, line 199: hmm, afaik sha384 is hardcoded [in CoCo AS](https://github.com/confidential-containers/trustee/blob/b7defdada4909f70b860285b752b0fe1f5c55e60/attestation-service/src/token/simple.rs#L23). I understand that ITA needs this to be sha256/sha512 in this case, but do we need changes on the protocol level to support both variants?

mythi: The `rcar_client.rs` changes in this PR will be dropped in the next push. This will be handled as part of https://github.com/confidential-containers/trustee/issues/242 work landing soon.

pawelpros: Reverted changes from rcar_client

mythi (approved):

mkulke (approved):

PR #711, opened 2024-09-06
Rust 1.81 fixes

mythi:
The `stable` toolchain in our CI isn't that good choice after all, since it can change over night. Anyway, this unblocks CI again and gets everything running w/ Rust 1.81.


api-server-rest/build.rs
```diff
@@ -101,6 +118,17 @@ fn main() -> std::io::Result<()> {
         .run()
         .expect("Generate ttrpc protocol code failed.");
 
+    // Fix clippy warnings of code generated from ttrpc_codegen
+    replace_text_in_file(
+        "src/ttrpc_proto/attestation_agent_ttrpc.rs",
+        "#![allow(box_pointers)]\n",
```
Xynnn007, line 124: btw I have been fixing this upstream https://github.com/containerd/ttrpc-rust/pull/235. But seem it would take a long time so we need this PR.

Xynnn007 (approved):

arronwy (approved):
Thanks @mythi LGTM!

PR #722, opened 2024-09-16
release: Publish vendored code

fidencio:
Currently we're not publishing the vendored code as part of the release, which leads to all the downstream which do not have access to internet whilst building their components to do this process manually.

Let's be nice to the downstreams, let's publish the vendors ourselves.

Here's one release that I cut, on my own branch, showing the result of this action: https://github.com/fidencio/guest-components/releases/tag/v0.0.0-test-vendored-tarball

### DIFF
--- a/.github/workflows/vendor_release.yml
+++ b/.github/workflows/vendor_release.yml
@@ -1,0 +1,24 @@
+name: Generate vendored content for the release
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  generate-and-publish-vendored-code:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Generate and publish cargo vendored tarball
+        run: |
+          tag=$(echo $GITHUB_REF | sed 's|refs/tags/||')
+          tarball_name="guest-components-${tag}-vendor.tar.gz"
+          mkdir -p .cargo
+          cargo vendor --locked >> .cargo/config.toml
+
+          tar -cvzf ${tarball_name} vendor .cargo/config.toml 
+          gh release upload ${tag} ${tarball_name}
+        env:
+          GH_TOKEN: ${{ github.token }}


fidencio:
There's no need to have it merged and then cut a 0.10.1, as we can easily update the vendored code manually for v0.10.0.

.github/workflows/vendor_release.yml
```diff
@@ -0,0 +1,31 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Generate cargo vendored tarball
+        run: |
+          tag=$(echo $GITHUB_REF | sed 's|refs/tags/||')
+          tarball_name="guest-components-${tag}-vendor.tar.gz"
+          vendor_dir_list=""
+          for i in $(find . -name 'Cargo.lock'); do
```
mkulke, line 19: SInce this is a workspace, shouldn't there be exactly 1 Cargo.lock file in the project root?

mkulke: I'd assume it's unlikely that we would revert it back to multiple individual projects. However what we might want to cover is different feature toggles, or would the default features be sufficient?

fidencio: > However what we might want to cover is different feature toggles, or would the default features be sufficient?

That's a very good question, and I would start with what I proposed, as https://doc.rust-lang.org/cargo/commands/cargo-vendor.html does not mention anything about specific features.

.github/workflows/vendor_release.yml
```diff
@@ -0,0 +1,24 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Generate and publish cargo vendored tarball
+        run: |
+          tag=$(echo $GITHUB_REF | sed 's|refs/tags/||')
+          tarball_name="guest-components-${tag}-vendor.tar.gz"
+          [ -d .cargo ] || mkdir .cargo
+          cargo vendor >> .cargo/config.toml
```
mkulke, line 19: `mkdir -p` , but I assume we do want to abort if that folder exists, otherwise we'd archive unexpected stuff.

`cargo vendor --locked` probably makes sense here?

fidencio: https://github.com/fidencio/guest-components/releases/tag/v0.0.2-test-vendored-tarball

mkulke (approved):

dcmiddle (approved):

PR #712, opened 2024-09-06
AA: kbs: Add supported hash algorithms to Request

jodh-intel:
Add a `supported-hash-algorithms` list to the `Request` to allow the
returned `Challenge` to select an appropriate TEE-specific algorithm
from the list.

Signed-off-by: James O. D. Hunt <james.o.hunt@intel.com>

PR #723, opened 2024-09-16
AA: kbs: Improve handling of invalid RCAR JSON

jodh-intel:
Create a new error to handle the scenario where the RCAR JSON specifies the selected hash algorithm
using the wrong data type (any type other than 'string').

### DIFF
--- a/attestation-agent/kbs_protocol/src/client/rcar_client.rs
+++ b/attestation-agent/kbs_protocol/src/client/rcar_client.rs
@@ -56,6 +56,26 @@ async fn get_request_extra_params() -> serde_json::Value {
     extra_params
 }
 
+fn get_hash_algorithm(extra_params: serde_json::Value) -> Result<HashAlgorithm> {
+    let algorithm = match extra_params.get(SELECTED_HASH_ALGORITHM_JSON_KEY) {
+        Some(selected_hash_algorithm) => {
+            let name = selected_hash_algorithm
+                .as_str()
+                .ok_or(Error::UnexpectedJSONDataType(
+                    "string".into(),
+                    selected_hash_algorithm.to_string(),
+                ))?
+                .to_lowercase();
+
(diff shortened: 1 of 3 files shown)


jodh-intel:
This PR addresses @mkulke's comment: https://github.com/confidential-containers/guest-components/pull/712#discussion_r1751765402.

mkulke (approved):

Xynnn007 (approved):

attestation-agent/kbs_protocol/src/client/rcar_client.rs
```diff
@@ -56,6 +56,26 @@ async fn get_request_extra_params() -> serde_json::Value {
     extra_params
 }
 
+async fn get_hash_algorithm(extra_params: serde_json::Value) -> Result<HashAlgorithm> {
```
Xynnn007, line 59: we might not need this to be `async`

jodh-intel: @Xynnn007 - true, this function doesn't actually _need_ to be async, but I feel the code is cleaner + safer if we use async everywhere we can reasonably do so.

If I drop the `async`, I either:

- block the calling task implicitly (maybe not an issue now, but potentially problematic if the implementation changes (future bug potential)).
- block it explicitly with `spawn_blocking()` (extra code noise for no obvious benefit).

Xynnn007: My personal preference is to mark it as non-async. Once it needs to be async, we can then mark it as async version.

btw, if a function is a concrete implementation of an async-trait function, I can live with situations where even no async code is inside the function body.

wdyt?

jodh-intel: I think we've spent enough time discussing a 5 byte change. Branch updated :smile:

PR #726, opened 2024-09-26
Fix configuration file default value and make error information more detailed

Xynnn007:
Please see the commit message of this PR.

cc @stevenhorsman

### DIFF
--- a/attestation-agent/attester/src/sgx_dcap/mod.rs
+++ b/attestation-agent/attester/src/sgx_dcap/mod.rs
@@ -68,7 +68,7 @@ impl Attester for SgxDcapAttester {
                     report_data.as_ptr() as *const sgx_report_data_t,
                 ) {
                     Ok(_) => occlum_quote,
-                    Err(e) => bail!("generate quote: {e}"),
+                    Err(e) => bail!("generate quote: {e:?}"),
                 }
             }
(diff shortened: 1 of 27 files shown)


mkulke (commented):
thanks @Xynnn007 for looking into this so quickly, can we maybe also as a drive-by fix make a missing `credentials` setting in the CDH conf default to `[]`?

at the moment users have to specify explicitly, even if a user does not use the feature.

```
credentials = []
```

mythi:
> This default value of KBS cert would make the KBS client to read it as a PEM cert, thus always fails to launch. 

The unit tests probably need updating to get this path covered?

Xynnn007:
> at the moment users have to specify explicitly, even if a user does not use the feature.
> 
> ```
> credentials = []
> ```

Without this section it could be fine. See the unit test for config example https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/hub/src/config.rs#L249-L261



> The unit tests probably need updating to get this path covered?

Updated. See https://github.com/confidential-containers/guest-components/pull/726/files#diff-51b4317c4f4b87e6825be8fe3761800028d718fff2c48542e78904a6fc82aba1R303

mkulke (approved):

PR #727, opened 2024-09-27
CDH | Fix ttrpc memory bug and gRPC lock bug

### LINKED ISSUE #688: CDH | Multiple CDH instances in ttrpc CDH process
Currently we are using a [macro](https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/hub/src/bin/ttrpc-cdh.rs#L45-L51) to create structures that handles different API of CDH. Because of the limit of ttrpc, we should create a separate `Server` struct for every server, this would occupy 4~ memory than only one `Server` struct.

Let's fix this by multiplex one `Server` struct after https://github.com/containerd/ttrpc-rust/pull/235 gets merged.

Xynnn007:
Please see the commit message for details.

Fixes #688


mythi:
this PR could also undo/revert e3c3fef, right?

Xynnn007:
> this PR could also undo/revert [e3c3fef](https://github.com/confidential-containers/guest-components/commit/e3c3fefd4caabd2c614cb5f5bca574863aae400e), right?

right. Added the commit

attestation-agent/attestation-agent/src/bin/ttrpc-aa/ttrpc_protocol/attestation_agent.rs
```diff
@@ -9,7 +9,6 @@
 #![allow(unused_attributes)]
 #![cfg_attr(rustfmt, rustfmt::skip)]
 
-#![allow(box_pointers)]
```
mythi, line 12: do we have a code path in CI that is not built since we did not catch this part earlier?

Xynnn007: I am afraid not. We should turn on `ttrpc` feature in the ci also to cover this. Let me do a fix in this PR also.

fitzthum (approved):
LGTM

Xynnn007:
Hey @mythi Could you help to get another look?

ChengyuZhu6 (approved):
LGTM thanks @Xynnn007

mythi (approved):
LGTM. I wonder if rebasing is needed to get a clean merge (OTOH, github does not complain about merge conflicts so I guess it's good)

PR #728, opened 2024-09-27
attestation-agent: default to ttrpc in Makefile

mkulke:
CDH already does this and ASR doesn't work with GRPC. This change shouldn't break any existing invocations of the makefile with ttrpc=true, only builds that rely on grpc being implicitly picked. Due to the CDH's default and ASR being ttrpc-only I don't think we break anything downstream.


Xynnn007 (approved):

ChengyuZhu6 (approved):
It makes sense to me. Thanks @mkulke

fitzthum (approved):
LGTM.

PR #731, opened 2024-09-29
CI: Publish binaries with ORAS

mkulke:
Pushing artifacts as binaries to the project's GHCR. The build job is split between AA and CDH+ASR. AA has specific build and runtimerequirements depending on the TEE, while the CDH+ASR are generic per arch.

Hence $AA is tagged with $sha-$tee ($arch is implicit in $tee) while CDH+ASR are tagged with $sha-$arch.

## Why

Peerpod as a downstream project currently builds GC binaries as part of their pipeline. This is rather costly, the project is written in golang and needs to maintain a complex rust build infra for GC that often drifts in terms of rust-versions, feature-toggles and makes it hard to keep up to date with recent GC. There's also considerable overhead in terms of build-time.

It makes sense to have canonical cache of build-artifacts that can be consumed downstream.

### DIFF
--- a/.github/actions/install-intel-dcap/action.yml
+++ b/.github/actions/install-intel-dcap/action.yml
@@ -9,7 +9,7 @@ runs:
     id: install-tdx-dependencies
     shell: bash
     run: |
-      sudo curl -sL https://download.01.org/intel-sgx/sgx_repo/ubuntu/intel-sgx-deb.key | sudo gpg --dearmor --output /usr/share/keyrings/intel-sgx.gpg
-      sudo echo 'deb [arch=amd64 signed-by=/usr/share/keyrings/intel-sgx.gpg] https://download.01.org/intel-sgx/sgx_repo/ubuntu ${{ inputs.ubuntu-version }} main' | sudo tee /etc/apt/sources.list.d/intel-sgx.list
+      curl -sL https://download.01.org/intel-sgx/sgx_repo/ubuntu/intel-sgx-deb.key | sudo gpg --dearmor --output /usr/share/keyrings/intel-sgx.gpg
+      echo 'deb [arch=amd64 signed-by=/usr/share/keyrings/intel-sgx.gpg] https://download.01.org/intel-sgx/sgx_repo/ubuntu ${{ inputs.ubuntu-version }} main' | sudo tee /etc/apt/sources.list.d/intel-sgx.list
       sudo apt-get update
       sudo apt-get install -y --no-install-recommends libtdx-attest-dev
--- a/.github/actions/install-intel-dcap/action.yml
+++ b/.github/actions/install-intel-dcap/action.yml
@@ -1,0 +1,15 @@
+name: 'Install Intel DCAP libraries'
+inputs:
+  ubuntu-version:
+    required: true
+runs:
+  using: "composite"
+  steps:
+  - name: Install TDX dependencies
+    id: install-tdx-dependencies
+    shell: bash
+    run: |
+      sudo curl -sL https://download.01.org/intel-sgx/sgx_repo/ubuntu/intel-sgx-deb.key | sudo gpg --dearmor --output /usr/share/keyrings/intel-sgx.gpg
+      sudo echo 'deb [arch=amd64 signed-by=/usr/share/keyrings/intel-sgx.gpg] https://download.01.org/intel-sgx/sgx_repo/ubuntu ${{ inputs.ubuntu-version }} main' | sudo tee /etc/apt/sources.list.d/intel-sgx.list
+      sudo apt-get update
+      sudo apt-get install -y --no-install-recommends libtdx-attest-dev
--- a/.github/workflows/publish-artifacts.yml
+++ b/.github/workflows/publish-artifacts.yml
@@ -1,0 +1,186 @@
+name: Publish artifacts to ORAS
+
+on:
+  push:
+    branches:
+    - main
+
+env:
+  RUST_TOOLCHAIN: 1.76.0
+
+jobs:
+  publish-aa:
+    permissions:
+      contents: read
+      packages: write
+      id-token: write
+    strategy:
+      matrix:
+        tee:
+        - none
+        - amd
+        - az-cvm-vtpm
+        - tdx
+        - se
+        - cca
+        arch:
+        - x86_64
+        - s390x
+        exclude:
+        - tee: amd
+          arch: s390x
+        - tee: az-cvm-vtpm
+          arch: s390x
+        - tee: tdx
+          arch: s390x
+        - tee: se
+          arch: x86_64
+        - tee: cca
+          arch: s390x
+        include:
+        - tee: none
+          arch: x86_64
+          libc: musl
+        - tee: none
+          arch: s390x
+          libc: gnu
+        - tee: amd
+          arch: x86_64
+          libc: musl
+        - tee: az-cvm-vtpm
+          arch: x86_64
+          libc: gnu
+        - tee: tdx
+          arch: x86_64
+          libc: gnu
+        - tee: se
+          arch: s390x
+          libc: gnu
+        - tee: cca
+          arch: x86_64
+          libc: musl
+    runs-on: ${{ matrix.arch == 's390x' && 's390x' || 'ubuntu-22.04' }}
+    env:
+      TEE_PLATFORM: ${{ matrix.tee }}
+      LIBC: ${{ matrix.libc }}
+      REGISTRY: ghcr.io
+      IMAGE_NAME: ${{ github.repository }}
+    steps:
+    - name: Log in to the Container registry
+      uses: docker/login-action@65b78e6e13532edd9afa3aa52ac7964289d1a9c1
+      with:
+        registry: ${{ env.REGISTRY }}
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+
+    - uses: oras-project/setup-oras@v1
+      with:
+        version: 1.2.0
+
+    - uses: actions/checkout@v4
+
+    - uses: actions-rust-lang/setup-rust-toolchain@v1
+      with:
+        toolchain: ${{ env.RUST_TOOLCHAIN }}
+        target: ${{ matrix.arch }}-unknown-linux-${{ matrix.libc }}
+        override: true
+
+    - name: Install tpm dependencies
+      if: matrix.tee == 'az-cvm-vtpm'
+      run: |
+        sudo apt-get install -y --no-install-recommends libtss2-dev
+
+    - name: Install tdx dependencies
+      if: matrix.tee == 'tdx'
+      run: |
+        sudo curl -sL https://download.01.org/intel-sgx/sgx_repo/ubuntu/intel-sgx-deb.key | sudo gpg --dearmor --output /usr/share/keyrings/intel-sgx.gpg
+        sudo echo 'deb [arch=amd64 signed-by=/usr/share/keyrings/intel-sgx.gpg] https://download.01.org/intel-sgx/sgx_repo/ubuntu jammy main' | sudo tee /etc/apt/sources.list.d/intel-sgx.list
+        sudo apt-get update
+        sudo apt-get install -y --no-install-recommends libtdx-attest-dev
+
+    - uses: actions/checkout@v4
+
+    - name: Build
+      run: make ./target/${{ matrix.arch }}-unknown-linux-${{ matrix.libc}}/release/attestation-agent
+
+    - name: Publish to ORAS
+      env:
+        ARCH: ${{ matrix.arch == 'x86_64' && 'amd64' || matrix.arch }}
+      run: |
+        mkdir oras
+        cd oras
+        cp ../target/${{ matrix.arch }}-unknown-linux-${{ matrix.libc}}/release/attestation-agent .
+        tar cJf attestation-agent.tar.xz attestation-agent
+        arch_tag="${{ github.sha }}-${{ matrix.tee }}_${{ matrix.arch }}"
+        image="${REGISTRY}/${IMAGE_NAME}/attestation-agent"
+        tag="${{ github.sha }}-${{ matrix.tee }}"
+        arch_tag="${tag}_${ARCH}"
+        oras push "${image}:${arch_tag}" attestation-agent.tar.xz
+        # We need to create the platform annotations with docker, since oras 1.2 doesn't support
+        # pushing with platform yet.
+        docker manifest create "${image}:${tag}" --amend "${image}:${arch_tag}"
+        docker manifest annotate --arch "$ARCH" --os linux "${image}:${tag}" "${image}:${arch_tag}"
+        docker manifest push "${image}:${tag}"
+
+  publish-cdh-and-asr:
+    permissions:
+      contents: read
+      packages: write
+      id-token: write
+    strategy:
+      matrix:
+        arch:
+        - x86_64
+        - s390x
+        include:
+        - arch: x86_64
+          libc: musl
+        - arch: s390x
+          libc: gnu
+    runs-on: ${{ matrix.arch == 's390x' && 's390x' || 'ubuntu-22.04' }}
+    env:
+      LIBC: ${{ matrix.libc }}
+      REGISTRY: ghcr.io
+      IMAGE_NAME: ${{ github.repository }}
+    steps:
+    - name: Log in to the Container registry
+      uses: docker/login-action@65b78e6e13532edd9afa3aa52ac7964289d1a9c1
+      with:
+        registry: ${{ env.REGISTRY }}
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+
+    - uses: oras-project/setup-oras@v1
+      with:
+        version: 1.2.0
+
+    - uses: actions-rust-lang/setup-rust-toolchain@v1
+      with:
+        toolchain: ${{ env.RUST_TOOLCHAIN }}
+        target: ${{ matrix.arch }}-unknown-linux-${{ matrix.libc }}
+        override: true
+
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y --no-install-recommends \
+          libdevmapper-dev \
+          protobuf-compiler
+
+    - uses: actions/checkout@v4
+
+    - name: Build CDH
+      run: make ./target/${{ matrix.arch }}-unknown-linux-${{ matrix.libc}}/release/confidential-data-hub
+
+    - name: Build ASR
+      run: make ./target/${{ matrix.arch }}-unknown-linux-${{ matrix.libc}}/release/api-server-rest
+
+    - name: Publish to ORAS
+      run: |
+        mkdir oras
+        cd oras
+        cp ../target/${{ matrix.arch }}-unknown-linux-${{ matrix.libc}}/release/{confidential-data-hub,api-server-rest} .
+        tar cJf confidential-data-hub.tar.xz confidential-data-hub
+        tar cJf api-server-rest.tar.xz api-server-rest
+        oras push ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}/confidential-data-hub:${{ github.sha }}-${{ matrix.arch }} confidential-data-hub.tar.xz
+        oras push ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}/api-server-rest:${{ github.sha }}-${{ matrix.arch }} api-server-rest.tar.xz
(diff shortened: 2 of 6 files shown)


.github/workflows/publish-artifacts.yml
```diff
@@ -0,0 +1,174 @@
...
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+
+    - uses: oras-project/setup-oras@v1
+      with:
+        version: 1.0.0
+
+    - uses: actions/checkout@v4
+
+    - uses: actions-rust-lang/setup-rust-toolchain@v1
```
mythi, line 72: Would it be possible we agree on one approach to get the toolchain installed and use that. This is the third approach in this repo alone and enclave-cc also uses `dtolnay/rust-toolchain`.

.github/workflows/publish-artifacts.yml
```diff
@@ -0,0 +1,174 @@
...
+      run: |
+        sudo apt-get install -y --no-install-recommends libtss2-dev
+
+    - name: Install tdx dependencies
+      if: matrix.tee == 'tdx'
+      run: |
+        sudo curl -sL https://download.01.org/intel-sgx/sgx_repo/ubuntu/intel-sgx-deb.key | sudo gpg --dearmor --output /usr/share/keyrings/intel-sgx.gpg
+        sudo echo 'deb [arch=amd64 signed-by=/usr/share/keyrings/intel-sgx.gpg] https://download.01.org/intel-sgx/sgx_repo/ubuntu jammy main' | sudo tee /etc/apt/sources.list.d/intel-sgx.list
+        sudo apt-get update
+        sudo apt-get install -y --no-install-recommends libtdx-attest-dev
```
mythi, line 90: we've started to have the vTPM/DCAP installation copied in many places so perhaps we could make them reusable in workflows using 
```suggestion
      uses: "./.github/workflows/lib-install-dcap.yaml"
```

mkulke: oof, I don't know. I'd personally prefer some duplication instead of excessive abstraction and people having to jump through workflows to replicate what's happening. for installing tpm headers this would be definitely overkill but having a reusable wf for the tdx dep might be tolerable. I suppose such a workflow needs to be parameterized per ubuntu distro?

mkulke: ah, wait. that doesn't make sense. a workflow will run in its own worker machine, so you'd be installing packages in the child wf , but it wouldn't be available in the calling workflow

.github/workflows/publish-artifacts.yml
mythi: we have that tracked in https://github.com/confidential-containers/confidential-containers/issues/230 but what would be the best option to use?

.github/workflows/publish-artifacts.yml
mkulke: but we can use a "composite" action for sharing steps, added a commit.

mythi: I have no strong opinions. My comment was based on the fact that the DCAP installation is a bit broken and I was hoping to do the fixing in one place only. OTOH, `libtdx-attest-dev` is eventually going away so maybe I'll just wait for that to happen instead..

.github/workflows/publish-artifacts.yml
```diff
@@ -0,0 +1,173 @@
+name: Publish artifacts to ORAS
+
+on:
+  push:
+    branches:
+    - main
+
+env:
+  RUST_TOOLCHAIN: 1.80.0
```
stevenhorsman, line 9: I like the idea of pining the rust version (IIRC most of the other builds are using stable and image-rs is using 1.76?). This is probably outside the scope of this PR, but I wondered if having the versions in a shared place (e.g. versions.yaml) would be helpful for sync these up. 

Also was there are particular reason for picking 1.80.0 over 1.81.0 as the latest stable?
I'll note that in kata-containers we are currently using rust 1.76.0 as Red Hat couldn't bump to a more recent versions for their downstream work. This isn't necessarily a reason to downgrade, I just wanted to share the reasons behind kata being back-level.

.github/workflows/publish-artifacts.yml
```diff
@@ -0,0 +1,173 @@
...
+        - none
+        - amd
+        - az-cvm-vtpm
+        - tdx
+        - se
+        - cca
+        include:
+        - tee: none
+          libc: musl
+          arch: x86_64
```
stevenhorsman, line 29: Is having a `gnu` `s390x` version of `none` for non-SE s390x possible? I think that combination is the one we use in the peer pods generic s390x podvm image?

.github/workflows/publish-artifacts.yml
```diff
@@ -0,0 +1,173 @@
...
+      LIBC: ${{ matrix.libc }}
+      REGISTRY: ghcr.io
+      IMAGE_NAME: ${{ github.repository }}
+    steps:
+    - name: Take a pre-action for self-hosted runner
+      if: matrix.arch == 's390x'
+      run: |
+        if [ -f "${HOME}/script/pre_action.sh" ]; then
+          "${HOME}/script/pre_action.sh" cc-guest-components
+        fi
```
stevenhorsman, line 57: @BbolroC can confirm, but I think we use the `ACTIONS_RUNNER_HOOK_JOB_STARTED` and `ACTIONS_RUNNER_HOOK_JOB_COMPLETED` envs on the s390x runner now rather than needing to call the pre/post install action scripts directly.

.github/workflows/publish-artifacts.yml
```diff
@@ -0,0 +1,173 @@
...
+    - name: Log in to the Container registry
+      uses: docker/login-action@65b78e6e13532edd9afa3aa52ac7964289d1a9c1
+      with:
+        registry: ${{ env.REGISTRY }}
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+
+    - uses: oras-project/setup-oras@v1
+      with:
+        version: 1.0.0
```
stevenhorsman, line 68: I think we need to use oras 1.2.0+ to get multi-arch support for the pull at least, I'm not sure about the push, but it might be worth updating it to align with kata and CAA's version anyway?

.github/workflows/publish-artifacts.yml
```diff
@@ -0,0 +1,173 @@
...
+      LIBC: ${{ matrix.libc }}
+      REGISTRY: ghcr.io
+      IMAGE_NAME: ${{ github.repository }}
+    steps:
+    - name: Take a pre-action for self-hosted runner
+      if: matrix.arch == 's390x'
+      run: |
+        if [ -f "${HOME}/script/pre_action.sh" ]; then
+          "${HOME}/script/pre_action.sh" cc-guest-components
+        fi
```
stevenhorsman, line 128: Same comment here as above

.github/workflows/publish-artifacts.yml
```diff
@@ -0,0 +1,173 @@
...
+    - name: Log in to the Container registry
+      uses: docker/login-action@65b78e6e13532edd9afa3aa52ac7964289d1a9c1
+      with:
+        registry: ${{ env.REGISTRY }}
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+
+    - uses: oras-project/setup-oras@v1
+      with:
+        version: 1.0.0
```
stevenhorsman, line 139: Can we bump this to 1.2.0 as mentioned above?

stevenhorsman (commented):
A few comments/questions. Overall it looks good though

.github/workflows/publish-artifacts.yml
mkulke: yeah, it's a bit unfortunate that we don't have this centralized somewhere. we could also consider a composite action that specifies the rust version in a single place at least for the CI actions. It's not that critical to be 100% consistent, since rust doesn't have a runtime, but it's annoying when something breaks.

tha latest stable 1.81.0 is broken for some dependency that we use I think.

.github/workflows/publish-artifacts.yml
mkulke: so this step can be removed?

stevenhorsman: Yeah, that's my understanding

.github/workflows/publish-artifacts.yml
stevenhorsman: Cool - that sounds like a good reason to not pick 1.81.0!

.github/workflows/publish-artifacts.yml
mkulke: well, in fact it's not so easy. at the moment we have a 1:1 mapping from tee to arch. now we'd have a none-x64_64 and a none-s390x. we can rename the TEE or are there other options?

stevenhorsman: Yeah, I guess for `TEE_PLATFORM`= none we need to also append the arch and ensure we have the matching logic in kata and CAA where we consume them. We can then maybe revisit this when the `--platform` feature in later oras releases is fully implemented.

.github/workflows/publish-artifacts.yml
stevenhorsman: FYI: I've found https://github.com/confidential-containers/trustee/pull/453 which shows this is implemented in the coco org.

.github/workflows/publish-artifacts.yml
mkulke: ok, I ventured a bit into the oci rabbithole. there is a way to get those platform annotations onto the arch images w/ docker, and then we can publish a multi-arch image that can be pulled w/ `oras pull --plaform linux/s390x ...`

we could even misuse the `variant` segment of the platform for TEEs (`linux/s390x/none`) but I'm not sure we want to go there, this should be the variant of the cpu arch I think (arm/v7).

.github/actions/install-intel-dcap/action.yml
```diff
@@ -0,0 +1,15 @@
...
+    required: true
+runs:
+  using: "composite"
+  steps:
+  - name: Install TDX dependencies
+    id: install-tdx-dependencies
+    shell: bash
+    run: |
+      sudo curl -sL https://download.01.org/intel-sgx/sgx_repo/ubuntu/intel-sgx-deb.key | sudo gpg --dearmor --output /usr/share/keyrings/intel-sgx.gpg
+      sudo echo 'deb [arch=amd64 signed-by=/usr/share/keyrings/intel-sgx.gpg] https://download.01.org/intel-sgx/sgx_repo/ubuntu ${{ inputs.ubuntu-version }} main' | sudo tee /etc/apt/sources.list.d/intel-sgx.list
```
mythi, line 13: thanks a lot for this! (I think it's better like this)

```suggestion
      curl -sL https://download.01.org/intel-sgx/sgx_repo/ubuntu/intel-sgx-deb.key | sudo gpg --dearmor --output /usr/share/keyrings/intel-sgx.gpg
      echo 'deb [arch=amd64 signed-by=/usr/share/keyrings/intel-sgx.gpg] https://download.01.org/intel-sgx/sgx_repo/ubuntu ${{ inputs.ubuntu-version }} main' | sudo tee /etc/apt/sources.list.d/intel-sgx.list
```

.github/workflows/publish-artifacts.yml
```diff
@@ -0,0 +1,187 @@
...
+          libc: gnu
+        - tee: tdx
+          arch: x86_64
+          libc: gnu
+        - tee: se
+          arch: s390x
+          libc: gnu
+        - tee: cca
+          arch: x86_64
+          libc: musl
```
stevenhorsman, line 61: Optional suggestion, that I've used when the matrix combinations get messy and difficult (for me) to read:
```
        platforms: [
          { tee: none, arch: x86_64, libc: musl},
          { tee: amd, arch: x86_64, libc: musl},
          { tee: az-cvm-vtpm, arch: x86_64, libc: musl},
          { tee: tdx, arch: x86_64, libc: musl},
          { tee: cca, arch: x86_64, libc: musl},
          { tee: none, arch: s390x, libc: gnu},
          { tee: se, arch: s390x, libc: gnu},
        ]
```
then you can use `matrix.platforms.arch/tee/libc` in the references

stevenhorsman (commented):
One optional comment, but it's looking good

stevenhorsman (approved):

.github/workflows/publish-artifacts.yml
mkulke: interesting, can you extend the example to a `strategy` block?

mkulke: great, that is way more legible!

.github/workflows/publish-artifacts.yml
mkulke: i found that the rust-lang action also sets up the cache automagically, so I reverted to this.

dcmiddle:
Not trying to be pedantic, just checking my understanding...
This isn't publishing _to oras_. It is publishing _using the oras cli_, which pushes to a specified registry, e.g. ghcr.
Is that right?

mkulke:
Edit: I added build attestations for those artifacts, which is a rather new gh feature. we might want to use them when consuming the binaries:

```bash 
gh attestation verify oci://ghcr.io/mkulke/guest-components/api-server-rest@sha256:251cfcbd73a625b1224b7f5b8d639ab93779650e8c42aaa5bdf4636fa00d7922 -o mkulke
Loaded digest sha256:251cfcbd73a625b1224b7f5b8d639ab93779650e8c42aaa5bdf4636fa00d7922 for oci://ghcr.io/mkulke/guest-components/api-server-rest@sha256:251cfcbd73a625b1224b7f5b8d639ab93779650e8c42aaa5bdf4636fa00d7922
Loaded 1 attestation from GitHub API
✓ Verification succeeded!

sha256:251cfcbd73a625b1224b7f5b8d639ab93779650e8c42aaa5bdf4636fa00d7922 was attested by:
REPO                     PREDICATE_TYPE                  WORKFLOW
mkulke/guest-components  https://slsa.dev/provenance/v1  .github/workflows/publish-artifacts.yml@refs/heads/mkulke/test
```

we can assert some facts about the build status:

```
gh attestation verify oci://ghcr.io/mkulke/guest-components/api-server-rest@sha256:251cfcbd73a625b1224b7f5b8d639ab93779650e8c42aaa5bdf4636fa00d7922 -o mkulke --format json | jq .[].verificationResult.signature.certificate
{
  "certificateIssuer": "CN=sigstore-intermediate,O=sigstore.dev",
  "subjectAlternativeName": "https://github.com/mkulke/guest-components/.github/workflows/publish-artifacts.yml@refs/heads/mkulke/test",
  "issuer": "https://token.actions.githubusercontent.com",
  "githubWorkflowTrigger": "push",
  "githubWorkflowSHA": "65ecab01e3f3e3e89653977b15078e02b9a46698",
  "githubWorkflowName": "Publish artifacts to ORAS",
  "githubWorkflowRepository": "mkulke/guest-components",
  "githubWorkflowRef": "refs/heads/mkulke/test",
  "buildSignerURI": "https://github.com/mkulke/guest-components/.github/workflows/publish-artifacts.yml@refs/heads/mkulke/test",
  "buildSignerDigest": "65ecab01e3f3e3e89653977b15078e02b9a46698",
  "runnerEnvironment": "github-hosted",
  "sourceRepositoryURI": "https://github.com/mkulke/guest-components",
  "sourceRepositoryDigest": "65ecab01e3f3e3e89653977b15078e02b9a46698",
  "sourceRepositoryRef": "refs/heads/mkulke/test",
  "sourceRepositoryIdentifier": "651443953",
  "sourceRepositoryOwnerURI": "https://github.com/mkulke",
  "sourceRepositoryOwnerIdentifier": "273280",
  "buildConfigURI": "https://github.com/mkulke/guest-components/.github/workflows/publish-artifacts.yml@refs/heads/mkulke/test",
  "buildConfigDigest": "65ecab01e3f3e3e89653977b15078e02b9a46698",
  "buildTrigger": "push",
  "runInvocationURI": "https://github.com/mkulke/guest-components/actions/runs/11148465609/attempts/1",
  "sourceRepositoryVisibilityAtSigning": "public"
}
```

I'm curious how this will look on private runners @BbolroC @stevenhorsman did you use that yet on s390x?

mkulke:
> Not trying to be pedantic, just checking my understanding... This isn't publishing _to oras_. It is publishing _using the oras cli_, which pushes to a specified registry, e.g. ghcr. Is that right?

yes, albeit there's some conventions/annotations attached when using the oras cli to push to a registry. technically correct would be "pushing to OCI w/ oras"

.github/workflows/publish-artifacts.yml
```diff
@@ -0,0 +1,188 @@
+name: Publish artifacts to ORAS
```
dcmiddle, line 1:
```suggestion
name: Publish artifacts with ORAS
```

.github/workflows/publish-artifacts.yml
```diff
@@ -0,0 +1,188 @@
...
+      with:
+        ubuntu-version: jammy
+      if: matrix.platform.tee == 'tdx'
+
+    - uses: actions/checkout@v4
+
+    - name: Build
+      run: make ./target/${{ env.RUST_TARGET }}/release/attestation-agent
+
+    - name: Publish to ORAS
```
dcmiddle, line 71:
```suggestion
    - name: Publish with ORAS
```

.github/workflows/publish-artifacts.yml
```diff
@@ -0,0 +1,188 @@
...
+
+    - uses: actions/checkout@v4
+
+    - name: Build CDH
+      run: make ./target/${{ env.RUST_TARGET }}/release/confidential-data-hub
+
+    - name: Build ASR
+      run: make ./target/${{ env.RUST_TARGET }}/release/api-server-rest
+
+    - name: Publish CDH + ASR to ORAS
```
dcmiddle, line 156:
```suggestion
    - name: Publish CDH + ASR with ORAS
```

Makefile
```diff
@@ -21,8 +21,8 @@ else ifeq ($(TEE_PLATFORM), fs)
   ATTESTER = none
 else ifeq ($(TEE_PLATFORM), tdx)
   ATTESTER = tdx-attester
-else ifeq ($(TEE_PLATFORM), az-tdx-vtpm)
-  ATTESTER = az-tdx-vtpm-attester
+else ifeq ($(TEE_PLATFORM), az-cvm-vtpm)
```
mythi, line 24: while we are at it, perhaps change this one too https://github.com/confidential-containers/guest-components/blob/17f9899c226e5097d1157d86b37f8e5459bf39bc/.github/workflows/aa_cc_kbc.yml#L42

mythi: actually, this seems different and a bit confusing. That is for `KBC` (which does not seem to be used in the Makefile :thinking:) looks like the github workflows did not get updated in c41e8c1022cff748abb0214cd23b83c2e342d11b

stevenhorsman:
> I'm curious how this will look on private runners @BbolroC @stevenhorsman did you use that yet on s390x?

Not as far as I know. @BbolroC is (also) on public holiday today, but might have more info. I think we give it a try and see how it looks.

Makefile
mkulke: yeah, I think the toggles are used consistently across the repo, some of them are historic artifacts. KBC, ATTESTERS and now TEE_PLATFORM are all variations of the same toggle with subtle differences. This probably should be cleaned up holistically in docs, makefiles and workflows at some point, but I'm hesitant to put more drive-by fixes into this PR.

I'll open another PR to add the azure tdx line that workflow.

dcmiddle (approved):

.github/actions/install-intel-dcap/action.yml
mythi: @mkulke I think I saw you accepting this suggestion but it's now missing?

mkulke:
> > I'm curious how this will look on private runners @BbolroC @stevenhorsman did you use that yet on s390x?
> 
> Not as far as I know. @BbolroC is (also) on public holiday today, but might have more info. I think we give it a try and see how it looks.

@stevenhorsman it looks like oras installer doesn't support the ibm arch, but I suppose it's a bug, since the project publishes s390x packages, now?

image

stevenhorsman:
> @stevenhorsman it looks like oras installer doesn't support the ibm arch, but I suppose it's a bug, since the project publishes s390x packages, now?

Yeah, maybe just an issue with the action?

stevenhorsman:
It looks from https://github.com/oras-project/setup-oras/blob/4e826cc60c14f9b275cce0ad4ed390c040de5099/src/lib/release.ts#L95-L100 like it is expecting arch of `s390`, not `s390x`

mkulke:
> It looks from https://github.com/oras-project/setup-oras/blob/4e826cc60c14f9b275cce0ad4ed390c040de5099/src/lib/release.ts#L95-L100 like it is expecting arch of `s390`, not `s390x`. We might need some help from Choi to confirm what the settings are, but we might want a patch to support both

https://github.com/oras-project/setup-oras/pull/57

stevenhorsman:
> https://github.com/oras-project/setup-oras/pull/57

Nice - beat me too it :)

PR #734, opened 2024-10-02
drop deprecated eaa_kbc + move gh actions to Ubuntu 24.04

mythi:


mythi:
(I will wait and rebase to #731 before moving away from draft)

mythi:
@mkulke added 3a97cb4 here

Xynnn007 (approved):
lgtm. I do not know if kata's guest image uses ubuntu 24.04 now. If not, some potential problems like compatibility would not be found.

mythi:
>I do not know if kata's guest image uses ubuntu 24.04 now. If not, some potential problems like compatibility would not be found.

The alternative for this PR is to change `ubuntu-latest` to point back to `ubuntu-22.04` but my preference is to jump to `ubuntu-24.04`. We would not be able to cover all distros for the guest image anyways. AFAICS peer-pods isn't using ubuntu at all.

dcmiddle (approved):
LGTM
fwiw, I scanned for any missed references to the removed components and tests but didn't find anything.
There's an open PR in confidential-containers to remove instruction references to eaa kbc. 
(The website might have some content cleanup to do).

PR #737, opened 2024-10-03
CI: Add azure tdx feature to aa_kbc build

mkulke:
This attester/kbc combination was missing so far.


mythi:
@mkulke I looked into this a bit deeper after my initial comment and I believe this is "no op". This adds to `make KBC=` but the `Makefile` does not use it for anything.

I'm currently checking if we need `aa_cc_kbc.yaml` at all since it has a lot of similarities/overlap with `aa_basic.yaml`

PR #739, opened 2024-10-08
gha: Do not use oras-project/setup-oras for s390x

BbolroC:
`oras-project/setup-oras` (raised by #731) does not support s390x like:

```
Error: Error: unsupported architecture: s390x
```

https://github.com/confidential-containers/guest-components/actions/runs/11223420868/job/31198020471
https://github.com/confidential-containers/guest-components/actions/runs/11223420868/job/31198020098

But the package is already installed and maintained by the self-hosted runner admin.
Let's skip installing it for the platform.


mkulke (approved):
Ah, that's great. we don't have to wait for the action to be fixed then

PR #740, opened 2024-10-08
ci: skip oras setup for s390x builds

mkulke:
We missed the 2nd invocation of setup-oras in the prior PR #739

CDH and ASR builds do not need to setup oras on s390x self-hosted runners, since the setup is currently broken and it's provided on the self-hosted runners anyway.


BbolroC (approved):
LGTM thanks @mkulke !

PR #755, opened 2024-10-15
Add trustee-client - a simple client for Trustee

uril:
Gather evidence, attest and get secrets from Trustee

Currently it simply uses AA's kbs_protocol client and attesters. If needed, it can be enhanced later.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ members = [
     "confidential-data-hub/storage",
     "image-rs",
     "ocicrypt-rs",
+    "trustee-client",
 ]
 
 [workspace.dependencies]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6566,6 +6566,23 @@ dependencies = [
 ]
 
 [[package]]
+name = "trustee-client"
+version = "0.10.0"
+dependencies = [
+ "anyhow",
+ "base64",
+ "clap",
+ "config",
+ "env_logger",
+ "kbs_protocol",
+ "log",
+ "rstest",
+ "serde",
+ "serde_json",
+ "tokio",
+]
+
+[[package]]
 name = "try-lock"
 version = "0.2.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/README.md
+++ b/README.md
@@ -24,6 +24,9 @@ Confidential Data Hub.
 [coco-keyprovider](attestation-agent/coco_keyprovider/)
 CoCo Keyprovider. Used to encrypt the container images.
 
+[trustee-client](trustee-client)
+A simple client to fetch secrets from Trustee
+
 ## Tools
 
 [secret-cli](confidential-data-hub/secret)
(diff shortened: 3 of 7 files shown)


Cargo.toml
```diff
@@ -17,6 +17,7 @@ members = [
     "confidential-data-hub/storage",
     "image-rs",
     "ocicrypt-rs",
+    "trustee-client",
```
fitzthum, line 20: I think this name might be too general. This tool only exercises a subset of the Trustee APIs. Btw see [this proposal](https://github.com/confidential-containers/trustee/issues/529) which might end up with a similar name.

Maybe this could be called `cc-kbc-resource-getter` or `trustee-resource-getter` or `trustee-resource-client` or something else

uril:
Thanks @fitzthum 
I can put the Trustee URL + Certificate in the command line.
If needed a configuration file can be added later.
I noticed that AA and CDH each use a configuration file so I added it here too.

Cargo.toml
uril: Thanks, I'll take a look at that proposal and think of names.

ChengyuZhu6:
Good job. While this PR functions similarly to a kbs-client. I agree with Tobin's comment that it would be better to move it to the trustee repo if the client is focused on expanding the [kbs-client](https://github.com/confidential-containers/trustee/tree/main/tools/kbs-client)

Xynnn007 (commented):
Nice work. I noticed that the sub-crate aims to implement a binary for dep `kbs_protocol`. Probably we could make it a binary of `kbs_protocol` instead of a separate crate. [Example](https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/attester/Cargo.toml#L43-L45) for this

Xynnn007:
> Btw, it's interesting that we're in the guest-components repo. I think that's fine and maybe having some kind of client in both guest-components and trustee repos will help us decouple the tests a little bit.

The server (trustee) and client (guest-components) codes are in two repos. If we want to have some test for compability of the two repo, it would always require us to pull the other's code/image/... in another repo

uril:
@ChengyuZhu6 @Xynnn007 thanks for looking.

The idea is to have a simple program in guest-components repository that can get resources from Trustee.

I thought of adding it as a binary under kbs_protocol, but if more features are added later would it stay under kbs_protocol or move ?

Xynnn007:
> I thought of adding it as a binary under kbs_protocol, but if more features are added later would it stay under kbs_protocol or move ?

We can see the concrete situation then. Up to now, it seems that we can keep the feature aligning with `kbs_protocol`'s.

fitzthum:
Yeah seems fine to keep it in this repo and I think having a binary under kbs protocol makes sense. It should be relatively easy to move this there. We can move it again if the scope increases, but it seems like the goal of this really is to have an extremely narrow scope.

uril:
Thanks. I'll send another PR under kbs_protocol and close this one.
Is trustee-agent an acceptable name ? If not I'll go with the suggested trustee-resource-getter/trustee-resource-client.

Xynnn007:
All names sound not bad to me, as now `trustee`, `kbs protocol server`, `(cc) kbs` are referring to a same thing. I think we can know what this tool is at the first glance at the name

Issue #674, opened 2024-08-15
[RFC] Add mesh VPN support to the CDH


portersrc:
CoCo currently has no first-class support for securing the communication between enclaves. This responsibility has been left to the application, but CoCo could provide this optional feature. This RFC outlines what changes in the confidential data hub could support such a feature.

The following high-level diagram shows how this could work:
<img width="708" alt="image" src="https://github.com/user-attachments/assets/6bf5d11d-285e-42b3-82e8-968e39f527fb">
Orange nodes are where the main changes would be needed, including in the CDH. The main points are:
1. A "mesh daemon", in this case Nebula [7], would provide the fundamental mesh support (peer discovery, establishing routes, encryption). It runs inside the guest VM. This would be packaged with guest-components.
2. CDH must be enlightened with (a) general mesh setup support/structs and (b) specifically support to start a nebula mesh.
3. If this option is enabled by the user/workload owner, then kata-agent would invoke the CDH's mesh setup at start.
4. trustee must be able to supply any needed secrets for such a mesh [4].
5. The encrypted mesh would co-exist alongside the untrusted kubernetes overlay network, and both would be exposed and usable by the workload pod.

---

# Proposed changes to guest-components
- Add a mesh VPN daemon, Nebula, to guest-components.
- Augment CDH with an encrypted mesh folder (alongside kms, secret, and storage)
- Within that mesh folder, provide a Nebula implementation. Mesh set up entails:
    - Requesting a key, cert, and CA cert from trustee
    - Starting the mesh daemon

# External changes
Changes outside of guest-components that would be needed for this approach include:
- kata-agent needs to be able to conditionally invoke the CDH's mesh setup at start-up
- trustee needs to be able to serve mesh-related secrets [4]

# Alternative designs and frameworks
Alternative designs that were considered, each with varying merits:
- Traditional VPN implementations such as Wireguard, OpenVPN, IPSec (strongswan, libreswan)
- Istio + Envoy

# References
[1] [August 15, 2024 community meeting with VPN talk](https://docs.google.com/document/d/1E3GLCzNgrcigUlgWAZYlgqNTdVwiMwCRTJ0QnJhLZGA/edit#heading=h.gxexiy5hso9k)
[2] [May 9, 2024 community meeting with VPN talk](https://docs.google.com/document/d/1E3GLCzNgrcigUlgWAZYlgqNTdVwiMwCRTJ0QnJhLZGA/edit#heading=h.h297yxi38oyl)
[...]

PR #763, opened 2024-10-23
Add mesh VPN support to the CDH

portersrc:
RFC issue is [here](https://github.com/confidential-containers/guest-components/issues/674)

More changes are required before this should be moved from draft state:
- trustee plugin and nebula support [here](https://github.com/confidential-containers/trustee/pull/539
- possible kata-containers support (no PR opened for this yet) (may depend on initdata)
- packaging (e.g. how to build and include nebula itself)
- integration testing


confidential-data-hub/hub/src/bin/ttrpc-cdh-tool.rs
```diff
@@ -105,6 +108,15 @@ struct PullImageArgs {
     bundle_path: String,
 }
 
+#[derive(Debug, Args)]
+#[command(author, version, about, long_about = None)]
+struct InitOverlayNetworkArgs {
+    #[arg(short, long)]
+    pod_name: String,
+    #[arg(short, long)]
+    lighthouse_pub_ip: String,
```
fitzthum, line 117: I think this could maybe come from the CDH config file (with init-data coming soon) rather than be passed from the caller.

confidential-data-hub/hub/src/hub.rs
```diff
@@ -93,6 +93,18 @@ impl DataHub for Hub {
...
+    async fn init_overlay_network(
+        &self,
+        pod_name: String,
+        lighthouse_pub_ip: String,
+    ) -> Result<Vec<u8>> {
+        info!("init overlay network called");
+        overlay_network::init(pod_name, lighthouse_pub_ip).await?;
+        // FIXME remove return value for this interface if we don't need
+        // anything here.
+        Ok(Vec::<u8>::new())
```
fitzthum, line 106: Agree with this comment. Just return a result here probably

confidential-data-hub/overlay-network/src/overlay_network/nebula.rs
```diff
@@ -0,0 +1,230 @@
...
+use std::net::Ipv4Addr;
+use std::process::Command;
+
+pub struct NebulaMesh {
+    pod_name: String,
+    lighthouse_ip: String,
+}
+
+// FIXME This should be a shared struct, if possible, with trustee's nebula
+//       plugin. It's that plugin's custom protocol.
```
fitzthum, line 23: Not really possible. Even our attesters have structs that need to match the verifiers in the AS

confidential-data-hub/overlay-network/src/overlay_network/nebula.rs
```diff
@@ -0,0 +1,230 @@
...
+
+const CA_CERT_PATH: &str = "/tmp/nebula/ca.crt";
+const POD_CERT_PATH: &str = "/tmp/nebula/pod.crt";
+const POD_KEY_PATH: &str = "/tmp/nebula/pod.key";
+const LIGHTHOUSE_CONFIG_PATH: &str = "/tmp/nebula/lighthouse-config.yaml";
+const WORKER_CONFIG_PATH: &str = "/tmp/nebula/config.yaml";
+
+// FIXME These should be configurable
+const LIGHTHOUSE_IP: Ipv4Addr = Ipv4Addr::new(192, 168, 100, 100);
+const LIGHTHOUSE_MASK: Ipv4Addr = Ipv4Addr::new(255, 255, 255, 0);
```
fitzthum, line 41: Yeah, try to get these from the CDH config file. Otherwise we have sort of a single-use feature.

confidential-data-hub/overlay-network/src/overlay_network/nebula.rs
```diff
@@ -0,0 +1,230 @@
...
+
+impl NebulaMesh {
+    /// Initialize a nebula mesh. The general approach is as follows:
+    /// - Calculate what the mesh IP will be for this worker.
+    /// - Ask trustee for its nebula credentials
+    /// - Start the nebula daemon.
```
fitzthum, line 47: nit: put comment before function name

confidential-data-hub/overlay-network/src/overlay_network/nebula.rs
```diff
@@ -0,0 +1,230 @@
...
+        if is_lighthouse {
+            mesh_ip = LIGHTHOUSE_IP;
+            which_config = LIGHTHOUSE_CONFIG_PATH;
+        } else {
+            mesh_ip = self.generate_mesh_ip()?;
+            which_config = WORKER_CONFIG_PATH;
+        }
+
+        // FIXME: kbs hard-coded to localhost is wrong. This should be based on
+        //        ResourceUri? Where does it come from?
```
fitzthum, line 61: config file

confidential-data-hub/overlay-network/src/overlay_network/nebula.rs
```diff
@@ -0,0 +1,230 @@
...
+const LIGHTHOUSE_IP: Ipv4Addr = Ipv4Addr::new(192, 168, 100, 100);
+const LIGHTHOUSE_MASK: Ipv4Addr = Ipv4Addr::new(255, 255, 255, 0);
+
+impl NebulaMesh {
+    /// Initialize a nebula mesh. The general approach is as follows:
+    /// - Calculate what the mesh IP will be for this worker.
+    /// - Ask trustee for its nebula credentials
+    /// - Start the nebula daemon.
+    pub async fn init(&self) -> Result<()> {
+        let is_lighthouse: bool = self.lighthouse_ip.is_empty();
```
fitzthum, line 49: so we're going to use one of the pods as the lighthouse?

confidential-data-hub/overlay-network/src/overlay_network/nebula.rs
```diff
@@ -0,0 +1,230 @@
...
+
+        let (mesh_ip, which_config);
```
fitzthum, line 51: to be more rust-like, you could do `let (mesh_ip, which_config) if is_lighthouse {` and then have each arm return a tuple

confidential-data-hub/overlay-network/src/overlay_network/nebula.rs
```diff
@@ -0,0 +1,230 @@
...
+            which_config = LIGHTHOUSE_CONFIG_PATH;
+        } else {
+            mesh_ip = self.generate_mesh_ip()?;
+            which_config = WORKER_CONFIG_PATH;
+        }
+
+        // FIXME: kbs hard-coded to localhost is wrong. This should be based on
+        //        ResourceUri? Where does it come from?
+        let prefix_len: u32 = self.netmask_to_prefix_len(LIGHTHOUSE_MASK);
+        let neb_cred_uri: String = format!(
```
fitzthum, line 63: would it make sense to move this through 85ish into it's own method?

confidential-data-hub/overlay-network/src/overlay_network/config_templates.rs
fitzthum, line 1: it could be cool to make these configs more object oriented. rather than just storing a string config, you could have a struct that stores a string and knows how to write itself to a certain place and such. not a requirement but could be nifty

fitzthum (commented):
A few comments on the first half of the PR

confidential-data-hub/overlay-network/src/overlay_network/nebula.rs
portersrc: It's sitting pretty above `init` right now. Do you want above inside the function? Or before `impl NebulaMesh`? Or maybe I missed something.

confidential-data-hub/hub/src/bin/ttrpc-cdh-tool.rs
portersrc: I like this idea. Here's [an example](https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/example.config.json) from the repo. So you're thinking a new top-level key, e.g. "overlay-network"?

confidential-data-hub/overlay-network/src/overlay_network/nebula.rs
portersrc: ... yea. Thoughts? We looked at two designs: (1) the lighthouse is a coco pod (we went with this for now); (2) the lighthouse runs and lives alongside trustee.

confidential-data-hub/overlay-network/src/overlay_network/config_templates.rs
portersrc: I think the config as a giant string (like we have here) (and like what I think you're still suggesting) is good. Reasons include readability, and updating it (to change a default value, or to change something due to a nebula update). But if it's part of a struct that can manage how and where it's written out, etc. (rather than burying those details in the `NebulaMesh::init` function), then it's better than how it is now.

confidential-data-hub/overlay-network/src/overlay_network/nebula.rs
fitzthum: idk what i was thinking tbh

confidential-data-hub/overlay-network/src/overlay_network/nebula.rs
fitzthum: I would do #2. Trustee should already be at some static ip known to all pods and it won't move or go down during the lifecycle of the workload. This would also allow you to more easily have two completely different workloads sharing the same overlay network.

portersrc: I like the merits of #2. Maybe the thing to do is strip the ability to launch the lighthouse here (which will simplify things); then create a new issue and PR for lighthouse support in trustee. Thoughts?

portersrc: I stripped lighthouse support from this PR (and thus from guest-components). It will need some kind of support from the trustee side now.

confidential-data-hub/hub/src/bin/ttrpc-cdh-tool.rs
portersrc: I updated the example configs and also CdhConfig to include these fields.

portersrc:
Do we want a compile-time feature flag for the overlay-network AND a CdhConfig setting to enable it? We should definitely ahve the config setting. I'm wondering if the compile-time feature confuses the matter (though it has its merits).

fitzthum:
> Do we want a compile-time feature flag for the overlay-network AND a CdhConfig setting to enable it? We should definitely have the config setting. I'm wondering if the compile-time feature confuses the matter (though it has its merits).

We probably should have a feature so the code can be completely removed for trypophobes.

confidential-data-hub/hub/src/config.rs
```diff
@@ -61,6 +62,9 @@ pub struct CdhConfig {
     #[serde(default = "ImageConfig::from_kernel_cmdline")]
     pub image: ImageConfig,
 
+    #[serde(default)]
+    pub overlay_network: OverlayNetworkConfig,
```
cclaudio, line 69: If you have this as `pub overlay_network: Some(OverlayNetworkConfig)` and the `OverlayNetworkConfig struct` as below, I think you would not need the `enable` flag as rust would require nebula to be defined at compilation time. Later, if someone adds support for another overlay network we could have them under cargo feature flags.

```rust
pub struct OverlayNetworkConfig {
    pub nebula: NebulaConfig,
}
```
With this I think you should be able to clean up multiple other places in this PR. E.g. you may not need the `validate` function and the `overlay_network` cargo feature flag.

confidential-data-hub/overlay-network/src/nebula/mod.rs
```diff
@@ -0,0 +1,237 @@
...
+#[derive(Debug, Serialize, Deserialize)]
+pub struct NebulaPluginResponse {
+    pub node_crt: Vec<u8>,
+    pub node_key: Vec<u8>,
+    pub ca_crt: Vec<u8>,
+}
+
+const NEBULA_BIN: &str = "/opt/overlay-network/nebula";
+const LIGHTHOUSE_PORT: u32 = 4242; // sync with WORKER_CONFIG_TEMPLATE
+const CA_CERT_PATH: &str = "/tmp/nebula/ca.crt";
```
cclaudio, line 36: I don't know if CDH is multi-threaded or if we will have support for multiple nebula VPNs, but if so it might be better to use [tempdir_in](https://github.com/confidential-containers/trustee/blob/main/kbs/src/plugins/implementations/nebula_ca.rs#L418)

confidential-data-hub/overlay-network/src/nebula/mod.rs
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    /// The nebula daemon requires a CA cert, pod cert, and pod key. The
+    /// nebula-plugin in the trustee provides these, so this function queries
+    /// it with the necessary parameters.
+    async fn request_key_and_certs(&self, kbs_url: Ipv4Addr) -> Result<NebulaPluginResponse> {
+        let mesh_ip = self.generate_mesh_ip()?;
+        let overlay_netmask: Ipv4Addr = self.config.overlay_netmask.parse()?;
+        let prefix_len: u32 = self.netmask_to_prefix_len(overlay_netmask);
+        let neb_cred_uri: String = format!(
```
cclaudio, line 68: IP and pod-name are required, but it also have some nice optional fields. I would be great if we can support them as well
https://github.com/confidential-containers/trustee/blob/main/kbs/src/plugins/implementations/nebula_ca.rs#L55

cclaudio: That makes me thinkg if we should add the nebula version somewhere to be able to check for compatibility in the future

confidential-data-hub/hub/protos/api.proto
```diff
@@ -42,6 +42,15 @@ message ImagePullResponse {
...
 
+message InitOverlayNetworkRequest {
+    string pod_name = 1;
+    string lighthouse_pub_ip = 2;
+}
+
+message InitOverlayNetworkResponse {
+    int32 return_code = 1;
+}
+
```
fitzthum: I think we probably want to enable this entirely via the cdh config / init-data rather than through the kata agent. You may still want to get the pod name from the agent, though.

confidential-data-hub/overlay-network/src/nebula/mod.rs
portersrc: maybe we handle this if/when we try to tackle multi-VPN?

confidential-data-hub/overlay-network/src/nebula/mod.rs
portersrc: I remember you added these fields. I'm seeing:
```
    /// Optional: how long the cert should be valid for.
    /// The default is 1 second before the signing cert expires.
    /// Valid time units are seconds: "s", minutes: "m", hours: "h".
    duration: Option<String>,
    /// Optional: comma separated list of groups.
    groups: Option<String>,
    /// Optional: comma separated list of ipv4 address and network in CIDR notation.
    /// Subnets this cert can serve for
    subnets: Option<String>,
```
I'm wondering now if `duration` should even be something that the worker node requests. Should it instead be set up on the trustee side? For `groups` and `subnets`, I'm tempted to punt to a future PR. For example, I'm not clear on the constraints for the subnets option (and I haven't tested either of these features yet).

portersrc: > That makes me think if we should add the nebula version somewhere to be able to check for compatibility in the future

I also wonder if nebula does this as part of its own protocol.

confidential-data-hub/hub/protos/api.proto
portersrc: What is the thing that should actually kick the initialization of the overlay network now, though? The CDH is a driver that registers services to listen for requests. For example, it might receive a request to pull an image. Or, in the case of the overlay network, it would receive a request from the kata-agent to start. If kata-agent isn't doing that, what should? In other words, after we register the service [here](https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/hub/src/bin/ttrpc-cdh.rs#L69), who actually makes this API call?

PR #764, opened 2024-10-23
Add some debug logs and change some log messages

gauravkuredhat:
This PR is for testing purpose. I have added some debug logs on some of rust files.
Merging is not required.

### DIFF
--- a/attestation-agent/attestation-agent/src/lib.rs
+++ b/attestation-agent/attestation-agent/src/lib.rs
@@ -17,50 +17,17 @@ mod eventlog;
 pub mod token;
 
 use eventlog::{Content, EventLog, LogEntry};
-use log::{debug, info, warn};
+use log::{debug, info, warn, error};
 use token::*;
 
 use crate::config::Config;
 
-/// Attestation Agent (AA for short) is a rust library crate for attestation procedure
-/// in confidential containers. It provides kinds of service APIs related to attestation,
-/// including the following
-/// - `get_token`: get attestation token from remote services, e.g. attestation services.
-/// - `get_evidence`: get hardware TEE signed evidence due to given runtime_data, s.t.
-/// report data.
-/// - `extend_runtime_measurement`: extend the runtime measurement. This will extend the
-/// current hardware runtime measurement register (if any) or PCR for (v)TPM (under
-/// development) platforms
-/// with a runtime event.
-/// - `check_init_data`: check if the given data slice matches the current confidential
-/// computing environment's host data field, e.g. MRCONFIGID for TDX, HOSTDATA for SNP.
-///
-/// # Example
-///
-/// ```no_run
-/// use attestation_agent::AttestationAgent;
-/// use attestation_agent::AttestationAPIs;
-///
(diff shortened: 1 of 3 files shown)


attestation-agent/attestation-agent/src/config/mod.rs
```diff
@@ -50,7 +50,7 @@ impl Default for EventlogConfig {
         Self {
             eventlog_algorithm: HashAlgorithm::Sha384,
             init_pcr: DEFAULT_PCR_INDEX,
-            enable_eventlog: false,
```
Xynnn007, line 53: Now only tdx platform supports eventlog recording thus this defaulting to `true` might cause bugs on other platforms.

attestation-agent/attestation-agent/src/config/mod.rs
```diff
@@ -94,7 +94,7 @@ impl TryFrom<&str> for Config {
             .add_source(config::File::with_name(config_path))
             .set_default("eventlog_config.eventlog_algorithm", DEFAULT_EVENTLOG_HASH)?
             .set_default("eventlog_config.init_pcr", DEFAULT_PCR_INDEX)?
-            .set_default("eventlog_config.enable_eventlog", "false")?
```
Xynnn007, line 97: same as before

attestation-agent/attestation-agent/src/lib.rs
```diff
@@ -17,50 +17,17 @@ mod eventlog;
...
-/// let mut aa = AttestationAgent::new(None).unwrap();
-///
-/// let _quote = aa.get_evidence(&[0;64]);
-/// ```
-
-/// `AttestationAPIs` defines the service APIs of attestation agent that need to make requests
-///  to the Relying Party (Key Broker Service) in Confidential Containers.
-///
-/// For every service API, the `kbc_name` and `kbs_uri` is necessary, `kbc_name` tells
-/// attestation agent which KBC module it should use and `kbs_uri` specifies the KBS address.
```
Xynnn007, line 54: Why deleting these inlined documents?

attestation-agent/Makefile
```diff
@@ -115,7 +115,7 @@ else
 endif
 
 build:
-	cd attestation-agent && $(RUST_FLAGS) cargo build $(release) --no-default-features --features "$(features)" $(binary) $(LIBC_FLAG)
+	cd attestation-agent && RUST_LOG=debug $(RUST_FLAGS) cargo build $(release) --no-default-features --features "$(features)" $(binary) $(LIBC_FLAG)
```
Xynnn007, line 118: I suspect that `RUST_LOG=debug` only works for runtime rather than building time. https://rust-lang-nursery.github.io/rust-cookbook/development_tools/debugging/config_log.html?highlight=rust_log#enable-log-levels-per-module

Xynnn007 (commented):
Hi @gauravkuredhat , thanks a lot for this contribution! 

Sorry for a late reviewing. This would do great help for debugging and logging. Do you have any plan to keep on this?

btw, you might need a DCO and take a rebase.

PR #768, opened 2024-10-28
CDH | add one-shot CDH

Xynnn007:
One-shot CDH is a different version of CDH that will run and do the operation one time then exit. This is useful in scenarios where CDH cannot run as a long term daemon.


mkulke (approved):
nice addition, this should be useful for debugging 👍

fitzthum (approved):
Looks good.

We should add a link to this in the tools section of the README.

ChengyuZhu6 (approved):
LGTM

PR #788, opened 2024-11-06
kbs_protocol: update KBS config for test_client tests

mythi:
client::rcar_client::test::test_client pulls kbs:latest image for tests but the local kbs-config.toml is oudated and tests won't run.

Update the config to get CI tests passing again.

### DIFF
--- a/attestation-agent/kbs_protocol/test/kbs-config.toml
+++ b/attestation-agent/kbs_protocol/test/kbs-config.toml
@@ -1,6 +1,30 @@
+[http_server]
 insecure_http = true
-insecure_api = true
 sockets = ["0.0.0.0:8085"]
 
-[attestation_token_config]
-attestation_token_type = "CoCo"
+[attestation_token]
+insecure_key = true
+
+[policy_engine]
+policy_path = "/opa/confidential-containers/kbs/policy.rego"
+
+[attestation_service]
+type = "coco_as_builtin"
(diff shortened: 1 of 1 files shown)


attestation-agent/kbs_protocol/test/kbs-config.toml
```diff
@@ -1,6 +1,29 @@
...
+    [attestation_service.rvps_config]
+    store_type = "LocalFs"
+    remote_addr = ""
+
+[admin]
+insecure_api = true
+
+[[plugins]]
+name = "resource"
+type = "LocalFs"
```
Xynnn007, line 29: Need another line in the end `dir_path = "/opt/confidential-containers/kbs/repository"` to specify the resource dir.

Xynnn007: I put a PR to specify a default value if this field is not given to align with [the doc](https://github.com/confidential-containers/trustee/blob/main/kbs/docs/config.md#resource-configuration) where `dir_path` is not required. https://github.com/confidential-containers/trustee/pull/564

mythi: thanks for the PR. I had this passing in my local tests and then decided to drop `dir_path` just before pushing since I noticed it was the same as the "default". That did not work well :-)

fitzthum (approved):
`LGTM`

PR #791, opened 2024-11-08
Add trustee-agent - a simple tool to fetch secrets from Trustee

uril:
The trustee attester is a tool to attest and fetch secrets from Trustee.

Trustee attester is using attestation-agent's kbs_protocol client and attesters to gather hardware-based confidential-computing evidence and send it over to Trustee.

This PR replaces #755. Binary renamed and moved under kbs_protocol.
Also Trustee URL and optional certificate are given in the command line and not in a configuration file.

### DIFF
--- a/.github/workflows/trustee-attester.yml
+++ b/.github/workflows/trustee-attester.yml
@@ -1,0 +1,93 @@
+name: trustee-attester basic build
+
+on:
+  pull_request:
+    paths:
+      - 'attestation-agent/**'
+      - '.github/workflows/trustee-attester.yml'
+      - 'Cargo.toml'
+      - 'Cargo.lock'
+  create:
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  trustee_attester_ci:
+    name: Check
+    defaults:
+      run:
+        working-directory: ./attestation-agent/kbs_protocol/
+    strategy:
+      fail-fast: false
+      matrix:
+        rust:
+          - stable
+        instance:
+          - ubuntu-24.04
+          - s390x
+        include:
+          - instance: ubuntu-24.04
+            cargo_test_opts: "-p kbs_protocol --bin trustee-attester --no-default-features --features background_check,passport,openssl,all-attesters,bin"
+          - instance: s390x
+            cargo_test_opts: "-p kbs_protocol --bin trustee-attester --no-default-features --features background_check,passport,openssl,se-attester,bin"
+    runs-on: ${{ matrix.instance }}
+    steps:
+      - name: Code checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+
+      - name: Install Rust toolchain (${{ matrix.rust }})
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: ${{ matrix.rust }}
+          override: true
+          components: rustfmt, clippy
+
+      - name: Install protoc
+        run: |
+          sudo apt-get update && sudo apt-get install -y protobuf-compiler
+
+      - uses: ./.github/actions/install-intel-dcap
+        with:
+          ubuntu-version: noble
+        if: matrix.instance == 'ubuntu-24.04'
+
+      - name: Install TPM dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libtss2-dev
+        if: matrix.instance == 'ubuntu-24.04'
+
+      - name: Install dm-verity dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libdevmapper-dev
+
+      - name: Run cargo build
+        uses: actions-rs/cargo@v1
+        with:
+          command: build
+          args: ${{ matrix.cargo_test_opts }}
+
+      - name: Run cargo test
+        uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: ${{ matrix.cargo_test_opts }}
+
+      - name: Run cargo fmt check
+        uses: actions-rs/cargo@v1
+        with:
+          command: fmt
+          args: --all -- --check
+
+      - name: Run rust clippy
+        uses: actions-rs/cargo@v1
+        with:
+          command: clippy
+          args: ${{ matrix.cargo_test_opts }}
--- a/attestation-agent/kbs_protocol/Cargo.toml
+++ b/attestation-agent/kbs_protocol/Cargo.toml
@@ -10,7 +10,9 @@ anyhow.workspace = true
 async-trait.workspace = true
 attester = { path = "../attester", default-features = false }
 base64.workspace = true
+clap = { workspace = true, features = ["derive"], optional = true }
 crypto = { path = "../deps/crypto", default-features = false }
+env_logger = { workspace = true, optional = true }
 jwt-simple.workspace = true
 kbs-types.workspace = true
 log.workspace = true
@@ -36,6 +38,10 @@ tokio = { workspace = true, features = [ "rt", "macros", "fs", "process" ]}
 [build-dependencies]
 ttrpc-codegen = { workspace = true, optional = true }
 
+[[bin]]
+name = "trustee-attester"
+required-features = ["bin"]
+
 [features]
 default = ["background_check", "passport", "rust-crypto", "all-attesters"]
 
@@ -56,3 +62,5 @@ se-attester  = ["attester/se-attester"]
 
 rust-crypto = ["reqwest/rustls-tls", "crypto/rust-crypto"]
 openssl = ["reqwest/native-tls-vendored", "crypto/openssl"]
+
+bin = ["tokio/rt", "tokio/macros", "clap", "env_logger"]
--- a/attestation-agent/kbs_protocol/src/bin/trustee-attester/README.md
+++ b/attestation-agent/kbs_protocol/src/bin/trustee-attester/README.md
@@ -1,0 +1,33 @@
+# Trustee attester #
+
+A tool to attest and fetch secrets from Trustee
+
+Trustee attester is a part of [confidential-containers](https://github.com/confidential-containers)
+[guest-components](https://github.com/confidential-containers/guest-components)
+project but can be used for confidential VMs as well.
+
+Trustee attester is using attestation-agent's kbs_protocol client and
+attesters to gather hardware-based confidential-computing evidence
+and send it over to Trustee.
+
+A resource with exact same path must be uploaded to Trustee before trustee-attester runs.
+
+
+## Build: ##
+
+```bash
+cargo build -p kbs_protocol --bin trustee-attester --no-default-features
+--features "background_check,passport,<openssl|rust-crypto>,bin,<attesters-list>"
+```
+
+## Run: ##
+
+```bash
+$ trustee-attester --url <Trustee-URL> [--cert-file <path>] get-resource --path <resource-path>
+```
+
+## Example: ##
+
+```bash
+$ trustee-attester --url http://localhost:50000 get-resource --path default/keys/dummy
+```
--- a/attestation-agent/kbs_protocol/src/bin/trustee-attester/main.rs
+++ b/attestation-agent/kbs_protocol/src/bin/trustee-attester/main.rs
@@ -1,0 +1,90 @@
+// Copyright (c) 2023 by Alibaba.
+// Copyright (c) 2024 Red Hat, Inc
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Attest and fetch confidential resources from Trustee
+
+use anyhow::Result;
+use base64::engine::general_purpose::STANDARD;
+use base64::Engine;
+use clap::{Parser, Subcommand};
+use log::debug;
+use std::fs;
+use std::path::PathBuf;
+
+use kbs_protocol::evidence_provider::NativeEvidenceProvider;
+use kbs_protocol::KbsClientBuilder;
+use kbs_protocol::KbsClientCapabilities;
+use kbs_protocol::ResourceUri;
+
+#[derive(Parser)]
+struct Cli {
+    /// Trustee URL of format <protocol>://<host>:<port>
+    #[clap(long, value_parser)]
+    url: String,
+
+    /// Trustee https certificate file path (PEM format)
+    #[clap(long, value_parser)]
+    cert_file: Option<PathBuf>,
+
+    #[clap(subcommand)]
+    command: Commands,
+}
+
+#[derive(Subcommand)]
+enum Commands {
+    /// Get confidential resource
+    #[clap(arg_required_else_help = true)]
+    GetResource {
+        /// KBS Resource path of format <repository>/<type>/<tag>
+        /// Document: https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/docs/KBS_URI.md
+        #[clap(long, value_parser)]
+        path: String,
+    },
+}
+
+#[tokio::main(flavor = "current_thread")]
+async fn main() -> Result<()> {
+    env_logger::init_from_env(env_logger::Env::new().default_filter_or("info"));
+
+    let cli = Cli::parse();
+
+    let url = cli.url;
+    let cert_file = cli.cert_file;
+
+    debug!("url {}", url);
+    debug!("cert_file {:?}", cert_file);
+
+    // Native evidence provider
+    let evidence_provider = Box::new(NativeEvidenceProvider::new()?);
+
+    // a kbs_protocol client with evidence_provider
+    let mut client_builder = KbsClientBuilder::with_evidence_provider(evidence_provider, &url);
+
+    // if a certificate is given, use it
+    if let Some(cf) = cert_file {
+        debug!("Reading certificate from cert_file {}", cf.display());
+        let cert = fs::read_to_string(cf)?;
+        client_builder = client_builder.add_kbs_cert(&cert)
+    }
+
+    // Build the client. This client is used throughout the program
+    let mut client = client_builder.build()?;
+
+    match cli.command {
+        Commands::GetResource { path } => {
+            // resource_path should start with '/' but not with '//'
+            let resource_path = match path.starts_with('/') {
+                false => format!("/{}", path),
+                true => path,
+            };
+            let resource = ResourceUri::new("", &resource_path)?;
+            let resource_bytes = client.get_resource(resource).await?;
+
+            println!("{}", STANDARD.encode(resource_bytes));
+        }
+    };
+
+    Ok(())
+}


attestation-agent/kbs_protocol/src/bin/trustee-agent/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+Trustee agent is a part of [confidential-containers](https://github.com/confidential-containers)
+[guest-components](https://github.com/confidential-containers/guest-components)
+project but can be used for confidential VMs as well.
+
+
+
+Build with:
+    cargo build -p kbs_protocol --bin trustee-agent --no-default-features \\
+          --features "background_check,passport,<openssl|rusttls>,bin,<attesters-list>"
```
Xynnn007, line 15:
```suggestion
This can be formatted into a markdown style

Build with:
```bash
cargo build -p kbs_protocol --bin trustee-agent --no-default-features \
          --features "background_check,passport,<openssl|rusttls>,bin,<attesters-list>"
```
```

File: attestation-agent/kbs_protocol/src/bin/trustee-agent/README.md [hunk 2]
```diff
@@ -0,0 +1,23 @@
...
+
+
+Run:
+  $ trustee-agent --url <Trustee-URL> [--cert-file <path>] get-resource --path <resource-path>
```
Xynnn007, line 19: Also this

```suggestion
```bash
./target/debug/trustee-agent --url <Trustee-URL> [--cert-file <path>] get-resource --path <resource-path>
```
```

File: attestation-agent/kbs_protocol/src/bin/trustee-agent/README.md [hunk 3]
```diff
@@ -0,0 +1,23 @@
...
+
+Example:
+  $ trustee-agent --url http://localhost:50000 get-resource --path default/keys/dummy
```
Xynnn007, line 22:
```suggestion
```bash
./target/debug/trustee-agent --url http://localhost:50000 get-resource --path default/keys/dummy
```
```

**Review by Xynnn007 — COMMENTED** [review]
Thanks @uril ! The code looks pretty good and I like it! Some last comments from my side

File: attestation-agent/kbs_protocol/src/bin/trustee-agent/README.md (hunk 2 above)
Comment (Xynnn007) [reply attestation-agent/kbs_protocol/src/bin/trustee-agent/README.md:19]: There should also be another ```   the next line. It can not be correctly printed here. THe same to the following

**Comment by Xynnn007** [comment]
`cargo fmt` needs to be performed under the codebase to make the CI happy.

File: attestation-agent/kbs_protocol/src/bin/trustee-agent/main.rs [hunk 4]
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    // Build the client. This client is used throughout the program
+    let mut client = client_builder.build()?;
+
+    match cli.command {
+        Commands::GetResource { path } => {
+            let resource_uri = format!("kbs:///{}", path);
+            let resource_bytes = client
+                .get_resource(serde_json::from_str(&format!("\"{resource_uri}\""))?)
```
mkulke, line 80: I don't think we need this json conversion. can you create a uri with `kbs_protocol::ResourceURI::new()?`

attestation-agent/kbs_protocol/src/bin/trustee-agent/main.rs
```diff
@@ -0,0 +1,88 @@
...
+                .await?;
+
+            println!("{}", STANDARD.encode(resource_bytes));
```
mkulke, line 83: I'd make this configurable. base64 by default is fine, but sometimes you want to pipe the output into another process or a file and then it's clunky

mkulke (commented):
some nits

is "agent" the appropriate name here? i'd associate agent with a long-running process that is doing something on behalf of me, like attestation-agent or ssh-agent

attestation-agent/kbs_protocol/src/bin/trustee-agent/main.rs
```diff
@@ -0,0 +1,88 @@
...
+    command: Commands,
+}
+
+#[derive(Subcommand)]
+enum Commands {
+    /// Get confidential resource
+    #[clap(arg_required_else_help = true)]
+    GetResource {
+        /// KBS Resource path, e.g my_repo/resource_type/123abc
+        /// Document: https://github.com/confidential-containers/attestation-agent/blob/main/docs/KBS_URI.md
```
beraldoleal, line 42: This repo is archived, maybe use this instead?

https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/docs/KBS_URI.md

attestation-agent/kbs_protocol/Cargo.toml
```diff
@@ -10,7 +10,9 @@ anyhow.workspace = true
 async-trait.workspace = true
 attester = { path = "../attester", default-features = false }
 base64.workspace = true
+clap = { workspace = true, features = ["derive"], optional = true }
 crypto = { path = "../deps/crypto", default-features = false }
+env_logger.workspace = true
```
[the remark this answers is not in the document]
uril: Yes, done.

attestation-agent/kbs_protocol/src/bin/trustee-agent/README.md
uril: Thanks for the suggestion. I added this and some headers

attestation-agent/kbs_protocol/src/bin/trustee-agent/README.md
uril: I'm not sure what you mean with "another ``` the next line."
I added the ```bash to the patch.

I left it without "./target/debug/" as the program can be also run from e.g. /usr/bin or /usr/local/bin (after cargo install)

attestation-agent/kbs_protocol/src/bin/trustee-agent/README.md
uril: same as above

attestation-agent/kbs_protocol/src/bin/trustee-agent/main.rs
uril: Fixed, thanks.

attestation-agent/kbs_protocol/src/bin/trustee-agent/main.rs
uril: You're correct. I replaced it with new() as suggested.

attestation-agent/kbs_protocol/src/bin/trustee-agent/main.rs
uril: This configuration can be added later.
If not printed as base64, it's printed as a vector of bytes.
I can print it as String, and it may be more convenient to the user, but my concern is
that some bytes may be non-printable.
// let secret = String::from_utf8(resource_bytes)?;

attestation-agent/kbs_protocol/src/bin/trustee-agent/main.rs
uril: As for the name, the program does do the attestation and resource retrieval on your behalf, but it's not a long running process. It runs on the guest, similar to the attestation-agent (which is a long lived service).
I suggested the name trustee-client, but it is already taken for management of Trustee.
The name trustee-resource-getter was also suggested and is most accurate but it's long.
Would you prefer one of trustee-kbc, trustee-cc-kbc ?

uril:
> `cargo fmt` needs to be performed under the codebase to make the CI happy.

Thanks @Xynnn007, it found two issues.

attestation-agent/kbs_protocol/src/bin/trustee-agent/README.md
```diff
@@ -0,0 +1,28 @@
...
+project but can be used for confidential VMs as well.
+
+Trustee agent is using attestation-agent's kbs_protocol client and
+attesters to gather hardware-based confidential-computing evidence
+and send it over to Trustee.
+
+A resource with exact same path must be uploaded to Trustee before get_resource is called.
+
+
+##Build:
```
Xynnn007, line 15:
```suggestion
## Build:
```

attestation-agent/kbs_protocol/src/bin/trustee-agent/README.md
```diff
@@ -0,0 +1,28 @@
+#Trustee agent
```
Xynnn007, line 1:
```suggestion
# Trustee agent
```

attestation-agent/kbs_protocol/src/bin/trustee-agent/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+
+##Build:
+```bash
+cargo build -p kbs_protocol --bin trustee-agent --no-default-features
+--features "background_check,passport,<openssl|rust-crypto>,bin,<attesters-list>"
+
+##Run:
+```bash
+$ trustee-agent --url <Trustee-URL> [--cert-file <path>] get-resource --path <resource-path>
```
Xynnn007, line 22:
```suggestion
$ trustee-agent --url <Trustee-URL> [--cert-file <path>] get-resource --path <resource-path>
```

attestation-agent/kbs_protocol/src/bin/trustee-agent/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+
+##Example:
+```bash
+$ trustee-agent --url http://localhost:50000 get-resource --path default/keys/dummy
```
Xynnn007, line 27:
```suggestion
## Example
$ trustee-agent --url http://localhost:50000 get-resource --path default/keys/dummy
```

attestation-agent/kbs_protocol/src/bin/trustee-agent/main.rs
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    // Build the client. This client is used throughout the program
+    let mut client = client_builder.build()?;
+
+    match cli.command {
+        Commands::GetResource { path } => {
+            let resource_url = String::new(); // leave empty, use cli.url
+            let resource_path = format!("/{}", path); // path must start with a '/'
+            let resource = ResourceUri::new(&resource_url, &resource_path)?;
```
Xynnn007, line 80: If the `resource_url` here is always an empty string, we can directly use `""` (`&str`) instead of always "new" a string and refer it
```suggestion
            let resource_path = format!("/{}", path); // path must start with a '/'
            let resource = ResourceUri::new("", &resource_path)?;
```

Xynnn007 (commented):
Thanks. Some more nits about the readme document. The final rendering shape can be seen on the page https://github.com/uril/coco-guest-components/tree/trustee-agent/attestation-agent/kbs_protocol/src/bin/trustee-agent. You can take a look if it meets your ideal shape when debugging

attestation-agent/kbs_protocol/src/bin/trustee-agent/main.rs
mkulke: yeah, I think it's correct to output b64 by default, but it would be useful if we had a `--raw` flag or something, in case we want to pipe/redirect the secret somewhere. It's not required, though.

attestation-agent/kbs_protocol/src/bin/trustee-agent/main.rs
mkulke: I'm uninspired, too. "trustee-key-release" or "trustee-util" maybe?

uril: How about  "trustee-attester" ?

attestation-agent/kbs_protocol/src/bin/trustee-agent/README.md
uril: Yes, thanks, I fixed it with ## BUILD: ##

attestation-agent/kbs_protocol/src/bin/trustee-agent/README.md
uril: ````bash`   starts the block and ````` closes it

attestation-agent/kbs_protocol/src/bin/trustee-agent/README.md
uril: Fixed, thanks

attestation-agent/kbs_protocol/src/bin/trustee-agent/README.md
uril: I added ```   "close" the code-block

attestation-agent/kbs_protocol/src/bin/trustee-agent/main.rs
uril: I've renamed it, thanks

uril:
> Thanks. Some more nits about the readme document. The final rendering shape can be seen on the page https://github.com/uril/coco-guest-components/tree/trustee-agent/attestation-agent/kbs_protocol/src/bin/trustee-agent. You can take a look if it meets your ideal shape when debugging

Thanks, that was a good tip. I fixed the markdown syntax. 
Note that the rename broke this link.

attestation-agent/kbs_protocol/src/bin/trustee-agent/main.rs
uril: Okay

attestation-agent/kbs_protocol/src/bin/trustee-attester/README.md
```diff
@@ -0,0 +1,33 @@
+# Trustee attester #
```
Xynnn007, line 1: The preview looks good, while we do not need the `#` in the end of the line when making headings. https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#headings

attestation-agent/kbs_protocol/src/bin/trustee-attester/main.rs
```diff
@@ -0,0 +1,87 @@
...
+use kbs_protocol::evidence_provider::NativeEvidenceProvider;
+use kbs_protocol::KbsClientBuilder;
+use kbs_protocol::KbsClientCapabilities;
+use kbs_protocol::ResourceUri;
+
+#[derive(Parser)]
+struct Cli {
+    /// Trustee URL of format <protocol>://<host>:<port>
+    #[clap(long, value_parser)]
+    // url: Option<String>,
```
Xynnn007, line 25: Is this something left for debug?

Xynnn007 (approved):
Last comments, otherwise looks good to me. Please pay attention to @mkulke 's comments

attestation-agent/kbs_protocol/src/bin/trustee-attester/main.rs
```diff
@@ -0,0 +1,87 @@
...
+        client_builder = client_builder.add_kbs_cert(&cert)
+    }
+
+    // Build the client. This client is used throughout the program
+    let mut client = client_builder.build()?;
+
+    match cli.command {
+        Commands::GetResource { path } => {
+            let resource_path = format!("/{}", path); // path must start with a '/'
+            let resource = ResourceUri::new("", &resource_path)?;
```
mkulke, line 79: nit: this will bail with an obscure error message if the user-specified path already starts with a `/`, so I would suggest to either:

a) validate the path format with a regex, so we can safely prepend the /

b) prepend the / only if it's not present: `format!("/{}", path.trim_start_matches('/'))` (even if it looks a bit ridiculous)

mkulke (approved):
a small nitpick, other lgtm

attestation-agent/kbs_protocol/src/bin/trustee-attester/main.rs
uril: Yes, removed.

attestation-agent/kbs_protocol/src/bin/trustee-attester/main.rs
mkulke: 👍 for booleans we can stick to if/else statements though, they also are expressions in rust, i.e. you can return from a branch: let x = if 1 != 2 { "foo" } else { "bar" }

mkulke (changes requested):
sorry I just noticed this: we shouldn't merge the PR without CI for the tool.

we don't need to publish binaries, but there should at least be compile/lint/format checks for the code.

you can take a look at the AA build for reference: https://github.com/confidential-containers/guest-components/blob/main/.github/workflows/aa_basic.yml

uril:
Added a commit with CI. Thanks for recommending aa_basic.yml.

.github/workflows/trustee-attester.yml
```diff
@@ -0,0 +1,99 @@
...
+  create:
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  trustee_attester_ci:
+    if: github.event_name != 'push'
```
mkulke, line 27: why this limitation?

.github/workflows/trustee-attester.yml
```diff
@@ -0,0 +1,99 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        rust:
+          - stable
+        instance:
+          - ubuntu-24.04
+        include:
+          - instance: ubuntu-24.04
+            cargo_test_opts: "-p kbs_protocol --bin trustee-attester --no-default-features --features background_check,passport,openssl,all-attesters,bin"
```
mkulke, line 41: i don't think this is a matrix. we only run the suite on ubuntu 24.04 w/ rust stable

.github/workflows/trustee-attester.yml
```diff
@@ -0,0 +1,99 @@
...
+          components: rustfmt, clippy
+
+      - name: Install protoc
+        run: |
+          sudo apt-get update && sudo apt-get install -y protobuf-compiler
+
+      - uses: ./.github/actions/install-intel-dcap
+        with:
+          ubuntu-version: noble
+        if: matrix.instance == 'ubuntu-24.04'
```
mkulke, line 64: we only run on ubuntu 24.04

.github/workflows/trustee-attester.yml
```diff
@@ -0,0 +1,99 @@
...
+
+      - name: Install TPM dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libtss2-dev
+        if: matrix.instance == 'ubuntu-24.04'
```
mkulke, line 70: we only run this on ubuntu 24.04

.github/workflows/trustee-attester.yml
mkulke, line 41: why --no-default-features?

.github/workflows/trustee-attester.yml
```diff
@@ -0,0 +1,99 @@
...
+      - name: Install dm-verity dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libdevmapper-dev
+
+      - name: Run cargo build
+        uses: actions-rs/cargo@v1
+        with:
+          command: build
+          args: ${{ matrix.cargo_test_opts }}
```
mkulke, line 81: if we want to share the opts, we can set an ENV on the job like this:

image

the env can be reference as `${{ env.CARGO_TEST_OPTS }}`

.github/workflows/trustee-attester.yml
uril: Same as in aa_basic.yml

.github/workflows/trustee-attester.yml
uril: I removed the s390x entry of the list, for now, so only ubuntu is left there.
Possibly someone will re-add it later.

.github/workflows/trustee-attester.yml
mkulke: I think that's a bug. let's remove it here.

.github/workflows/trustee-attester.yml
mkulke: I see, maybe add this as a comment above the matrix statement. otherwise, it's unclear to the reader of that workflow.

uril: Build only with the " --features" that are specified.
Same for dependent crates, only use features that are specified.

uril: I re-added s390x. It was added to many other .yml files

mkulke (approved):

fitzthum:
@uril please keep an eye on the new workflow after we merge this. We may need a follow-up with some fixes.

PR #792, opened 2024-11-08
CDH/KMS: remove ehsm from defaut features

Xynnn007:
eHSM crate is no longer maintained and archived. Also, it requires openssl to be built thus cannot be built with musl.

This patch removes it from default features of `kms` crate. If one wants to build CDH with eHSM support, `KMS_PROVIDER=ehsm` can be used when executing Makefile of CDH.


mythi:
> eHSM crate is no longer maintained and archived

It might be good to add this to the docs too?

mkulke (approved):

PR #794, opened 2024-11-08
cdh: use b64url encoding in sealed-secrets JWS

mkulke:
The payload of the JWS should be encoded/decoded with b64url and no padding.


fitzthum (approved):
`LGTM`

Xynnn007 (commented):
If we want to change the base64 coding alphabet, we need also change the following things to keep it all consistent

The base64 encoding and examples in the doc: https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/docs/SEALED_SECRET.md
The generation code: https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/secret/src/secret/mod.rs#L70

mkulke:
> If we want to change the base64 coding alphabet, we need also change the following things to keep it all consistent. Seems that those are `STANDARD` for now. Btw is there any reason to change the code alphabet? I am not familiar with the convention of base64.

afaik, all of the JW* family use b64-url encoding. some characters b64 set are hard to process (e.g. as query params in urls)  

image

> The base64 encoding and examples in the doc: https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/docs/SEALED_SECRET.md The generation code: https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/secret/src/secret/mod.rs#L70

thx. will fix the docs. but the generation code should be fixed in this PR, no?

Xynnn007:
> thx. will fix the docs. but the generation code should be fixed in this PR, no?

Yes, it should be fixed also in the PR imo

Xynnn007 (approved):
LGTM

mkulke:
> > thx. will fix the docs. but the generation code should be fixed in this PR, no?
> 
> Yes, it should be fixed also in the PR imo

ah, sorry, I was being ambiguous. I meant it _is_ already fixed (b64 is aliased to b64_url_nopad). is there any other place that would need to be adjusted?

image

Xynnn007:
@mkulke No else found from myside. Sry too as my first glance did not catch the part that you already had fixed.

PR #807, opened 2024-11-15
AA | Fix the idempotence of eventlog 

### LINKED ISSUE #803: AA | Idempotence and robust of AAEL (AA Eventlog)
Now when [we enable eventlogging](https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/attestation-agent/config.example.toml#L37) for AA, the launch of AA would create an AAEL file and extend the INIT event to hardware.

In some cases the AA would is not always available and would restart. This would let the hardware status be extended twice but the AAEL file does not record the corresponding records. We need a patch for AA to detect whether the AAEL file already exists. If so, the INIT event will not be extended to hardware and AA just append new entries to the AAEL file.

Another corner case that needs to be considered is if AA crashes during the AAEL recording process, how to reorganize the previously unrecorded entry and RTMR status when AA starts next time to protect atomicity

Xynnn007:
This patch fixes the idempotence of eventlog. Before this, when AA restarts and eventlog is activated, the originally recorded aael will be truncated and the INIT event will be recorded repeatedly. This patch will check whether there is an existing AAEL when AA is restarted. If so, it will skip creating and recording the INIT event.

At the same time, a synchronization mechanism is used to ensure that RTMR expansion will not occur repeatedly after AA abnormally interrupts execution.

Close #803


attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -158,31 +187,27 @@ impl Attester for TdxAttester {
...
+
+    async fn get_rtmr_value(&self, pcr_index: u64) -> Result<Vec<u8>> {
+        let td_report = Self::get_report()?;
+        let rtmr_index = Self::pcr_to_rtmr(pcr_index) as usize;
+
+        let mut rtmr_u8 = Vec::new();
+        let rtmr = &td_report.tdinfo.rtmr[rtmr_index * 6..(rtmr_index + 1) * 6];
+        for i in rtmr {
+            rtmr_u8.extend_from_slice(&i.to_le_bytes());
+        }
```
mkulke, line 207: it might be good to add unit tests for this part, you can use td_report fixtures and maybe move this part into an internal fn to not have the get_report() side effect.

api-server-rest/openapi/api.json
```diff
@@ -1,5 +1,5 @@
 {
-  "openapi": "3.0.3",
+  "openapi": "3.1.0",
```
mkulke, line 2: I guess this is unrelated to this PR?

Xynnn007: Yes. It updates the openapi automatically generated file in a separate commit. if needed i can move it to another pr

mkulke: no, it's fine. I noticed the same, locally.

dcmiddle:
fyi, @mythi @binxing for abnormal logger behavior and implications for how logs and events are managed.

attestation-agent/attester/src/tdx/mod.rs
Xynnn007: Nice idea. Let me add this.

mkulke (approved):

PR #812, opened 2024-11-20
cdh: move image crate into hub/image module

mkulke:
the image crate is used exclusively by cdh, so we don't need to maintain it as individual crate, helping with naming conflicts and reducing build complexity.

note: we want to do this more module-like crates like `kms`, or `secret`, but this can be done in individual PRs to keep.

I think a good target picture would be a single crate for CDH w/ bin & lib, that can be consumed in other crates


Xynnn007 (approved):
Good and thanks for this! Let's if CI is happy.

mythi (approved):

PR #814, opened 2024-11-20
cdh: move storage and secret crates into cdh modules

mkulke:
the storage and secret crate is used exclusively by cdh, so we don't need to we don't need to maintain those as individual crates, helping with naming conflicts and reducing build complexity.

the storage crate has a dep on the secret crate, so we have to cover them in a single change.


confidential-data-hub/hub/Cargo.toml
```diff
@@ -43,42 +48,50 @@ log.workspace = true
 prost = { workspace = true, optional = true }
 protobuf = { workspace = true, optional = true }
 resource_uri.path = "../../attestation-agent/deps/resource_uri"
-secret.path = "../secret"
-storage.path = "../storage"
+rand.workspace = true
```
Xynnn007, line 51: please move `rand` before `resource_uri` to follow dictionarial order

Xynnn007 (approved):
Cool. Thanks for this! Only one nit

fitzthum (approved):
`LGTM`

PR #816, opened 2024-11-20
cdh: move kms crate to cdh module

mkulke:
the kms crate is used exclusively by cdh, so we don't need to maintain it as individual crate, which helps with naming conflicts and reduces build complexity.

note: this currently builds on pr #814


Xynnn007 (approved):
LGTM. Thanks!

fitzthum (approved):
LGTM

Xynnn007:
Oh. Now KBS imports kms as a dependency [here](https://github.com/confidential-containers/trustee/blob/main/Cargo.toml#L37) as backend impl of KBS. Now if we want to import the denpendency in trustee we need to use `confidential-data-hub`, which will also import `image-rs` and a lot of other dependencies.

So probably we'd better move kms crate out to `deps` as it would be used outside cdh.

fitzthum:
Yup, tbh I think we gotta revert this or it's going to be quite annoying to update Trustee to the latest guest-components.

mythi:
> Yup, tbh I think we gotta revert this or it's going to be quite annoying to update Trustee to the latest guest-components.

I spent a bit of time last week trying to re-structure `confidential-data-hub` so that only parts of it are visible outside but that got me to a rabbit hole. Currently `lib.rs` seems to enable pretty much everything but maybe there's a way to split it for the binaries' needs and shared functionality.

mkulke:
we should attempt to avoid circular dependencies, if the kms code is required in both trustee and g-c shouldn't it be in trustee?

mythi:
> if the kms code is required in both trustee and g-c shouldn't it be in trustee?

currently, trustee (repo) depends on both `kbs_protocol` and `kms/confidential-data-hub`.

fitzthum:
I was thinking about the `kbs_protocol` dependency the other day. Maybe this is a terrible idea, but one option is to split shared code into a different repo. We've already done this for `kbs-types` and in some ways it could make sense for the kbs-protocol to have a similar scope. Unfortunately the `kbs_protocol` currently depends on the attesters, which makes things messy.

Anyway in the short term I just want to bump Trustee to use the latest GC. Currently we have [this](https://github.com/confidential-containers/trustee/blob/main/Cargo.toml#L38) which will take some re-factoring or we could back-out this change in gc. I think it probably is best to import as little as possible from gc for the kms, hence we might want kms to be a separate package.

PR #820, opened 2024-11-25
AA/kbs_protocol: Update to 0.2.0 to fix JWE decryption logic due to RFC7516

Xynnn007:
~Per RFC7516, the AEAD's auth tag should be included inside the JWE body. We fix this to align with trustee side~

https://github.com/confidential-containers/trustee/pull/597

~~Let's wait after https://github.com/confidential-containers/trustee/pull/597 gets merged.~~

This patch does a bunch of fixes per RFC7516.
1. AEAD Auth Tag is now expcilitly included inside the `tag` part.
2. Add TEE key pair support for `RSA OAEP` and `ECDH-ES+A256KW`(by default)
3. Mark `RSA PKCS#1 v1.5` padding scheme as deprecated as it is not recommended.

### DIFF
--- a/attestation-agent/deps/crypto/src/native/aes256gcm.rs
+++ b/attestation-agent/deps/crypto/src/native/aes256gcm.rs
@@ -8,9 +8,36 @@
 use anyhow::*;
 use openssl::symm::Cipher;
 
+use crate::AeadCipher;
+
 const TAG_LENGTH: usize = 16;
 
-pub fn decrypt(encrypted_data: &[u8], key: &[u8], iv: &[u8]) -> Result<Vec<u8>> {
+pub fn encrypt_with_aad_detached_tag(
+    key: &[u8],
+    data: &[u8],
+    iv: &[u8],
+    aad: &[u8],
+) -> Result<AeadCipher> {
+    let cipher = Cipher::aes_256_gcm();
+    let mut tag = vec![0; TAG_LENGTH];
+    let ciphertext = openssl::symm::encrypt_aead(cipher, key, Some(iv), aad, data, &mut tag)
+        .map_err(|e| anyhow!("{e:?}"))?;
+    Ok(AeadCipher { tag, ciphertext })
+}
+pub fn decrypt_with_aad_detached_tag(
+    key: &[u8],
+    encrypted_data: &[u8],
+    iv: &[u8],
+    aad: &[u8],
+    tag: &[u8],
+) -> Result<Vec<u8>> {
+    let cipher = Cipher::aes_256_gcm();
+
+    openssl::symm::decrypt_aead(cipher, key, Some(iv), aad, encrypted_data, tag)
+        .map_err(|e| anyhow!("{e:?}"))
+}
+
+pub fn decrypt(key: &[u8], encrypted_data: &[u8], iv: &[u8]) -> Result<Vec<u8>> {
     let cipher = Cipher::aes_256_gcm();
     if encrypted_data.len() < TAG_LENGTH {
         bail!("Illegal length of ciphertext");
@@ -21,7 +48,7 @@ pub fn decrypt(encrypted_data: &[u8], key: &[u8], iv: &[u8]) -> Result<Vec<u8>> 
         .map_err(|e| anyhow!(e.to_string()))
 }
 
-pub fn encrypt(data: &[u8], key: &[u8], iv: &[u8]) -> Result<Vec<u8>> {
+pub fn encrypt(key: &[u8], data: &[u8], iv: &[u8]) -> Result<Vec<u8>> {
     let cipher = Cipher::aes_256_gcm();
     let mut tag = [0u8; TAG_LENGTH];
     let mut ciphertext = openssl::symm::encrypt_aead(cipher, key, Some(iv), &[], data, &mut tag)
@@ -34,14 +61,41 @@ pub fn encrypt(data: &[u8], key: &[u8], iv: &[u8]) -> Result<Vec<u8>> {
 mod tests {
     use rstest::rstest;
 
-    use super::{decrypt, encrypt};
+    use super::{decrypt, decrypt_with_aad_detached_tag, encrypt, encrypt_with_aad_detached_tag};
 
     #[rstest]
-    #[case(b"plaintext1", b"0123456789abcdefghijklmnopqrstuv", b"unique nonce")]
-    #[case(b"plaintext2", b"hijklmnopqrstuv0123456789abcdefg", b"unique2nonce")]
-    fn en_decrypt(#[case] plaintext: &[u8], #[case] key: &[u8], #[case] iv: &[u8]) {
-        let ciphertext = encrypt(plaintext, key, iv).expect("encryption failed");
-        let plaintext_de = decrypt(&ciphertext, key, iv).expect("decryption failed");
+    #[case(b"0123456789abcdefghijklmnopqrstuv", b"plaintext1", b"unique nonce")]
+    #[case(b"hijklmnopqrstuv0123456789abcdefg", b"plaintext2", b"unique2nonce")]
+    fn en_decrypt(#[case] key: &[u8], #[case] plaintext: &[u8], #[case] iv: &[u8]) {
+        let ciphertext = encrypt(key, plaintext, iv).expect("encryption failed");
+        let plaintext_de = decrypt(key, &ciphertext, iv).expect("decryption failed");
+        assert_eq!(plaintext, plaintext_de);
+    }
+
+    #[rstest]
+    #[case(
+        b"0123456789abcdefghijklmnopqrstuv",
+        b"plaintext1",
+        b"unique nonce",
+        b"test-aad"
+    )]
+    #[case(
+        b"hijklmnopqrstuv0123456789abcdefg",
+        b"plaintext2",
+        b"unique2nonce",
+        b"test-aad"
+    )]
+    fn en_decrypt_with_aad(
+        #[case] key: &[u8],
+        #[case] plaintext: &[u8],
+        #[case] iv: &[u8],
+        #[case] aad: &[u8],
+    ) {
+        let ciphertext =
+            encrypt_with_aad_detached_tag(key, plaintext, iv, aad).expect("encryption failed");
+        let plaintext_de =
+            decrypt_with_aad_detached_tag(key, &ciphertext.ciphertext, iv, aad, &ciphertext.tag)
+                .expect("decryption failed");
         assert_eq!(plaintext, plaintext_de);
     }
 }
--- a/attestation-agent/deps/crypto/src/native/ec.rs
+++ b/attestation-agent/deps/crypto/src/native/ec.rs
@@ -1,0 +1,194 @@
+// Copyright (c) 2024 Alibaba Cloud
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use anyhow::{anyhow, bail, Context, Result};
+use openssl::{
+    aes::{self, AesKey},
+    bn::{BigNum, BigNumContext},
+    derive::Deriver,
+    ec::{EcGroup, EcKey},
+    hash::{Hasher, MessageDigest},
+    nid::Nid,
+    pkey::{PKey, Private},
+};
+use zeroize::Zeroizing;
+
+use crate::{
+    ec::{Curve, KeyWrapAlgorithm},
+    AES_GCM_256_KEY_BITS,
+};
+
+#[derive(Clone, Debug)]
+pub enum EcKeyPair {
+    P256(P256EcKeyPair),
+}
+
+impl Default for EcKeyPair {
+    fn default() -> Self {
+        Self::P256(P256EcKeyPair::default())
+    }
+}
+
+impl EcKeyPair {
+    fn private_key(&self) -> &PKey<Private> {
+        match self {
+            Self::P256(p256) => p256.private_key(),
+        }
+    }
+
+    pub fn curve(&self) -> Curve {
+        match self {
+            Self::P256(_) => Curve::P256,
+        }
+    }
+
+    pub fn x(&self) -> Result<Vec<u8>> {
+        match self {
+            Self::P256(p256) => p256.x(),
+        }
+    }
+
+    pub fn y(&self) -> Result<Vec<u8>> {
+        match self {
+            Self::P256(p256) => p256.y(),
+        }
+    }
+
+    pub fn to_pkcs8_pem(&self) -> Result<Zeroizing<String>> {
+        let pem = self.private_key().private_key_to_pem_pkcs8()?;
+        let pem = String::from_utf8(pem).map_err(|_| anyhow!("invalid pem"))?;
+        Ok(Zeroizing::new(pem))
+    }
+
+    pub fn from_pkcs8_pem(pem: &str) -> Result<Self> {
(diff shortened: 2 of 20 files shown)


mkulke (commented):
I guess we should also bump `KBS_PROTOCOL_VERSION`. Is there a rationale how we increase the versions. would it be `0.2.0` or `0.1.1`. if it's a breaking change, not merely additive, I would assume we have to bump it to `0.2.0`.

Xynnn007:
@mkulke Updated the protocol version. Also update the test image of kbs

mythi:
> @mkulke Updated the protocol version. Also update the test image of kbs using code of [confidential-containers/trustee#597](https://github.com/confidential-containers/trustee/pull/597).

We don't need to change the test image if the protocol version changes so the `rcar_client.rs` changes can be dropped (the test is skipped until the server image comes with the matching protocol version)

Xynnn007:
> We don't need to change the test image if the protocol version changes so the `rcar_client.rs` changes can be dropped (the test is skipped until the server image comes with the matching protocol version)

This will make the CI red. The client side code now will use the `latest` kbs image of CoCo Community, which is `0.1.1` rather than `0.2.0`.

If you do not want to change the test image. It would make sense once I get approvals of this PR w/ the commit that changes test image built from the KBS side code. Then I can drop the commit. The CI will be red, but we all know what happened and get the PR merged.

mythi:
> This will make the CI red.

c89b96b3 has regression. There is no `ProtocolVersion` error anymore.

Xynnn007:
Once https://github.com/confidential-containers/trustee/pull/600 gets merged, we can restart the failed CI here and it will be green.

Xynnn007:
Ok. All tests passed. I think it can be merged now

deeglaze:
This still doesn't protect the protected headers.
See
https://datatracker.ietf.org/doc/html/rfc7516#appendix-A.4.5

And example https://datatracker.ietf.org/doc/html/rfc7516#section-3.3 particularly this item:

> Let the Additional Authenticated Data encryption parameter be ASCII(BASE64URL(UTF8(JWE Protected Header))).

Xynnn007:
@deeglaze Thanks for this! Let me do a deeper fix tomorrow.

Xynnn007:
Seems that the current `Response` (JWE Json) does not have `aad` field. Let's go back to https://github.com/virtee/kbs-types/pull/44 upstream first.

Xynnn007:
Let's wait for https://github.com/virtee/kbs-types/pull/45 to be merged, too

Xynnn007:
The lint error is not related to this PR. Fixed in https://github.com/confidential-containers/guest-components/pull/828

Xynnn007:
Hi, this PR is ready for review @deeglaze @mkulke @mythi

attestation-agent/deps/crypto/src/native/aes256gcm.rs
```diff
@@ -10,6 +10,19 @@ use openssl::symm::Cipher;
 
 const TAG_LENGTH: usize = 16;
 
+pub fn decrypt_with_aad(
+    encrypted_data: &[u8],
```
huoqifeng, line 14: @Xynnn007 just curious, any reason that the parameters order are not same for `encrypted_data/ciphertext`, `key`, `iv`, `aad`, `tag` in fn `decrypt_with_aad` and `aes256gcm_decrypt`?

Xynnn007: Oh. No special reasons. Let me change the order. Thanks for pointing out this

Xynnn007: Adds a commit for this.

huoqifeng (approved):
LGTM
Thanks! Ding

mkulke (approved):

Xynnn007:
Let me convert this to draft before upstream is ready.

Xynnn007:
Let's wait for the upstream fix to be merged and then I will take a rebase. Other parts are ready for review.

Xynnn007:
Changes happen since last approvals. PTAL @huoqifeng @mkulke 

cc @fitzthum @mythi

attestation-agent/deps/crypto/src/rust/aes256gcm.rs
```diff
@@ -4,11 +4,28 @@
...
+    iv: &[u8],
+    aad: &[u8],
+    tag: &[u8],
+) -> Result<Vec<u8>> {
+    let decrypting_key = Key::<Aes256Gcm>::from_slice(key);
+    let cipher = Aes256Gcm::new(decrypting_key);
+    let nonce = Nonce::from_slice(iv);
+    let mut plaintext = encrypted_data.to_vec();
+    cipher
+        .decrypt_in_place_detached(nonce, aad, &mut plaintext, tag.into())
```
fitzthum, line 22: Why are we using the `detached` version of this function?

attestation-agent/deps/crypto/src/rust/aes256gcm.rs
```diff
@@ -37,11 +54,11 @@ mod tests {
...
 
     #[rstest]
-    #[case(b"plaintext1", b"0123456789abcdefghijklmnopqrstuv", b"unique nonce")]
-    #[case(b"plaintext2", b"hijklmnopqrstuv0123456789abcdefg", b"unique2nonce")]
-    fn en_decrypt(#[case] plaintext: &[u8], #[case] key: &[u8], #[case] iv: &[u8]) {
-        let ciphertext = encrypt(plaintext, key, iv).expect("encryption failed");
-        let plaintext_de = decrypt(&ciphertext, key, iv).expect("decryption failed");
+    #[case(b"0123456789abcdefghijklmnopqrstuv", b"plaintext1", b"unique nonce")]
+    #[case(b"hijklmnopqrstuv0123456789abcdefg", b"plaintext2", b"unique2nonce")]
+    fn en_decrypt(#[case] key: &[u8], #[case] plaintext: &[u8], #[case] iv: &[u8]) {
```
fitzthum, line 59: Should we add test case for the AEAD method?

attestation-agent/deps/crypto/src/native/ec.rs
```diff
@@ -0,0 +1,190 @@
...
+        let private_key = PKey::private_key_from_pem(pem.as_bytes())?;
+        let ec_key = private_key.ec_key().context("must be a ec key")?;
+        if ec_key
+            .group()
+            .curve_name()
+            .ok_or(anyhow!("failed to get curve name"))?
+            == Nid::X9_62_PRIME256V1
+        {
+            return Ok(Self::P256(P256EcKeyPair { private_key }));
+        }
```
fitzthum, line 75: I might do this as a match statement but no big deal.

attestation-agent/deps/crypto/src/native/ec.rs
```diff
@@ -0,0 +1,190 @@
...
+        public_key.affine_coordinates_gfp(private_key.group(), &mut x, &mut _y, &mut ctx)?;
+        let mut x = x.to_vec();
+        x.resize(32, b'0');
+        Ok(x)
+    }
+
+    pub fn y(&self) -> Result<Vec<u8>> {
+        let private_key = self.private_key.ec_key().context("must be a ec key")?;
+        let public_key = private_key.public_key();
+        let mut _x = BigNum::new()?;
```
fitzthum, line 182: Would it make more sense to have one function to get both x and y and then maybe have a wrapper around it that just returns if we need it? There is some duplication between these two functions.

attestation-agent/deps/crypto/src/rust/ec.rs
```diff
@@ -0,0 +1,177 @@
...
+}
+
+#[derive(Clone, Debug)]
+pub struct P256EcKeyPair {
+    secret_key: SecretKey,
+    public_key: P256PublicKey,
+}
+
+impl Default for P256EcKeyPair {
+    fn default() -> Self {
```
fitzthum, line 137: The openssl version has `new` instead of `default` any reason why they aren't the same?

attestation-agent/deps/crypto/src/rust/ec.rs
```diff
@@ -0,0 +1,177 @@
...
+}
+
+impl EcKeyPair {
+    pub fn curve(&self) -> Curve {
+        match self {
+            Self::P256(_) => Curve::P256,
+        }
+    }
+
+    pub fn secret_key(&self) -> &SecretKey {
```
fitzthum, line 40: For openssl implementation this method is called `private_key`. Any reason for this to be different?

attestation-agent/deps/crypto/src/rust/ec.rs
```diff
@@ -0,0 +1,177 @@
...
+        Ok(Self {
+            secret_key,
+            public_key,
+        })
+    }
+
+    pub fn x(&self) -> Result<Vec<u8>> {
+        let x = EncodedPoint::from(self.public_key)
+            .x()
+            .ok_or(anyhow!("invalid public key: without coordicate X"))?
```
fitzthum, line 165: nit: `coordicate` -> `coordinate` also below

attestation-agent/kbs_protocol/src/keypair.rs
```diff
@@ -3,80 +3,160 @@
...
-        self.keypair.decrypt(mode, cipher_text)
+    pub fn unwrap_cek(&self, header: &ProtectedHeader, cipher_text: Vec<u8>) -> Result<Vec<u8>> {
+        #[allow(deprecated)]
+        if &header.alg[..] == PaddingMode::PKCS1v15.as_ref() {
+            warn!("Use deprecated Rsa PKCSv1.5 algorithm!");
+            let TeeKey::Rsa(key) = &self.key else {
+                bail!("Unmatched key. Must be RSA key");
+            };
+
+            let cek = key.decrypt(PaddingMode::OAEP, cipher_text)?;
```
fitzthum, line 71: Should this padding mode be `PKCS1v15`?

attestation-agent/kbs_protocol/src/keypair.rs
```diff
@@ -3,80 +3,160 @@
...
+            let y = epk
+                .get("y")
+                .ok_or(anyhow!("Invalid JWE ProtectedHeader. Without `y`"))?
+                .as_str()
+                .ok_or(anyhow!("Invalid JWE ProtectedHeader. `y` is not a string"))?;
+
+            let y = URL_SAFE_NO_PAD.decode(y)?;
+
+            let TeeKey::Ec(key) = &self.key else {
+                bail!("Unmatched key. Must be RSA key");
```
fitzthum, line 110: This should probably be "Must be EC Key"

attestation-agent/kbs_protocol/src/keypair.rs
```diff
@@ -3,80 +3,160 @@
...
+                    y,
+                })
+            }
+        }
     }
 
     #[inline]
-    pub fn decrypt(&self, mode: PaddingMode, cipher_text: Vec<u8>) -> Result<Vec<u8>> {
-        self.keypair.decrypt(mode, cipher_text)
+    pub fn unwrap_cek(&self, header: &ProtectedHeader, cipher_text: Vec<u8>) -> Result<Vec<u8>> {
```
fitzthum, line 63: If this is now called `unwrap_cek` maybe we should change the `cipher_text` arg to be `wrapped_cek`.

attestation-agent/kbs_protocol/src/keypair.rs
```diff
@@ -3,80 +3,160 @@
...
+
+            let TeeKey::Ec(key) = &self.key else {
+                bail!("Unmatched key. Must be RSA key");
+            };
+
+            if crv != key.curve().as_ref() {
+                bail!("Unmatched curve: {}", crv);
+            }
+
+            let cek = key.unwrap_key(cipher_text, x, y, KeyWrapAlgorithm::EcdhEsA256Kw)?;
```
fitzthum, line 117: We are using `x` and `y` from the header here. Should we be using them from the public key that we store? The RSA methods don't take the public key from the header but for EC we do. What is the thinking behind this?

fitzthum (commented):
Ok, I finally reviewed this PR. I have maybe 2 comments that are important (possibly there is a little bug) and a bunch of less important suggestions.

attestation-agent/deps/crypto/src/symmetric.rs
```diff
@@ -43,11 +45,22 @@ pub fn decrypt(
...
     match wrap_type {
-        WrapType::Aes256Gcm => aes256gcm::decrypt(&ciphertext, &key, &iv),
-        WrapType::Aes256Ctr => aes256ctr::decrypt(&ciphertext, &key, &iv),
+        WrapType::Aes256Gcm => aes256gcm::decrypt(&key, &ciphertext, &iv),
+        WrapType::Aes256Ctr => aes256ctr::decrypt(&key, &ciphertext, &iv),
     }
 }
 
+/// Decrypt the given `ciphertext` with AES256-GCM algorithm.
+pub fn aes256gcm_decrypt(
```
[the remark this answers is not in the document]
Xynnn007: Nice idea. For extensibility, I did a little change with name `decrypt_aead`

attestation-agent/deps/crypto/src/rust/aes256gcm.rs
Xynnn007: The non-`detached` version implies the `tag` lies as the suffix of the cipher text, while jwe detaches `tag` as a separate claim in jwt.

attestation-agent/deps/crypto/src/rust/aes256gcm.rs
Xynnn007: Yes. Let me add this.

attestation-agent/deps/crypto/src/rust/ec.rs
Xynnn007: This is to keep the naming consistent with the underlying library. The Openssl series likes to call the private key secret key, while Rust calls it private key.

attestation-agent/kbs_protocol/src/keypair.rs
Xynnn007: Yes. Let me fix

attestation-agent/kbs_protocol/src/keypair.rs
Xynnn007: Thanks. Let me fix

attestation-agent/kbs_protocol/src/keypair.rs
Xynnn007: Nice idea.

attestation-agent/kbs_protocol/src/keypair.rs
Xynnn007: Due to https://datatracker.ietf.org/doc/html/rfc7518#section-4.6, The `x` and `y` here are of `ephemeral key`. They are used to make a Diffie Hellman protocol with `x` and `y` of the key we store to finally derive the decryption key to decrypt the ciphertext. The logic of obtaining decryption key with elliptic curve is different from that of RSA directly decrypting with private key.

Xynnn007:
Fixed the comments and take a rebase.

fitzthum (approved):
LGTM. Thanks for taking this on.

Last call for review before this is merged.

fitzthum:
Ok, let's try to get the Trustee side in soonish.

Issue #823, opened 2024-11-26
Empty outblob when using TSM to get TDX Quote


Xynnn007:
### Describe the bug

Thanks from @LiuSecone

When using the `evidence_getter` inside a TDX guest that has `/sys/kernel/config/tsm/report`, the evidence of TDX would be
```json
{
    "cc_eventlog": "AQAAA....",
    "quote": "",
    "aa_eventlog": null
}
```

This is not expected. If the reading of `outblob`, attester shoud raise an error and try back to `ioctl` way to get evidence. But the actual execution result treats the empty reading of `/sys/kernel/config/tsm/report/xxx/outblob` a success.

### How to reproduce

See description

### CoCo version information

v0.10.0

### What TEE are you seeing the problem on

None

Xynnn007:
cc @mythi

mythi:
> This is not expected. If the reading of `outblob`, attester shoud raise an error and try back to `ioctl` way to get evidence. But the actual execution result treats the empty reading of `/sys/kernel/config/tsm/report/xxx/outblob` a success.

We have not treated this as an error because there is no read error:

```
root@tdx-guest:~# dd if=/dev/random of=/sys/kernel/config/tsm/report/123/inblob bs=1 count=64
root@tdx-guest:~# cat /sys/kernel/config/tsm/report/123/outblob
root@tdx-guest:~# echo $?
```

(`echo hahaha` is incorrect reportdata and triggers the error shown in the original report)

I'm fine adding an `.is_empty()` check to `tdx-attester` but the `ioctl()` fallback is not ok since that interface is deprecated.

mythi:
> But the actual execution result treats the empty reading of `/sys/kernel/config/tsm/report/xxx/outblob` a success.

you might also wonder the root cause of an empty quote: TDX QGS is either too old (DCAP 1.21 or later is needed) or Qemu cannot connect to it (e.g., `vsock_loopback` module missing)

mythi:
Linux 7.1 will improve the error handling https://github.com/torvalds/linux/commit/0f409eaea53e49932cf92a761de66345c9a4b4be

we can get that backported to Kata if it helps

PR #825, opened 2024-11-27
tdx-attester: log error on empty TSM report

mythi:
Fixes: #823 

tdx_guest TSM provider covers a wide range of errors which trigger an errno on outblob read but can also return empty reports without error.

One such scenario seems to be when Qemu isn't connecting to TDX QGS properly (likely due to misconfiguration) but returns back with an empty buffer.

Notify users about this scenario and log an error on empty TSM report but don't turn it into a new error because there isn't any.

Note: when used with `evidence_getter`, a logger must be enabled to get the logs visible.

### DIFF
--- a/attestation-agent/attester/src/tdx/mod.rs
+++ b/attestation-agent/attester/src/tdx/mod.rs
@@ -87,6 +87,7 @@ impl Attester for TdxAttester {
             },
             |tsm| {
                 tsm.attestation_report(TsmReportData::Tdx(report_data.clone()))
+                    .inspect(|outblob| {if outblob.is_empty() {log::error!("TSM provider returned an empty quote without an error")}})
                     .context("TDX Attester: quote generation using TSM reports failed")
             },
         )?;


attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -87,6 +87,7 @@ impl Attester for TdxAttester {
             },
             |tsm| {
                 tsm.attestation_report(TsmReportData::Tdx(report_data.clone()))
+                    .inspect(|outblob| {if outblob.is_empty() {log::error!("TSM provider returned an empty quote without an error")}})
```
Xynnn007, line 90: If the quote here is empty, the attester side `error` info cannot be exposed to users when using CoCo. The verifier side will face error "base64 decode failed". If we can ensure the behavior here is only caused by misconfiguration of host side we can also raise an error to indicate "TSM provider returned an empty quote, please ensure the host side QGS is configured correctly".

Xynnn007:
Let me handle the CI failure which is not related to this PR.

attestation-agent/attester/src/tdx/mod.rs
mythi: > If we can ensure the behavior here is only caused by misconfiguration

I'm currently checking if Qemu is just handling this case wrong. I'd think the same could also happen due to other reasons too where Qemu only sees `qemu-system-x86_64: Failed to connect to '2:4050': Connection reset by peer` (which is what I get in my test setup with no QGS running).

In the mean time, perhaps we could also look to the verifier side to return meaningful error if the evidence is missing mandatory fields.

Xynnn007: Ok. Anyway, both attester and verifier side should raise error if no quote is included. For attester the error would be finally exposed to kata-agent and thus host side containerd making the deployer know something wrong with host side stack.

mythi: I did a bit of research and it looks the kernel has an issue of not handling the errors correctly. There's even a dedicated error code for `TDX_VP_GET_QUOTE_QGS_UNAVAILABLE`.

Xynnn007: How this error can be caught by AA?

mythi: I'm looking to submit the kernel fix which is something like:
```
diff --git a/drivers/virt/coco/tdx-guest/tdx-guest.c b/drivers/virt/coco/tdx-guest/tdx-guest.c
index abe3076f965e..de71a430a4b9 100644
--- a/drivers/virt/coco/tdx-guest/tdx-guest.c
+++ b/drivers/virt/coco/tdx-guest/tdx-guest.c
@@ -268,6 +268,13 @@ static int tdx_report_new(struct tsm_report *report, void *data)
                goto done;
        }
 
+       /* Quote data is included only when the status is OK */
+       if (quote_buf->status != GET_QUOTE_SUCCESS) {
+               pr_err("GetQuote failed, status:%llx\n", quote_buf->status);
+               ret = -EIO;
+               goto done;
+       }
+
```

Which then shows up as an error normally, like in my test case with `evidence_getter`:

```
get evidence failed: quote generation using TSM reports failed

Caused by:
    0: Failed to access TSM Report attribute: outblob (Input/output error (os error 5))
    1: Input/output error (os error 5)
```

Xynnn007: Nice!

mythi:
I'm closing this one but keeping #823 open to track the kernel/qemu fixes.

PR #833, opened 2024-12-06
AA: add ttrpc client tool

Xynnn007:
This tool is useful when we do debugging and test things for ttrpc-AA.


attestation-agent/attestation-agent/src/bin/ttrpc-aa-client.rs
```diff
@@ -0,0 +1,151 @@
...
+                .expect("request to AA");
+            let token = String::from_utf8(res.Token).unwrap();
+            println!("{token}");
+        }
+        Operation::ExtendRuntimeMeasurement(extend_runtime_measurement_args) => {
+            let req = ExtendRuntimeMeasurementRequest {
+                Domain: extend_runtime_measurement_args.domain,
+                Operation: extend_runtime_measurement_args.operation,
+                Content: extend_runtime_measurement_args.content,
+                ..Default::default()
```
mkulke, line 142: Can we add an (optional) arg for the register?

mkulke (commented):
thanks, this will be very useful for debugging. will this be covered by the CI?

attestation-agent/attestation-agent/src/bin/ttrpc-aa-client.rs
Xynnn007: Added.

Xynnn007:
@mkulke Now cargo clippy and fmt check will cover the tool code.

mkulke (approved):

Issue #840, opened 2024-12-10
image-rs: Panick when pulling image


fidencio:
### Describe the bug

```
Dec 10 07:09:08 jorge kata[261242]: time="2024-12-10T07:09:08.725962832Z" level=debug msg="reading guest console" console-protocol=unix console-url=/run/vc/vm/c80f9ac981266288f2163ac4777b9ca1f54cbf2790b197b160cdf1c141c9ec18/console.sock name=containerd-shim-v2 pid=261242 sandbox=c80f9ac981266288f2163ac4777b9ca1f54cbf2790b197b160cdf1c141c9ec18 source=virtcontainers subsystem=sandbox vmconsole="thread 'tokio-runtime-worker' panicked at /opt/cargo/git/checkouts/guest-components-1e54b222ad8d9630/075b9a9/image-rs/src/pull.rs:106:21:"
Dec 10 07:09:08 jorge kata[261242]: time="2024-12-10T07:09:08.726008239Z" level=debug msg="reading guest console" console-protocol=unix console-url=/run/vc/vm/c80f9ac981266288f2163ac4777b9ca1f54cbf2790b197b160cdf1c141c9ec18/console.sock name=containerd-shim-v2 pid=261242 sandbox=c80f9ac981266288f2163ac4777b9ca1f54cbf2790b197b160cdf1c141c9ec18 source=virtcontainers subsystem=sandbox vmconsole="index out of bounds: the len is 25 but the index is 25"
```

### How to reproduce

Pull the following encrypted image: quay.io/fidencio/prueba:encrypted
key: `/h0vEDaLozthTZ/R6rQ736g7XE7khFd2lNWT8mzCjPQ=`

Check kata-containers logs and you'll notice the panic.

### CoCo version information

CoCo v0.11.0

### What TEE are you seeing the problem on

None

mkulke:
```json
{
    "Name": "quay.io/fidencio/prueba",
    "Digest": "sha256:756b10356eacdadb63b00b96cf4b3a4bb9658c424c2fc07c87ce475d256bd87f",
    "RepoTags": [
        "encrypted"
    ],
    "Created": "2024-11-14T11:42:30.649705443+01:00",
    "DockerVersion": "",
    "Labels": {
        "org.opencontainers.image.ref.name": "ubuntu",
        "org.opencontainers.image.version": "20.04"
    },
    "Architecture": "amd64",
    "Os": "linux",
    "Layers": [
[...]

PR #842, opened 2024-12-11
image-rs: fix duplicated image layer detective logic

Xynnn007:
Before this commit, when we pull images who have two encrypted layers whose corresponding plaintext layers are same. For example `quay.io/fidencio/prueba:encrypted`, there will be an error like

```
index out of bounds: the len is 25 but the index is 25
```

This is caused by the deduplication logic inside image pull logic. On one hand, it will delete duplicated layers recorded inside image manifest, who reflectes the encrypted layers/blobs. On the other hand, it will delete duplicated layers recorded inside the config.json, who reflects the plaintext of the layers.

The image encryption logic will generate a random symmetric key for each layer. Thus even the same plaintext layer would be encrypted into different blobs/layers. Thus after deduplication, we might have more layers for image manifest.

This patch changes the deduplicating logic, by only check the layer digests inside image manifest, s.t. even if there are two same plaintext layers, we will pull and decrypt both of them. It's ok to do some optimization later if a fully analyzation is taken.

Fies #840

cc @fidencio @mkulke @arronwy

### DIFF
--- a/image-rs/src/pull.rs
+++ b/image-rs/src/pull.rs
@@ -162,6 +162,10 @@ impl<'a> PullClient<'a> {
         };
 
         let decryptor = Decryptor::from_media_type(&layer.media_type);
+
+        // There are two types of layers:
+        // 1. Compressed layer = Compress(Layer Data)
+        // 2. Encrypted+Compressed layer = Compress(Encrypt(Layer Data))
         if decryptor.is_encrypted() {
             let decrypt_key = tokio::task::spawn_blocking({
                 let decryptor = decryptor.clone();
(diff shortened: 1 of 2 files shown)


fidencio:
Let me give it a quick try on my side and, as always, thanks @Xynnn007!

Xynnn007:
Fixed the lint error.

fidencio (approved):
lgtm and works as expected, thanks @Xynnn007!

image-rs/src/pull.rs
```diff
@@ -162,6 +162,10 @@ impl<'a> PullClient<'a> {
         };
 
         let decryptor = Decryptor::from_media_type(&layer.media_type);
+
+        // There are two types of layers:
+        // 1. Comporessed layer = Compress(Layer Data)
```
mkulke, line 167:
```suggestion
        // 1. Compressed layer = Compress(Layer Data)
```

mkulke (approved):

PR #843, opened 2024-12-11
images-rs: clarify comments in image config file

fitzthum:
I said I would update these comments back when we refactored the config file and now I have finally done it.

No functional changes. Hopefully this helps. These comments are fairly important because they describe a couple of features that aren't widely documented.

### DIFF
--- a/image-rs/src/config.rs
+++ b/image-rs/src/config.rs
@@ -9,19 +9,20 @@ use std::path::{Path, PathBuf};
 
 use crate::snapshots::SnapshotType;
 
+/// By default use a work dir in `/run` because for confidential guests `/run`
+/// is typically in a `tmpfs` which is backed by encrypted memory.
 pub const DEFAULT_WORK_DIR: &str = "/run/image-rs/";
 
-/// Default policy file path.
+/// Default path to policy file in KBS or locally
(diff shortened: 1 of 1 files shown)


huoqifeng (approved):
LGTM
Thanks! @fitzthum

image-rs/src/config.rs
```diff
@@ -38,51 +39,45 @@ pub struct ImageConfig {
...
+    /// or it can be stored locally (somewhere in the rootfs).
     ///
-    /// At the same time, some enhencements based on CoCo is used, that is the
-    /// `keyPath` field can be filled with a KBS URI like `kbs:///default/key/1`
+    /// For example, `file:///etc/image-policy.json` for a local policy file
+    /// or `kbs:///default/policies/image-security-policy.json` for a policy
+    /// file that has been provisioned to a KBS.
     ///
-    /// This value defaults to `None`.
+    /// If this field is not set (which is the default) not image security policy
```
Xynnn007, line 61:
```suggestion
    /// If this field is not set (which is the default) no image security policy
```

Xynnn007 (approved):
Thanks for this! @fitzthum

mkulke (approved):

PR #844, opened 2024-12-12
ci: publish arm64 artifacts

seungukshin:
Publishing arm64 artifacts to the project's GHCR using cross-compilation. This is an interim step. The native compilation will be introduced when arm runner is set up for CI.

`publish-artifacts` workflows are executed in my forked repo. except s390x architecture. (https://github.com/seungukshin/guest-components/actions/runs/12298710877)

### DIFF
--- a/.github/workflows/publish-artifacts.yml
+++ b/.github/workflows/publish-artifacts.yml
@@ -18,13 +18,15 @@ jobs:
     strategy:
       matrix:
         platform: [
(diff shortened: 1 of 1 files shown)


mkulke (approved):
lgtm. nice that cross-compiling works ootb.

huoqifeng (approved):
LGTM
thank you! @seungukshin

PR #855, opened 2024-12-24
Chore update josekit version & update rust toolchain version to v1.83.0

Xynnn007:
please see concrete commit messages for each commit.

Close #851


fitzthum (approved):

Xynnn007:
Noticed that we updated the rust toolchain version.

mythi:
> Noticed that we updated the rust toolchain version.

did you intent to make it 1.80.x which is now the MSRV?

PR #860, opened 2024-12-30
AA/attester: support to read CCEL for kernel older than v6.4

Xynnn007:
linux kernels older than v6.4 does not support to read CCEL from acpi sysfs. This patch supports to read CCEL from dev/mem

This is accomplished by reading CCEL's physical address from ACPI table description. Then read the CCEL from /dev/mem.

ACPI table to include CCEL patch of linux kernel

https://github.com/torvalds/linux/commit/4f855dcead6c5be0a48a2779eeecb170ec144534

### DIFF
--- a/attestation-agent/attester/Cargo.toml
+++ b/attestation-agent/attester/Cargo.toml
@@ -59,7 +59,7 @@ all-attesters = [
 # quotes. It's an unconditional dependency for tdx-attester since that is the only way to
 # generate TDX quotes with upstream kernels.
 tsm-report = ["tempfile"]
-tdx-attester = ["scroll", "tsm-report", "tdx-attest-rs"]
+tdx-attester = ["scroll", "tsm-report", "tdx-attest-rs", "tokio/io-util"]
 sgx-attester = ["occlum_dcap"]
 az-snp-vtpm-attester = ["az-snp-vtpm"]
 az-tdx-vtpm-attester = ["az-snp-vtpm-attester", "az-tdx-vtpm"]
--- a/attestation-agent/attester/src/tdx/ccel.rs
+++ b/attestation-agent/attester/src/tdx/ccel.rs
@@ -1,0 +1,101 @@
+// Copyright (c) 2024 Alibaba Cloud
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use anyhow::{bail, Context};
+use scroll::Pread;
+use tokio::io::{AsyncReadExt, AsyncSeekExt};
+
+use crate::tdx::TdxAttester;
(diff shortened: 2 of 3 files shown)


mythi (changes requested):
> linux kernels older than v6.4 does not support to read CCEL from acpi sysfs.

do you have a particular kernel version in mind?  We haven't clearly said how old we support I think going back to ~2 years is already a bit too much, IMO. kata-runtime only supports the configfs-tsm based get quote which implies the minimum version is 6.7.

Xynnn007:
> do you have a particular kernel version in mind? We haven't clearly said how old we support I think going back to ~2 years is already a bit too much, IMO. kata-runtime only supports the configfs-tsm based get quote which implies the minimum version is 6.7.

Yes. We want to run some components, s.t. AA on aliyun OS which is based on kernel older than 6.7 now. The latest patch of config-tsm things are not supported. btw, this is not directly related to kata.

mythi:
What's the version you are targeting?

Xynnn007:
> What's the version you are targeting?

It does not make sense to say the kernel version. We use a relatively old kernel version based on 5.10, but did a lot of patches on that.

mythi:
The reason for my question is that 6.2 added the tdx guest driver so I guess there's a short time window in history where this could be justified somehow but sounds like that is not the case here. 

>  but did a lot of patches on that.

I think you need to add https://github.com/torvalds/linux/commit/4f855dcead6c5be0a48a2779eeecb170ec144534 to it

Xynnn007:
> The reason for my question is that 6.2 added the tdx guest driver so I guess there's a short time window in history where this could be justified somehow but sounds like that is not the case here. 
> 
> >  but did a lot of patches on that.
> 
> I think you need to add https://github.com/torvalds/linux/commit/4f855dcead6c5be0a48a2779eeecb170ec144534 to it

right. We have pushed forward this, but it takes some time. Before that, i think it makes sense to get this merge. This will not be related to tdx guest device but only acpi table of guest memory.

mythi:
IMO, merging this is not justified.

Xynnn007:
It might not be good to include the code in CoCo directly.

Xynnn007:
@mythi This patch will not break any of current behavior.

mythi:
> @mythi This patch will not break any of current behavior.

I'm not concerned about that. The problem is that the upstream project needs to maintain workarounds needed by some special downstream configurations. That's dead code for the official configurations and we don't even know how long this is going to be needed. I agree it's not a lot of code but if we start accepting workarounds it's hard to say no next time a similar case happens.

PR #862, opened 2025-01-02
move to LazyLock

mythi:


fitzthum (approved):
LGTM.

Are we officially tracking our msrv somewhere?

Xynnn007:
@fitzthum nowhere as I know. Up to now only the GH workflow ci yamls have define the rust toolset versions but we do not have a place to publicly tell.

- msrv update policy https://github.com/confidential-containers/confidential-containers/issues/258
- where to tell developers the msrv. we could add to README of each sub-projects.

PR #864, opened 2025-01-03
AA/attester: move CCEL read logic to eventlog_rs crate

Xynnn007:
We move CCEL reading logic to `eventlog_rs` crate to keep the logic clean in gc.


mythi:
It looks this just moves the already rejected functionality to a separate crate that is pulled here. It does not change my review opinion about the change.

PR #884, opened 2025-01-21
image-rs: fix broken link

fitzthum:
I can't remember if we ever had a separate image-rs repo. Either way, our link checker has recently started complaining about this, and indeed the link doesn't exist.


mkulke (approved):

mythi (approved):
LGTM too.

IIRC Guest-components was renamed from image-rs. The redirect works for me still but it's better to make CI happy asap.

PR #885, opened 2025-01-22
[WIP] AMD SNP: add derived-key endpoint

eldios:
Add custom feature for AMD SNP to get a `derived-key` via a dedicated REST API endpoint.


Xynnn007:
Hi @eldios , what is the DerivedKey for?

eldios:
It's used as a unique artifact that is bound to the machine and a specific trait you can include (policy, measurement, machine ID, etc..) and that only lives in the Enclave and thus is supposed secure and reproducible (in the same setup/condition).

fitzthum:
Key derivation is a somewhat generic idea in confidential computing, but I'm not sure if platforms other than SNP really support this.

On SNP key derivation is not widely used, mainly because you can only generate a particular key from inside a particular enclave, which is tricky for orchestration. That said, it could potentially be useful for workloads that want to store some data then will be access by the same workload (on the same node) later.

I'm not against exposing this functionality.

attestation-agent/protos/attestation-agent.proto
```diff
@@ -54,6 +60,7 @@ message GetTeeTypeResponse {
 }
 
 service AttestationAgentService {
+    rpc GetDerivedKey(GetDerivedKeyRequest) returns (GetDerivedKeyResponse) {};
```
Xynnn007, line 63: Could we add some documents or notes about the RPC, like
what is it for?
what does it require from the underlying TEE?
what attributes of the key would/would not have?

Xynnn007 (commented):
Hi @eldios , could you help to squash the commits with some commit messages? It would help review

attestation-agent/protos/attestation-agent.proto
mythi: This seems rather intrusive change (given the API impact) to enable an SNP specific feature.  Might be easier to just make `/dev/sev_guest`  available to the workload and have the ioctl logic implemented there.

Xynnn007: I am not sure whether the interface that Lele wants to add is to use the TEE hardware features as a trusted cryptographic seed source to derive keys for other cryptographic operations. If so, SGX/TPM have similar interfaces to derive key, and I think it can be further designed as a high-level interface, using different hardware features at the bottom layer.

mythi: Right, it would be good to understand (CoCo) use-case first.

fitzthum: one twist is that I'm not sure configfs supports this stuff

mateo-moon:
Interested in reviving this — it's the building block for a real need. On stock `kata-qemu-snp`, workload containers can't reach `/dev/sev-guest` (absent from the container devfs even when privileged), so `SNP_GET_DERIVED_KEY` is unreachable from workload code and the only path to a stable, measurement-bound key is this AA endpoint.

I opened confidential-containers/guest-components#1678 proposing to generalize this into a TEE-agnostic `Attester::get_derived_key(context)` (SNP backed by the `sev` crate here, TDX/SGX/CCA via their native mechanisms) plus a CDH `seal`/`unseal` consumer that uses the derived key as a local KEK for `secure_mount` — i.e. KBS-independent persistent-volume sealing. Would you be open to rebasing this PR toward that abstract shape, or to me helping carry it? Happy to coordinate rather than open a competing PR.

eldios:
> I opened #1678 proposing to generalize this into a TEE-agnostic `Attester::get_derived_key(context)` (SNP backed by the `sev` crate here, TDX/SGX/CCA via their native mechanisms) plus a CDH `seal`/`unseal` consumer that uses the derived key as a local KEK for `secure_mount` — i.e. KBS-independent persistent-volume sealing. Would you be open to rebasing this PR toward that abstract shape, or to me helping carry it? Happy to coordinate rather than open a competing PR.

Sure I'm happy to revive this PR and bring it forward in a more agnostic way. Let me give it a stab and bring it in the right direction.

@mateo-moon would you mind checking my profile and reach out in one of the many possible ways listed there? thanks!

mythi:
> it's the building block for a real need. On stock `kata-qemu-snp`, workload containers can't reach `/dev/sev-guest` (absent from the container devfs even when privileged), so `SNP_GET_DERIVED_KEY` is unreachable from workload code and the only path to a stable, measurement-bound key is this AA endpoint.

It's done trivially via [NRI device-injector](https://github.com/containerd/nri/blob/main/plugins/device-injector/README.md) to unblock you. 

```yaml
   mounts.noderesource.dev/container.c0: |+
      - source: "configfs"
        destination: "/sys/kernel/config"
        type: "configfs"
        options:
          - "nosuid"
          - "noexec"
          - "nodev"
    devices.noderesource.dev/container.c0: |+
      - path: /dev/tdx_guest
        type: c
        major: 10
        minor: 258
```

```
$ kubectl exec -it coco-guest-dev -- /bin/ls -la /dev/tdx_guest
crw-------    1 root     root       10, 258 Sep 14 06:29 /dev/tdx_guest
$ kubectl exec -it coco-guest-dev -- /bin/ls -la /sys/kernel/config/tsm/report
total 0
drwxr-xr-x    2 root     root             0 Sep 14 06:28 .
drwxr-xr-x    3 root     root             0 Sep 14 06:28 ..
```

So this PR is not needed.

eldios:
> > it's the building block for a real need. On stock `kata-qemu-snp`, workload containers can't reach `/dev/sev-guest` (absent from the container devfs even when privileged), so `SNP_GET_DERIVED_KEY` is unreachable from workload code and the only path to a stable, measurement-bound key is this AA endpoint.
> 
> It's done trivially via [NRI device-injector](https://github.com/containerd/nri/blob/main/plugins/device-injector/README.md) to unblock you.
> 
> ```yaml
>    mounts.noderesource.dev/container.c0: |+
>       - source: "configfs"
>         destination: "/sys/kernel/config"
>         type: "configfs"
>         options:
>           - "nosuid"
>           - "noexec"
>           - "nodev"
>     devices.noderesource.dev/container.c0: |+
>       - path: /dev/tdx_guest
>         type: c
>         major: 10
>         minor: 258
> ```
> 
> ```
> $ kubectl exec -it coco-guest-dev -- /bin/ls -la /dev/tdx_guest
> crw-------    1 root     root       10, 258 Sep 14 06:29 /dev/tdx_guest
> $ kubectl exec -it coco-guest-dev -- /bin/ls -la /sys/kernel/config/tsm/report
> total 0
> drwxr-xr-x    2 root     root             0 Sep 14 06:28 .
> drwxr-xr-x    3 root     root             0 Sep 14 06:28 ..
> ```
> 
> So this PR is not needed.

the two can easily cohexist.

That said last time I tried with AMD-SNP this approach didn't work (the device would not answer).

Unfortunately I never tried with TDX and I don't have access to AMD SNP machines right now so I'm stuck on developing this PR further.

mateo-moon:
Thanks @mythi for pointing me toward NRI, and @eldios for being willing to pick this up again! I went back and tested it on a real AMD SNP node with CoCo `kata-qemu-snp` (Kata 4.0.0). The device did answer in this setup: injecting `/dev/sev-guest` let the workload call `SNP_GET_DERIVED_KEY` successfully, without privileged mode or a configfs mount.

A few results relevant to this discussion:

- Key confirmations matched after a container restart and a compatible fresh VM launch on the same node. Changing the measured kernel command line changed the key.
- With guest policy enforced, deliberately injecting the device into a forbidden container was rejected by the guest. Changing the authorized container's command under the unchanged policy was also rejected.
- Changing the agent policy changed the key, even with the same launch measurement. SNP always mixes in `HOST_DATA`, where CoCo binds the InitData hash, including the agent policy. That corrected my earlier concern about policy binding.

There are still boundaries to keep in mind: two containers authorized under one policy got the same base key, and a different image with matching startup settings was accepted under the unchanged policy in this guest-pull configuration. Image authorization and per-consumer key isolation need their own enforcement; this endpoint as currently proposed would not solve those either.

So, after these tests, I think NRI together with guest policy is enough to solve the SNP device-access/key-derivation issue I brought here. My earlier “only path” claim was wrong, and I don't think we need to revive this PR for my use case. Sorry for the detour, @eldios, and thanks both for the discussion!

eldios:
Ah nice! I guess some of the latest fixes changed the behavior after my tests long ago then!

Thanks for verifying and coming back with the results!

eldios:
closing this for now since it seems that INR works just as good.

If anybody needs this specifically, feel free to reopen and contact me, I'd be happy to help with the work

PR #900, opened 2025-02-10
chore: update rand from 0.8.5 to 0.9.0

Xynnn007:
New rand crate changes
- sub module distributions renamed to distr
- api thread_rng renamed to rng
- OsRng now only supports TryRngCore (probably catch an error)

Manually & Close #898


fitzthum (approved):
LGTM. Thanks for the update

Issue #901, opened 2025-02-10
image-rs fails to mount images with many layers


squarti:
### Describe the bug

If an image has many layers, the mount command may fail with the following error:

```
failed to mount "overlay" to "/tmp/test1/rootfs", with error: ENOENT: No such file or directory
```

The problem happens because mount has a 4K limit for the options param. 

The number of layers image-rs can handle is significantly reduced when compare with containerd because layers a stored in path which includes the layer digest (ie. /run/image-rs/layers/sha256_055008870f0b0eef21dc8f9be651f07c3f52fa33724d899f8d14bded0a4fa38d).

### How to reproduce

Pull an image with many layers (for instance: quay.io/silenio_quarti/layers:latest)

```
cargo run --example test -- quay.io/silenio_quarti/layers /tmp/test1
```

sample `example/test.rs`:
```
use image_rs::image::ImageClient;
use std::path::Path;
use std::env;

#[tokio::main]
async fn main() {
    let args: Vec<String> = env::args().collect();
    let image = args[1].as_str();
    let rootfs = args[2].as_str();
    println!("Pulling {} to {}", image, rootfs);
    let mut image_client = ImageClient::default();
    image_client
            .pull_image(image, Path::new(rootfs), &None, &None)
            .await
            .expect("failed to download image");
    println!("done.");
}
```

### CoCo version information

0.11.0

### What TEE are you seeing the problem on

None

### Failing command and relevant log output

```shell
output:

Pulling quay.io/silenio_quarti/layers to /tmp/test1
thread 'main' panicked at image-rs/examples/test.rs:15:14:
failed to download image: failed to mount "overlay" to "/tmp/test1/rootfs", with error: ENOENT: No such file or directory


dmesg output showing truncated layer:


[348019.484642] overlayfs: failed to resolve '/run/image-rs/layers/sha256_5c5893bcd73b59801ebc9624b503221ac0646c2f964294889208041a62d': -2


Actual layer:

root@build-podvm:~/guest-components/image-rs# ls -la /run/image-rs/layers/ | grep sha256_5c5893bcd73b59801ebc9624b503221ac0646c2f964294889208041a62d
drwxr-xr-x  3 root root   80 Feb 10 18:06 sha256_5c5893bcd73b59801ebc9624b503221ac0646c2f964294889208041a62dea68c
```

PR #902, opened 2025-02-10
image-rs: use index for layers store path

squarti:
Use index for layer store path instead of layer digest

Fixes: #901


mkulke (commented):
awesome bug report, thanks. Can you describe in more detail what you are doing in your fix in the PR description?

squarti:
A possible issue with the approach taken is that pull client cannot run in parallel since the layer index is managed in memory.  Although, this already seems to be an issue since the meta store is only saved at the end of pull_image.

Xynnn007 (commented):
Thanks for the fix. Btw I did not find any place that shows `mount` does not support data longer than 4096 bytes although the behavior shows this.


Another thing I worry about is synchronization issues. Now `ImageClient` can respond to multiple `pull_image` requests asynchronously. Each `pull_image` request will create a `PullClient`. Assuming multiple pull_image requests are initiated at the same time, it will be possible to obtain multiple identical or close `layers_index`, and overwriting might occur.

The previous method of using hash values ​​to name layers can avoid this risk assuming that the hashes do not collide. This is possible if an auto-incrementing number is used.

If number is used, do we need to consider using PullClient as part of ImageClient and perform operations such as `async_pull_layers` and `pull_manifest` exclusively and atomically?

squarti:
Would moving the counter to `ImageClient` address these issues?

ChengyuZhu6:
I'm thinking would it be possible to use the footprint instead of the full hash value as the layer store name? For example, changing from `run/image-rs/layers/sha256_055008870f0b0eef21dc8f9be651f07c3f52fa33724d899f8d14bded0a4fa38d` to `run/image-rs/layers/sha256_055008870f0b`?

Xynnn007:
> I'm thinking would it be possible to use the footprint instead of the full hash value as the layer store name? For example, changing from `run/image-rs/layers/sha256_055008870f0b0eef21dc8f9be651f07c3f52fa33724d899f8d14bded0a4fa38d` to `run/image-rs/layers/sha256_055008870f0b`?

I have considered it before, but considering that the probability of collision will increase exponentially after truncating the hash value, it may cause unexpected bugs in extreme cases. If we use the self-increasing index as proposed, layer collisions can be absolutely avoid.

mkulke:
is there a way to write this code in a way that we can unit-test the construction of the layer storage paths? this would be good also for specifying what we are doing.

Xynnn007 (approved):
LGTM. @mkulke PTAL

image-rs/src/image.rs
```diff
@@ -93,23 +94,16 @@ pub struct ImageClient {
 
     /// The config
     pub(crate) config: ImageConfig,
+
+    /// The image layer store
+    pub(crate) layer_store: Arc<LayerStore>,
```
mkulke, line 99: I don't think the we need the Arc around the whole layer store, it should be sufficient to only wrap the `AtomicUsize`. Please see the attached patch. We can pass around the layer store via ref

[arc_for_atomicusize_only.patch](https://github.com/user-attachments/files/18797207/arc_for_atomicusize_only.patch)

image-rs/src/layer_store.rs
```diff
@@ -0,0 +1,79 @@
...
+    /// Next layer index
+    layers_index: AtomicUsize,
+}
+
+impl LayerStore {
+    /// Retrieves the next layer index by computing the maximum index from the
+    /// existing layers in data dir.
+    pub fn get_layer_index(data_dir: PathBuf) -> AtomicUsize {
+        let mut next = 0;
+        if data_dir.exists() {
```
mkulke, line 19: what if data_dir doesn't exist, do we need to error out in this case? if no, and if 0 is a proper value, then we can early exit (and safe a level of indentation):

```rust
if (!data_dir.exists() {
  return AtomicUsize::new(0);
}
...
```

image-rs/src/layer_store.rs
```diff
@@ -0,0 +1,79 @@
...
+            if let Ok(paths) = fs::read_dir(data_dir) {
```
mkulke, line 20: what if this op fails, shouldn't we propagate the error, panic or something else? if we want to swallow it for some reason, we can early exit and maybe log it and save a level of indentation.

```rust
let Ok(paths) = fs::read_dir(data_dir) else {
   // maybe log?
   return AtomicUsize::new(0);
}
```

image-rs/src/layer_store.rs
```diff
@@ -0,0 +1,79 @@
...
+                for entry_path in paths.filter_map(|result| result.ok()) {
```
mkulke, line 21: I understand we are ignoring [IO errors](https://doc.rust-lang.org/beta/std/fs/struct.ReadDir.html#errors) here, is this intentional?

image-rs/src/layer_store.rs
```diff
@@ -0,0 +1,79 @@
...
+                    if let Some(name) = entry_path.file_name().to_str() {
```
mkulke, line 22: what if the path can't be converted to str? Should we handle the None case?

image-rs/src/layer_store.rs
```diff
@@ -0,0 +1,79 @@
...
+                        let n = name.parse::<usize>().unwrap_or(0);
```
mkulke, line 23: do we expect a parsing error here?

image-rs/src/image.rs
squarti: I am thinking whether we need Arc. AtomicUsize should be thread safe.

squarti: Now I remember why I changed from lifetime to Arc.  There is validation [error](https://github.com/confidential-containers/guest-components/actions/runs/13330930001/job/37234635058?pr=902) which I do not quite understand.

Any suggestions how to handle it?

mkulke: While types like `AtomicUsize` provide thread-safe operations, sharing them across concurrently running tasks (like our image pulling algo) requires shared ownership semantics. `Arc` ensures that the data remains valid until all tasks are finished using it.

image-rs/src/layer_store.rs
squarti: The file name should be an integer which should not fail to convert since it is a valid Unicode value.  If for some reason, the file names cannot be converted, it will not conflict with real layer directories.

image-rs/src/layer_store.rs
squarti: Similar to above. If there is a parsing error, the file name is not an integer. It should not conflict with real layers.

image-rs/src/layer_store.rs
squarti: Added logging.

image-rs/src/layer_store.rs
squarti: Added logging.

ChengyuZhu6 (approved):
LGTM, thanks @squarti . While I'm concerned about the ci failure with `ubuntu-24.04-arm`.

fitzthum:
CI Green. @mkulke are we good to merge this?

mkulke (commented):
at the moment we're just logging the errors instead of bailing out. since this is a sensitive area of the code, I would prefer if we would propagate the errors and fail on anything that we encounter and do not expect, i.e. `get_layers_index()` should return a `Result<>` since it's a fallible operation.

image-rs/src/layer_store.rs
squarti: I changed to propagate the error

image-rs/src/layer_store.rs
squarti: I changed to propagated the error

image-rs/src/layer_store.rs
```diff
@@ -0,0 +1,103 @@
...
+            let entry_path = path?;
+            // these should not fail since each file name must be an index.
+            // If for some reason, file name is not an index it will not
+            // cause any conflicts with new and existing layers.
+            if let Some(name) = entry_path.file_name().to_str() {
+                let n = name.parse::<usize>().unwrap_or(0);
+                if n >= next {
+                    next = n + 1;
+                }
+            }
```
mkulke, line 35:
```suggestion
            let n: usize = entry_path
                .file_name()
                .to_str()
                .ok_or(anyhow!("entry path isn't valid utf"))?
                .parse()?;
            if n >= next {
                next = n + 1;
            }
```

image-rs/src/image.rs
```diff
@@ -153,6 +147,7 @@ impl ImageClient {
         let config = ImageConfig::try_from(work_dir.join(CONFIGURATION_FILE_NAME).as_path())
             .unwrap_or_else(|_| ImageConfig::new(work_dir.clone()));
         let meta_store = MetaStore::try_from(work_dir.join(METAFILE).as_path()).unwrap_or_default();
+        let layer_store = LayerStore::new(work_dir).unwrap_or_default();
```
mkulke, line 150:
```suggestion
        let layer_store = LayerStore::new(work_dir).unwrap_or_else(|e| {
            error!("failed to construct layer store", e);
            LayerStore::default()
        };
```

mkulke (commented):
thanks for accommodating, a few nits.

it's not great that we swallow the errors at the upper layers now, but apparently that is also what we do with the meta store. It's still an improvement because we don't get something undefined on malformed images, but a predictable and unusuable 0-ed out layerstore. if we unwrap we still need to log the errors though (see suggestion), otherwise it won't be fun to debug.

image-rs/src/layer_store.rs
squarti: Added error handling as suggest by @mkulke

image-rs/src/layer_store.rs
squarti: Added error handling as suggest by @mkulke

mkulke (approved):

PR #905, opened 2025-02-12
attester: tdx: make libtdx-attest optional

mythi:
TDX Guest driver support is problematic: configfs-tsm currently
does not allow to request bare TD reports. It's always a signed
TD quote that needs exits to QGS for signing. However, for the
initdata checks, we only need the TD report.

Linux still supports the /dev/tdx_guest ioctl() for getting the
report. Since the early days, we have used libtdx-attest wrappers
for the ioctls(). However, getting the C libraries and headers
installed in various places has turned out to be problematic.
Furthermore, it isn't even necessary for uses cases like
kbs-client (where only the configfs-tsm GetQuote is needed).


attestation-agent/attestation-agent/Cargo.toml
```diff
@@ -76,6 +76,7 @@ all-attesters = [
     "cca-attester",
 ]
 tdx-attester = ["kbs_protocol?/tdx-attester", "attester/tdx-attester"]
+tdx-attester-oot = ["kbs_protocol?/tdx-attester-oot", "attester/tdx-attester", "attester/tdx-attest-dcap-ioctls"]
```
fitzthum, line 79: Can we add a comment about what this is. Is `oot` the best name for the feature? This is referring to a non-upstream kernel? Maybe we should just call it `-libtdx` or something.

fitzthum (approved):
LGTM but one nit about the name of the feature

attestation-agent/attestation-agent/Cargo.toml
mythi: I renamed it to `tdx-attester-libtdx` and added a README entry.

PR #910, opened 2025-02-15
[Build] Stop build.rs from triggering due to protoc output for hub

portersrc:
Currently if we clone the repo and issue `make` from `confidential-data-hub`, the `build` target fires, which runs this:
```cd hub && $(RUST_FLAGS) cargo build $(release) --no-default-features --features "$(features)" $(binary) $(LIBC_FLAG)```

During normal development, repeatedly running this command will take ~10s. This is because the build.rs scripts are designed to produce .rs files from .proto files, which causes cargo to see dirty/stale .rs source and thus rebuild. This PR only triggers those build.rs steps when the proto files change. This should reduce that normal development flow (i.e. `make` from `confidential-data-hub/`) to the expected 0.5s.

### DIFF
--- a/attestation-agent/kbs_protocol/build.rs
+++ b/attestation-agent/kbs_protocol/build.rs
@@ -27,6 +27,8 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
             Ok(())
         }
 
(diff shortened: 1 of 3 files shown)


Xynnn007 (approved):
LGTM

PR #911, opened 2025-02-15
[Build] Change default LIBC to musl in cdh Makefile

portersrc:
To the best of my knowledge, we want to use musl by default. The top-level Makefile of guest-components will pass LIBC=musl (if it's not already set). But the CDH Makefile defaults to gnu (perhaps some 2-year-old vestige). A higher-level concern is perhaps how these Makefiles should be invoked, but if we expect musl by default, we should adjust this with this PR.


fitzthum (approved):
I think this is ok. Seems like this should be the default and I think these are the platforms that need `gnu`, but let's double-check.

Xynnn007 (approved):
Nice. `musl` was not set because feature `ehsm` encountered `openssl` which cannot be built with musl. Now with https://github.com/confidential-containers/guest-components/pull/792 CDH can be built with musl.

PR #913, opened 2025-02-18
image-rs: enable rustls-native-certs when using rustls

Xynnn007:
`oci-client/rustls-tls-native-roots` feature will allow oci-client to read local CA certs including self-signed certs in the rootfs when using rustls suites.

Before this patch, if we pull an image with `kata-cc-rustls-tls` feature from self-signed registry whose CA is updated with `update-ca-certificates`, there will be an error

```
failed to pull manifest error sending request for url (https://127.0.0.1:5000/v2/alpine/manifests/latest)
```

With this patch it will succeed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3024,7 +3024,7 @@ dependencies = [
  "nix 0.29.0",
  "nydus-api",
  "nydus-service",
- "oci-client 0.14.0",
+ "oci-client",
  "oci-spec",
  "ocicrypt-rs",
  "prost 0.13.5",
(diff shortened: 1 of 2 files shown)


mythi (approved):
> pull an image with `kata-cc-rustls-tls` feature from self-signed registry whose CA is updated with `update-ca-certificates`, there will be an error

LGTM, thanks! With `kata-cc-native-tls` this works automatically but the hardest part is to get these custom certs added to the 'kata w/ coco' rootfs.

Xynnn007:
Ok, let me add another commit to support the same for cosign signature seeing that https://github.com/sigstore/sigstore-rs/pull/437 gets merged.

fitzthum (approved):
LGTM

PR #917, opened 2025-02-20
resource-uri: handle two-part resource paths

pmores:
Two-part resource paths have so far resulted in an error.  This however seems incorrect since the repository part of a resource path is specified as optional in the kbs-protocol definition(*).

This change could conceivably break a caller that relies on ResourceUri to bail out if a resource path only has two parts.  Arguably such a caller would itself be faulty though, given the kbs-protocol definition.

(*) https://github.com/confidential-containers/trustee/blob/main/kbs/docs/kbs_attestation_protocol.md#secret-resource

### DIFF
--- a/attestation-agent/deps/resource_uri/src/lib.rs
+++ b/attestation-agent/deps/resource_uri/src/lib.rs
@@ -57,17 +57,21 @@ impl TryFrom<url::Url> for ResourceUri {
 
         let path = &value.path()[1..];
         let values: Vec<&str> = path.split('/').collect();
-        if values.len() == 3 {
-            Ok(Self {
-                kbs_addr: addr,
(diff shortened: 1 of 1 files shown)


pmores:
Please do review carefully since I'm very new here and chances of me overlooking something are definitely solidly non-zero.

fitzthum (approved):
Seems good. One nit.

Xynnn007:
Please see my comment https://github.com/confidential-containers/trustee/pull/712#issuecomment-2673211086

pmores:
Closing this since a decision was taken to go in the opposite direction, i.e. make the repository part of resource path mandatory (implemented by trustee [PR #720](https://github.com/confidential-containers/trustee/pull/720)).

PR #919, opened 2025-02-21
cdh: allow Trustee to bump Aliyun KMS plugin

mythi:
Not an ideal fix but allows trustee to keep both of its guest-components dependencies pointing to the same version.

More details: https://github.com/confidential-containers/guest-components/pull/816#issuecomment-2537669124


Xynnn007 (approved):

fitzthum (approved):
Ok

Issue #918, opened 2025-02-20
Flaky Tests


fitzthum:
Anyone who has submitted a PR recently has probably noticed that the `image-rs` tests take a few tries. I think this is just because we aren't always able to pull the image. This seems to happen most frequently on s390x but maybe it is elsewhere too. It's not a huge inconvenience but maybe there is a way we can fix this?

cc @BbolroC

fitzthum:
For example https://github.com/confidential-containers/guest-components/actions/runs/13429528372/job/37546759608?pr=916

BbolroC:
Thanks for the notice. I will figure it out soon.

mythi:
> Interestingly, that PR didn’t trigger the CI job we’re looking at.

CDH workflow triggers were not updated when the pull functionality got moved there: changes to `image-rs/*` do not trigger CDH tests even if CDH now has a dependency to `image-rs`.

Xynnn007:
> CDH workflow triggers were not updated when the pull functionality got moved there: changes to `image-rs/*` do not trigger CDH tests even if CDH now has a dependency to `image-rs`.

That is the case.

BbolroC:
> > Interestingly, that PR didn’t trigger the CI job we’re looking at.
> 
> CDH workflow triggers were not updated when the pull functionality got moved there: changes to `image-rs/*` do not trigger CDH tests even if CDH now has a dependency to `image-rs`.

Then, I think it would be great time to extend our dependency graph. 😉

BbolroC:
Update:

The tests are passing after cleaning up `/run/image-rs/` before running the test. Shout-out to @Xynnn007 ❤

fitzthum:
Ok, we fixed the tests that were failing every time, but there are also some flaky jobs.

For example, https://github.com/confidential-containers/guest-components/actions/runs/13515374662/job/37762988211#step:20:474

We're pulling a container from the aliyun container service here. I wonder if we want to copy these containers to another registry? We could also keep that registry but skip that test on bare metal runners that might have more trouble connecting to it.

mythi:
> I wonder if we want to copy these containers to another registry?

+1

Xynnn007:
Let me help with this.

PR #921, opened 2025-02-21
ci: allow CDH ci to be triggered by image-rs changes

Xynnn007:
image-rs now is a dependency of CDH, thus any change upon image-rs should be test on CDH too.

related to #918

### DIFF
--- a/.github/workflows/cdh_basic.yml
+++ b/.github/workflows/cdh_basic.yml
@@ -5,12 +5,14 @@ on:
     branches:
       - "main"
     paths:
+      - 'image-rs/**'
       - 'confidential-data-hub/**'
       - '.github/workflows/cdh_basic.yml'
       - 'Cargo.toml'
       - 'Cargo.lock'
   pull_request:
     paths:
+      - 'image-rs/**'
       - 'confidential-data-hub/**'
       - '.github/workflows/cdh_basic.yml'
       - 'Cargo.toml'


BbolroC (approved):
lgtm thanks @Xynnn007 !

fitzthum:
I think this may be causing some issues in the CI like https://github.com/confidential-containers/guest-components/actions/runs/13465974015/job/37631791816?pr=922

Xynnn007:
This is not caused by the commit. Test locally and the commit https://github.com/confidential-containers/guest-components/commit/8a5fa7fade95310799c4e3563c5b100420faff97 and https://github.com/confidential-containers/guest-components/commit/29711496a38bad2cb9597d8954b306435f95a3e4 togetherly causes this. 


But there is nothing wrong with the commits themselves. It is the repeated build process that occurred in the test, which caused the same binary file to be found when the mv command was executed. I'll make a PR to fix this.

Xynnn007:
https://github.com/confidential-containers/guest-components/pull/923

mythi:
> But there is nothing wrong with the commits themselves. It is the repeated build process that occurred in the test,

TBH I'm slightly worried about the CI workflow triggers (we had two incidents last week) not covering everything. What is the repeated build process that was not run in #910?

mythi:
> > But there is nothing wrong with the commits themselves. It is the repeated build process that occurred in the test,
> 
> TBH I'm slightly worried about the CI workflow triggers (we had two incidents last week) not covering everything. What is the repeated build process that was not run in #910?

Looking at #923 I believe the challenge is that the test execution is partly defined by `Makefile`s. Perhaps we should prefer raw workflow rules for all the builds and tests.

PR #922, opened 2025-02-21
deps: update to protobuf 3.7.1

fitzthum:
Since we weren't explicitly setting protobuf to 3.5.1, Cargo recently started giving us a new version.

Normally this would be fine, but we have some protos that we don't build every time. Specifically, the protos for the image service are only rebuilt when you build with a certain feature. As such, these protos were still expecting version 3.5.1 even though we were using 3.7.1.

To fix, I regenerated these protos with the new version of protobuf. I have also pinned protobuf to 3.7.1 to avoid similar issues in the future.

I have also updated all the other proto files even though I think most of them will be rebuilt at build time. We might want to adopt just one approach for building these things.

Finally, I updated ttrpc beacuse the existing version did not work with protobuf 3.7.1


ryansavino (approved):
LGTM. Thanks @fitzthum. This was causing me a large headache when building the kata tarballs. We may need to do some dependency adjusting in kata, but this should help.

Cargo.toml
```diff
@@ -42,7 +42,7 @@ log = "0.4.25"
 nix = "0.29"
 openssl = "0.10"
 prost = "0.13"
-protobuf = "3.5.1"
+protobuf = "=3.7.1"
```
Xynnn007, line 45: I think we might not need the `=` here to fix the version?

Xynnn007 (commented):
Thanks for this. https://github.com/confidential-containers/guest-components/pull/923 will help to fix the musl error. Not sure if s390x error is by chance.

Cargo.toml
fitzthum: It's not required to fix things today, but I am thinking we should pin the version so that we don't run into this problem again. Otherwise the build will break again next time there is an update. Hopefully depandabot will tell us when a new version is available and we can update manually.

In the future we might want to change how the image_pull protos are built to resolve this

Xynnn007 (approved):
lgtm. We would need a rebase

PR #923, opened 2025-02-23
CI: delete musl build for CDH

Xynnn007:
Due to https://github.com/confidential-containers/guest-components/commit/29711496a38bad2cb9597d8954b306435f95a3e4 musl build will be triggered when the platform is not s390x and powerpc during the `Build and install` step. Thus we do not need this new round to do a duplicated musl build test.


mythi (approved):
LGTM to get the CI going again.

PR #928, opened 2025-02-25
Revert moving KMS into CDH

fitzthum:
See discussion [here](https://github.com/confidential-containers/guest-components/pull/816#issuecomment-2678907212)

I also included a commit from https://github.com/confidential-containers/guest-components/pull/922 so that I could successfully rebuild the lock file. I will drop that soon.

I had to massage a few things with the hub Cargo.toml.

I am open to other approaches here, but I am going to use this locally at least to keep moving on the multi-device work, which requires me to bump guest-components in Trustee.


mythi:
> I am open to other approaches here,

This is probably the best approach. I hacked `confidential-data-hub` a bit to get the dependencies cleaned up for Trustee update but could not get that to work easily. That could be an alternative.

In addition to the revert, we could think about dropping `ehsm` since the upstream is no longer maintained.

fitzthum:
cc @mkulke

portersrc (approved):
lgtm for now
I had to rebase the mesh PR on this in order to move forward

mkulke (approved):
lgtm. retriggered the failing s390x job

fitzthum:
Rebasing

PR #933, opened 2025-02-27
resource-uri: make repository part of resource path mandatory in docs

pmores:
This is to align KBS URI documentation with trustee PR #720(*) which makes the repository part of resource path mandatory.  Note that no code change seems necessary here since the actual 'ResourceUri' implementation has in fact already treated the repository part as mandatory, bailing out if it was missing.

(*) https://github.com/confidential-containers/trustee/pull/720


Xynnn007 (approved):
LGTm. Thanks!

dcmiddle (approved):

PR #941, opened 2025-03-10
coco_keyprovider: abondon shadow dependency

Xynnn007:
shadow is actually not used in the project. This will resolve the CI error like https://github.com/confidential-containers/guest-components/actions/runs/13758440224/job/38469596953?pr=940


fitzthum (approved):
LGTM

I wonder why we added this in the first place.

Xynnn007:
we once use this to add some build infomation to the main function but soon abandoned

PR #944, opened 2025-03-17
image-rs | Add Registry Configuration

### LINKED ISSUE #896: Support user-defined container registry config in image-rs
https://github.com/containers/image/blob/main/docs/containers-registries.conf.5.md provides a standard of container registry configuration. It allows users to control including but not limited to the following settings
- Registry remapping
- Registry level blocking
- Registry level insecure HTTP support

This feature could help to deal with the following user cases
- registries that is not running in HTTPS, like https://github.com/confidential-containers/guest-components/issues/861
- pulling images with special network situations, like connecting to a mirrored registry. like https://github.com/confidential-containers/guest-components/issues/601

Xynnn007:
Related to #896


image-rs/src/registry/mod.rs
```diff
@@ -0,0 +1,494 @@
...
+        // The original rule is to add registries to handle an unqualified image.
+        // Crate `oci-spec`'s parser logic of [`oci_client::Reference`] will automatically
+        // add `docker.io` to unqualified images.
+        // Thus this call will try to add tasks for image pull request whose registry
+        // is `docker.io`.
+        //
+        // TODO: This is sometimes not a good idea, as the user might want to
+        // pull images like `docker.io/foo/bar` directly but current implementation will add
+        // tries from registries in `unqualified_search_registries` We can handle this later.
+        if image_reference.registry() != "docker.io" {
```
fitzthum, line 299: Would it help the above to check if the registry starts with "docker.io" rather than exactly equaling it?

fitzthum (approved):
LGTM. Couldn't really find anything to complain about an these seem like nice features. I made one comment.

These transformations will be applied before the image policy is checked, right? That's what we want. Might be worth noting it in the docs.

image-rs/src/registry/mod.rs
Xynnn007: The exactly equaling will help to avoid cases where registry is like `docker.io.com` rather than `docker.io`. The source code to add `docker.io` is [here](https://github.com/youki-dev/oci-spec-rs/blob/main/src/distribution/reference.rs#L310)

mythi:
[containerd has its own](https://github.com/containerd/containerd/blob/main/docs/hosts.md) syntax for something similar.

image-rs/src/image.rs
```diff
@@ -197,22 +240,48 @@ impl ImageClient {
...
+        let mut client_config = ClientConfig::default();
+        if task.use_http {
+            client_config.protocol = ClientProtocol::Http;
+        }
+
+        if let Some(no_proxy) = &self.config.skip_proxy_ips {
+            client_config.no_proxy = Some(no_proxy.clone())
+        }
+
+        if let Some(https_proxy) = &self.config.image_pull_proxy {
```
portersrc, line 264: If I understand this correctly, this is kind of curious. A user could specify an https proxy in their `ImageConfig` even though, with the new registry config that you've added, they might want to pull from a local registry with http, for example. Maybe it's better to at least issue a warning of misconfiguration here? (Otherwise what happens? The `oci` crate will happily oblige and try to go through the https proxy before using the local registry you specified?)

portersrc (approved):
lgtm
just 1 question about how this interacts with `ImageConfig` now

fitzthum:
> [containerd has its own](https://github.com/containerd/containerd/blob/main/docs/hosts.md) syntax for something similar.

Interesting. I'm not sure that we necessarily have to be tied to containerd here.

image-rs/src/image.rs
Xynnn007: Yes, this is indeed a corner case, but it does fit the logic of the configuration. I'll add a warning here

Xynnn007: Added the warning https://github.com/confidential-containers/guest-components/pull/944/files#diff-d753fb19cbbddd2cb199db08ea89c5ebe30e51c0edf69dc4414c36390eae9c44R276-R281

Xynnn007:
> [containerd has its own](https://github.com/containerd/containerd/blob/main/docs/hosts.md) syntax for something similar.

Our previous registry auth and image signature all use the standards under the [containers](https://github.com/containers/image/tree/main/docs) repository. We can leave the containerd format as a TODO and make up for it in subsequent work if necessary.


btw, added an integration test commit. Let's see if it passes

mythi:
> > [containerd has its own](https://github.com/containerd/containerd/blob/main/docs/hosts.md) syntax for something similar.
> 
> Interesting. I'm not sure that we necessarily have to be tied to containerd here, but it might be nice.

My main thinking was just that there's no standard spec here. This PR implements a Redhat spec and users coming from the containerd world cannot reuse their configs with CoCo.

Xynnn007:
> My main thinking was just that there's no standard spec here. This PR implements a Redhat spec and users coming from the containerd world cannot reuse their configs with CoCo.

Yes. We can support for that when needed.

Xynnn007:
i think so

PR #945, opened 2025-03-18
image-rs: delete duplicated unit test case

### LINKED ISSUE #918: Flaky Tests

Xynnn007:
the original aliyun registry is personal so the network is not always good thus causing flaky CIs. We consider removing it, because the test case is duplicated with the one of

```
ghcr.io/confidential-containers/test-container-image-rs:cosign-signed
```

and with key

```
test_data/signature/cosign/cosign3.pub
```

Fixes #918

### DIFF
--- a/image-rs/src/signature/policy/cosign/mod.rs
+++ b/image-rs/src/signature/policy/cosign/mod.rs
@@ -279,19 +279,6 @@ mod tests {
         &format!("\
             {{\
                 \"type\": \"sigstoreSigned\",\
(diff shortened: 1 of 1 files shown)


fitzthum (approved):
Great. Thanks

mkulke (approved):

PR #957, opened 2025-03-21
Attest Multiple Devices

fitzthum:
Currently we assume that a guest will only run one attester. This PR allows multiple attesters to run in one guest, allowing us to attest environments where there is a CPU attester but also device attesters. This PR does not support any particular devices, but it introduces a sample device attester to show how it can be done.

Two important pieces of this approach are: 

1) It's non hierarchical. We assume that all attesters run independently. There is no binding between devices. For various reason I think this is a good idea.

2) We add a new identifier for devices known as the device class. This is a string that an attester provides. An attester can set this however they want, but generally it should be a generic term of the type of device that is being attested. This will ultimately show up in the attestation token that represents this guest.

Of course this PR depends on changes in kbs-types and Trustee


fitzthum:
I'll need a little bit of time to sort out failures in all of the deployment cases that I did not test, but feel free to review. I do have this tested e2e locally.

Xynnn007 (commented):
Thanks for this! I have had a quick glance upon the PR, it is impressive. But before diving into details, could you share some idea of the [view](https://github.com/confidential-containers/trustee/issues/650#issuecomment-2750617538)? It might lead to different design pattern

attestation-agent/attestation-agent/src/bin/ttrpc-aa-client.rs
```diff
@@ -101,15 +101,15 @@ pub async fn main() {
...
+            let req = GetTeeTypesRequest {
                 ..Default::default()
             };
             let res = client
-                .get_tee_type(context::with_timeout(TIMEOUT), &req)
+                .get_tee_types(context::with_timeout(TIMEOUT), &req)
                 .await
                 .expect("request to AA");
-            println!("{}", res.tee);
+            println!("{}", res.tees);
```
Xynnn007: Emmm. Still think about the hierarchical design. In the current fashion, it will say something like
```
tdx, nvidia-gpu, some-nic
```

in the hierarchical design, it will say (any name for the first term)
```
platform/cpu/basic: tdx+ccel+aael
devices: [nvidia-gpu, some-nic]
```
or
```
platform/cpu/basic: azure-vtpm-tdx(tdx+vtpm)
devices: [nvidia-gpu, some-nic]
```


wdyt?

attestation-agent/attestation-agent/src/lib.rs
```diff
@@ -72,23 +72,40 @@ pub trait AttestationAPIs {
...
+                    if attester.supports_runtime_measurement() {
+                        info!("Found {:?} for runtime attestation.", tee);
+                        if let Some(_already_found_attester) = runtime_attester {
+                            bail!("Only one attester should support runtime attestation")
+                        }
+                        runtime_attester = Some(attester.clone());
+                    }
+                }
+                runtime_attester.ok_or(anyhow!("No runtime attester found."))?
+            };
```
Xynnn007, line 105: If we reorginaize the code into a hierarchical one, we could avoid the situation judgement here. As `supports_runtime_measurement` actually means `platform/cpu/basic` or something similar

attestation-agent/attestation-agent/src/lib.rs
```diff
@@ -201,12 +226,20 @@ impl AttestationAPIs for AttestationAgent {
     /// Perform the initdata binding. If current platform does not support initdata
     /// binding, return `InitdataResult::Unsupported`.
     async fn bind_init_data(&self, init_data: &[u8]) -> Result<InitDataResult> {
-        self.attester.bind_init_data(init_data).await
+        let mut init_data_result = InitDataResult::Unsupported;
+
+        for (tee, attester) in &self.attesters {
+            match attester.bind_init_data(init_data).await? {
```
Xynnn007, line 232: Probably initdata is similar to runtime attestation, s.t. only one (cpu/basic/platform) supports that .

attestation-agent/attestation-agent/src/bin/ttrpc_dep/server.rs
```diff
@@ -108,30 +108,35 @@ impl AttestationAgentService for AA {
...
+
+        let mut tee_strings = vec![];
+        for tee in tees {
+            let tee_string = serde_json::to_string(&tee)
+                .map_err(|e| {
+                    error!("AA (ttrpc): get tee type failed:\n {e:?}");
+                    let mut error_status = ::ttrpc::proto::Status::new();
+                    error_status.set_code(Code::INTERNAL);
+                    error_status
+                        .set_message(format!("[ERROR:{AGENT_NAME}] AA-KBC get tee type failed"));
```
Xynnn007, line 128: Not related to the current PR. We might need to drop `KBC` term some time

Xynnn007 (commented):
After reading the entire code, I still feel that the hierarchical design is better. We can encapsulate Evidence in a hierarchical manner based on the original code. Each Tee name can be directly used as the submod name of EAR, and then correspond to its own claim. I left some design details in the specific code. Please share your opinions.

fitzthum:
One high-level thing. 

> Each Tee name can be directly used as the submod name of EAR, and then correspond to its own claim.

Regardless of if we do the hierarchical approach or the non-hierarchical approach, we probably want the submod names to be generic so that the EAR token remains generic. Currently this is one of the reasons the TEE class is useful. You end up with submods called `cpu0`, `gpu0`, `gpu1`, `nic0`, etc, which allows the resource policy to remain generic i.e. a policy can release a secret as long as there is some valid cpu and some valid gpu. The user doesn't need to care what platform they are on although they could check that if they want to. 

As for the hierarchical discussion, one more point here is that our attesters are fundamentally not hierarchical. We aren't passing any information from one to another. We simply run whichever attesters are detected (in random order). On the other end of things, the EAR token is not hierarchical. It just has a list of submods. So why insert a hierarchy in between those two things?

attestation-agent/attestation-agent/src/lib.rs
fitzthum: Yes. Although I am thinking that we should introduce init-data at the scope of the AA rather than attach it to TEE evidence; as in, add it as a field in the attestation message of the RCAR. I think we maybe should do something similar for the eventlog, but we can worry about that later.

This is related to the hierarchical question. Rather than have one root TEE which keeps track of the init-data and event log, I think it makes more sense to have that info at the scope of the AA, and then allow attesters and verifiers to use the info if they want to.

Xynnn007: Yes, here it actually just means choosing a tee to bind or check the initdata hash, as for the initdata hash or plaintext itself, it is still processed by AA. I think this is consistent with your idea. A tee "specifically marked" TEE is responsible for that.

Xynnn007:
> Regardless of if we do the hierarchical approach or the non-hierarchical approach, we probably want the submod names to be generic so that the EAR token remains generic. Currently this is one of the reasons the TEE class is useful. You end up with submods called `cpu0`, `gpu0`, `gpu1`, `nic0`, etc, which allows the resource policy to remain generic i.e. a policy can release a secret as long as there is some valid cpu and some valid gpu. The user doesn't need to care what platform they are on although they could check that if they want to.

I found some examples in [the spec](https://www.ietf.org/archive/id/draft-fv-rats-ear-05.html#section-3.2.1). The granularity is very small. For example, the Arm CCA example is "CCA Realm" and "CCA Platform", both together of which are semantically equivalent to verifying the static metrics and signature of tdx.

I think we can modularize these submodules at the solution level based on the current CoCo verification. For example, TDX is not named purely TDX or CPU. In CoCo, it is actually a complete solution of TDX+CCEL+AAEL. We can define a series of ids, such as `coco-tdx-ccel-aael` or a more standardized URI form `submodule://confidentialcontainers.org/remote-attestation/platform/tdx-ccel-aael`.

It will be useful to distinguish `submodule://confidentialcontainers.org/remote-attestation/platform/snp` and `submodule://confidentialcontainers.org/remote-attestation/platform/azure-snp-vtpm`. With the name we can easily know what they mean, s.t. one is for baremetal SNP boot chain measurement, and the other is Azure's SNP + vTPM oot chain measurement.

Apart from this submodule, then we can include submodules for devices like GPUs and NICs.



> As for the hierarchical discussion, one more point here is that our attesters are fundamentally not hierarchical. We aren't passing any information from one to another. We simply run whichever attesters are detected (in random order). On the other end of things, the EAR token is not hierarchical. It just has a list of submods. So why insert a hierarchy in between those two things?

Yes, it is not a hierarchical structure. This may be just a terminology problem. It is just a unique/special submodule/attester/verifier with additional initdata binding and runtime attestation functions. We need to treat it as a special case in the code, so it is better to give it a separate field in the whole process to make the coding easier. Other than that, it is no different from other attester/verifiers.

fitzthum:
Ok, I don't want to be too stubborn here. There are just a lot of pieces to fit together. To me a lot of this flows from the EAR support that we've already done. Anyway, it's good to discuss this change in depth because it is significant. Let me try to be as concise as possible.

> I think we can modularize these submodules at the solution level based on the current CoCo verification. For example, TDX is not named purely TDX or CPU. In CoCo, it is actually a complete solution of TDX+CCEL+AAEL. We can define a series of ids, such as coco-tdx-ccel-aael or a more standardized URI form 

Unfortunately the EAR spec is vague on how the submodule names are generated. I think this is a bit of an oversight because it makes attestation tokens from difference sources less interchangeable.

To me one of the design goals for our EAR implementation is to have the resource policies be decoupled from the hardware. Today, an admin can write simple policies like this.
```
package policy
import rego.v1

default allow = false

allow if {
    input["submods"]["cpu"]["ear.status"] != "contraindicated"
}
```
The big upside of AR4SI (the standard defining the trust vectors in EAR) is that this will work on any platform. You can also write more specific conditions like `input["submods"]["cpu"]["ear.status.executables"] == 2` that are still not hardware dependent. Of course you can still check which platform was attested with something like `input["submods"]["cpu"]["ear.veraison.annotated-evidence"]["snp"]` (I have a pending PR to replace that veraison key btw). It's important to expose the specific platform values, but if the submod were the platform name, you would have to have a different policy for each platform, which is the opposite of what we want with EAR.

I think when we talk about multiple devices, we should keep using a generic submod keys. I agree with what you are saying about how our attesters represent more information than just the platform, but to me this just seems like a potential problem with the names of the verifiers. Remember that the verifier name will still be part of the submod, so people can distinguish between the different platforms.

> This may be just a terminology problem. It is just a unique/special submodule/attester/verifier with additional initdata binding and runtime attestation functions. We need to treat it as a special case in the code, so it is better to give it a separate field in the whole process to make the coding easier.

In some ways the `cpu0` submod name would function as the "base," but it would be a little bit more generic. I think the only place where we need to identify the base/root of the guest is in the policy and when the KBS extracts the TEE-pubkey. For the most part the differences between the types of devices can be captured by the attesters/verifiers themselves.

For init-data, we call `bind_init_data` on every attester. Device attesters that don't support init_data, can simply do nothing here. On the verifier side, the init-data can be tracked by the AA/KBS and then provided as expected init-data to every verifier. If the verifier doesn't support init-data, it simply ignores this. If it does support init-data, it returns an init-data TCB claim. If an admin wants to check the init-data, they can write a policy that checks the tcb claims associated with `cpu0`. If cpu doesn't support init-data, the policy fails.

This is how the report data currently works and the init-data is almost already setup like this. We just have a couple gaps to fill. Note that for the report-data we don't need to check the report data in the policy, but the KBS does extract the TEE pubkey from the report data [here](https://github.com/confidential-containers/trustee/blob/main/kbs/src/token/mod.rs#L18). Going back to the above, if we change the naming scheme for submods, we will need to change this logic i.e. by adding a path for each potential platform.

The eventlog can work in a similar way. The interface is a little different in the guest. Rather than call runtime measurement on every device, I add a function so that the eventlog can check which devices have runtime measurement support. Still, the idea is that it's the attester itself that defines the behavior. On the verifier side, the eventlog doesn't result in its own TCB claims, but verifiers that want to can use the eventlog to generate other claims. Like I said in one of the comments, I think we should move the eventlog out of the TEE evidence in the future, but this isn't too big of a deal.

As I see it, there is really only one special-case, which is for s390x. Here we want to handle the report data differently based on the CPU TEE. This doesn't fit very well with what I just described, but it's easy enough to work around and I think the s390x code might be changing in the future anyway.

tl;dr I don't think we need to distinguish a base device in the Rust code. Rather, attesters and verifiers can implement various behavior that flow into the tcb claims and ultimately the policy. I can understand how it seems simpler to have a base device, but I think it's a bit artificial and potentially limiting in the future.

Xynnn007:
Let me go back and re-examine our opinions and try to translate it into some concrete points.

1. We want AS policies to be able to export EARs that are decoupled from hardware, preventing KBS or other EAR consumers from setting policies for specific platforms.
2. The location of the key in the EAR should not be related to the hardware architecture
3. How to design the code for handling initdata and runtime measurement.
4. How to interactive between KBS and AS.


About 1. 

I am also with this view. However, in the specific implementation process, I found that different architectures do have different meanings. When we consider mapping different architectures to the same AR4SI structure, we need to carefully consider the content items that need to be verified under the current architecture. I thought too much about this problem, and we can solve this problem by writing AS policies more carefully.

About 2.

This is also fine for current implementation because it works now. An optional method is to put it in the EAR extension. This is more semantically correct because the semantics of AS is to verify all TEE evidences and endorse the public key after integration. Hiding it in the AR4SI corresponding to a specific submods does not fit the context.

About 3.

This is why I have been thinking about the hierarchical relationship. At present, it seems that the rtmr and initdata functions are basically held by only one hardware (there will be no hardware in the foreseeable future, even if it is tpm, it will form a common solution with a certain TEE, just like vtpm+snp), that is, the CPU. We don’t need to check it by indexing the first one or traversing it every time, just extract it separately and put it in the structure or code. Or index it by a specific key.

About 4.

The current implementation is to submit all TEE evidence to AS for verification. This is inevitable because the KBS public key needs to be embedded in the Token, and the so-called KBS Token is actually the AS Token. There is a layer of abstraction missing here. I have no opinion on the current implementation, but I hope to refactor this later, that is, call AS verification multiple times to ensure the cleanliness of the AS interface. KBS will issue its own Token or certificate to the client. I will try to solve this work after initdata is implemented based on this design.

mythi:
> This is why I have been thinking about the hierarchical relationship. At present, it seems that the rtmr and initdata functions are basically held by only one hardware (there will be no hardware in the foreseeable future, even if it is tpm, it will form a common solution with a certain TEE, just like vtpm+snp), that is, the CPU. We don’t need to check it by indexing the first one or traversing it every time, just extract it separately and put it in the structure or code. Or index it by a specific key.

It seems to me that we could find a compromise if we go back to some old discussions and also see what else is being worked in the 'composite evidence' problem space. For example, [RATS Conceptual Message Wrapper](https://datatracker.ietf.org/doc/draft-ietf-rats-msg-wrap/) gives suggestions how to define a grouped evidence like what we are discussion here.

After reading [a very old thread](https://github.com/confidential-containers/confidential-containers/issues/176#issuecomment-1842917174) I noticed that it matches what I've had in mind behind my feedback: the top-level structure does not need to change (even if my [earlier suggestion](https://github.com/confidential-containers/trustee/issues/650#issuecomment-2631680720) was slightly different).

We may want to decouple `Tee`  (or `Tees`) from the attesters and simply have independent attesters to contribute the list of evidences included part of `tee-evidence` (e.g. would an attester for the runtime measurements / canonical eventlog work?) . In a closer look, `Tee` in the KBS protocol does not seem that useful if each attester can provide an identifiable tag for its evidence. It could still be there and filled with an id of a "lead attester" (following the concepts in [RATSd](https://github.com/veraison/ratsd/wiki)).

fitzthum:
> 1) I thought too much about this problem, and we can solve this problem by writing AS policies more carefully.

Totally agree. The AS policy is basically the heart of the attestation logic and currently it could be improved a little bit. I think it is probably a good design to have most of the logic there + attester + verifier and have the rest of things mostly be piping.

>  2) This is more semantically correct because the semantics of AS is to verify all TEE evidences and endorse the public key after integration.

EAR tokens support a `key-attestation` field, but it is Veraison extension (like `annotated-evidence`) so it didn't seem worth it to use. That extension seems to be part of the Appraisal, anyway, which is specific to each submod so it would be roughly equivalent. We could define our own extension at the top-level of the EAR. This might be a good idea. I need to think about it a little more.

For now, we should make sure that any device attesters/verifiers check the report data. As long as they do, all the devices will be endorsing the same public key as cpu0.

> 3) We don’t need to check it by indexing the first one or traversing it every time, just extract it separately and put it in the structure or code.

Yeah it does take a little bit of extra work to check the properties of each attester, but most of the time there will only be one attester anyway, maybe two. Honestly, though i don't think it matters that much if there is just one list with all the verifiers or a field plus a list. I don't really like differentiating the verifiers themselves as primary verifiers / device verifiers. That feels a little clunky, but it wouldn't be the end of the world. 

> 4) KBS will issue its own Token or certificate to the client if all verifications succeed.

This might intersect some of the discussions about using SPIFFE or key cloak. I am open to adopting a different type of token for the KBS, although i think the EAR token works decently for this.

fitzthum:
> After reading https://github.com/confidential-containers/confidential-containers/issues/176#issuecomment-1842917174 I noticed that it matches what I've had in mind behind my feedback: the top-level structure does not need to change (even if my https://github.com/confidential-containers/trustee/issues/650#issuecomment-2631680720 was slightly different).

The approach here is explicitly not using composite evidence. Composite evidence isn't about the format of the evidence so much as the process for attesting things. Basically you need to attest things in order, with the CPU coming last so that you can embed the other evidence into the CPU evidence. One of the main tenants of this design is to have independent attesters and verifiers, which is at odds with this. Now we should make sure that we all really believe that this is secure. As long as it is, independent attesters/verifiers are much simpler. To adopt a composite strategy, we would need to change how we invoke our attesters/verifiers somewhat significantly.

> We may want to decouple Tee (or Tees) from the attesters and simply have independent attesters to contribute the list of evidences included part of tee-evidence

One drawback of this PR is that it doesn't handle one attester which represents multiple devices very well. For instance if you have a bunch of the same GPU in a guest, you will only get one submod for all of them. In the future we would probably want to rework this a little bit so that there can be multiple instances of the same attester or one attester can return multiple sets of evidence. I'm not totally sure about the best approach here but it seems safe to iterate on.

fitzthum:
> Now we should make sure that we all really believe that this is secure

I am potentially having some doubts here 😬 

In particular, without composite evidence it might be possible for a man in the middle to remove a device from the evidence list, making the verifier unaware about the full TCB. There is no way to add a device, convincing Trustee that there is some secure device that isn't really there, but there may we a way to remove one.

attestation-agent/kbs_protocol/test/policy.rego
huoqifeng, line 1: Would it be helpful if to add some document for the design besides the test data in this PR?

fitzthum:
Superseded by #991

PR #962, opened 2025-04-01
attesters: improve sample warning

fitzthum:
It is very common for people to build the kbs-client without support for the platform that they want to attest. A lot of people either don't notice or understand our existing warning.

To clarify things a little bit, make the warning we print when enabling the sample attester more detailed.


Xynnn007 (approved):
LGTM.

PR #975, opened 2025-04-22
toolchain: add rust toolchain file

Xynnn007:
rust-toolchain will record the current cargo and rust toolchain version to build the project.

See https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file for more details.

cc @larrydewey

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,2 @@
+[toolchain]
+channel = "1.80.0"


Xynnn007:
> Should we use a toml file to match [with kata](https://github.com/kata-containers/kata-containers/blob/main/rust-toolchain.toml)?

Good idea. Let me change this.

fitzthum (approved):
LGTM

PR #980, opened 2025-04-27
AA: set default URL field for coco_as and kbs token field

Xynnn007:
When we do not explicitly set the fields in AA's config file like

```
[token_configs.coco_as]
url = "http://127.0.0.1:8000"

...

[token_configs.kbs]
url = "https://127.0.0.1:8080"
```

The launch of AA will fail saying that the launch config is not valid. This is not expected in scenarios no attestation token is needed for AA, and users only rely on AA's get_evidence APIs. Some other scenarios are those where the specified Attestation Token will never be accessed.

In such scenarios the url fields do not matters, thus we set default values. Note that the url fields are used lazily when the get_token API is called, so it is safe to set an empty value.

### DIFF
--- a/attestation-agent/attestation-agent/src/config/mod.rs
+++ b/attestation-agent/attestation-agent/src/config/mod.rs
@@ -95,6 +95,8 @@ impl TryFrom<&str> for Config {
             .set_default("eventlog_config.eventlog_algorithm", DEFAULT_EVENTLOG_HASH)?
             .set_default("eventlog_config.init_pcr", DEFAULT_PCR_INDEX)?
             .set_default("eventlog_config.enable_eventlog", "false")?
+            .set_default("token_configs.coco_as.url", "")?
+            .set_default("token_configs.kbs.url", "")?
             .build()?;
 
         let cfg = c.try_deserialize()?;


huoqifeng (approved):
LGTM
I think there should be no risk here.

mkulke:
> This is not expected in scenarios no attestation token is needed for AA, and users only rely on AA's get_evidence APIs

would it be better to make this field optional then?

Xynnn007:
@mkulke The parent struct `coco_as` and `kbs` is 

> would it be better to make this field optional then?

Oh yes. Let me add more patches and some unit tests for this.

Xynnn007:
@mkulke btw, do you think it is time for us to get rid of `AA_KBC_PARAMS`? I have put a PR to move all configurable things via to config file. https://github.com/kata-containers/kata-containers/pull/11197

mkulke:
> @mkulke btw, do you think it is time for us to get rid of `AA_KBC_PARAMS`? I have put a PR to move all configurable things via to config file. [kata-containers/kata-containers#11197](https://github.com/kata-containers/kata-containers/pull/11197)
> 
> cc @fitzthum @stevenhorsman

yup, in peerpods it was removed with the introduction of initdata.

fitzthum:
> btw, do you think it is time for us to get rid of AA_KBC_PARAMS? 

I think we've gotta keep support for this in the kernel command line. It is a big step in complexity for users to go from setting one kernel command line argument to creating the init-data. Even if we would prefer users to use init-data, we should keep the simple path around, at least to not break everyone who is currently using it. 

Also, not every platform truly supports init-data. The init-data implementation for s390x is a bit sketchy.

mkulke:
> > btw, do you think it is time for us to get rid of AA_KBC_PARAMS?
> 
> I think we've gotta keep support for this in the kernel command line. It is a big step in complexity for users to go from setting one kernel command line argument to creating the init-data. Even if we would prefer users to use init-data, we should keep the simple path around, at least to not break everyone who is currently using it.
> 
> Also, not every platform truly supports init-data. The init-data implementation for s390x is a bit sketchy.
> 
> @mkulke peer pods doesn't support setting AA_KBC_PARAMS via the command line? That would be a pretty big divergence from bare metal.

I just git-grepped the repo, there is no explicit support in the repo. Now, if you build/boot a kernel with that param present it might still somehow work via guest-components, but afaik there is no facility to either set the commandline at launch or build it into the image.

Xynnn007:
For PeerPod, the initdata is provisioned via CSP specific instance metadata service according to the [spec](https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/docs/initdata.md) but I am not sure if it is still up-to-date.

For baremetal deployments it will be unwise to import breakage changes. [IBM SE spec](https://www.ibm.com/docs/en/linuxonibm/pdf/l130se03.pdf) says the kernel cmdline is encapsulated into the guest images so deprecation of kernel cmdline will sure break current users. 

So let me move the kernel cmdline reading from kata-agent to CDH too, and if no CDH config file is given it will use legacy configs from kernel cmdline. This is also help to make initdata possible - if the configs are given via initdata, they will use the provisioned one.

In future if we get better initdata configuration way and s390x has perfect support for initdata (now it only injects but [not validate](https://github.com/confidential-containers/trustee/blob/main/deps/verifier/src/se/mod.rs#L33) or bind to any hardware register or vTPM PCR, so documents like https://www.ibm.com/docs/en/rhocp-ibm-z?topic=runtime-verifying-attestation-process might mislead users into thinking they are verified) we can think of getting rid of kernel cmdlines.

mkulke:
> For PeerPod, the initdata is provisioned via CSP specific instance metadata service according to the [spec](https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/docs/initdata.md) but I am not sure if it is still up-to-date.

yup and before the introduction of initdata `AA_KBC_PARAMS` has been provided as part of a configmap in kubernetes, I don't think anyone relied on a kernel cmdline parameter in peerpords (that is usually not something that you get can provide easily when launching a cloud vm). the AA_KBC_PARAMS has been dropped in favor of a "global" initdata parameter

fitzthum:
> So let me move the kernel cmdline reading from kata-agent to CDH too, and if no CDH config file is given it will use legacy configs from kernel cmdline. This is also help to make initdata possible - if the configs are given via initdata, they will use the provisioned one.

Sounds good. We should be able to mostly remove the kernel command line stuff from Kata (although we should probably keep some tests that use them).

PR #984, opened 2025-04-28
CDH & AA | Feature fix and ttrpc proto updates

Xynnn007:
Only some small fix and chore work. Please see concrete commit message.


fitzthum (approved):
LGTM

PR #985, opened 2025-04-28
AA/CDH: some fix-up work to support moving image-pull to CDH

Xynnn007:
This patch has some different patches to support moving image pull to CDH work. Please see the concrete commit messages.

cc @mkulke

### DIFF
--- a/attestation-agent/attestation-agent/src/config/kbs.rs
+++ b/attestation-agent/attestation-agent/src/config/kbs.rs
@@ -3,12 +3,12 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-use anyhow::Result;
+use anyhow::{bail, Result};
 use serde::Deserialize;
 
 use super::aa_kbc_params::AaKbcParams;
 
-#[derive(Clone, Debug, Deserialize)]
+#[derive(Clone, Debug, Deserialize, PartialEq)]
 pub struct KbsConfig {
     /// URL Address of KBS.
     pub url: String,
@@ -18,8 +18,14 @@ pub struct KbsConfig {
 }
 
 impl KbsConfig {
+    /// This function will try to read kbc and url from aa_kbc_params from env and kernel commandline.
+    /// If not given, or the kbc is not cc_kbc, it will return an error.
+    /// Because only cc_kbc will set kbs uri.
     pub fn new() -> Result<Self> {
         let aa_kbc_params = AaKbcParams::new()?;
+        if aa_kbc_params.kbc != "cc_kbc" {
+            bail!("specified aa_kbc_params.kbc is not kbs");
+        }
         Ok(Self {
             url: aa_kbc_params.uri,
             cert: None,
--- a/image-rs/src/config.rs
+++ b/image-rs/src/config.rs
@@ -13,7 +13,7 @@ use crate::snapshots::SnapshotType;
 
 /// By default use a work dir in `/run` because for confidential guests `/run`
 /// is typically in a `tmpfs` which is backed by encrypted memory.
-pub const DEFAULT_WORK_DIR: &str = "/run/image-rs/";
+pub const DEFAULT_WORK_DIR: &str = "/run/kata-containers/image/";
 
 /// Default path to policy file in KBS or locally
 pub const POLICY_FILE_PATH: &str = "kbs:///default/security-policy/test";
@@ -550,7 +550,7 @@ mod tests {
     #[test]
     fn test_image_config_from_file() {
         let data = r#"{
-            "work_dir": "/run/image-rs/",
+            "work_dir": "/run/kata-containers/image/",
             "default_snapshot": "overlay",
             "image_security_policy_uri": "file:///etc/image-policy.json",
             "authenticated_registry_credentials_uri": "file:///etc/image-auth.json",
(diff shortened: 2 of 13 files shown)


fitzthum:
I think we might need to keep the kernel commandline thing. I don't think we should expect everyone to set the init-data every time. Also, I think s390x doesn't really support init-data.

Xynnn007:
> I think we might need to keep the kernel commandline thing. I don't think we should expect everyone to set the init-data every time. Also, I think s390x doesn't really support init-data.

Updated with idea that CDH handles the kernel cmdlines.

attestation-agent/attestation-agent/src/config/kbs.rs
```diff
@@ -20,6 +20,9 @@ pub struct KbsConfig {
 impl KbsConfig {
     pub fn new() -> Result<Self> {
         let aa_kbc_params = AaKbcParams::new()?;
+        if aa_kbc_params.kbc != "kbs" {
+            bail!("specified aa_kbc_params.kbc is not kbs");
```
fitzthum, line 24: Does this mean that we aren't supporting stuff like `offline_fs_kbc` anymore?

image-rs/src/config.rs
```diff
@@ -251,7 +254,7 @@ impl ImageConfig {
     /// Try read configs from kernel command line, and other items will be set as default.
     pub fn from_kernel_cmdline() -> Self {
         let mut res = ImageConfig {
-            work_dir: PathBuf::from(DEFAULT_WORK_DIR.to_string()),
+            work_dir: PathBuf::from(DEFAULT_KATA_WORK_DIR.to_string()),
```
fitzthum, line 257: We are only using the kata work dir when the config is set via kernel command line? I don't quite understand the reason for that.

fitzthum (commented):
Looks good. I have two questions.

attestation-agent/attestation-agent/src/config/kbs.rs
Xynnn007: offline_fs_kbc is still supported in CDH side to get resource. AA's KBS config is only used to get attestation token from KBS.

image-rs/src/config.rs
Xynnn007: Yup. This is somehow a compatible way to do so. CDH works in
1. w/o launch config. It is what we are using in current CoCo, s.t. kata work dir as kata-agent defines.
2. w/ launch config, the config can be set by the user

attestation-agent/attestation-agent/src/config/kbs.rs
Xynnn007: it should be `cc_kbc`, updated.

image-rs/src/config.rs
fitzthum: Would it be simpler to just have the Kata path as the default regardless of which config option is used?

Do you have an example of when a user would want to configure this? Would that conflict with kata? I don't really understand this feature tbh. cc @stevenhorsman

Xynnn007: some existing users will make guest images with big container image layer stored insede the image.

stevenhorsman: Just to check if understand this - `DEFAULT_KATA_WORK_DIR` is only used if config is set via kernel, but we will be using initdata, so that won't get triggered going forwards and we would use `/run/image-rs` ?

Xynnn007: Yes. you both are right if cdh config in initdata is used it will use `/run/image-rs` as default.

The workdir `/run/image-rs/` exists in the image-rs repository for ~ 1 year and the older one `/var/run/image-rs` for ~ 2 years. I once thought that we should keep the path as-is, but it is fine to change is to align with kata's.

wdyt?

btw cc @ChengyuZhu6

ChengyuZhu6: I thik this is reasonable, as it addresses the issue of enabling CDH to download the image to the location without relying on initdata or configuring the image download directory.

fitzthum: From a usability standpoint I think it's good to minimize the number of things required in the init-data.

attestation-agent/attestation-agent/src/config/kbs.rs
```diff
@@ -18,8 +18,14 @@ pub struct KbsConfig {
 }
 
 impl KbsConfig {
+    /// This function will try to read kbc and url from aa_kbc_params from env and kernel commandline.
```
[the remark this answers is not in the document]
Xynnn007: You are right that it does not make sense here. I will fix here.

btw regretfully it is still used to deliver kbs address for [kbc module](https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/kms/src/plugins/kbs/mod.rs#L36-L37). It involves some extra work. Let me retire the env totally in another PR.

ChengyuZhu6 (approved):
LGTM Thanks @Xynnn007

Xynnn007:
I successfully tested guest-pull with this PR and [the PR on the kata side](https://github.com/kata-containers/kata-containers/pull/11197), but there was a problem in the error test, which was related to the error message returned by CDH RPC. I will fix it with another separate PR after this PR.

mkulke (approved):

Xynnn007:
there is no changes upon the documents thus we can ignore the broken CI https://github.com/confidential-containers/guest-components/actions/runs/14899112373/job/41847282087?pr=985

PR #990, opened 2025-05-01
Update SNP Attester

DGonzalezVillal:
Updated the SNP Attester to use version 6.0.0 of the VIRTEE SEV crate. Minor code adjustments were made to accommodate the library's new report-handling logic, but the overall functionality remains unchanged.

This will also add compatibility to the changes introduced in https://github.com/confidential-containers/trustee/pull/785, so both will need to be part of compatible releases.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -426,7 +426,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_with",
- "sev 4.0.0",
+ "sev 6.0.0",
  "sha2 0.10.9",
  "strum 0.27.1",
  "tdx-attest-rs",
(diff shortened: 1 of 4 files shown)


mythi:
It would be desirable to add support for `configfs_tsm` too (and perhaps have the option to not depend on external crates?). Any such plans?

DGonzalezVillal:
> It would be desirable to add support for `configfs_tsm` too (and perhaps have the option to not depend on external crates?). Any such plans?

I could open a separate PR for config-tsm, but I believe that's out of the scope of this PR.

I'm also of the belief that external crates are helpful. Writing structures for Attestation Reports and handling them seems out of scope for Kata. Using crates really simplifies this. I think finding better ways to keep dependencies synchronized across all components would be more beneficial. It's easier to bump a dependency and review minimal code than to make large codebase changes whenever a platform supplier (such as AMD) makes changes to their firmware.

mythi:
> > It would be desirable to add support for `configfs_tsm` too (and perhaps have the option to not depend on external crates?). Any such plans?
> 
> I could open a separate PR for config-tsm, but I believe that's out of the scope of this PR.

Thanks. Yeah, I was not necessarily expecting it to be included here. I just have not seen `snp-attester` maintainers around so I thought to ask here if you knew anything. 
 
> I'm also of the belief that external crates are helpful. Writing structures for Attestation Reports and handling them seems out of scope for Kata. Using crates really simplifies this. I think finding better ways to keep dependencies synchronized across all components would be more beneficial. It's easier to bump a dependency and review minimal code than to make large codebase changes whenever a platform supplier (such as AMD) makes changes to their firmware.

`SnpEvidence` does not have a hard dependency to the structures but I can understand it can be beneficial for getting `HOSTDATA`.

Xynnn007 (approved):
LGTM.

fitzthum (approved):
LGTM. Let's merge this once we have all the pieces ready.

PR #991, opened 2025-05-03
Multi-Device Attestation (Take 2)

fitzthum:
This is a much better version of https://github.com/confidential-containers/guest-components/pull/957

(I am keeping the original branch in place in case anyone is developing on top of it. This PR is significantly different).

This PR isn't quite done, but it's ready for some feedback. I will fix the missing pieces next week.

I have adopted a composite evidence approach. This was not as complex as I had first pictured and I think the other approach was not secure. I wrote big commit messages. Check them out. 

Two interesting improvements over the first version:

1) Since we now have a primary attester, I don't pluralize the Tee field in the kbs protocol request. I realized that the first iteration didn't really need this either. It would only come in handy if we needed different nonces for difference devices, but that doesn't seem necessary. This allowed me to drop a lot of changes from the first iteration and it means that we don't actually need to modify the KBS protocol (although I do have a couple tweaks to `kbs-types`).

2) Since the previous iteration had a very simple representation of the guest evidence (i.e. a vector), I wasn't very rigorous about defining structs for it. In this version I have created a couple new abstractions to make this more clear. These abstractions also allow us to centralize most of the complexity of evidence generation, rather than implementing it in each of the callers. i think overall the changes are much clearer.

### DIFF
--- a/attestation-agent/attester/src/az_snp_vtpm/mod.rs
+++ b/attestation-agent/attester/src/az_snp_vtpm/mod.rs
@@ -3,7 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-use super::{Attester, InitDataResult};
+use super::{Attester, InitDataResult, TeeEvidence};
 use anyhow::{bail, Context, Result};
 use az_snp_vtpm::{imds, is_snp_cvm, vtpm};
 use log::{debug, info};
@@ -31,7 +31,7 @@ struct Evidence {
 
 #[async_trait::async_trait]
 impl Attester for AzSnpVtpmAttester {
-    async fn get_evidence(&self, report_data: Vec<u8>) -> anyhow::Result<String> {
+    async fn get_evidence(&self, report_data: Vec<u8>) -> anyhow::Result<TeeEvidence> {
         let report = vtpm::get_report()?;
         let quote = vtpm::get_quote(&report_data)?;
         let certs = imds::get_certs()?;
@@ -43,7 +43,7 @@ impl Attester for AzSnpVtpmAttester {
             vcek,
         };
 
-        Ok(serde_json::to_string(&evidence)?)
+        Ok(serde_json::to_value(&evidence)?)
     }
 
     async fn bind_init_data(&self, init_data_digest: &[u8]) -> anyhow::Result<InitDataResult> {
--- a/attestation-agent/attester/src/bin/evidence_getter.rs
+++ b/attestation-agent/attester/src/bin/evidence_getter.rs
@@ -3,7 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-use attester::*;
+use attester::CompositeAttester;
 use clap::Parser;
 use std::io::Read;
 use tokio::fs;
@@ -29,9 +29,7 @@ async fn main() {
     let mut report_data = vec![0u8; 64];
 
     let cli = Cli::parse();
-
-    let tee = detect_tee_type();
-    let attester: BoxedAttester = tee.try_into().expect("create attester failed");
+    let attester = CompositeAttester::new().expect("Failed to initialize attester.");
 
     match cli {
         Cli::Stdio => std::io::stdin()
@@ -51,7 +49,7 @@ async fn main() {
     }
 
     let evidence = attester
-        .get_evidence(report_data)
+        .primary_evidence(report_data)
         .await
         .expect("get evidence failed");
     println!("{evidence}");
--- a/attestation-agent/attester/src/lib.rs
+++ b/attestation-agent/attester/src/lib.rs
@@ -41,7 +41,7 @@ pub mod tsm_report;
 #[cfg(feature = "se-attester")]
 pub mod se;
 
-pub type BoxedAttester = Box<dyn Attester + Send + Sync>;
+pub(crate) type BoxedAttester = Box<dyn Attester + Send + Sync>;
 
 impl TryFrom<Tee> for BoxedAttester {
     type Error = anyhow::Error;
@@ -78,7 +78,7 @@ pub enum InitDataResult {
 }
 
 #[async_trait::async_trait]
-pub trait Attester {
+pub(crate) trait Attester {
     /// Call the hardware driver to get the Hardware specific evidence.
     /// The parameter `report_data` will be used as the user input of the
     /// evidence to avoid reply attack.
--- a/attestation-agent/attester/src/lib.rs
+++ b/attestation-agent/attester/src/lib.rs
@@ -4,7 +4,12 @@
 //
 
 use anyhow::*;
-use kbs_types::Tee;
+use kbs_types::{Tee, TeePubKey};
+use serde::{Deserialize, Serialize};
+use serde_json::json;
+use std::collections::HashMap;
+
+use crypto::HashAlgorithm;
 
 pub mod sample;
 pub mod utils;
@@ -156,4 +161,163 @@ pub fn detect_tee_type() -> Tee {
          Attestation will continue using the fallback sample attester."
     );
     Tee::Sample
+}
+
+/// Get any additional TEEs that might be connected to the guest,
+/// such as a confidential device.
+pub fn detect_attestable_devices() -> Vec<Tee> {
+    // TODO
+    vec![]
+}
+
+/// The CompositeAttester struct is an interface to all the attesters
+/// that represent a confidential guest.
+pub struct CompositeAttester {
+    primary_attester_type: Tee,
+    primary_attester: BoxedAttester,
+    additional_attesters: HashMap<Tee, BoxedAttester>,
+}
+
+/// CompositeEvidence is the combined evidence from all the TEEs
+/// that represent the guest.
+#[derive(Serialize, Deserialize)]
+pub struct CompositeEvidence {
+    primary_evidence: String,
+    // The additional evidence is a map of Tee -> evidence,
+    // but we convert it to a string to avoid any inconsistencies
+    // with serialization. The string in this struct is exactly
+    // what is used to calculate the runtime data.
+    additional_evidence: String,
+}
+
+impl CompositeAttester {
+    pub fn new() -> Result<Self> {
+        let primary_tee = detect_tee_type();
+        let additional_tees = detect_attestable_devices();
+
+        let mut additional_attesters = HashMap::new();
+        for tee in additional_tees {
+            additional_attesters.insert(tee, tee.try_into()?);
+        }
+
+        Ok(Self {
+            primary_attester_type: primary_tee,
+            primary_attester: primary_tee.try_into()?,
+            additional_attesters,
+        })
+    }
+
+    pub fn tee_type(&self) -> Tee {
+        self.primary_attester_type
+    }
+
+    /// Get evidence for the guest.
+    /// If the guest has devices/TEEs beyond the CPU,
+    /// composite evidence will be generated.
+    /// The evidence from the additional devices will be collected first
+    /// so that it can be bound to the evidence of the primary device
+    /// via the report/runtime data.
+    pub async fn evidence(
+        &self,
+        tee_pubkey: TeePubKey,
+        nonce: String,
+        hash_algorithm: HashAlgorithm,
+    ) -> Result<String> {
+        let additional_evidence = match self.additional_attesters.is_empty() {
+            true => "".to_string(),
+            false => {
+                // Calculate the runtime data for devices, which does not include the
+                // device evidence.
+                let device_runtime_data = json!({
+                    "tee-pubkey": tee_pubkey,
+                    "nonce": nonce,
+                });
+
+                let device_runtime_data = serde_json::to_string(&device_runtime_data)
+                    .context("serialize runtime data failed")?;
+                let device_runtime_data = hash_algorithm.digest(device_runtime_data.as_bytes());
+
+                let additional_evidence = self.additional_evidence(device_runtime_data).await?;
+                serde_json::to_string(&additional_evidence)?
+            }
+        };
+
+        // Calculate the runtime data for the primary attester, which includes
+        // the device evidence retrieved above.
+        let primary_runtime_data = match self.primary_attester_type {
+            // SE handles the report data differently. As such, it does not support
+            // multi-device attestation.
+            Tee::Se => {
+                if !self.additional_attesters.is_empty() {
+                    bail!("Cannot attest multiple devices on s390x platform.")
+                }
+                nonce.into_bytes()
+            }
+            _ => {
+                let primary_runtime_data = json!({
+                    "tee-pubkey": tee_pubkey,
+                    "nonce": nonce,
+                    "additional-evidence": additional_evidence,
+                });
+                let primary_runtime_data = serde_json::to_string(&primary_runtime_data)
+                    .context("serialize runtime data failed")?;
+                hash_algorithm.digest(primary_runtime_data.as_bytes())
+            }
+        };
+        let primary_evidence = self.primary_evidence(primary_runtime_data).await?;
+
+        let guest_evidence = CompositeEvidence {
+            primary_evidence,
+            additional_evidence,
+        };
+        Ok(serde_json::to_string(&guest_evidence)?)
+    }
+
+    /// Get the evidence from the primary attester.
+    /// The caller is responsible for handling the report data.
+    pub async fn primary_evidence(&self, report_data: Vec<u8>) -> Result<String> {
+        self.primary_attester.get_evidence(report_data).await
+    }
+
+    /// Get the evidence from any additional attesters.
+    /// The caller is responsible for handling the report data.
+    pub async fn additional_evidence(&self, report_data: Vec<u8>) -> Result<HashMap<Tee, String>> {
+        let mut evidence = HashMap::new();
+
+        for (tee, attester) in &self.additional_attesters {
+            evidence.insert(*tee, attester.get_evidence(report_data.clone()).await?);
+        }
+
+        Ok(evidence)
+    }
+
+    /// Extend TEE specific dynamic measurement register
+    /// to enable dynamic measurement capabilities for input data at runtime.
+    /// This will be carried out via the primary attester only.
+    pub async fn extend_runtime_measurement(
+        &self,
+        event_digest: Vec<u8>,
+        register_index: u64,
+    ) -> Result<()> {
+        self.primary_attester
+            .extend_runtime_measurement(event_digest, register_index)
+            .await
+    }
+
+    /// Bind init data to the hardware evidence of the primary attester
+    /// either by checking that it matches an existing field or by extending
+    /// a PCR.
+    pub async fn bind_init_data(&self, init_data_digest: &[u8]) -> Result<InitDataResult> {
+        self.primary_attester.bind_init_data(init_data_digest).await
+    }
+
+    /// Get the current PCR value for some index.
+    /// This will be fulfilled via the primary attester.
+    /// It will raise an error if the primary attester
+    /// does not support runtime measurement.
+    pub async fn get_runtime_measurement(&self, pcr_index: u64) -> Result<Vec<u8>> {
+        self.primary_attester
+            .get_runtime_measurement(pcr_index)
+            .await
+    }
 }
--- a/attestation-agent/attester/src/lib.rs
+++ b/attestation-agent/attester/src/lib.rs
@@ -12,6 +12,7 @@ use std::collections::HashMap;
 use crypto::HashAlgorithm;
 
 pub mod sample;
+pub mod sample_device;
 pub mod utils;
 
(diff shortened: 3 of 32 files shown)


attestation-agent/attester/src/lib.rs
```diff
@@ -157,3 +162,162 @@ pub fn detect_tee_type() -> Tee {
...
+/// GuestEvidence is the combined evidence from all the TEEs
+/// that represent the guest.
+#[derive(Serialize, Deserialize)]
+pub struct GuestEvidence {
+    primary_evidence: String,
+    // The additional evidence is a map of Tee -> evidence,
+    // but we convert it to a string to avoid any inconsistencies
+    // with serialization. The string in this struct is exactly
+    // what is used to calculate the runtime data.
+    additional_evidence: String,
```
Xynnn007, line 212: In the serde_json crate, the serialization result is partially [canonicalized](https://www.rfc-editor.org/rfc/rfc8785) (in cases where only strings and literals are used), so if we do not use floats it might be enough to use `serde_json::Value` here?

attestation-agent/attester/src/lib.rs
```diff
@@ -157,3 +162,162 @@ pub fn detect_tee_type() -> Tee {
...
+        self.primary_attester_type
+    }
+
+    /// Get evidence for the guest.
+    /// If the guest has devices/TEEs beyond the CPU,
+    /// composite evidence will be generated.
+    /// The evidence from the additional devices will be collectd first
+    /// so that it can be bound to the evidence of the primary device
+    /// via the report/runtime data.
+    pub async fn evidence(
```
Xynnn007, line 242: What about moving this function together with other higher level "combined evidence things" up to `kbs_protocol` crate and still leave `attester` crate to have simple `get_evidence` APIs? This allows combinatorial logic to be part of the KBS protocol w/o changing the underlying simple api

attestation-agent/attester/src/az_snp_vtpm/mod.rs
```diff
@@ -31,7 +31,7 @@ struct Evidence {
 
 #[async_trait::async_trait]
 impl Attester for AzSnpVtpmAttester {
-    async fn get_evidence(&self, report_data: Vec<u8>) -> anyhow::Result<String> {
+    async fn evidence(&self, report_data: Vec<u8>) -> anyhow::Result<String> {
```
Xynnn007: Is this change necessary?imo that Rust does not use getter naming, which is for structure fields, and attester's `get_evidence` is more like a procedure call, so I think it looks okay. wdyt?

attestation-agent/attester/src/lib.rs
```diff
@@ -77,12 +77,14 @@ pub enum InitDataResult {
     Unsupported,
 }
 
+pub(crate) type TeeEvidence = serde_json::Value;
```
Xynnn007, line 82: I love this definition and change.

Xynnn007 (commented):
TBO I love this version! There are some comments. btw, 

1. do you think we need to define another type of `DeviceAttester` rather than reuse the `Attester` one?
2. How about renaming `Attesters` to `GuestAttester` as attesters is somehow overload from my personal view,

attestation-agent/attester/src/lib.rs
fitzthum: I went back and forth on this a little bit. I think it can work with a `HashMap` or a `BTreeMap` here, but with a String there might be more peace of mind. You don't have to spend any time thinking about whether it will work.

Also, note that I changed the return type of the attesters to be a Value. If we have a map here, we would probably want to change that Value back to a string to make sure we don't have any serialization inconsistencies with the evidence itself, which is totally arbitrary.

Anyway, I opted to make the evidence a value but turn the map of evidence into a String, but I am open to other ideas.

attestation-agent/attester/src/az_snp_vtpm/mod.rs
fitzthum: I can drop this. I am not totally convinced but doesn't really matter.

attestation-agent/attester/src/lib.rs
fitzthum: I think that's a good idea. It will simplify implementing the interface to the ASR/workload.

fitzthum:
> do you think we need to define another type of DeviceAttester rather than reuse the Attester one? It would not have runtime data or initdata APIs.

Yeah there are a few different ways to differentiate the type of attester. Currently it comes down to which detection function the attester is in. To me this seems the simplest, but I'm open to some other approaches. I think it's kind of up to the maintainers of kbs-types, actually. If they are happy with adding devices to the Tee enum (confidential devices are Tees, after all), then maybe it's best to use the same trait.

> How about renaming Attesters to GuestAttester as attesters is somehow overload from my personal view,

Good idea. While writing the commit messages I realized that it's confusing to have one struct that is a plural of another.

attestation-agent/attester/src/lib.rs
fitzthum: Although it does create some duplication because we have to implement the composite evidence logic in both the AA evidence provider and the native evidence provider.

Xynnn007: Oh, this may involve some details. When we re-examine combined attestation, we will find that the RPC interface provided by AA before is called `get_evidence`, which implies that there is only one TEE type. In the combined attestation scenario, this does not apply anymore. We should make changes to AA's AP. That is, we would have
1. `get_combined_evidence(runtime_data) -> Result<GuestEvidence>`: what CDH needs is this.
2. `get_evidence(tee: TEE, runtime_data) -> Result<TeeEvidence>`: a single type of tee evidence. Note that this tee enum indicates a CPU or a device evidences.
3. `get_tees`: it will return a bunch of device names and a cpu name if exist.

I think with 1 we can cover the gap. 2 & 3 can be done later?

fitzthum: The AA API basically needs to do two things. 1) Serve the evidence provider in the KBS protocol (native or ttrpc) 2) Expose the evidence to the workload via the ASR. There are some other miscellaneous things like the AA token provider, but those are the main ones.

I think we can handle all of this with 3 APIs. 

1) `get_composite_evidence(tee_pubkey, nonce, hash_alg)` - evidence provider uses this
2) `get_primary_evidence(runtime_data)`  - exposed to the workload. maybe we just call this `get_evidence` for compatibility
3) `get_additional_evidence(runtime_data)` - exposed to the workload.

The composite evidence endpoint must get the nonce and pubkey as input so that it can construct the runtime data for the primary attester and additional attesters. The other endpoints just take the runtime data directly, which fits with the current API we expose to the workload. I think it's easier for the user to have primary and additional endpoints rather than iterating through all the TEEs. They will have to setup the runtime data themselves either way.

Xynnn007: 2 & 3 looks good to me.

Overall it is good to me. As 1 is only the interfaces inbetween AA and CDH and will not influence high level users, we can improve it later if needed.

attestation-agent/attester/src/lib.rs
Xynnn007: Yup. String is always safe. Crates like https://crates.io/crates/olpc-cjson can make the serialization consistent but imports more dependency.

fitzthum:
Ok, first set of comments should be addressed and I have the tests passing.

There are still a couple of loose ends 
1) I am skipping the e2e test, which is a circular dependency
2) The AS token client code will change if we change the AS interface. There are also some longterm questions to answer about this mode.

Otherwise, I think things are in decent shape. PTAL. I will work on fixing up the Trustee PR.

Xynnn007 (approved):
LGTM.

attestation-agent/attester/src/lib.rs
mkulke: I personally don't love it so much, since this is just an alias to serialization library, and, if possible, we should not expose this in a the API (this leaves the option to later switch it to sth else, for e.g. more resource-constrained TEEs). We can maybe use a Newtype (`struct TeeEvidence(serde_json::Value)` the evidence.0 shouldn't be pub, so we don't leak imoplementation details)?

attestation-agent/attester/src/lib.rs
```diff
@@ -72,12 +79,26 @@ pub enum InitDataResult {
...
+pub(crate) type TeeClass = String;
```
mkulke, line 83: Smple type aliases can be confusing since they signal that the type is something more than it actually is. TeeClass is just a String without any verification. If the tee class should be a constrained value, then it can be an (open) enum or a Newtype `struct TeeClass(String)` that will only be produced by some validation. If it's not constrained it should be just a String I think.

attestation-agent/attester/src/lib.rs
```diff
@@ -157,3 +178,182 @@ pub fn detect_tee_type() -> Tee {
...
+    /// so that it can be bound to the evidence of the primary device
+    /// via the report/runtime data.
+    pub async fn composite_evidence(
+        &self,
+        tee_pubkey: TeePubKey,
+        nonce: String,
+        hash_algorithm: HashAlgorithm,
+    ) -> Result<String> {
+        let additional_evidence = match self.additional_attesters.is_empty() {
+            true => "".to_string(),
```
mkulke, line 249: Very nice. I understand we now tie the evidences together by mixing the "additional evidences" into the CPU Tee report_data.

Is there a reason why we do it this way and not the opposite way (mix the evidence of the CPU into the report data of the additional devices)?

attestation-agent/kbs_protocol/src/client/rcar_client.rs
```diff
@@ -334,6 +334,7 @@ mod test {
 
     const CONTENT: &[u8] = b"test content";
 
+    #[ignore]
```
mythi, line 337: The test knows to ignore protocol version error so you could just bump https://github.com/confidential-containers/guest-components/blob/8e39f5ac58066910c60c698d4dd5ff60654d07d6/attestation-agent/kbs_protocol/src/client/mod.rs#L51 to get the same result.

attestation-agent/attester/src/lib.rs
mythi: it could be dropped from this series if it's not strictly necessary?

attestation-agent/attester/src/lib.rs
```diff
@@ -157,3 +178,182 @@ pub fn detect_tee_type() -> Tee {
...
+    }
+
+    pub fn tee_type(&self) -> Tee {
+        self.primary_attester_type
+    }
+
+    /// Get evidence for the guest.
+    /// If the guest has devices/TEEs beyond the CPU,
+    /// composite evidence will be generated.
+    /// The evidence from the additional devices will be collectd first
```
mythi, line 239: nit: 
```suggestion
    /// The evidence from the additional devices will be collected first
```

mythi:
Overall, looks very good to me. None of my review comments are blockers so +1 for this take 2.

attestation-agent/kbs_protocol/src/client/rcar_client.rs
fitzthum: Hm good point. There's an interesting question here about whether we should bump the KBS protocol version in this PR. The first version of this PR relied on changes in the KBS protocol (to specify all the TEEs in the first part of the RCAR handshake), but this version does not. That said, multi-device guest-components won't be compatible with an older version of Trustee because the format of the evidence is different. I'm not sure exactly the best way to capture this.

attestation-agent/attester/src/lib.rs
fitzthum: Yeah good question. As a baseline the reason for binding the evidence together is to prevent it from being tampered with in transit. If our combined evidence was just a list of reports that all used the same runtime data, a mitm could remove one of the reports without anyone noticing. This would cause Trustee to skip validation of a device that may or may not meet requirements. If the mitm also had their own CVM with a device attached to it, they could take the report data from evidence they intercepted, request a report from their valid device, and insert that into the evidence. This would make Trustee think that the attesting guest had a device that it did not. Ofc they could also remove an invalid device an add a seemingly valid one.

If we put the hash of the additional evidence into the runtime data of the primary evidence, these attacks are prevented. There is no way to generate such evidence without actually having a guest that matches the evidence. This case reduces to the logic around attesting with a CPU only (tee key pair, measurement of AA, etc). We are basically just extending the CPU evidence with some other stuff that happens to be evidence for devices.

So what if instead of putting the additional devices into the primary evidence, we put the primary evidence into the additional evidence? In this case I think these attacks would still be doable. The attacker couldn't drop the primary evidence because it would be in the additional evidence, but they could drop or replace any of the additional evidence. It seems like this would be a little bit backwards; extending each piece of additional evidence to also cover the primary evidence. Regardless of the terminology, this wouldn't create one root for the evidence.

On a related note, I think the primary attester doesn't necessarily have to be the CPU, and actually the code doesn't strictly require that. It's probably best practice, though.

attestation-agent/attester/src/lib.rs
fitzthum: I am basically just following the style of the TeeParsedClaims alias in Trustee. I don't have strong feelings about this. I don't think simple type alises are necessarily a bad thing. It can make things more descriptive even if it adds an extra trip to `mod.rs` where you say wait it's just a string...

attestation-agent/attester/src/lib.rs
fitzthum: I'm not too worried about requiring attesters to use serde_json. Every single one does so already. The case of a resource constrained TEE seems out of scope since that will probably require changing a bunch of other stuff in guest components (i.e. all the other places we use serde_json).

attestation-agent/attester/src/lib.rs
```diff
@@ -78,9 +78,21 @@ pub enum InitDataResult {
...
 pub(crate) type TeeEvidence = serde_json::Value;
+pub(crate) type TeeClass = String;
 
 #[async_trait::async_trait]
 pub(crate) trait Attester {
+    /// Each attester should define which category of tee it represents.
+    /// This could be something like "cpu" or "gpu" or any other scheme
+    /// used by the attester.
+    /// Ultimately this class will represent a module in the attestation token
+    /// signed by the attestation service.
```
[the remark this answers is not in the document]
mythi: Would this mean the attesters can impact what categories the token ends up having? Perhaps the AS code is better for defining these.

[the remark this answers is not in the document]
fitzthum: The TEE class becomes a key part of the token. The idea isn't that attesters are setting random TEE Class values so much that they might select between a couple of different well-known values. For instance one attester could potentially handle a gpu or a nic developed by the same company.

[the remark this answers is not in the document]
Xynnn007: This scenario is possible, but I wonder if their detection logic will be different. For example, the NIC and GPU may both use heuristic detection of different devices under `/dev`. Thus leading to a decouple of the attester logic into different `Tee`s.

We can just package it into different Tee types (If we let attester to set `TeeClass` dynamically, we would meet `Tee: CompanyA, TeeType: Gpu/ Tee: CompanyA, TeeType: Nic` rather than `Tee: GpuCompanyA`/`Tee: NicCompanyA`. I thought the latter is better, wdyt?). If the underlying SDK is the same, it doesn’t matter, just like  SGX and TDX.

attestation-agent/attester/src/lib.rs
mkulke: but what a string represents should be reflected in the variable name, no? 

a type alias is useful to shorten complex types, but if we want to e.g. signify that a TeeClass cannot be an arbitrary string, but somehow constrained a newtype is a good solution (the newtype would be returned by a fn that asserts the constraint)

attestation-agent/attester/src/lib.rs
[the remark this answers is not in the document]
Xynnn007: Tee is used to allow the verification end to determine a specific Verifier. There are only two possible relationships between TeeType and Attester : 
1. A manufacturer uses one Attester corresponding to different hardware TeeTypes, 
2. An Attester corresponds to a single TeeType;

Companies can reorganize the code structure so that implementations can use the same codebase to produce different Attesters -- which we could help.


So the only possible reasonable usage is that there is additional metadata that wants to be added. Because this field itself is not protected by the integrity of Evidence, it can be included in the Request field of RCAR or directly in our encapsulated Evidence. So we can embed it in the Evidence format when necessary and let the corresponding Verifier plug-in handle it?



All above is from a worry that this field will not have any practical effect, because once we introduce a new parameter value, for AS, all current Evidence (all CPU) will inevitably need to enter a constant value "cpu", and because there is no actual implementation example here, so I don't know how it will be used. Maybe we can postpone the introduction of this field until it is really needed?

attestation-agent/attester/src/lib.rs
fitzthum: Well we have function return values and tuples where there is no name (and we don't want to change the name of the function). The Tee Class isn't contrained in a way that can be easily expressed in code (otherwise we'd use an enum). 

Anyway, this style is used by the verifiers in the AS already. Maybe we can have a discussion about changing that, but I am just trying to be consistent for now.

attestation-agent/attester/src/lib.rs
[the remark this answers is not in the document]
fitzthum: At a high-level, I think what an attester represents is a method of attestation rather than a specific device. Today we already have different generations of devices that share the same attester. It wouldn't surprise me if different types of devices will share the same type of attestation as well. In fact, that seems like a good idea. In the long run I don't think it's sustainable to infer the properties of a Tee from the TeeType alone. We will need some more metadata.

Now, as you say, this metadata could be internal to the Tee evidence (not the hw evidence, but the bundled hw evidence coming from the attester). In some ways there is very little different between that and putting the class into the composite evidence. I lean slightly towards putting it into the composite evidence because it seems more explicit to me i.e. when you look at the attester you can easily see what type of device it is; if you print out the composite evidence you can see the Tee class; implementers don't need to think as much about extending the evidence provided by the attester; it paves the way for creating another field such as TeeName in the future; etc.

ofc maybe the most significant tradeoff is just that it will take 1.5 days to pull out this field, which spans a number of places in the code, but I am happy to do that if you guys really don't like the TeeClass in the composite evidence.

fitzthum:
Ok, I have fixed up most of the feedback on both PRs, rebased, an added docs. I am trying to narrow scope as much as possible here, so if there is anything that can be left for a follow-up, I would like to do that. The `TeeClass` is still in the `CombinedEvidence` as I prefer it, but I might remove that.

Xynnn007:
> Ok, I have fixed up most of the feedback on both PRs, rebased, an added docs. I am trying to narrow scope as much as possible here, so if there is anything that can be left for a follow-up, I would like to do that. The `TeeClass` is still in the `CombinedEvidence` as I prefer it, but I might remove that. Time to start thinking seriously about merging these things.

From my point of view, most of them are fine, and the only one left is teeClass. I still hope to delete it, because it may be fine on the gc side, but it will add an extra parameter on the AS side. For the caller of AS, the TEE field is enough to indicate what TEE he wants to verify, and the additional class field will increase the usage burden; if our goal is to have a name in the submod, then we might as well map all the current Tee enums to the value of cpu0, and make a TODO comment, and then correct it later.

attestation-agent/attester/src/lib.rs
fitzthum: (I removed the TeeClass from this PR entirely btw)

attestation-agent/attester/src/lib.rs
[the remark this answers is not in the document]
fitzthum: I dropped the TeeClass from this PR.

Xynnn007 (approved):
Another approve for this! btw the https://github.com/confidential-containers/guest-components/pull/991#discussion_r2078732181 is a TODO for me (just a notification)

dcmiddle:
I haven't looked deeply enough to offer a useful +1, but I did want to say that the commit messages were really helpful in understanding the approach.

attestation-agent/attester/src/bin/evidence_getter.rs
```diff
@@ -3,7 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-use attester::*;
+use attester::Attesters;
```
mythi:
```suggestion
use attester::CompositeAttester;
```

same change needed below

attestation-agent/attester/src/lib.rs
```diff
@@ -157,3 +166,170 @@ pub fn detect_tee_type() -> Tee {
...
+
+/// The Attesters struct is an interface to all the attesters
+/// that represent a confidential guest.
+pub struct CompositeAttester {
+    primary_attester_type: Tee,
+    primary_attester: BoxedAttester,
+    additional_attesters: HashMap<Tee, BoxedAttester>,
+}
+
+/// GuestEvidence is the combined evidence from all the TEEs
```
mythi:
```suggestion
/// CompositeEvidence is the combined evidence from all the TEEs
```

PR #993, opened 2025-05-09
AA | Fix AAEL recording when there is an empty eventlog file

Xynnn007:
if there is already an empty aael file, AA is expected to truncate the original file and write to that file. The old code opens the file with read-only permission, thus cannot write.

This patch fixes this, also includes some unit tests to cover this.


fitzthum (approved):
LGTM

Xynnn007:
let me get this merged and ignore the broken link checker.

PR #996, opened 2025-05-09
ci: disable link checker temporarily

Xynnn007:
Before we get closer to a better solution of link checker, let's temporarily disable this link checker.

This is fair and we can do fix-ups to broken links after the link checker goes back well.


mkulke (approved):
Agree 100%. right now we have no link checks, no way to fix it _and_ a broken CI

fitzthum (approved):
LGTM

PR #1001, opened 2025-05-14
chore(dep): update tempfile to 3.20

Xynnn007:
The new version of tempfile changes API `into_path` to `keep`

Close #997


mythi (approved):
thanks!

This has the positive side-effect that with
```
-        .work_dir(work_dir.into_path())
+        .work_dir(work_dir.path().to_path_buf())
```

the tmpdirs actually get deleted after the tests :-)

PR #1013, opened 2025-05-31
test: test build cdh on s390x

Xynnn007:
This patch is used to test building process for CDH on s390x.

See the building error

https://github.com/kata-containers/kata-containers/actions/runs/15344193896/job/43176739919?pr=11197


Xynnn007:
the s390x build still works.

PR #1016, opened 2025-06-04
add nvidia-attester

mythi:
https://github.com/confidential-containers/confidential-containers/issues/278 wasn't clear whether we can use pure Rust in the nvidia-attester. I gave `nvml-wrapper` a try and that resulted in this PR. Build-tested only at the moment but any help to get this to run on real HW would be nice.

This PR targets to get simple experiments started using `evidence_getter` using `cargo build --no-default-features --features bin,nvidia-attester,tdx-attester --bin evidence_getter --release`.

TODO

- [x] testing
- [x] rebase to #991 
- [x] improve commit messages
- [x] ~[`nvml-wrapper` release](https://github.com/rust-nvml/nvml-wrapper/issues/114)~

### DIFF
--- a/attestation-agent/attester/Cargo.toml
+++ b/attestation-agent/attester/Cargo.toml
@@ -23,6 +23,7 @@ hex.workspace = true
 iocuddle = { version = "0.1.1", optional = true }
 kbs-types.workspace = true
 log.workspace = true
+nvml-wrapper = { git = "https://github.com/rust-nvml/nvml-wrapper", rev="7e0752f331" , optional = true, default-features = false, features = ["serde"]}
 occlum_dcap = { git = "https://github.com/occlum/occlum", tag = "v0.29.7", optional = true }
 pv = { version = "0.10.0", package = "s390_pv", optional = true }
 scroll = { version = "0.13.0", default-features = false, features = [
@@ -75,6 +76,7 @@ all-attesters = [
     "cca-attester",
     "se-attester",
     "tpm-attester",
+    "nvidia-attester",
 ]
 
 # tsm-report enables a module that helps attesters to use Linux TSM_REPORTS for generating
@@ -84,6 +86,7 @@ tsm-report = ["tempfile"]
 tdx-attester = ["scroll", "tsm-report", "iocuddle"]
 tdx-attest-dcap-ioctls = ["tdx-attest-rs"]
 sgx-attester = ["occlum_dcap"]
+nvidia-attester = ["nvml-wrapper"]
 az-snp-vtpm-attester = ["az-snp-vtpm"]
 az-tdx-vtpm-attester = ["az-snp-vtpm-attester", "az-tdx-vtpm"]
 snp-attester = ["sev"]
--- a/attestation-agent/attester/src/nvidia/mod.rs
+++ b/attestation-agent/attester/src/nvidia/mod.rs
@@ -1,0 +1,111 @@
+// Copyright (c) 2025 Intel Corporation
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use super::{Attester, TeeEvidence};
+use anyhow::{bail, Context, Result};
+use base64::{engine::general_purpose::STANDARD, Engine};
+use nvml_wrapper::{enums::device::DeviceArchitecture, Nvml};
+use serde::Serialize;
+
+const NVIDIA_NONCE_SIZE: usize = 32;
+
+pub fn detect_platform() -> bool {
+    // Return true iff one GPU is found and it has CC mode set.
+    match Nvml::init() {
+        Ok(nvml) => {
+            nvml.device_count().is_ok_and(|count| count == 1)
+                && nvml
+                    .device_by_index(0)
+                    .is_ok_and(|device| device.is_cc_enabled().unwrap_or_default())
+        }
+        Err(_) => false,
+    }
+}
+
+/// NRAS knows about "switch" and "gpu" but the expected evidence
+/// content is the same. nvidia-attester can compose a list of
+/// all CC enabled nvml/nscq devices using this evidence struct.
+#[derive(Serialize)]
+struct NvDeviceReportAndCert {
+    arch: DeviceArchitecture,
+    uuid: String,
+    evidence: String,
+    certificate: String,
+}
+
+#[derive(Serialize)]
+struct NvDeviceEvidence {
+    device_evidence_list: Vec<NvDeviceReportAndCert>,
+}
+
+#[derive(Debug, Default)]
+pub struct NvAttester {}
+
+#[async_trait::async_trait]
+impl Attester for NvAttester {
+    /// Generate evidence for the NVIDIA devices. A 32 byte nonce is taken from the first 32
+    /// report_data bytes. report_data shorter than 32 bytes is zero padded.
+    async fn get_evidence(&self, mut report_data: Vec<u8>) -> Result<TeeEvidence> {
+        let nvml = Nvml::init()?;
+        let devices = nvml.device_count()?;
+
+        let mut device_evidence_list = vec![];
+
+        if report_data.len() < NVIDIA_NONCE_SIZE {
+            report_data.resize(NVIDIA_NONCE_SIZE, 0);
+        }
+
+        let nonce: [u8; NVIDIA_NONCE_SIZE] = report_data[0..NVIDIA_NONCE_SIZE].try_into()?;
+
+        for index in 0..devices {
+            let device = nvml.device_by_index(index)?;
+
+            let report = device
+                .confidential_compute_gpu_attestation_report(nonce)
+                .context("Failed to get attestation report for device {index}")?;
+
+            let certificate = device
+                .confidential_compute_gpu_certificate()
+                .context("Failed to get certificate for device {index}")?;
+
+            let dev_arch = device
+                .architecture()
+                .context("Failed to get architecture for device {index}")?;
+
+            let dev_uuid = device
+                .uuid()
+                .context("Failed to get UUID for device {index}")?;
+
+            device_evidence_list.push(NvDeviceReportAndCert {
+                arch: dev_arch,
+                uuid: dev_uuid,
+                // NRAS expects hex as the encoding for the report so it's used
+                // here instead of base64. With that, the relying parties that
+                // use this evidence with NRAS do not have to re-encode it.
+                evidence: hex::encode(report.attestation_report),
+                certificate: STANDARD.encode(certificate.attestation_cert_chain),
+            });
+
+            device
+                .set_confidential_compute_state(true)
+                .context("Failed to set device {index} to ready state")?;
+
+            // Run final sanity check to ensure the device confidential computing mode is enabled,
+            // it's in a production environment, and accepting client requests.
+            if !device
+                .check_confidential_compute_status()
+                .is_ok_and(|status| status)
+            {
+                bail!("NVIDIA attester: device {index} CC status check failed")
+            }
+        }
+
+        let full_evidence = NvDeviceEvidence {
+            device_evidence_list,
+        };
+
+        serde_json::to_value(&full_evidence).context("Serialize NVIDIA evidence failed")
+    }
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -380,6 +380,7 @@ dependencies = [
  "kbs-types",
  "log",
  "num-traits",
+ "nvml-wrapper",
  "occlum_dcap",
  "picky-asn1 0.10.1",
  "picky-asn1-der 0.5.2",
@@ -3925,6 +3926,29 @@ checksum = "071dfc062690e90b734c0b2273ce72ad0ffa95f0c74596bc250dcfd960262841"
 dependencies = [
  "autocfg",
  "libm",
+]
+
+[[package]]
+name = "nvml-wrapper"
+version = "0.11.0"
+source = "git+https://github.com/rust-nvml/nvml-wrapper?rev=7e0752f331#7e0752f331d37a62a46ef2fb6329a33fae36dd8d"
+dependencies = [
+ "bitflags 2.9.0",
+ "libloading",
+ "nvml-wrapper-sys",
+ "serde",
+ "serde_derive",
+ "static_assertions",
+ "thiserror 1.0.69",
+ "wrapcenum-derive",
+]
+
+[[package]]
+name = "nvml-wrapper-sys"
+version = "0.9.0"
+source = "git+https://github.com/rust-nvml/nvml-wrapper?rev=7e0752f331#7e0752f331d37a62a46ef2fb6329a33fae36dd8d"
+dependencies = [
+ "libloading",
 ]
 
 [[package]]
@@ -7797,6 +7821,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3268f3d866458b787f390cf61f4bbb563b922d091359f9608842999eaee3943c"
 dependencies = [
  "bitflags 2.9.0",
+]
+
+[[package]]
+name = "wrapcenum-derive"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a76ff259533532054cfbaefb115c613203c73707017459206380f03b3b3f266e"
+dependencies = [
+ "darling",
+ "proc-macro2",
+ "quote",
+ "syn 2.0.99",
 ]
 
 [[package]]
--- a/attestation-agent/attestation-agent/Cargo.toml
+++ b/attestation-agent/attestation-agent/Cargo.toml
@@ -77,6 +77,7 @@ all-attesters = [
     "se-attester",
     "cca-attester",
     "tpm-attester",
+    "nvidia-attester",
 ]
 tdx-attester = ["kbs_protocol?/tdx-attester", "attester/tdx-attester"]
 tdx-attester-libtdx = [
@@ -102,6 +103,7 @@ hygon-dcu-attester = [
     "attester/hygon-dcu-attester",
 ]
 tpm-attester = ["kbs_protocol?/tpm-attester", "attester/tpm-attester"]
+nvidia-attester = ["kbs_protocol?/nvidia-attester", "attester/nvidia-attester"]
 
 # Either `rust-crypto` or `openssl` should be enabled to work as underlying crypto module
 rust-crypto = ["kbs_protocol?/rust-crypto"]
--- a/attestation-agent/attester/src/lib.rs
+++ b/attestation-agent/attester/src/lib.rs
@@ -43,6 +43,9 @@ pub mod se;
 #[cfg(feature = "tpm-attester")]
 pub mod tpm;
 
+#[cfg(feature = "nvidia-attester")]
+pub mod nvidia;
+
 pub type BoxedAttester = Box<dyn Attester + Send + Sync>;
 
 impl TryFrom<Tee> for BoxedAttester {
@@ -72,6 +75,8 @@ impl TryFrom<Tee> for BoxedAttester {
             Tee::Se => Box::<se::SeAttester>::default(),
             #[cfg(feature = "tpm-attester")]
             Tee::Tpm => Box::new(tpm::TpmAttester::new()?),
+            #[cfg(feature = "nvidia-attester")]
+            Tee::Nvidia => Box::<nvidia::NvAttester>::default(),
             _ => bail!("TEE is not supported!"),
         };
 
@@ -201,6 +206,11 @@ pub fn detect_tee_type() -> Tee {
 /// such as a confidential device.
 pub fn detect_attestable_devices() -> Vec<Tee> {
     let mut additional_devices = vec![];
+
+    #[cfg(feature = "nvidia-attester")]
+    if nvidia::detect_platform() {
+        additional_devices.push(Tee::Nvidia);
+    }
 
     if sample_device::detect_platform() {
         additional_devices.push(Tee::SampleDevice);
--- a/attestation-agent/kbs_protocol/Cargo.toml
+++ b/attestation-agent/kbs_protocol/Cargo.toml
@@ -67,6 +67,7 @@ hygon-dcu-attester = ["attester/hygon-dcu-attester"]
 cca-attester = ["attester/cca-attester"]
 se-attester = ["attester/se-attester"]
 tpm-attester = ["attester/tpm-attester"]
+nvidia-attester = ["attester/nvidia-attester"]
 
 rust-crypto = ["reqwest/rustls-tls", "crypto/rust-crypto"]
 openssl = ["reqwest/native-tls-vendored", "crypto/openssl"]


fitzthum:
> OK, good to know. Where is this thread?

@zvonkok and I discussed this at last weeks community meeting I think

mythi:
thanks! I'll check the recording since I missed the meeting

mythi:
FYI my PR to kbs-types: https://github.com/virtee/kbs-types/pull/62

fitzthum:
> one thing I noticed while testing was that  .check_confidential_compute_status() returns false if the is not ready...

I guess this PR isn't checking that now. Should we call `set_confidential_compute_state` at the end of this attester?

mythi:
> > one thing I noticed while testing was that  .check_confidential_compute_status() returns false if the is not ready...
> 
> I guess this PR isn't checking that now. 

I had it in the first version but dropped it after I got to test this on real HW. `true` means it [needs to be ready](https://github.com/rust-nvml/nvml-wrapper/blob/076e26174056f8e740e3fec495658bf850275a21/nvml-wrapper/src/device.rs#L819) as well. Due to the issue I reported in `nvml-wrapper` I cannot do any other status/state checks atm.

> Should we call `set_confidential_compute_state` at the end of this attester?

I was also thinking whether `get_evidence()` or `detect_platform()` could do all the init steps needed for attesation/CC. the attester does not know about the attestation result so it could do `set_confidential_compute_state` before getting the evidence as well?

fitzthum:
> the attester does not know about the attestation result so it could do set_confidential_compute_state before getting the evidence as well?

Yeah. As I mentioned on the call there is a small window for an attack here. An invalid device could return invalid evidence, then if it is enabled it might be able to overwrite the AA binary and replace that evidence with something that looks legit (evidence harvested from another guest with a valid device, for instance). Basically it could undermine the operation of the AA, which would be bad. We can prevent this by enabling the device only after the attestation token has been received. It doesn't really matter if the attestation "passes" as long as we have an attestation token that represents the guest correctly. In the future we may want to add a callback like this.

That said, in the short term I think we should just enable the device at the end of the attester. I don't think it's actually feasible for devices with bounce buffers to interfere with the AA.

mythi:
I'm OK with that. These things are currently missing because of https://github.com/rust-nvml/nvml-wrapper/issues/97 and I haven't spent much time debugging it yet.

attestation-agent/attester/Cargo.toml
```diff
@@ -26,6 +26,7 @@ serde.workspace = true
 serde_json.workspace = true
 serde_with.workspace = true
 sev = { version = "6.2.1", default-features = false, features = ["snp"], optional = true }
+nvml-wrapper = { version = "0.11.0", optional = true, default-features = false, features = ["serde"]}
```
Xynnn007: Let's move the dep to follow alphabet order for better maintaince

attestation-agent/attester/Cargo.toml
```diff
@@ -37,6 +38,7 @@ hyper = { version = "0.14", features = ["full"], optional = true }
 hyper-tls = { version = "0.5", optional = true }
 tokio = { version = "1", features = ["full"], optional = true }
 tempfile = { workspace = true, optional = true }
+env_logger = { workspace = true, optional = true }
```
Xynnn007: Let's move the dep to follow alphabet order for better maintaince

attestation-agent/attester/src/bin/evidence_getter.rs
```diff
@@ -49,8 +50,17 @@ async fn main() {
...
     println!("{evidence}");
+
+    let additional_evidence = attesters
+        .additional_evidence(report_data)
+        .await
+        .expect("get additional evidence failed");
+
+    for (tee, evidence) in additional_evidence {
+        println!("{tee:?} => {evidence}");
+    }
```
Xynnn007: One thing to note that this will change the behavior of this binary. With the original one we can
```shell
./evidence_getter | <handle-evidence>
```
While now it cannot. But this is ok for me as it is only a debugging tool.

attestation-agent/attester/src/nvidia/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+//
+
+use super::{Attester, TeeEvidence};
+use anyhow::{Context, Result};
+use base64::Engine;
+use nvml_wrapper::{enums::device::DeviceArchitecture, Nvml};
+use serde::Serialize;
+
+pub fn detect_platform() -> bool {
+    // TODO: loop all devices here to check their CC status?
```
Xynnn007: Do we have answer for this?

attestation-agent/attester/src/nvidia/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+        let devices = nvml.device_count()?;
+
+        let mut device_evidence_list = vec![];
+        let b64_engine = base64::engine::general_purpose::STANDARD;
+
+        for index in 0..devices {
+            let device = nvml.device_by_index(index)?;
+
+            // Take the first half of the reportdata hash as the nonce (the hash includes the KBS protocol nonce)
+            let nonce: [u8; 32] = report_data[0..32].try_into()?;
```
Xynnn007: Let's add a `warn!` here if the `report_data.len() != 32`. It can be padding or truncated. Also we do not need to mention kbs protocol here as they are decoupled.

attestation-agent/attester/src/nvidia/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+
+            // TODO: refactor once NMVL knows about other CC device types
```
Xynnn007:
```suggestion
            // TODO: add device type distinguishment once NVML knows about other CC device types
```

Xynnn007 (commented):
Some nits and a question. Do we need to install some dependency else out of bind to make this run - or the nvml dependency will be compiled into the final binary?

If we need something else, we need to mention in a doc

attestation-agent/attester/src/nvidia/mod.rs
mythi: dropped this completely for now

attestation-agent/attester/src/nvidia/mod.rs
mythi: No but I think a check for "CC capable" would be sufficient here. WDYT?

attestation-agent/attester/src/bin/evidence_getter.rs
mythi: updated

attestation-agent/attester/Cargo.toml
mythi: dropped this commit completely

attestation-agent/attester/Cargo.toml
mythi: fixed

mythi:
> Do we need to install some dependency else out of bind to make this run - or the nvml dependency will be compiled into the final binary?

it builds OK without the nvml lib. it needs to be available runtime (which is the expectation for all of the nvidia driver/libs stuff) and the check for it is included as part of `detect_platform()`.

attestation-agent/attester/src/nvidia/mod.rs
mythi: added but the more fundamental question is can we re-purpose report data for this. my ITA PR shows another problem: it needs to know how nonce became the report data hash to provide it as the "nonce" for the verifier.

attestation-agent/attester/src/nvidia/mod.rs
Xynnn007: Yes. We do not need to return true if the GPUs/devices are not confidential

attestation-agent/attester/src/nvidia/mod.rs
Xynnn007: Not sure the concrete question is, but this is a low level API that only receives the raw report data. The origin should be some upper semantics to handle.

dcmiddle:
> I'm OK with that. These things are currently missing because of [rust-nvml/nvml-wrapper#97](https://github.com/rust-nvml/nvml-wrapper/issues/97) and I haven't spent much time debugging it yet.

I put up a fix https://github.com/rust-nvml/nvml-wrapper/pull/104

attestation-agent/attester/src/nvidia/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+        let nvml = Nvml::init()?;
+        let devices = nvml.device_count()?;
+
+        let mut device_evidence_list = vec![];
+        let b64_engine = base64::engine::general_purpose::STANDARD;
+
+        for index in 0..devices {
+            let device = nvml.device_by_index(index)?;
+
+            if report_data.len() < 32 {
```
cclaudio: nit. `report_data` does not seem to be modified. I think we can move it to outside the loop

attestation-agent/attester/src/nvidia/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+                warn!("NVIDIA Attester: Report data must be at least 32 bytes");
+            }
+
+            let nonce: [u8; 32] = report_data[0..32].try_into()?;
+
+            let report = device
+                .confidential_compute_gpu_attestation_report(nonce)
```
cclaudio, line 66: We may want to include the nonce/report_data in the evidence as well so that the verifier can check the report freshness.  Maybe it could be a new field in the `NvDeviceReportAndCert` struct

attestation-agent/attester/src/nvidia/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+                .context("Failed to get certificate for device {index}")?;
+
+            let dev_arch = device
+                .architecture()
+                .context("Failed to get architecture for device {index}")?;
+
+            let dev_uuid = device
+                .uuid()
+                .context("Failed to get UUID for device {index}")?;
+
```
cclaudio, line 80: The function [nvtrust::verify_attestation_report()](https://github.com/NVIDIA/nvtrust/blob/main/guest_tools/gpu_verifiers/local_gpu_verifier/src/verifier/cc_admin_utils.py#L493) verifies the report signature, but before that it performs some checks:
- match nonce in the report with the one generated by cc admin
- match the driver version and vbios version in the report with the ones fetched from the driver

We may also want to fetch the `driver_version` and `vbios_version` from the driver and include that in the Evidence

cclaudio (commented):
I am developing a nvidia verifier for trustee based on this PR.  I have some comments based on that.

attestation-agent/attester/src/nvidia/mod.rs
fitzthum: The nonce can be calculated from the report_data, which the KBS will have access to.

attestation-agent/attester/src/nvidia/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+            device_evidence_list.push(NvDeviceReportAndCert {
+                arch: dev_arch,
+                uuid: dev_uuid,
+                evidence: b64_engine.encode(report.attestation_report),
+                certificate: b64_engine.encode(certificate.attestation_cert_chain),
+            });
+        }
+
+        let full_evidence = NvDeviceEvidence {
+            device_evidence_list,
```
cclaudio, line 106: AFAIK, sending an attestation request with evidences for multiple devices is not supported yet and it seems relatively more complex. E.g. the verifier/policy-engine would need to decide what to do with the entire request if one or more devices fail the attestation.

What if, for now, we support attestation requests containing only one device? The attester will need to decide what to do with device that fail attestation anyway. Later we could add more complexity to the verifier if necessary

fitzthum: Yeah the scope isn't quite right here. Each attester/verifier will result in one EAR submod, which is really supposed to represent one device. In the future I think we should tweak things so that we can instantiate multiple attesters of the same type. In the short term I am ok with only allowing one device or collapsing all the devices together into one submod. Maybe the first would be better since we don't really support multiple GPUs right now anyway (at least not in coco)

attestation-agent/attester/src/nvidia/mod.rs
mythi: `driver_version` and `vbios_version` are already part the evidence. The additional checks are something you'd add to your CoCo policies

attestation-agent/attester/src/nvidia/mod.rs
mythi: > we don't really support multiple GPUs right now anyway (at least not in coco)

I'm not aware of any limitations with this. Can you elaborate? My plan with this attester is to include the evidence from every nvidia GPUs the guest has. AFAICS, NRAS API supports "evidence list" also

attestation-agent/attester/src/nvidia/mod.rs
mythi: updated, thanks

mythi:
> I am developing a nvidia verifier for trustee based on this PR, I am planning to post it soon.

Are you going to start with something NRAS based or trying to connect to OCSP/RIM services directly in the verifier?

attestation-agent/attester/src/nvidia/mod.rs
cclaudio: @mythi right, the nvtrust defines policies for the `driver_version` and `vbios_version` (entries 15 and 16 of this [table](https://github.com/NVIDIA/nvtrust/blob/main/guest_tools/attestation_troubleshooting_guide.md)).

I think the question here is if we want the verifier to validate that information in the evidence bundle. I have the impression that nvtrust is also validating that -- maybe to make sure that the claims match what was requested. Do we want to do that?

attestation-agent/attester/src/nvidia/mod.rs
fitzthum: I think there are some limits to the existing CDI support. I would expect to support multiple devices in the midterm.

Since each attester gets turned into one submod (with a name like `gpu0`), I think it would probably be best to have each attester represent one device in the long term. I guess this is negotiable tho.

zvonkok:
Let's ensure we do the right thing here. On the Verifier side, we should consider Trustee as a broker that will reach out to the OCSP and RIM Service to perform the attestation. There will be no legal/sane way to download RIM files without proper license. There will be a "free" license providing access RIM and OCSP (only via direct to NRAS cloud services mode of attestation) and cannot be used for production and does not come with attestation service SLA.

attestation-agent/attester/src/nvidia/mod.rs
zvonkok: I can provide a snippet on how to check if GPU has CC=on, if we run with the GPU initrd/image this check is done anyway upon boot-up. We can make sure nothing changed during runtime.

attestation-agent/attester/src/nvidia/mod.rs
zvonkok: We support multiple GPUs in PPCIE mode so we will have GPU and NvSwtich attestation.

mythi: OK, I may have misunderstood nvtrust README that states: "There are only two supported GPU usage configurations: ALL GPUs are in CC mode. Each GPU can be assigned to one Confidential VM. In this scenario, use the CC verifier. ALL GPUs are in PPCIe mode. All GPUs must be assigned one Confidential VM. In this scenario, use the PPCIE verifier."

So if we cannot attest switches, we must accept single GPU configurations only?

mythi:
> Let's ensure we do the right thing here. On the Verifier side, we should consider Trustee as a broker that will reach out to the OCSP and RIM Service to perform the attestation. There will be no legal/sane way to download RIM files without proper license. 

AFAUI, this is what @cclaudio has in mind so we should be aligned: for starters, developers would manually store the relevant collateral in RVPS and the `nvidia-verifier` would use them from there.

> There will be a "free" license providing access to RIM and OCSP (only via direct to NRAS cloud services mode of attestation) and cannot be used for production and does not come with attestation service SLA.

Which of the [options](https://github.com/NVIDIA/nvtrust/tree/main/guest_tools#nvidia-attestation-service-integration-options) you are referring to here? 

FWIW, I was initially thinking we could start `nvidia-verifier` with NRAS based verification since it'd be easier to implement and then move on to options 2/3 gradually but I don't have any objections to jump there directly.

attestation-agent/attester/src/nvidia/mod.rs
mythi: I wonder if the link you provided is for the local verification flows. In our case, `nvidia-verifier` would make the version claims available (similar to entries 15/16) for the policies.

attestation-agent/attester/src/nvidia/mod.rs
zvonkok: Yes with Hopper. Blackwell is another story. We can start with one GPU and extend from there. We need the NvSwitches attestation sooner or later.

attestation-agent/attester/src/nvidia/mod.rs
mythi: I have this covered. `nvml-wrapper` just failed for this but it's fixed now. Will push an update to this shortly.

attestation-agent/attester/src/nvidia/mod.rs
mythi: thanks for the clarification. I agree and will re-work the code to follow this.

mythi:
> > one thing I noticed while testing was that  .check_confidential_compute_status() returns false if the is not ready...
> 
> I guess this PR isn't checking that now. Should we call `set_confidential_compute_state` at the end of this attester?

this is added now.

zvonkok:
@fitzthum This will be gone when we are enforcing the CC license. Do not rely  on that.

zvonkok:
@mythi Sorry I meant Option 2 with a local attester/verifier.

attestation-agent/attester/src/nvidia/mod.rs
mythi: > but this is a low level API that only receives the raw report data. The origin should be some upper semantics to handle.

is there a need to use runtime data hash here or could we just pass the raw nonce bytes for the device attesters?

Xynnn007: Only a bytes slice, i.e. nonce/hash/challenge. For devices they do not need to know what the slice mean, only include it exactly same inside the evidence. 

btw, the device attester itself should consider the input slice length and do trimming or padding if needed.

attestation-agent/attester/src/nvidia/mod.rs
mythi: > What if, for now, we support attestation requests containing only one device?

this is now reflected in `detect_platform()` but I'd prefer to keep the evidence list as part of `NvDeviceEvidence`.

fitzthum: I don't have a strong preference in the short term, but down the road having one submod for multiple devices is not ideal.

mythi: verifier `evaluate()` could be made to return a list of claims of same or different classes?

fitzthum: yeah that's not a bad idea. We'd have to have to shuffle a few things around but I think it would work.

bpradipt:
> Q: should we move the `nvidia-verifier` discussion to a Trustee RFC?

As I understand there are no licensing issues with the attester implementation.

The contention is on the verifier side on having NRAS integration as well as a standalone verifier. In that case can be move the verifier discussions to the Trustee? Or am I missing something ?

@zvonkok @fitzthum @mythi

attestation-agent/attester/src/nvidia/mod.rs
```diff
@@ -0,0 +1,106 @@
...
+
+            let dev_uuid = device
+                .uuid()
+                .context("Failed to get UUID for device {index}")?;
+
+            device_evidence_list.push(NvDeviceReportAndCert {
+                arch: dev_arch,
+                uuid: dev_uuid,
+                evidence: b64_engine.encode(hex::encode(report.attestation_report)),
+                certificate: b64_engine.encode(certificate.attestation_cert_chain),
```
cclaudio: There is no need to encode with both b64 and hex, that complicates the decoding since b64::decode() generates &[u8] but hex::decode() takes a string as argument. Can we use only `hex::encode()` for both the envidence and the certificate?

cclaudio:
I just submitted a nvidia-verifier PR
https://github.com/confidential-containers/trustee/pull/890

attestation-agent/attester/src/nvidia/mod.rs
mythi: this was mostly transparent to me as I pass it on ITA directly and it meets their input validation. do you think it's better to have `certificate` hex encoded as well?

Xynnn007: Oh. This double encoding is not good. Let's use only one and either is fine for me.

@mythi we can do some wrapper logic on the trustee side before sending to ITA.

mythi: It was like that originally but doing it here made things much simpler. Let me do some checking on what I really need for ITA.

Xynnn007: Yes. tbo ITA might also need to think over the double encoding

mythi: I think it's on NRAS side that takes b64

Xynnn007: Yeah, such logic should be handled inside, usually not exposed to callers ; )

mythi: What logic? ITA should not be touching the GPU evidence since it's sent to NRAS. I remember `b64_engine.encode(report.attestation_report)` alone did not work. All docs are vague on this but ITA Go client suggested that hex encoding was needed. As I understand it, NRAS forces this double encoding.

cclaudio: I suggested to use hex encode, but actually base64 would be better because the resulting encoded data is smaller when using base64. Hence less bytes to transmit over network :-)

Yes, I would base64 encode both the evidence and the certificate.

I am testing the `nvidia-verifier` with this PR ... I will let you know more later

mythi: Right now I'm not planning to change this because it seems it's what NRAS expects.

fitzthum: It seems pretty clear cut that we shouldn't be encoding the evidence twice here. Three people have suggested that. If NRAS needs an extra layer of encoding, we can just do that in the adapter on the Trustee side. No reason to carry that design choice here, especially since we will have verifiers that do not use NRAS.

mythi: > No reason to carry that design choice here, especially since we will have verifiers that do not use NRAS.

my thinking is it's better to do it in one place instead of two other places separately so `nvidia-verifier` without NRAS is the minority here. 

the compromise would be to do only `hex::encode(report.attestation_report)` to avoid any extra b64 decode and error handling

mythi: dropped the double encoding and added a comment why hex is chosen

attestation-agent/attester/src/nvidia/mod.rs
```diff
@@ -0,0 +1,108 @@
...
+        let nvml = Nvml::init()?;
+        let devices = nvml.device_count()?;
+
+        let mut device_evidence_list = vec![];
+
+        if report_data.len() < 32 {
+            warn!("NVIDIA Attester: Report data must be at least 32 bytes");
+        }
+
+        let nonce: [u8; 32] = report_data[0..32].try_into()?;
```
Xynnn007: note that if it is longer than 32 bytes, it will be truncated. also, if it is less than 32 bytes, the indexing `[]0..32]` here will panic. We can use soome similar logic as https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/attester/src/tdx/mod.rs#L127

Xynnn007 (commented):
One nit and others lgtm. Let me give an approval after the fix.

Also, the next target is to integrate this into guest image. Like discussed in the trustee meeting, we might need to change AA to be released in glibc for dynamic link issues.

Do you think we should open an issue to track it, or in some other way?

attestation-agent/attester/src/nvidia/mod.rs
mythi: updated

attestation-agent/attester/src/nvidia/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+#[async_trait::async_trait]
+impl Attester for NvAttester {
+    async fn get_evidence(&self, mut report_data: Vec<u8>) -> Result<TeeEvidence> {
+        let nvml = Nvml::init()?;
+        let devices = nvml.device_count()?;
+
+        let mut device_evidence_list = vec![];
+
+        if report_data.len() < NVIDIA_NONCE_SIZE {
+            warn!("NVIDIA Attester: Report data must be at least {NVIDIA_NONCE_SIZE} bytes");
```
Xynnn007: The term `must` hints that it will panic but it does not? Also, if it is longer than `NVIDIA_NONCE_SIZE` bytes, do we need to raise a warn also?
```suggestion
            warn!("NVIDIA Attester: Report data shorter than {NVIDIA_NONCE_SIZE} bytes will be padded with `\0`.");
```

mythi: > Also, if it is longer than `NVIDIA_NONCE_SIZE` bytes, do we need to raise a warn also?

we know it's longer than that in the case of KBS protocol / RCAR client so getting a warn always is not very nice IMO.

Xynnn007: `attester` is a crate, and can be called by other callers than CoCo, so it might not be sure that the caller knows about the padding/truncation. btw we already warns about the shorter case, so we can either warn both, or warn neither

mythi: I moved all of this to doc comments.

Xynnn007 (approved):
LGTM! thanks @mythi

mythi:
> Do you think we should open an issue to track it, or in some other way?

Perhaps we could just to it in the PR that bumps guest-components for the next Kata/CoCo release. We have the same case with CDH, right?

fitzthum (approved):
Let's roll

PR #1025, opened 2025-06-13
attester: tdx-attester: extend RTMRs

mythi:
Linux 6.16 adds the RTMR ABI that lets userspace to extend TEE RTMRs. Initially the support is added for TDX and follows https://github.com/torvalds/linux/blob/master/Documentation/ABI/testing/sysfs-devices-virtual-misc-tdx_guest

Change tdx-attester to use the ABI defined by the spec to do RTMR extend.

TODO:
- [x] testing

### DIFF
--- a/attestation-agent/attester/src/tdx/mod.rs
+++ b/attestation-agent/attester/src/tdx/mod.rs
@@ -20,6 +20,7 @@ mod rtmr;
 
 const TDX_REPORT_DATA_SIZE: usize = 64;
 
+const TDX_RTMR_PATH: &str = "/sys/devices/virtual/misc/tdx_guest/measurements";
 const TDX_GUEST_IOCTL: &str = "/dev/tdx_guest";
 
 pub fn detect_platform() -> bool {
@@ -49,16 +50,18 @@ fn get_quote_ioctl(report_data: &[u8]) -> Result<Vec<u8>> {
     }
 }
 
-// Return true if the TD environment can extend runtime measurement,
-// else false. The best guess at the moment is that if "TSM reports"
-// is available, the TD runs Linux upstream kernel and is _currently_
-// not able to do it.
+// Return true if the TD environment can extend runtime measurement.
+// The best guess at the moment is that if tdx-attest-dcap-ioctls
+// is enabled, runtime measurement is possible. It's also possible
+// if TDX_RTMR_PATH exits. The two are mutually exclusive.
 fn runtime_measurement_extend_available() -> bool {
-    if Path::new("/sys/kernel/config/tsm/report").exists() {
-        return false;
+    cfg_if::cfg_if! {
+            if #[cfg(feature = "tdx-attest-dcap-ioctls")] {
+                true
+            } else {
+                Path::new(TDX_RTMR_PATH).exists()
+        }
     }
-
-    true
 }
 
 #[derive(Serialize, Deserialize)]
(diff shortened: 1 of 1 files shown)


Xynnn007:
Nice to see that we are using the new feature of kernel. One thing not clear to me is whether we have a plan to update the kernel version of kata-side? https://github.com/kata-containers/kata-containers/blob/main/versions.yaml#L199

mythi:
I'm going to make the version bump once 6.16 is released

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -50,11 +51,9 @@ fn get_quote_ioctl(report_data: &[u8]) -> Result<Vec<u8>> {
 }
 
 // Return true if the TD environment can extend runtime measurement,
-// else false. The best guess at the moment is that if "TSM reports"
-// is available, the TD runs Linux upstream kernel and is _currently_
-// not able to do it.
+// else false.
 fn runtime_measurement_extend_available() -> bool {
-    if Path::new("/sys/kernel/config/tsm/report").exists() {
+    if Path::new(TDX_RTMR_PATH).exists() {
```
Xynnn007: Will this influence scenarios where we are using ioctl?

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -187,10 +186,17 @@ impl Attester for TdxAttester {
...
             hex::encode(extend_data)
         );
 
+        #[cfg(not(feature = "tdx-attest-dcap-ioctls"))]
+        std::fs::write(
+            Path::new(TDX_RTMR_PATH).join(format!("rtmr{rtmr_index}:sha384")),
+            extend_data,
+        )
+        .context("TDX Attester: failed to extend RTMR")?;
+
```
Xynnn007, line 174: I found that we have multiple `cfg` conditions here for ioctl/non-ioctl. Could you help to make two separate same-name functions and just call it here but enable one of them with feature flag?

attestation-agent/attester/src/tdx/mod.rs
mythi: right, I was not thinking the logic here very clearly. will re-work a bit.

attestation-agent/attester/src/tdx/mod.rs
mythi: I'll re-think this along with the updated `runtime_measurement_extend_available()` logic.

fitzthum:
I think we should also provide an option for people to use this for `bind_init_data` but that could potentially be future work.

attestation-agent/attester/src/tdx/mod.rs
mythi: @Xynnn007 do you think it's OK as it is now?

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -187,10 +190,19 @@ impl Attester for TdxAttester {
...
 
+        let rtmr_path = Path::new(TDX_RTMR_PATH);
+        if rtmr_path.exists() {
+            std::fs::write(
+                rtmr_path.join(format!("rtmr{rtmr_index}:sha384")),
+                extend_data,
+            )
+            .context("TDX Attester: failed to extend RTMR")?;
+        }
+
```
Xynnn007: If we are using `tdx-attest-dcap-ioctls` this code will also be executed, which is not expected?

mythi: It is expected: I think it's harmless to check the path exists (kinda what we do with TSM reports too)

Xynnn007: We'd better use also feature flag for this. A corner case is with io-ctl but coincidently the path exists.

mythi: okay, I thought about that corner base but that's almost impossible. Anyway, I'll rework on this further after my break...

mythi: the logic is now updated

mythi: @Xynnn007 PTAL

fitzthum (approved):
LGTM

mythi:
Finished testing. With a v6.16+ kernel, I get AA logs as follows:

```
[2025-08-25T13:14:48Z DEBUG attester::tdx] TDX Attester: extend RTMR3: 2330183d89750c6a374ac9827d33a33636e7876733544584c8ccc88df0df87eada37e5d71d0773502234c2b5fbc7448e                                            
[2025-08-25T13:14:48Z DEBUG ttrpc_aa::ttrpc_dep::server] AA (ttrpc): extend runtime measurement succeeded.
```

and without:

```
[2025-08-25T13:22:18Z ERROR ttrpc_aa::ttrpc_dep::server] AA (ttrpc): extend runtime measurement failed:  TDX Attester: Cannot extend runtime measurement on this system
```

Kata 3.20.0+ provides v6.16 confidential guest kernel so this should be ready for merge.

PR #1026, opened 2025-06-13
bump rust toolchain to 1.85.1

mythi:
Change how rust is configured an used in our CI. The changes are:

* bump the toolchain version pinning from 1.80.0 to 1.85.1
* move to use actions-rust-lang toolchain action since actions-rs is no
  longer maintained
* drop 'stable' from CI test matrix since it's a moving target and sometimes
  makes CI to fail in case a new stable rust version adds new linter checks
  that fail with our code. This should move to nighly clippy check or similar.

Update: add another commit that adds a weekly check for `cargo clippy` using rust stable.

### DIFF
--- a/.github/workflows/rust_stable_check.yml
+++ b/.github/workflows/rust_stable_check.yml
@@ -1,0 +1,55 @@
+name: Run clippy checks for Rust stable
+
+on:
+  # run every Sunday
+  schedule:
+    - cron: '0 3 * * 6'
+
+permissions:
+  contents: read
+
+jobs:
+  clippy:
+    strategy:
+      fail-fast: false
+      matrix:
+        package:
+          - api-server-rest
+          - coco_keyprovider
+          - crypto
+          - resource_uri
+          - kbs_protocol
+          - confidential-data-hub
+          - kms
+          - image-rs
+          - ocicrypt-rs
+        include:
+          - package: attestation-agent
+            feature-parameters: "--no-default-features --features rust-crypto,kbs,all-attesters,bin,ttrpc"
+          - package: kbc
+            feature-parameters: "--no-default-features --features openssl,cc_kbc,all-attesters"
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Code checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+
+      - name: Install Rust toolchain
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          toolchain: stable
+          override: true
+          components: rustfmt, clippy
+
+      - name: Install protoc
+        run: sudo apt update && sudo apt install -y protobuf-compiler
+
+      - name: Install TPM dependencies
+        run: sudo apt install -y libtss2-dev
+
+      - name: Install dm-verity dependencies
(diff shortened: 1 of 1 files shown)


mythi:
Ready for review.

Xynnn007 (commented):
Thanks. All look good. Only I think we can drop stable and introduce nightly in a same PR/commit. Otherwise, there is no CI for stable, which leads to the accumulation of many potential lint errors.

.github/workflows/aa_cc_kbc.yml
```diff
@@ -82,13 +73,9 @@ jobs:
           make ATTESTER=${{ matrix.attester }} && make install
 
       - name: Run rust lint check
-        uses: actions-rs/cargo@v1
-        with:
-          command: clippy
-          args: -p kbc --all-targets --features cc_kbc,${{ matrix.attester }},rust-crypto -- -D warnings
+        run: |
```
mkulke: nit: one-liners don't need a block string (`| ...`)

.github/workflows/aa_cc_kbc.yml
```diff
@@ -49,22 +46,16 @@ jobs:
...
     steps:
       - name: Code checkout
         uses: actions/checkout@v4
         with:
           fetch-depth: 1
 
-      - name: Install Rust toolchain (${{ matrix.rust }})
-        uses: actions-rs/toolchain@v1
+      - name: Install Rust toolchain
+        uses: actions-rust-lang/setup-rust-toolchain@v1
```
mkulke, line 57: eventually we need to pin reusable actions to their SHA's if we adopt more supply-chain checks, we might want to start with that:
```suggestion
        uses: actions-rust-lang/setup-rust-toolchain@9d7e65c320fdb52dcd45ffaa68deb6c02c8754d9 # v1.12.0
```

.github/workflows/aa_cc_kbc.yml
mythi: I will fix this. I think I just copied this from the step above and started to use the same syntax

.github/workflows/aa_cc_kbc.yml
mythi: I could do this in a follow-up PR

.github/workflows/ocicrypt_rs_build.yml
```diff
@@ -45,104 +39,71 @@ jobs:
...
         run: |
           sudo apt install -y protobuf-compiler libprotobuf-dev
 
       - name: Run cargo build
-        uses: actions-rs/cargo@v1
-        with:
-          command: build
-          args: -p ocicrypt-rs --all-features
+        run:
+          cargo build -p ocicrypt-rs --all-features
```
mkulke: afaiu this will not work. the statement needs to be in the same line without > or | prefixes

```suggestion
        run: cargo build -p ocicrypt-rs --all-features
```

mythi: oh, `actionlint` did not complain about it but will change things like this.

mythi: fixed

fitzthum (approved):
LGTM

PR #1028, opened 2025-06-17
kbs_protocol: update rcar_client test policy rules

mythi:
Trustee multi-device support changed the EAR token cpu submod path to include an index.

Update rcar_client test policy rules accordingly to get the test passing. Without the change, the test fails with an 'unauthorized' error.


mythi:
It seems OK now:
```
test client::rcar_client::test::test_client ... ok
```

mythi:
@Xynnn007 @mkulke would be good to get this merged asap so that tests can run again.

PR #1031, opened 2025-06-18
AA | Add initdata plaintext support

### LINKED ISSUE #1002: [AA] | Promote `--initdata` parameter
Currently, we provide AA with a startup parameter called `--initdata`, which actually receives the digest value of initdata to verify whether the parameter is successfully bound to the current platform initdata at startup.

However, according to this design, we need an additional component to parse the content of initdata.toml, calculate its hash, and then pass the hash value to the AA startup parameter. Currently, kata-agent and process-user-data need to implement this repeated logic. 

After some discussions with @mkulke , we hope to move this work to AA, that is,

1. External components only need to extract configuration files from initdata.toml

2. AA receives path to initdata.toml as a launch parameter when it starts, then calculates the hash and executes actions such as bind_init_data

Besides and personally, our parameter `--initdata` can accept a base64-encoded initdata digest hash value. This interface is relatively raw and simple, and I think it still has use cases, so I propose to keep the original function of the `--initdata` parameter and add the `--initdata-toml` parameter to achieve the above function.

cc @fitzthum

Xynnn007:
This patch adds initdata toml support to KbsProtocol.

Close #1002

### DIFF
--- a/attestation-agent/attestation-agent/src/bin/grpc-aa/main.rs
+++ b/attestation-agent/attestation-agent/src/bin/grpc-aa/main.rs
@@ -37,13 +37,13 @@ struct Cli {
     #[arg(short, long)]
     config_file: Option<String>,
 
-    /// Initdata to be verified by AA. If initdata check failed, AA will failed to launch.
+    /// Initdata digest to be verified by AA. If initdata check failed, AA will failed to launch.
     /// The initdata should be base64 standard encoding.
     ///
     /// Example:
-    /// `--initdata AAAAAAAAAAAA`
+    /// `--initdata_digest AAAAAAAAAAAA`
     #[arg(short, long)]
-    initdata: Option<String>,
+    initdata_digest: Option<String>,
 }
 
 #[tokio::main]
@@ -54,8 +54,8 @@ pub async fn main() -> Result<()> {
     let attestation_socket = cli.attestation_sock.parse::<SocketAddr>()?;
 
     let mut aa = AttestationAgent::new(cli.config_file.as_deref()).context("start AA")?;
-    if let Some(initdata) = cli.initdata {
-        info!("Initdata is given by parameter, try to check.");
+    if let Some(initdata) = cli.initdata_digest {
+        info!("Initdata digest is given by parameter, try to check.");
         let initdata = base64::engine::general_purpose::STANDARD
             .decode(&initdata)
             .context("base64 decode initdata")?;
--- a/attestation-agent/attestation-agent/src/bin/grpc-aa/main.rs
+++ b/attestation-agent/attestation-agent/src/bin/grpc-aa/main.rs
@@ -6,7 +6,7 @@
 mod server;
 
 use anyhow::*;
-use attestation_agent::{AttestationAPIs, AttestationAgent};
+use attestation_agent::{initdata::Initdata, AttestationAPIs, AttestationAgent};
 use base64::Engine;
 use clap::Parser;
 use log::{debug, info};
@@ -40,10 +40,28 @@ struct Cli {
     /// Initdata digest to be verified by AA. If initdata check failed, AA will failed to launch.
     /// The initdata should be base64 standard encoding.
     ///
+    /// Note that this is an alternative to `--initdata_toml`.
+    ///
+    /// If both parameters `--initdata_toml` and `initdata_digest` are provided, parameter
+    /// `--initdata_toml` takes precedence.
+    ///
     /// Example:
     /// `--initdata_digest AAAAAAAAAAAA`
     #[arg(short, long)]
     initdata_digest: Option<String>,
+
+    /// Path to the Initdata TOML file to be verified by AA. If initdata check failed, AA will failed to launch.
+    /// The initdata should be base64 standard encoding.
+    ///
+    /// Note that this is an alternative to `--initdata_digest`.
+    ///
+    /// /// If both parameters `--initdata_toml` and `initdata_digest` are provided, parameter
+    /// `--initdata_toml` takes precedence.
+    ///
+    /// Example:
+    /// `--initdata_toml /path/to/initdata.toml`
+    #[arg(short = 't', long)]
+    initdata_toml: Option<String>,
 }
 
 #[tokio::main]
@@ -54,15 +72,28 @@ pub async fn main() -> Result<()> {
     let attestation_socket = cli.attestation_sock.parse::<SocketAddr>()?;
 
     let mut aa = AttestationAgent::new(cli.config_file.as_deref()).context("start AA")?;
-    if let Some(initdata) = cli.initdata_digest {
-        info!("Initdata digest is given by parameter, try to check.");
+
+    let mut initdata_digest = None;
+    if let Some(initdata_toml_path) = cli.initdata_toml {
+        info!("Initdata TOML file is given by parameter");
+        let initdata_toml =
+            std::fs::read_to_string(&initdata_toml_path).context("read initdata toml file")?;
+        let (_, digest) = Initdata::parse_and_get_digest(&initdata_toml)?;
+        aa.set_initdata_toml(initdata_toml);
+        initdata_digest = Some(digest);
+    } else if let Some(initdata) = cli.initdata_digest {
+        info!("Initdata digest is given by parameter");
         let initdata = base64::engine::general_purpose::STANDARD
             .decode(&initdata)
             .context("base64 decode initdata")?;
-        let res = aa
-            .bind_init_data(&initdata)
-            .await
-            .context("The initdata supplied by the parameter is inconsistent with that of the current platform.")?;
+        initdata_digest = Some(initdata);
+    }
+
+    if let Some(initdata_digest) = initdata_digest {
+        let res = aa.bind_init_data(&initdata_digest).await.context(
+        "The initdata supplied by the parameter is inconsistent with that of the current platform.",
+    )?;
+
         match res {
             attester::InitDataResult::Ok => info!("Check initdata passed."),
             attester::InitDataResult::Unsupported => {
--- a/attestation-agent/attestation-agent/src/lib.rs
+++ b/attestation-agent/attestation-agent/src/lib.rs
@@ -5,10 +5,9 @@
 
 use anyhow::{bail, Context, Result};
 use async_trait::async_trait;
-use attester::CompositeAttester;
-use crypto::HashAlgorithm;
-use kbs_types::{Tee, TeePubKey};
-use std::{str::FromStr, sync::Arc};
+use attester::{detect_attestable_devices, detect_tee_type, BoxedAttester};
+use kbs_types::Tee;
+use std::{collections::HashMap, str::FromStr, sync::Arc};
 use tokio::sync::{Mutex, RwLock};
 
 pub use attester::InitDataResult;
@@ -64,18 +63,8 @@ pub trait AttestationAPIs {
     async fn get_evidence(&self, runtime_data: &[u8]) -> Result<Vec<u8>>;
 
     /// Get TEE hardware evidence from all additional attesters with runtime data
-    /// included.
+    /// included. If no additional attester is configured, it will return an empty vector.
     async fn get_additional_evidence(&self, runtime_data: &[u8]) -> Result<Vec<u8>>;
-
-    /// Get the composite evidence from a confidential guest.
-    /// This includes the primary and additional evidence, which
-    /// are bound together by the runtime data.
-    async fn get_composite_evidence(
-        &self,
-        tee_pubkey: TeePubKey,
-        nonce: String,
-        hash_algorithm: HashAlgorithm,
-    ) -> Result<Vec<u8>>;
 
     /// Extend runtime measurement register
     async fn extend_runtime_measurement(
@@ -94,10 +83,12 @@ pub trait AttestationAPIs {
 
 /// Attestation agent to provide attestation service.
 pub struct AttestationAgent {
+    primary_tee: Tee,
     config: RwLock<Config>,
-    attester: Arc<CompositeAttester>,
     eventlog: Option<Mutex<EventLog>>,
     initdata: Option<String>,
+    primary_attester: Arc<BoxedAttester>,
+    additional_attesters: HashMap<Tee, BoxedAttester>,
 }
 
 impl AttestationAgent {
@@ -105,7 +96,7 @@ impl AttestationAgent {
         let config = self.config.read().await;
         if config.eventlog_config.enable_eventlog {
             let eventlog = EventLog::new(
-                self.attester.clone(),
+                self.primary_attester.clone(),
                 config.eventlog_config.eventlog_algorithm,
                 config.eventlog_config.init_pcr,
             )
@@ -132,13 +123,21 @@ impl AttestationAgent {
         debug!("Using config: {config:#?}");
         let config = RwLock::new(config);
 
-        let attester = Arc::new(CompositeAttester::new()?);
+        let primary_tee = detect_tee_type();
+        let additional_tees = detect_attestable_devices();
+
+        let mut additional_attesters = HashMap::new();
+        for tee in additional_tees {
+            additional_attesters.insert(tee, tee.try_into()?);
+        }
 
         Ok(AttestationAgent {
+            primary_tee,
             config,
-            attester,
             eventlog: None,
             initdata: None,
+            additional_attesters,
+            primary_attester: Arc::new(primary_tee.try_into()?),
         })
     }
 
@@ -194,8 +193,8 @@ impl AttestationAPIs for AttestationAgent {
     /// data included.
     async fn get_evidence(&self, runtime_data: &[u8]) -> Result<Vec<u8>> {
         let evidence = self
-            .attester
-            .primary_evidence(runtime_data.to_vec())
+            .primary_attester
+            .get_evidence(runtime_data.to_vec())
             .await?;
         Ok(evidence.to_string().into_bytes())
     }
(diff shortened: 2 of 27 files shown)


fitzthum:
I'm not really a fan of changing the Attester lib interface and that seems mostly unrelated to supporting plaintext initdata.

It looks like you are putting the init-data into the tee evidence. I think it might make more sense to introduce init-data as a top-level field in the kbs-protocol. 

Anyway, I'll stop bugging you until it's not a draft.

Xynnn007:
The best way to integrate initdata is to extend kbs protocol's `Attestation` field into
```
{
  "tee-evidence": xx,
  "initdata": xx,
  "runtime-data": xx
}
```

If we don't want to modify the kbs protocol, putting it in `tee-evidence` is also an option (as shown in this PR). So what's your taste?


Also, combined Evidence is a concept on the KBS Protocol. Since it does not modify the underlying Attester hardware plug-in, it is better to keep its code in the kbs_ptorocol library to keep the AA interface clean. I have proposed this point of view https://github.com/confidential-containers/guest-components/pull/991#discussion_r2074634212

attestation-agent/attestation-agent/src/bin/grpc-aa/main.rs
```diff
@@ -40,10 +40,23 @@ struct Cli {
...
+    /// Path to the Initdata TOML file to be verified by AA. If initdata check failed, AA will failed to launch.
+    /// The initdata should be base64 standard encoding.
+    ///
+    /// Note that this is an alternative to `--initdata_digest`, and only one of them should be used.
+    /// If this option is provided, the file will be read and its content will be used for verification.
+    ///
+    /// Example:
+    /// `--initdata_toml /path/to/initdata.toml`
+    #[arg(short = 't', long)]
+    initdata_toml: Option<String>,
```
Apokleos, line 64: Apologies first of all. 
Perhaps I lack the related context or history of initdata, I just base on the Specification of `InitData`, and get its design of initdata which supports not only files with format `toml` but also `json`. To be honest, I am confused about it why we only support toml in code ? It came from the practice or to prevent some potential issues ?

Based on my understanding of InitData, what's about just using the `initdata` for the initdata plaintext and  `initdata-digest` for its digest ?

attestation-agent/kbs_protocol/src/client/mod.rs
```diff
@@ -46,6 +46,9 @@ pub struct KbsClient<T> {
 
     /// token
     pub(crate) token: Option<Token>,
+
+    /// initdata
```
Apokleos: I think it'd be better to make the initdata digest or the initdata content more concrete.

attestation-agent/kbs_protocol/src/evidence_provider/aa_ttrpc.rs
```diff
@@ -41,31 +41,42 @@ impl AAEvidenceProvider {
...
                 &req,
             )
             .await
             .map_err(|e| Error::AAEvidenceProvider(format!("call ttrpc failed: {e}")))?;
+        let evidence = serde_json::from_slice(&res.Evidence)
+            .map_err(|e| Error::AAEvidenceProvider(format!("illegal evidence format: {e}")))?;
+        Ok(evidence)
+    }
+
+    /// Get additional evidence with as runtime data (report data, challege)
```
Apokleos:
```suggestion
    /// Get additional evidence with runtime data (report data, challege)
```

attestation-agent/attestation-agent/src/bin/grpc-aa/main.rs
```diff
@@ -70,6 +83,24 @@ pub async fn main() -> Result<()> {
...
+            "The initdata supplied by the parameter is inconsistent with that of the current platform.",
+        )?;
+        match res {
+            attester::InitDataResult::Ok => info!("Check initdata passed."),
+            attester::InitDataResult::Unsupported => {
+                info!("Platform does not support initdata checking. Jumping.")
+            }
+        }
+        aa.set_initdata(initdata);
+    }
```
Apokleos: Should this code block be wrapped into a generic method? As I notice other place also use the same process handling logic.

Apokleos (commented):
Thx Ding @Xynnn007 good job! comments left, some of them might not true issues.

Xynnn007:
@fitzthum There is really no right or wrong here, just different design ideas. I must admit that my code is not the best when I rechecked. Let me try to summrize and share my ideas for you to choose from. In the KBS protocol, we have a simple evidence field, which we used to carry a single CPU evidence before. With the advent of device attestation, combined evidence took its place. 

In any case, we have to define an new wrapper Evidence abstraction format at a certain level, which contains
1. the CPU and device evidence. 
2. optional plaintext `runtime data` and `initdata`.

There are two design principles
1. decoupling. That means lower level crates do not need to have parameters that belongs to higher level abstration.
2. keeping the abstraction of the production and consumption ends at the same level. That means if an abstraction is AA level, it should be handled AS level.

 
Note that `TeeKeys` and `Nonce` should be included in the content of `runtime data`, and shielded from the underlying `Attester` plugin to follow decoupling.

There are some design ideas for this. 

1. Let `attester` crate itself provide the Evidence format abstraction, which is the `CompositeEvidence` you contributed before. We need to add two additional optional fields to the current CompositeEvidence, namely `runtime data` and `initdata fields`, to receive runtime data json and initdata toml. In this way, KbsProtocol organizes runtime data by itself (such as putting in teekey and nonce) to call `attester` crate and obtain evidence. On the symmetric side, AS provides the interface of verify compositeEvidence. In order to keep the interface abstraction of the producer (AA) and the consumer (AS) at the same level, the interface provided by AA should be `get_composite_evidence(Option<runtime_data>, Option<init_data>)`, and AS should provide `VerifyCompositeEvidence(evidence, Option<runtime_data>, Option<initdata>)`. A new CompositeEvidenceVerifier should be added in `verifier` crate.
2. Let `CompositeEvidence` be an abstration of `attestation-agent` crate level. That means AA provides `get_composite_evidence(Option<runtime_data>, Option<init_data>)` and calls `get_evidence` of `attesters`. AA manages different type, nested structure and generation logic. At the same level, AS provides `VerifyCompositeEvidence(evidence, Option<runtime_data>, Option<initdata>)`. AS manages different type, nested structure and extraction logic.
3. Let `CompositeEvidence` be kbs protocol level. That means `kbs_protocol` calls AA's `get_evidence` and `get_additional_evidences/get_device_evidences` API. At the same level, KBS manages different type, nested structure and extraction logic. KBS calls AS's `verify_evidence` API. This PR I nearly follow this way but not decouple enough because I did not bring in runtime data abstraction to cover teekeys and nonce.

I hope this makes sense to take a 1 from 3 choice, and before we are getting tagged, we can make a decision.

attestation-agent/attestation-agent/src/bin/grpc-aa/main.rs
Xynnn007: Oh yes. The original [spec](https://github.com/confidential-containers/trustee/blob/main/kbs/docs/initdata.md#examples-for-different-encodings) of initdata declared supporting different encodings, i.e. initdata toml and initdata json, but we choose initdata toml in kata/coco.

Xynnn007:
> The runtime data is tricky because while the components of the runtime data, the tee pubkey and the nonce, are clearly part of the kbs-protocol, with composite evidence, we need that information to generate the evidence.

Yes, this is totally fine, and it might not be that difficult imo. For an `attester` (we discuss solution 3), it doesn't need to know what is contained in the runtime data, so `kbs_protocol` hash them (specifically `{teekeys, nonce}`) as input to the additional `attester` to obtain evidence A. Then merge evidence A back to `{teekeys, nonce, A}` to get primary evidence B. Then `kbs_protocol` wraps both A and B into a CompositeEvidence. This can ensure that the attester library does not know any Kbs protocol semantics, but only a slice derived from a hash calculation, and also ensure that the Kbs protocol can correctly obtain and identify tee-keys and nonces.


Eventlog itself belongs to the logic of evidence but not hardware evidence. I think what you mean is that multiple platforms will support eventlog and there is no need to sink it into a specific tee plug-in. This is fine. A better way is to wrap eventlog and hw evidence as an evidence format. In theory, this format should be independent of the lower-level attester plugin and the upper-level kbs protocol.

> Anyway, the thing that I have the strongest opinion about is that the plaintext-init-data should be part of the kbs-protocol rather than part of the evidence.

I think I have done the same as your idea. If you take a look at the PR, you can find that the `Attestation` is currently

```json
{
    "evidence": {
        $Composite-Evidence,
        "initdata_toml": $initdata-toml
    },
    "teekeys": $tee-public-key,
    "nonce": $nonce,
}
```

Though I think
```json
{
    "evidence": {
        $Composite-Evidence,
        "initdata_toml": $initdata-toml,
        "runtime_data": {
            "teekeys": $tee-public-key,
            "nonce": $nonce,
        }
    }
}
```

or 
```json
{
    "evidence": $Composite-Evidence,
    "initdata_toml": $initdata-toml,
    "runtime_data": {
        "teekeys": $tee-public-key,
        "nonce": $nonce,
     }
}
```

is better for maintaince.

mkulke:
> > The runtime data is tricky because while the components of the runtime data, the tee pubkey and the nonce, are clearly part of the kbs-protocol, with composite evidence, we need that information to generate the evidence.
> 
> Yes, this is totally fine, and it might not be that difficult imo. For an `attester` (we discuss solution 3), it doesn't need to know what is contained in the runtime data, so `kbs_protocol` hash them (specifically `{teekeys, nonce}`) as input to the additional `attester` to obtain evidence A. Then merge evidence A back to `{teekeys, nonce, A}` to get primary evidence B. Then `kbs_protocol` wraps both A and B into a CompositeEvidence. This can ensure that the attester library does not know any Kbs protocol semantics, but only a slice derived from a hash calculation, and also ensure that the Kbs protocol can correctly obtain and identify tee-keys and nonces.
> 
> Eventlog itself belongs to the logic of evidence but not hardware evidence. I think what you mean is that multiple platforms will support eventlog and there is no need to sink it into a specific tee plug-in. This is fine. A better way is to wrap eventlog and hw evidence as an evidence format. In theory, this format should be independent of the lower-level attester plugin and the upper-level kbs protocol.
> 
> > Anyway, the thing that I have the strongest opinion about is that the plaintext-init-data should be part of the kbs-protocol rather than part of the evidence.
> 
> I think I have done the same as your idea. If you take a look at the PR, you can find that the `Attestation` is currently
> 
> ```json
> {
>     "evidence": {
>         $Composite-Evidence,
>         "initdata_toml": $initdata-toml
>     },
>     "teekeys": $tee-public-key,
>     "nonce": $nonce,
> }
> ```
> 
> Though I think
> 
> ```json
> {
>     "evidence": {
>         $Composite-Evidence,
>         "initdata_toml": $initdata-toml,
>         "runtime_data": {
>             "teekeys": $tee-public-key,
>             "nonce": $nonce,
>         }
>     }
> }
> ```
> 
> or
> 
> ```json
> {
>     "evidence": $Composite-Evidence,
>     "initdata_toml": $initdata-toml,
>     "runtime_data": {
>         "teekeys": $tee-public-key,
>         "nonce": $nonce,
>      }
> }
> ```
> 
> is better.

to me the 3rd variant (evidence + init-data + runtime-data on the top level) makes the most sense. and while we're bikeshedding... maybe we want to have consistent key names, too. afaiu, we need to bump the kbs protocol, so we should be able to change them.

 I don't think we need to spell out the file ending and can just call the the key `init-data` (and change the hyphenisation to `runtime-data`). If we really want to have the fact that the body is toml in the protocol semantics (idk, for future extensibility maybe), we can make it explicit:
 
 ```json
"init-data": { "format": "toml", "body": "..." }
```

**Comment by fitzthum** [comment]
> The structure we finally reached a consensus on is actually the concept of Evidence at the Kbs Protocol level;

Yes, I've got this. My question is whether this means you will be making a change to the kbs-protocol doc to describe the composite evidence and if we will be requiring anyone who uses the kbs-protocol to use our composite evidence scheme.

**Comment by Xynnn007** [comment]
> Yes, I've got this. My question is whether this means you will be making a change to the kbs-protocol doc to describe the composite evidence and if we will be requiring anyone who uses the kbs-protocol to use our composite evidence scheme.

Theoretically, it is not mandatory. But we may need to mention the runtime-data and init-data fields in the protocol design document to verify the binding relationship on the Rats Verifier.

In addition, the kbs_protocol library and kbs are examples of the KBS protocol implementation, so it is not necessary to require the document itself.

**Comment by fitzthum** [comment]
> In addition, the kbs_protocol library and kbs are examples of the KBS protocol implementation, so it is not necessary to require the document itself.

This seems a little bit inconsistent. If we implement in the kbs_protocol crate, it seems like it should be covered by the KBS protocol doc. That said, I don't think composite evidence should be a mandatory part of the protocol.

Anyway let's do the code parts first and then we can think about the doc.

**Comment by Xynnn007** [comment]
The lint error can be fixed once https://github.com/confidential-containers/guest-components/pull/1042 gets merged.

File: attestation-agent/kbs_protocol/src/client/mod.rs (hunk 2 above)
Comment (Xynnn007) [reply attestation-agent/kbs_protocol/src/client/mod.rs]: thanks, fixed

File: attestation-agent/kbs_protocol/src/evidence_provider/aa_ttrpc.rs (hunk 3 above)
Comment (Xynnn007) [reply attestation-agent/kbs_protocol/src/evidence_provider/aa_ttrpc.rs]: Thanks, fixed

File: attestation-agent/attestation-agent/src/bin/grpc-aa/main.rs (hunk 4 above)
Comment (Xynnn007) [reply attestation-agent/attestation-agent/src/bin/grpc-aa/main.rs]: Nice idea. I merged the logic inside the AA lib to make the parsing logic a blackbox

File: attestation-agent/attestation-agent/src/bin/grpc-aa/main.rs [hunk 5]
```diff
@@ -70,6 +83,12 @@ pub async fn main() -> Result<()> {
             }
         }
     }
+    if let Some(initdata_toml_path) = cli.initdata_toml {
```
mkulke: would it make sense to consolidate the handling of both options (toml file and digest), so make it more obvious. something like this:

```rust
let mut initdata_digest = cli.initdata_digest; 
if let Some(toml) = cli.initdata_toml {
  if initdata_digest.is_some() {
    bail!("initdata digest already provided");
  }
  let digest = calculate_initdata_digest(toml);
  initdata_digest = Some(digest);
}

if let Some(digest) = initdata_digest {
   aa.bind_init_data(digest).await?;
   ...
}
```

mkulke (commented):
Is there a legitimate use case for `--initdata_digest` or do we generally always want to set the initdata body, so that AA can attach it to the KBS exchange?

attestation-agent/attestation-agent/src/bin/grpc-aa/main.rs
Xynnn007: Yes, you are right that we should have exclusive logic. Let me do this.

Xynnn007:
> Is there a legitimate use case for --initdata_digest or do we generally always want to set the initdata body, so that AA can attach it to the KBS exchange?

Sometimes inside a CVM (not CoCo), the initdata digest (hostdata field) does not follow initdata spec. This is for this case.

attestation-agent/attestation-agent/src/bin/grpc-aa/main.rs
Xynnn007: I added some logic for the two configurations to have override priority to avoid direct errors. Relative information will be print via CLI help. PTAL

attestation-agent/attestation-agent/src/lib.rs
```diff
@@ -130,14 +123,50 @@ impl AttestationAgent {
...
+    /// an error.
+    ///
+    /// Either initdata is not support on the platform or the initdata
+    /// matches, it will return an `Ok`.
+    pub async fn set_initdata_toml(&mut self, initdata_toml: String) -> Result<()> {
+        let (_, digest) =
+            Initdata::parse_and_get_digest(&initdata_toml).context("parse initdata from toml")?;
+        self.initdata = Some(initdata_toml);
+        debug!("Initdata digest: {digest:?}");
+        let res = self.bind_init_data(&digest).await.context(
```
mkulke: maybe it makes sense to have only a single place where we call bind_init_data()? we can move this up to the caller, where we already call bind_init_data() w/ the digest when `--initdata_digest` is used. maybe we don't need this function at all then? since it's just

```rust
let (_, digest) = Initdata::parse_and_get_digest(&initdata_toml).context("parse initdata from toml")?;
self.initdata = Some(initdata_toml);
```

Xynnn007: Oh yes. The first version would be that form but with suggestion https://github.com/confidential-containers/guest-components/pull/1031#discussion_r2162892288 I tried to bring similar code into one place so the form is now.

Xynnn007: Concretely we maintain two parts of same code for both grpc and ttrpc

```rust
        info!("Initdata TOML file is given by parameter, try to check.");
        let initdata =
            std::fs::read_to_string(&initdata_toml).context("read initdata toml file")?;
        let (_, digest) = attester::Initdata::parse_and_get_digest(&initdata)
            .context("parse initdata from toml")?;
        info!("Initdata digest: {:?}", digest);
        let res = aa.bind_init_data(&digest).await.context(
            "The initdata supplied by the parameter is inconsistent with that of the current platform.",
        )?;
        match res {
            attester::InitDataResult::Ok => info!("Check initdata passed."),
            attester::InitDataResult::Unsupported => {
                info!("Platform does not support initdata checking. Jumping.")
            }
        }
        aa.set_initdata(initdata);
```

now both side

```rust
        let initdata_toml =
            std::fs::read_to_string(&initdata_toml_path).context("read initdata toml file")?;
        aa.set_initdata_toml(initdata_toml).await?;
```

I am not sure if this change meets @Apokleos 's need, but if you have better idea I can fix asap.

mkulke: I think it's fine/doesn't hurt to move parts of it into its own function, my suggestion was to have just 1 place where we call `bind_init_data(digest)`, ideally at a very obvious place, not nested too deeply in a fn call stack (otherwise it might get dropped by accident in refactorings, etc...) since it's a key/critical function.

set_initdata() could just return the digest and we proceed as with `--initdata_digest`. the fn name is probably not very accurate, but it's already doing 3 things (set, hash, bind), then it would do 2 things (set + hash) 

image

if code duplication for grpc and ttrpc is a concern then there could be a `process_init_data(init_data: Initdata)` fn, that accepts and handles both cases:

```rust
enum Initdata {
   Toml(String),
   Digest(String),
}
```

Xynnn007: Good idea. Let me do as the suggestion.

Xynnn007: The `process_init_data(init_data: Initdata)` needs to receive a parsed initdata, so either the function api receives the cli and aa instance, or we keep the conditional judgement logic outside the function.

I choose the later, and just a `set_initdata_toml` function.

mkulke (approved):

attestation-agent/kbs_protocol/src/evidence_provider/native.rs
```diff
@@ -3,40 +3,78 @@
...
-        self.0
-            .composite_evidence(tee_pubkey, nonce, hash_algorithm)
+    async fn primary_evidence(&self, runtime_data: Vec<u8>) -> Result<TeeEvidence> {
+        self.primary_attester
+            .get_evidence(runtime_data)
             .await
             .map_err(|e| Error::GetEvidence(e.to_string()))
     }
 
+    async fn get_additional_evidences(&self, runtime_data: Vec<u8>) -> Result<String> {
```
fitzthum: nit: sadly the plural of evidence is evidence, so we should probably change all these.

attestation-agent/kbs_protocol/src/client/rcar_client.rs
```diff
@@ -97,6 +98,13 @@ fn get_hash_algorithm(extra_params: serde_json::Value) -> Result<HashAlgorithm>
     Ok(algorithm)
 }
 
+fn serialize_canon_json<T: Serialize>(value: T) -> anyhow::Result<Vec<u8>> {
```
fitzthum: nit: maybe call this `serialize_json_canonically`

fitzthum (approved):
A couple small nits, but this looks good.

attestation-agent/kbs_protocol/src/evidence_provider/native.rs
Xynnn007: Oops. Let me fix this

Xynnn007:
Let me take a rebase after https://github.com/virtee/kbs-types/pull/65 gets merged.

PR #1036, opened 2025-06-23
CDH: add registy configuration item for example config

Xynnn007:
Now the registry configuration takes effect via CDH configuration file, but we did not show it in the example config.

### DIFF
--- a/confidential-data-hub/example.config.toml
+++ b/confidential-data-hub/example.config.toml
@@ -117,13 +117,28 @@ image_security_policy_uri = "kbs:///default/security-policy/test"
 # By default this value is not set.
 authenticated_registry_credentials_uri = "kbs:///default/credential/test"
 
+# Registry configuration supports defining registry blocking, mirroring,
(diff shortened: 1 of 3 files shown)


confidential-data-hub/example.config.toml
```diff
@@ -117,13 +117,28 @@ image_security_policy_uri = "kbs:///default/security-policy/test"
 # By default this value is not set.
 authenticated_registry_credentials_uri = "kbs:///default/credential/test"
 
+# Registry configuration supports define registry blocking, mirroring,
```
bpradipt: nit: s/define/defining

bpradipt (approved):
/lgtm
Just a minor typo..

confidential-data-hub/example.config.toml
Xynnn007: Thanks, fixed together with another part.

fitzthum (approved):
LGTM

PR #1037, opened 2025-06-23
CDH: adds support to read auth URI from the env

Xynnn007:
Now the registry auth is a little different from other secrets. It is not confidential because the host side of CoCo needs the auth to pull manifest.

In peerpod scenario, we need a fixed path of registry auth file inside the guest to be set from the control tunnel. But after moving guest pull from kata-agent to CDH, the auth uri can only be set via CDH.toml via initdata. This would make burden to users. To keep the previous behavior, we use this env to override the value.

cc @stevenhorsman @mkulke

### DIFF
--- a/confidential-data-hub/hub/src/config.rs
+++ b/confidential-data-hub/hub/src/config.rs
@@ -20,6 +20,9 @@ cfg_if::cfg_if! {
     }
 }
 
+const CDH_DEFAULT_IMAGE_AUTHENTICATED_REGISTRY_CREDENTIALS: &str =
+    "CDH_DEFAULT_IMAGE_AUTHENTICATED_REGISTRY_CREDENTIALS";
+
 #[derive(Clone, Deserialize, Debug, PartialEq)]
 pub struct KbsConfig {
     pub name: String,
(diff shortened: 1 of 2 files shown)


confidential-data-hub/hub/src/config.rs
stevenhorsman, line 1: Is it worth adding a test to validate the precedence of the config file over the env? e.g.
expanding the `read_config` test to check that is `CDH_DEFAULT_IMAGE_AUTHENTICATED_REGISTRY_CREDENTIALS=file:///test` is set and `authenticated_registry_credentials_uri = "kbs:///default/credential/test"`, then `config.image.authenticated_registry_credentials_uri` is `kbs:///default/credential/test`?

confidential-data-hub/hub/src/config.rs
```diff
@@ -94,6 +97,13 @@ impl CdhConfig {
             }
         };
 
+        if let std::result::Result::Ok(env) =
+            env::var(CDH_DEFAULT_IMAGE_AUTHENTICATED_REGISTRY_CREDENTIALS)
+        {
+            info!("Read authenticated registry credentials URI from env: {env}");
+            config.image.authenticated_registry_credentials_uri = Some(env);
```
mkulke, line 104: this will overwrite the config file. If we want to have the config file to take precedence, we need to move this up. But I think the usual behaviour is ENV > config-file, so we might just want to change the remark in example.config.toml

Xynnn007: Agreed. This would help to make peerpod logic a black box.

confidential-data-hub/hub/src/config.rs
Xynnn007: Yes, thanks. added!

mkulke (approved):

PR #1042, opened 2025-06-27
CDH: execute config unit test serially

Xynnn007:
Now we support to set the registry configuration file path via env, and this case is being test in unit test.

We need to ensure the tests are executed serially and not concurrently to avoid an env setting in test case 1 to influence in test case 2.


BbolroC (approved):
LGTM unless there is a clippy error. Thanks @Xynnn007 !

PR #1045, opened 2025-07-02
Add Hygon DCU device attestation support

Xynnn007:
This patch brings in Hygon DCU device attestation support, together with some fixes for csv attester.

### DIFF
--- a/attestation-agent/attester/src/csv/mod.rs
+++ b/attestation-agent/attester/src/csv/mod.rs
@@ -3,69 +3,57 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+//! # Hygon CSV Attester
+//!
+//! By default the attester will only get the attestation report from the hardware.
+//! In some cases the verifier might not have network access. [`CSV_INCLUDE_CERT_CHAIN_ENV`]
+//! env can be set to `true`, the attester will include the HSK and CEK cert in the
+//! report and send to the verifier.
+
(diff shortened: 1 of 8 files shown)


mkulke (approved):

attestation-agent/attester/src/csv/mod.rs
```diff
@@ -48,14 +43,17 @@ impl Attester for CsvAttester {
...
-        let cert_data = download_hskcek_from_kds(&report_signer.sn).await?;
+        // Fill in HSK_CEK with 0
+        let cert_size =
+            std::mem::size_of::<ca::Certificate>() + std::mem::size_of::<csv::Certificate>();
+        let cert_data = vec![0; cert_size];
         let mut cert_data = &cert_data[..];
-        let hsk = ca::Certificate::decode(&mut cert_data, ()).unwrap();
-        let cek = csv::Certificate::decode(&mut cert_data, ()).unwrap();
+        let hsk = ca::Certificate::decode(&mut cert_data, ())?;
+        let cek = csv::Certificate::decode(&mut cert_data, ())?;
```
fitzthum: Maybe it would be better to make these members of the CertChain struct optional rather than create these empty certs.

Also, it might make sense to allow the attester to download the certs optionally. For some users I actually think the VM will be more access to the inernet than Trustee will.

fitzthum (commented):
Looks good. One note about the CSV thing.

attestation-agent/attester/src/csv/mod.rs
Xynnn007: Makes sense. Let me mark this is Optional and control the behavior due to env.

PR #1047, opened 2025-07-04
 image-rs: Support registry config setting via initdata

### LINKED ISSUE #1034: [image-rs]: add support for etc-hosts injection into guest via initdada/cdh.toml for local registry image guest pulling
Currently, in our Confidential Containers (CoCo) setup, Guests face challenges pulling images from the cluster's local registry (e.g., Harbor) when using custom domain names. This issue prevents basic image fetching, which is a fundamental functional requirement for CoCo Pods.

- Problem Description
Our image registry uses a custom domain, such as `trust.hub`, as seen in image tags like `trust.hub:9443/registry.k8s.io/ubuntu:24.04`. In the CoCo scenario, images are pulled directly by the Guest VM (guest pull). However, the Guest cannot resolve this `trust.hub` domain through public DNS, leading to failed image pulls.

- Proposed Solution
To resolve this, we have tested it with injecting the Host's `/etc/hosts` file into the Guest. This file will contain the necessary custom domain mapping, specifically an entry like `172.20.1.2 trust.hub`. This allows the Guest to correctly resolve the` trust.hub` domain to the local(host) registry's IP address, enabling successful image pulls.

In CoCo, one possible solution for dynamicly injecting `hosts` into CVM is `InitData`. Within `InitData`, we can inject related configs(hosts and ca certifications) into guest and make it access local registry successfully.

Apokleos:
image-rs: Support registry config setting via initdata
    
    Currently, our registry configuration, which allows defining blocking,
    mirroring, and remapping rules, relies on a configuration file. This
    file can either be built into the rootFS or retrieved via KBS. However,
    this approach has proven inflexible and unsuitable for several scenarios.
    
    To address these limitations, directly embedding registry configuration
    information within ImageConfig via a new field(registry_config) is introduced.
    This allows us to leverage initdata to pass the registry configuration.
    
    This commit introduces the necessary logic to process the registry_config field.
    If registry_config is present in the ImageConfig, it will be used to configure
    the registry. Crucially, if registry_config is absent, it will gracefully fall
    back to the existing registry_configuration_uri logic, ensuring backward compatibility
    and a smooth transition.
    
    fixes #1034
    
    Signed-off-by: alex.lyn <alex.lyn@antgroup.com>


image-rs/src/image.rs
```diff
@@ -274,15 +274,22 @@ impl ImageClient {
...
+            match &self.registry_handler {
+                Some(handler) => handler.process(reference).map_err(|source| {
+                    PullImageError::IllegalRegistryConfigurationFormat { source }
+                })?,
+                None => vec![ImagePullTask {
+                    image_reference: reference,
+                    use_http: false,
+                    task_type: TaskType::Origininal,
+                }],
+            }
```
Xynnn007: The registry handler is initialized when the ImageClient is being created. You might not need to change this file, but add some logic near https://github.com/confidential-containers/guest-components/blob/main/image-rs/src/builder.rs#L149-L158 . 

- If `registry_config` is given in the config, use this config and skip `registry_configuration_uri` field
- If `registry_config` is not given, use `registry_configuration_uri`

Xynnn007 (commented):
Nice feature @Apokleos . I think it is a good way to go. Later we can even add other things like image policy content to the config. This could help cases where no kbs is given.

image-rs/src/config.rs
```diff
@@ -126,6 +127,14 @@ pub struct ImageConfig {
     #[serde(default = "Option::default")]
     pub registry_configuration_uri: Option<String>,
 
+    /// Registry configuration supports define registry blocking, mirroring,
```
fitzthum: nit: remove the word `define`

Apokleos:
Thx for your suggestion @fitzthum @Xynnn007 I have updated the codes, PTAL.

image-rs/src/registry/mod.rs
```diff
@@ -113,6 +113,11 @@ enum MatchResult {
 }
 
 impl RegistryHandler {
+    pub fn new(mut config: Config) -> Result<Self> {
+        config.validate_and_tidy().unwrap();
```
Xynnn007: let's avoid `unwrap()`s here, but `?`

image-rs/src/config.rs
```diff
@@ -126,6 +127,14 @@ pub struct ImageConfig {
     #[serde(default = "Option::default")]
     pub registry_configuration_uri: Option<String>,
 
+    /// Registry configuration supports registry blocking, mirroring and remapping rules. 
+    /// This field points to a registry configuration file, which can either be stored locally
+    /// in the rootfs or retrieved from initdata.
+    ///
+    /// This value defaults to `None`.
+    #[serde(default = "Option::default")]
+    pub registry_config: Option<RegistryConfig>,
```
Xynnn007, line 136: It would do great help to add some unit tests for parsing the config. Could you add some near [here](https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/hub/src/config.rs#L196-L203)?

image-rs/src/builder.rs
```diff
@@ -146,15 +146,23 @@ impl ClientBuilder {
...
 
-        let registry_handler = match &self.config.registry_configuration_uri {
-            Some(uri) => {
-                info!("getting registry configuration from {uri} ...");
-                let registry_configuration = resource_provider.get_resource(uri).await?;
-                let registry_handler = RegistryHandler::from_vec(registry_configuration)
-                    .map_err(|source| BuilderError::InvalidRegistryConfiguration { source })?;
-                Some(registry_handler)
+        let registry_handler = if let Some(config) = &self.config.registry_config {
+            info!("getting registry configuration plaintext ...");
```
fitzthum: This is slightly misleading since all of the options for the registry config are "plaintext." Better to say something like "using registry configuration from CDH config file"

fitzthum (approved):
One nit but otherwise LGTM. Maybe we can sneak this into the release.

image-rs/src/config.rs
Xynnn007: Or, leave a todo comment here. I can belp with this.

Apokleos: thx Ding, tests added

fitzthum:
Installation issue with the test. I think a rebase might help. @Apokleos

PR #1050, opened 2025-07-10
Add TPM attester

bpradipt:
Also include TEE quote in sample attester to help with testing

Drive-by-changes: Fix evidence_getter and add addtional logging to help with debugging.

Heavily inspired from
- https://github.com/Xynnn007/guest-components/tree/main/attestation-agent/attester/src/tpm
- https://github.com/kinvolk/azure-cvm-tooling/tree/main/az-cvm-vtpm/src/vtpm


Xynnn007:
cc @jialez0

bpradipt:
Note: before marking it ready, I will remove the merge commits. Having draft helps us to get an idea of the non-x86 CI runs.
All of them fails due to missing TPM deps, but wondering why as tpm-attester feature is not used for those runs. Debugging.

alicefr:
@bpradipt which architectures can actually use a TPM? I doubt s390x for example

bpradipt:
> @bpradipt which architectures can actually use a TPM? s390x?
I think it's available for all archs. At least the emulator and the tpm-tss2 packages are available across all archs.
In this CI, we can restrict for x86_64 only.

attestation-agent/attester/src/tpm_utils.rs
```diff
@@ -0,0 +1,312 @@
...
+    };
+
+    let engine = base64::engine::general_purpose::STANDARD;
+
+    drop(context);
+
+    Ok(TpmQuote {
+        signature: engine.encode(signature.marshall()?),
+        message: engine.encode(attest.marshall()?),
+        pcrs: read_all_pcrs(pcr_algorithm)?,
```
alicefr: can we please drop this field?

attestation-agent/attester/src/bin/evidence_getter.rs
```diff
@@ -48,9 +53,7 @@ async fn main() {
         }
     }
 
-    let evidence = attester
-        .primary_evidence(report_data)
-        .await
-        .expect("get evidence failed");
+    let evidence = attester.get_evidence(vec![]).await?;
```
uril: get_evidence(report_data)

attestation-agent/attester/src/tpm_utils.rs
bpradipt: We'll need the PCRs in the quote

bpradipt:
This should be ready for review. I'll work on the verifier next.

attestation-agent/attester/Cargo.toml
```diff
@@ -21,7 +21,7 @@ cfg-if.workspace = true
 crypto.path = "../deps/crypto"
 hex.workspace = true
 iocuddle = { version = "0.1.1", optional = true }
-kbs-types.workspace = true
+kbs-types = { "git" = "https://github.com/virtee/kbs-types.git", rev = "e3cc706" }
```
mkulke: Do we want to wait for a kbs-types release?

attestation-agent/attester/src/tpm/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+    detect_tpm_device().is_some()
+}
+
+#[async_trait::async_trait]
+impl Attester for TpmAttester {
+    /// Get evidence for the TPM attester.
+    async fn get_evidence(&self, report_data: Vec<u8>) -> Result<TeeEvidence> {
+        let tpm_device = detect_tpm_device().ok_or_else(|| anyhow!("No TPM device found"))?;
+        info!("[TPM Attester] Using TPM device: {}", tpm_device);
+        std::env::set_var("TCTI", format!("device:{}", tpm_device));
```
mkulke: Can we avoid setting that env and use a field in the TpmAttester struct instead? (set_env is [marked unsafe](https://doc.rust-lang.org/std/env/fn.set_var.html) in more recent rust releases)

attestation-agent/Makefile
```diff
@@ -131,6 +131,7 @@ lint-s390x:
 lint-aarch64:
 	cd attestation-agent && cargo clippy --no-default-features --features openssl,rust-crypto,cca-attester,kbs,coco_as,bin,grpc,ttrpc
 
+
```
Xynnn007: This might be a typo or formatting?

attestation-agent/Makefile
```diff
@@ -140,6 +141,7 @@ uninstall:
 clean:
 	cargo clean
 
+
```
Xynnn007: Also for this.

attestation-agent/attester/src/tpm/mod.rs
bpradipt: Oh ok. Thanks for pointing it out.

attestation-agent/attester/Cargo.toml
```diff
@@ -46,6 +46,11 @@ hyper = { version = "0.14", features = ["full"], optional = true }
 hyper-tls = { version = "0.5", optional = true }
 tokio = { version = "1", features = ["full"], optional = true }
 tempfile = { workspace = true, optional = true }
+tss-esapi = { version = "7.5", optional = true }
+openssl = { workspace = true, optional = true }
+rsa = { version = "0.9.6", features = ["pkcs5", "sha2"]}
```
Xynnn007: Could we use the same version of the workspace?
```suggestion
rsa = { workspace = true, features = ["pkcs5", "sha2"]}
```

attestation-agent/attester/Cargo.toml
```diff
@@ -46,6 +46,11 @@ hyper = { version = "0.14", features = ["full"], optional = true }
...
+num-traits = { version = "0.2" }
+env_logger = "0.11"
```
Xynnn007, line 49: Also for this. Can we use the same version of workspace?

```suggestion
env_logger.workspace = true
```

attestation-agent/attester/src/bin/evidence_getter.rs
```diff
@@ -24,12 +25,16 @@ enum Cli {
 }
 
 #[tokio::main]
-async fn main() {
+async fn main() -> anyhow::Result<()> {
```
Xynnn007: This is a simple binary, so I think it is fine to panic inside

attestation-agent/attester/src/tpm/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+        let data = if report_data.len() > 64 {
```
Xynnn007: Probably we could use the same logic as [tdx](https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/attester/src/tdx/mod.rs#L123), s.t.
`.resize(64, 0)`

attestation-agent/attester/src/tpm/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+use anyhow::*;
+use base64::Engine;
+use log::info;
+use serde::{Deserialize, Serialize};
+use tss_esapi::traits::Marshall;
+
+/// Evidence structure for the TPM Attester.
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Evidence {
+    pub svn: String,
```
Xynnn007: What does this `svn` mean?

attestation-agent/attester/src/tpm/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+        std::env::set_var("TCTI", format!("device:{}", tpm_device));
+        extend_pcr(event_digest, register_index).map_err(|e| anyhow!("Failed to extend PCR: {e}"))
+    }
+
+    /// Bind init data for the TPM attester (extends PCR 8).
+    async fn bind_init_data(&self, init_data_digest: &[u8]) -> Result<InitDataResult> {
+        let tpm_device = detect_tpm_device().ok_or_else(|| anyhow!("No TPM device found"))?;
+        info!("[TPM Attester] Using TPM device: {}", tpm_device);
+        std::env::set_var("TCTI", format!("device:{}", tpm_device));
+        extend_pcr(init_data_digest.to_vec(), 8)
```
Xynnn007: let's declare `8` as a const variable in the head of the module?

Xynnn007 (commented):
Thanks @bpradipt Some initial comments

mythi:
> Drive-by-changes: Fix evidence_getter 

perhaps get #1016 merged as it brings also adds the device evidence prints (along with the attester fixes)

attestation-agent/attester/src/tpm/mod.rs
bpradipt: Oops.. I'll remove it. This is a left over from the earlier version of using the sample device

attestation-agent/attester/src/bin/evidence_getter.rs
bpradipt: Ok.

attestation-agent/attester/src/tpm_utils.rs
```diff
@@ -0,0 +1,316 @@
...
+/// 2. If /dev/tpm0 exists, use that.
+/// 3. If /dev/tpm1 exists, use that.
+/// 4. Otherwise, return None.
+pub fn detect_tpm_device() -> Option<String> {
+    if let Result::Ok(dev) = std::env::var("TPM_DEVICE") {
+        if std::path::Path::new(&dev).exists() {
+            log::info!("TPM device detected from TPM_DEVICE env: {}", dev);
+            return Some(dev);
+        } else {
+            log::warn!("TPM_DEVICE env set to '{}', but device does not exist", dev);
```
mkulke: I would make that an error. if TPM_DEVICE it set, a user probably doesn't want a graceful fallback to another device.

attestation-agent/attester/src/tpm_utils.rs
```diff
@@ -0,0 +1,316 @@
...
+
+/// Detect the TPM device to use.
+///
+/// Priority:
+/// 1. If the TPM_DEVICE environment variable is set, use that.
+/// 2. If /dev/tpm0 exists, use that.
+/// 3. If /dev/tpm1 exists, use that.
+/// 4. Otherwise, return None.
+pub fn detect_tpm_device() -> Option<String> {
+    if let Result::Ok(dev) = std::env::var("TPM_DEVICE") {
```
mkulke: Maybe scope the env to AA: `AA_TPM_DEVICE`?

attestation-agent/attester/src/bin/evidence_getter.rs
bpradipt: @Xynnn007 one reason to not panic was to make it easier to figure out what's wrong for the user, instead of seeing a stack trace.

Xynnn007: Ok. tbo I do not have strong preference here, so I am ok with your idea!

attestation-agent/attester/src/tpm_utils.rs
bpradipt: Makes sense. I'll change it in the next iteration

attestation-agent/attester/src/tpm_utils.rs
bpradipt: ok

attestation-agent/attester/Cargo.toml
```diff
@@ -83,6 +89,8 @@ csv-attester = ["hygon", "codicon", "hyper", "hyper-tls", "tokio"]
 hygon-dcu-attester = ["hygon"]
 hygon = ["csv-rs"]
 cca-attester = ["tsm-report"]
-se-attester = ["pv"]
+se-attester  = ["pv"]
```
Xynnn007: I think this line is not related to the PR

attestation-agent/attester/src/lib.rs
```diff
@@ -40,6 +40,10 @@ pub mod tsm_report;
 #[cfg(feature = "se-attester")]
 pub mod se;
 
+#[cfg(feature = "tpm-attester")]
+pub mod tpm;
+pub mod tpm_utils;
```
Xynnn007: Let's move `tpm_utils` under `tpm` sub directory.

Xynnn007 (commented):
One thing left about the endorsement verification chain. Now the report only have ak public inside. The endorsement of the AK (EK) annd also endorsement of EK has not been added to the report. Thus a middleman can forge an AK. imo We'd better explicitly claim this risk in code and an issue

bpradipt:
> One thing left about the endorsement verification chain. Now the report only have ak public inside. The endorsement of the AK (EK) annd also endorsement of EK has not been added to the report. Thus a middleman can forge an AK. imo We'd better explicitly claim this risk in code and an issue

@Xynnn007 should I add the EK pub cert to the evidence ? or you have something else in mind ?

attestation-agent/attester/Cargo.toml
bpradipt: I'll remove it. I guess this got auto formatted via IDE.

bpradipt:
> > It still mentions a [guide](https://tpm2-software.github.io/tpm2-tss/getting-started/2019/12/18/Remote-Attestation.html) (not sure it is helpful). So what I want, abstractly, is a reasonable and complete solution to show the trust chain from trust anchor (e.g. a pre-configured CA cert on trustee side) to the quote. If it is difficult to implement in this single PR, we should record this and highlight that the endorsement chain is not complete with risks of MITM attack. and we can perfect it later
> 
> thx, I didn't see this guide before. I think it's very valuable and applicable for our situation. Generally, in TPM-land there is strong emphasis on using TPM policies (~ combinations of PCR values) that allow you to perform certain actions, like signing a quote or decrypting something. This is a form of local attestation (e.g. to unseal a disk encryption key) that we might be able to ignore for our purpose.
> 
> We can attach both AKpub and EKpub to the evidence and then we can verify that the EK has been signed by some vendor/root CA. But we also need to bind the AK to the EK and that is not trivial I think, since the EK is not a signing key. instead, how I think it works (on a high level)
> 
> * the verifier performs [make_credential](https://docs.rs/tss-esapi/latest/tss_esapi/struct.Context.html#method.make_credential) using the EKpub, the AKname (just a hash I think) a custom_secret, resulting in encrypted_secret S
> * the attester performs [activate_credential](https://docs.rs/tss-esapi/latest/tss_esapi/struct.Context.html#method.make_credential) which would produce custom_secret S'
> * the verifier would have to check S == S'
> 
> this would prove that the AK signing the quote sits in the same TPM as the EK. It will be interesting how we map this onto the KBS RCAR protocol.
> 

@fitzthum might have some ideas.

fitzthum:
Let's take a little step back first. At a high level we are talking about how we can validate that the quote we get from the TPM is genuine. This is kind of tricky. If the TPM were actually backed by an enclave, we could bind the AKpub to the hardware evidence of the enclave, like @mkulke mentions. This provides an important property, which is that the AKpub can't be changed in transit by a MITM. Without that guarantee, I'm not sure that providing the Akpub or the EKpub in the evidence gives us much. 

But wait, people do remote attestation with TPMs all the time. Can't we do what they do? I think the crux of remote TPM attestation is that there is some a priori agreement about the AKpub between the TPM and the remote party. A couple methods for doing this have been mentioned, such as sealing the AK to the EK. Either way, one important property is that the verifier knows what AKpub to expect when receiving a quote. One approach we could take would be to add a configuration parameter to the AS that allows the admin to specify the AKpub. This would assume that they have provisioned the TPM out-of-band.

We could also add support for provisioning the AK. In this case I think Trustee would need a list of allowed EKpubs. It would then generate a random AK and seal that to the EK. This would be provided somewhere in the KBS protocol (possibly as an additional field to the challenge). Ofc with vTPMs the EK can be dynamic. Provisioning the EK is probably out of scope here, but there has been some talk about doing that via Trustee when using the SVSM as a vTPM. In that case we could potentially tie all of these pieces together, although there are a lot of details to iron out.

tldr; I'm not sure there is any reason to have AKpub or EKpub in the evidence. As it stands, users don't really have any way to reason about the trustworthiness of the quote.

cc @cclaudio

jialez0:
I’d like to share some thoughts on this issue.

To address the trustworthiness of EK (Endorsement Key) and AK (Attestation Key) public key & certificates, one intuitive solution is to enhance Trustee with the ability to distribute EK and AK certificates. Since EK certificate distribution is often tied to the vendor in practice, we might prioritize enabling Trustee to distribute AK certificates.

To support AK certificate distribution, I believe the RCAR protocol is not an ideal carrier. For a TPM platform, the AK certificate needs to be persistently stored (typically in a designated NV storage area), and the credential provisioning usually only needs to happen once during its validity period. In contrast, the RCAR protocol is designed to run dynamically on every attestation event. Instead, I propose adding a new TPM CA plugin to KBS that provides a dedicated interface for this purpose.

Specifically, I suggest implementing AK certificate distribution in Trustee as follows:

1. Add a `tpm_ca` plugin to KBS with two APIs:
   - `/ak-credential`: The request must include the EK certificate, AK public key, and AK name. The response is the "challenge" value generated by calling `make_credential` with the provided inputs.
   - `/ak-cert`: The request must include the "credential" value obtained by running `activate_credential` using the challenge from the previous step. The response is an AK certificate signed by KBS.

2. The `tpm_ca` plugin should support configuration of an allowed list of trusted EK certificates. The `/ak-credential` endpoint must reject requests from TPM platforms whose EK certificates are not on this allowlist.

With this plugin in place, during platform initialization, the client (which could be integrated into the Attestation Agent or be a separate component) can generate a persistent AK key pair, then interact with the KBS `tpm_ca` plugin to obtain a signed AK certificate, which is then persistently stored in NV memory. For the remainder of the AK certificate’s validity period, remote attestation evidence only needs to include the AK certificate, Quote, and PCRs. Trustee can verify the authenticity of the evidence by validating the AK certificate’s signature.

For TPM platforms that have already been configured with AK certificates via other user-trusted mechanisms, we should also add a configurable trust anchor list of AK root certificates in Trustee’s TPM verifier—so these cases are supported as well. (By default, Trustee would trust the root certificate of its own `tpm_ca`.)

Once this plugin is implemented, supporting EK certificate distribution in the future would simply require adding a similar endpoint, such as `/ek-credential`.

By implementing this design, Trustee can fully cover the TPM trust chain from a technical capability standpoint. I believe this is essential—otherwise, Trustee’s support for TPM may appear to be a security solution with noticeable gaps.

cc @bpradipt @Xynnn007 @fitzthum @mkulke

mythi:
> structured proposal to strengthen the trust chain for TPM-based remote attestation in Trustee

would be better to have RFC issues for these.

mkulke:
@jialez0 those are some interesting points, but maybe beyond the scope of a simple TPM attester (at least in the initial revision). In the current code the AK is created ad-hoc and ephemeral, which I think is OK in principle.

my understanding was that for most physical TPM devices EKs are baked into the silicon and rooted to a vendor CA. to the user, a vTPM issued by a cloud provider, looks essentially the same. the HW root of trust is replaced by trust in the CSP's infra and the EK is rooted in a CSP CA.

I think elements of your "part 2" ideas still apply to a simple TPM attester, though. I also fear that we cannot somehow shoehorn the TPM remote attestation ceremony into our KBS RCAR protocol:

Regarding the ephemeral AK: can we assume we can create as many AKs as we want? Are there race conditions that we need to consider (get_evidence() being called concurrently).

For privacy reasons the AK is not chained to a CA via a simple PKI infrastructure. (AK used be called Attestation Identity Key, I think). This is a nuisance for verification, since the EK <=> AK link needs to be established via a secret+AKname(+challenge) that the verifier encrypts using EKpub, which is then decrypted by the TPM's EK and sent back to the verifier to assert. 

This is essentially a new request-challenge-response protocol that we need to follow to establish trust in an AK. Assuming we have established this binding $somehow and Trustee keeps a record of a trustworthy AK, the RCAR procedure would be straightforward (assert that quote has been signed by AK).

I'm not sure whether we could multiplex this second, prior-to-RCAR exchange between TPM attester into RCAR, so a trustee plugin/extension might be required, but I don't think it has to be very complicated and maybe the attester only needs to issue it once and then the AK is cached as trustworthy by Trustee. The added exchange amounts to this

```
POST /auth_ak(EKpub, AKpub, AKname)
response (w/ http session) => (cred_blob, enc_secret, nonce) 
```

and then the decrypted `secret` would need to be sent back as part of the evidence to establish trust in the AK. maybe we can use the ` "extra-params": {}` field in the `/kbs/v0/auth` [Request](https://github.com/confidential-containers/trustee/blob/main/kbs/docs/kbs_attestation_protocol.md#request) struct and similarly in the [Challenge](https://github.com/confidential-containers/trustee/blob/main/kbs/docs/kbs_attestation_protocol.md#challenge) response?

fitzthum:
Well let's start with the first option, which doesn't seem to require any changes to this PR. On the verifier side we can add a mechanism to check the AKpub. I share some of @mkulke's concerns about the second part, but happy to see an RFC/code with some of the specifics.

Xynnn007:
Agreed to have part 1, s.t. explicitly setting AK endorsement regardless how it comes. This would resolve what we are facing now.

And we can discuss more about EK part in a separate thread.

bpradipt:
@fitzthum @mkulke @Xynnn007 @jialez0 Updated the PR to align with part-1, ie using pre-provisioned AK for signing the quote.

attestation-agent/attester/src/tpm/utils.rs
```diff
@@ -0,0 +1,293 @@
...
+/// Detect the TPM device to use.
+/// Priority: 1. AA_TPM_DEVICE env var, 2. /dev/tpm[0..2]
+pub fn detect_tpm_device() -> Option<String> {
+    // Check environment variable first
+    if let Ok(dev) = env::var("AA_TPM_DEVICE") {
+        return if std::path::Path::new(&dev).exists() {
+            log::info!("TPM device detected from AA_TPM_DEVICE env: {}", dev);
+            Some(dev)
+        } else {
+            log::warn!(
```
mkulke: nit: I would still prefer fail with an error if `AA_TPM_DEVICE` is set to a specific device, which doesn't exist, instead of falling back to another device

mkulke (approved):

attestation-agent/attester/src/tpm/utils.rs
bpradipt: Yeah my bad. You had mentioned this earlier as well. Fixed it

attestation-agent/attester/Cargo.toml
```diff
@@ -46,6 +46,15 @@ hyper = { version = "0.14", features = ["full"], optional = true }
...
 tokio = { workspace = true, features = ["rt", "macros", "fs", "io-util"] }
 tempfile = { workspace = true, optional = true }
+tss-esapi = { version = "7.5", optional = true }
+openssl = { workspace = true, optional = true }
+rsa = { workspace = true, features = ["pkcs5", "sha2"]}
+num-traits = { version = "0.2" }
+env_logger = "0.11"
+picky-asn1 = "0.10.1"
+picky-asn1-x509 = { version = "0.14.2", optional = true }
+picky-asn1-der = { version = "0.5.2", optional = true }
```
Xynnn007: Can we organize the dependencies in alphabet order?

attestation-agent/attester/Cargo.toml
```diff
@@ -46,6 +46,15 @@ hyper = { version = "0.14", features = ["full"], optional = true }
 hyper-tls = { version = "0.5", optional = true }
 tokio = { workspace = true, features = ["rt", "macros", "fs", "io-util"] }
 tempfile = { workspace = true, optional = true }
+tss-esapi = { version = "7.5", optional = true }
+openssl = { workspace = true, optional = true }
+rsa = { workspace = true, features = ["pkcs5", "sha2"]}
+num-traits = { version = "0.2" }
+env_logger = "0.11"
+picky-asn1 = "0.10.1"
```
Xynnn007: Sry I did not find anywhere `picky-asn1`, `rsa`, `openssl` is used in rust code?

attestation-agent/attester/src/tpm/utils.rs
```diff
@@ -0,0 +1,293 @@
...
+    let ak_pk = SubjectPublicKeyInfo::new_rsa_key(modulus, public_exponent);
+    let ak_pk_bytes = picky_asn1_der::to_vec(&ak_pk)?;
+    Ok(ak_pk_bytes)
+}
+
+/// Detect the TPM device to use.
+/// Priority: 1. AA_TPM_DEVICE env var, 2. /dev/tpm[0..2]
+pub fn detect_tpm_device() -> Option<String> {
+    // Check environment variable first
+    if let Ok(dev) = env::var("AA_TPM_DEVICE") {
```
Xynnn007: Let's define this as a const and with some notes.

attestation-agent/attester/src/tpm/utils.rs
```diff
@@ -0,0 +1,293 @@
...
+        return if std::path::Path::new(&dev).exists() {
```
Xynnn007: There is a `return ... if {} else { return}` structure in code. Could we use
```rust
match std::path::Path::new(&dev).exists() {
  true => ...
  false => ...
}
```

attestation-agent/attester/src/tpm/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+
+#[async_trait::async_trait]
+impl Attester for TpmAttester {
+    /// Get evidence for the TPM attester.
+    async fn get_evidence(&self, mut report_data: Vec<u8>) -> Result<TeeEvidence> {
+        // Ensure report_data is exactly 64 bytes,
+        // truncating if longer or padding with zeros if shorter
+        report_data.resize(TPM_REPORT_DATA_SIZE, 0);
+        let ak_public_bytes = read_ak_public_key(&self.tpm_device, self.ak_handle_raw)?;
+        let tpm_quote = get_quote(&self.tpm_device, self.ak_handle_raw, &report_data, "SHA256")?;
```
Xynnn007: Could we define `SHA256` as a const with some notes?

attestation-agent/attester/src/tpm/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+
+        let evidence = Evidence {
+            report_data: base64::engine::general_purpose::STANDARD.encode(&report_data),
```
Xynnn007: I have a question that whether this `report_data` field is duplicated. As the verifier side we will only check the binding from the content inside tpm_quote, and the integrity of `report_data` here is not protected.

attestation-agent/attester/src/tpm/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+        // Use the stored tpm_device path.
+        extend_pcr(&self.tpm_device, init_data_digest.to_vec(), PCR_SLOT_8)
+            .map_err(|e| anyhow!("Failed to extend PCR for init data: {e}"))?;
+        Ok(InitDataResult::Ok)
+    }
+
+    /// Get runtime measurement (PCR value) for the given index.
+    async fn get_runtime_measurement(&self, index: u64) -> Result<Vec<u8>> {
+        // Use the stored tpm_device path.
+        let pcrs = read_all_pcrs(&self.tpm_device, "SHA256")?;
```
Xynnn007: Same as before, we can define `SHA256` as a const

attestation-agent/attester/Cargo.toml
bpradipt: picky-asn1 is used in utils.

attestation-agent/attester/src/tpm/mod.rs
bpradipt: I kept it for consistency sake as the other attesters uses report_data specifically. If you prefer to remove it let me know

bpradipt: @Xynnn007 I have incorporated all your feedback, except this one. Let me know what you prefer and I'll make necessary changes.
Want to get this in quickly ;)

Xynnn007: Oh yes, I prefer to remove it. An example is [TDX](https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/attester/src/tdx/mod.rs#L66-L74) it does not have this field, but the field is extracted from the quote (for TPM, it should be Evidence.[extra_data](https://docs.rs/tss-esapi/7.6.0/tss_esapi/structures/struct.Attest.html#method.extra_data)) by verifier.

bpradipt: @Xynnn007 updated the PR. PTAL.

Xynnn007 (approved):
lgtm, thanks @bpradipt

fitzthum:
I missed reviewing this one, but it looks ready to merge. We can fix anything that comes up while working on the verifier later.

PR #1052, opened 2025-07-14
CDH: Update secure mount for BlockDevice 

Xynnn007:
This patch mainly reconstructs the block device module of secure mount. It includes the following contents
1. Rust-ify luks2 related operations
2. Provided several typical temporary/persistent volume application scenarios
3. Optimized the code structure
4. Optimized the protobuf definition

Please check the specific commit message

### DIFF
--- a/.github/workflows/image_rs_build.yml
+++ b/.github/workflows/image_rs_build.yml
@@ -67,6 +67,11 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y libdevmapper-dev
         if: matrix.instance == 'ubuntu-24.04'
+      
+      - name: Install cryptsetup
+        run: |
+          sudo apt-get install -y libcryptsetup-dev pkg-config
+        if: matrix.instance == 'ubuntu-24.04'  || matrix.instance == 'ubuntu-24.04-arm'
 
       - name: Install cross-compliation support dependencies
         run: |
--- a/image-rs/scripts/build_confidential_data_hub.sh
+++ b/image-rs/scripts/build_confidential_data_hub.sh
@@ -19,8 +19,8 @@ CDH_DIR=$SCRIPT_DIR/../../confidential-data-hub
 
 pushd $CDH_DIR
 
-make RESOURCE_PROVIDER=none KMS_PROVIDER=none RPC="${RPC}"
-make DESTDIR="${SCRIPT_DIR}/${RPC}" install
+make RESOURCE_PROVIDER=none KMS_PROVIDER=none RPC="${RPC}" LIBC=gnu
+make DESTDIR="${SCRIPT_DIR}/${RPC}" install LIBC=gnu
 
 file "${SCRIPT_DIR}/${RPC}/confidential-data-hub"
 popd


Xynnn007:
The underlying `libcryptsetup-rs` leverages FFI to a dynamic library of `cryptsetup` than a static lib. Thus it cannot be built by `musl`. So the way to resolve error would be make CDH dynamic linked by `gnu` than `musl`. This would also require us

1. install some extra runtime dependencies on the kata rootfs side

protos/protos/confidential-data-hub/api.proto
```diff
@@ -29,9 +29,7 @@ message SecureMountRequest {
     string mount_point = 4;
 }
 
-message SecureMountResponse {
-    string mount_path = 1;
-}
+message SecureMountResponse {}
```
fitzthum, line 32: I can maybe imagine cases where the exact mount point isn't known until mount time and the CDH would want to report this to the caller, but we can worry about that later I guess.

confidential-data-hub/hub/src/storage/drivers/luks2.rs
```diff
@@ -0,0 +1,88 @@
...
+use libcryptsetup_rs::consts::vals::EncryptionFormat;
+use libcryptsetup_rs::{CryptInit, CryptParamsLuks2, CryptParamsLuks2Ref};
+use zeroize::Zeroizing;
+
+/// Algorithm of the integrity hash
+const HMAC_SHA256: &str = "hmac-sha256";
+
+const SECTOR_SIZE: u32 = 4096;
+
+pub fn encrypt_device(
```
fitzthum: It might be nice to do these functions as methods on a device struct. Not a requirement.

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
```diff
@@ -3,43 +3,98 @@
...
+    /// Format the block device with a filesystem type.
+    /// If not set, the block device will be mounted as a raw device.
+    #[serde(rename = "filesystemType")]
+    pub filesystem_type: Option<FsType>,
+
+    /// Indicates whether the block device is ephemeral.
+    /// If set to `true`, there will perform a filesystem format before mounting, if
+    /// any `filesystem_type` is specified.
+    /// If set to `false`, the block device will be mounted as is.
+    pub ephemeral: Option<String>,
```
fitzthum: I wonder if we should call this something more literal, like `format_filesystem`. Whether it is ephemeral storage actually depends on a few different things, such as if we format the fs, how we handle the encryption key (see above option), and how the storage is actually provided by the host. So it might be more clear to name this field more specifically. I don't have strong feelings about this though.

Also not sure this needs to be an option. We can probably just use a bool with a default value.

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
```diff
@@ -50,46 +105,178 @@ pub trait Interpreter {
...
+                let dev_path = format!("/dev/mapper/{}", devmapper_name);
+
+                self.cryptsetup_pairs
+                    .push((device_path.clone(), devmapper_name.clone()));
+                dev_path
+            }
+        };
+
+        // 4. if ephemeral is set, format the block device with the filesystem type,
+        // or direct make a symlink to the device path
```
fitzthum: Slightly confusing because the ephemeral check mentioned first here is down on line 223

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
```diff
@@ -50,46 +105,178 @@ pub trait Interpreter {
...
+        let Some(fs_type) = parameters.filesystem_type else {
+            symlink(&dev_path, mount_point).await.map_err(|source| {
+                BlockDeviceError::CreateSymlinkFailed {
+                    source,
+                    source_path: dev_path.to_string(),
+                    target_path: mount_point.to_string(),
+                }
+            })?;
+
+            self.symbol_files.push(mount_point.to_string());
```
fitzthum: I only see us push one location to this Vec. Do we need a vector? Isn't the link name always going to be the mount point?

Also, I would change the name of this field to be `symlink_names`, since symbol file tends to mean something else more generally.

confidential-data-hub/docs/use-cases/secure-mount-with-block-device.md
```diff
@@ -75,3 +79,116 @@ $ lsblk |grep "encrypted_disk"
...
   └─encrypted_disk_OEyEj   253:2    0 968.6M  0 crypt /mnt/test-path
 ```
+
+### Best Practices
+
+#### Ephemeral Storage
+
+Ephemeral storage is a storage that is not persistent.
+
+1. Mount a block device file, encrypt it using `luks2` with an ephemeral key, and mount the plaintext device to `/mnt/dev-path` as a file.
```
fitzthum: You might want to make it clear that these are two different options not two steps.

confidential-data-hub/docs/use-cases/secure-mount-with-block-device.md
```diff
@@ -75,3 +79,116 @@ $ lsblk |grep "encrypted_disk"
...
+    "flags": [],
+    "mount_point": "/mnt/directory-path"
+}
+```
+
+#### Persistent Storage
+
+1. Mount a `luks2` encrypted block device file and return the plaintext device path, mounting the plaintext to `/mnt/dev-path` as a device file. The decryption key is obtained from KBS.
+
+We can prepare an encrypted block device file before mounting it. Using `"passphare"` as passphare to encrypt the block device file with luks2.
```
fitzthum: nit: `passphare` -> `passphrase`

fitzthum (commented):
Nice PR. I made a few comments.

protos/protos/confidential-data-hub/api.proto
Xynnn007: I took this into consideration. An upper-level caller will often explicitly specify the mount point of the final target. This is easy for a file system directory, but for a block device, a different path may be generated (for example, a random device name under /dev/mapper under luks2). In actual implementation, it is highly recommended to create a link symbol to map the randomly generated device to a specified mount point. This helps to abstract it into a deterministic black box for use by upper-level callers (such as kata-agent)

confidential-data-hub/hub/src/storage/drivers/luks2.rs
Xynnn007: Good idea. Let me make `integrity` into a struct member and try to decrease the number of the parameters.

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
Xynnn007: oh. I have lost some more notes upon this. If `ephemeral` is true, the original device will also be encrypted which might cause data erasing and overwritting.

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
Xynnn007: let me fix the notes here.

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
Xynnn007: Yes, for future extension that potentially more symlink files would be generated when doing secure mount.

confidential-data-hub/docs/use-cases/secure-mount-with-block-device.md
Xynnn007: Oh. This is actually a case, neither two steps or two options.

Xynnn007:
btw, the most important issue is the musl build. https://github.com/confidential-containers/guest-components/pull/1052#issuecomment-3072418527

wdyt?

.github/workflows/cdh_basic.yml
```diff
@@ -54,10 +54,14 @@ jobs:
         run: |
           sudo apt-get update && sudo apt-get install -y protobuf-compiler
 
+      - name: Install cryptsetup
+        run: |
+          sudo apt-get update && sudo apt-get install -y libcryptsetup-dev pkg-config
```
mythi, line 60: no need to update again:
```suggestion
        run: sudo apt install -y libcryptsetup-dev pkg-config
```

(ditto for other places)

attestation-agent/deps/crypto/src/rand.rs
```diff
@@ -0,0 +1,7 @@
+use rand::Rng;
+
+pub fn random_key<const N: usize>() -> Vec<u8> {
```
mythi: AFAUI this intents to be a generic helper but the name sounds specific to the use LUKS2 case

```suggestion
pub fn random_bytes<const N: usize>() -> Vec<u8> {
```

(or make it local?)

mythi (commented):
this needs re-thinking:
> `Binary file added BIN +20 MB confidential-data-hub/hub/test_files/luks2-disk`

some minor suggestions. will take another tomorrow

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
fitzthum: Yes, with ephemeral mounts there will be a new random key. Any existing data will not be readable.

I am just saying that we have more than one option that relate to ephemeral storage. This one and the encryption key parameter above. Thus, maybe this should be called something more specific rather than `ephemeral`.

Xynnn007: Let me re-design this with a more clear parameter.

.github/workflows/cdh_basic.yml
Xynnn007: Thanks, fixed.

attestation-agent/deps/crypto/src/rand.rs
Xynnn007: it's relative common. Let me rename the function. Thanks for the suggestion

Xynnn007:
@fitzthum I change the API with more explicit work mode (named `action`). This would help to understand the blackbox much better without being annoyed by the complex `Option`s. ptal

confidential-data-hub/docs/use-cases/secure-mount-with-block-device.md
Xynnn007: Fixed the document due to new API design. ptal

mythi:
> > Binary file added BIN +20 MB confidential-data-hub/hub/test_files/luks2-disk
> 
> This is for unit test to mount an encrypted volume. I tried 16 MB~ but luks2 does not support that small, 20MB is enough.
> 
> An option is to mark it as git-lfs?

I was thinking the option where the test generates it runtime. Something like file create + nix fallocate.

.github/workflows/cdh_basic.yml
```diff
@@ -54,10 +54,14 @@ jobs:
...
+
       - name: Build and install
         run: |
           mkdir -p ${HOME}/.local/bin
-          make RESOURCE_PROVIDER=kbs,sev && make install PREFIX=${HOME}/.local
+          LIBC=gnu make RESOURCE_PROVIDER=kbs,sev && LIBC=gnu make install PREFIX=${HOME}/.local
```
mkulke:
```suggestion
          LIBC=gnu make RESOURCE_PROVIDER=kbs,sev && LIBC=gnu make install PREFIX=${HOME}/.local
```

confidential-data-hub/hub/src/storage/drivers/luks2.rs
```diff
@@ -0,0 +1,161 @@
+// Copyright (c) 2024 Intel
```
bpradipt, line 1: nit: is this correct ?

Xynnn007: Ah yes, I copy-and-paste the original file from Intel a lot. let me add a new line of alibaba to it if needed

Xynnn007: Thanks, added

ChengyuZhu6 (approved):
LGTM

fitzthum (approved):
LGTM

seems like we are generally ok with ditching musl here, but need to work out the details

Xynnn007:
let me put a PR for installing dependencies to the rootfs too and merge both at the same time

confidential-data-hub/docs/use-cases/secure-mount-with-block-device.md
```diff
@@ -40,30 +42,31 @@ $ confidential-data-hub
...
-    - `encryptType`: The encryption type. Currently, only LUKS is supported.
-    - `encryptKey`: Encryption key. It can be a sealed secret or a resource uri. If not set, it means that the device is unencrypted and a random 4096-byte key will be generated to encrypt the device.
-    - `dataIntegrity`: Enables dm-integrity to protect data integrity. Note that enabling data integrity will reduce IO performance by more than 30%.
+    - `deviceId`: The device number, formatted as "MAJ:MIN". At least one of `device_id` or `device_path` must be set. If both are set, `device_id` will be used.
+    - `devicePath`: The path of the source device path. At least one of `device_id` or `device_path` must be set. If both are set, `device_id` will be used.
+    - `action`: **Required** High-level action the blockdevice module performs. Different actions requires different parameters. See [Actions](#actions) for more details.
+    - `encryptionType`: **Required**. The encryption type. Currently, only `luks2` is supported. Different encryption types require different parameters. See [Encryption Types](#encryption-types) for more details.
+
 
 5. Make a request to CDH
```
manuelh-dev, line 77: should this not be `ttrpc-cdh-tool`

confidential-data-hub/docs/use-cases/secure-mount-with-block-device.md
```diff
@@ -40,30 +42,31 @@ $ confidential-data-hub
...
-        "volume_type": "BlockDevice",
-        "options": {
-            "deviceId": "7:0",
-            "encryptType": "LUKS",
-            "dataIntegrity": "true"
-        },
-        "flags": [],
-        "mount_point": "/mnt/test-path"
-    }
+    "volume_type": "BlockDevice",
```
manuelh-dev: that should be block-device, otherwise you're getting `Failed to recognize the storage type: Matching variant not found`

confidential-data-hub/docs/use-cases/secure-mount-with-block-device.md
```diff
@@ -75,3 +78,150 @@ $ lsblk |grep "encrypted_disk"
 └─encrypted_disk_OEyEj_dif 253:1    0 968.6M  0 crypt
```
manuelh-dev: I see this:
root@2f681695313a:/# lsblk |grep "encrypted_disk"
root@2f681695313a:/# lsblk
NAME                                   MAJ:MIN RM  SIZE RO TYPE  MOUNTPOINTS
loop0                                    7:0    0  100M  0 loop
`-7af68a6a-a9f0-47c9-b8db-7527d3bcd715 253:0    0   84M  0 crypt /tmp/test-path

confidential-data-hub/docs/use-cases/secure-mount-with-block-device.md
Xynnn007: We now have both ttrpc version and grpc version CDH, thus have ttrpc-cdh-tool and grpc-cdh-tool. Do you have any suggestions to make it clearer?

confidential-data-hub/docs/use-cases/secure-mount-with-block-device.md
Xynnn007: Thanks, fixed.

confidential-data-hub/docs/use-cases/secure-mount-with-block-device.md
manuelh-dev: Maybe something like this?

5. Make a request to CDH using the client-tool
For instance, using the `ttrpc` version of the client tool, call:
```
$ ttrpc-cdh-tool secure-mount ...
...
```

confidential-data-hub/docs/use-cases/secure-mount-with-block-device.md
Xynnn007: Let me update the document. Thanks for this.

Xynnn007:
Hi @manuelh-dev Thanks for your comments! I have adapted your suggestions about the code. About the bug, I will take a look then.

confidential-data-hub/docs/use-cases/secure-mount-with-block-device.md
Xynnn007: Thanks, fixed!

manuelh-dev:
> Hi @manuelh-dev Thanks for your comments! I have adapted your suggestions about the code. About the bug, I will take a look then.

Great, thank you. @fitzthum and I were looking over this yesterday as well - maybe it makes sense to add a test to `confidential-data-hub/hub/src/storage/drivers/luks2.rs` with `integrity : true`.

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
```diff
@@ -1,95 +1,396 @@
...
+    };
+
+    Ok(Zeroizing::new(key))
+}
+
+/// High-level action the blockdevice module performs.
+#[derive(Serialize, Deserialize, PartialEq, Debug)]
+#[serde(rename_all = "camelCase")]
+#[serde(tag = "action")]
+pub enum Action {
```
fitzthum: Hmm. This abstraction looks kind of familiar. Was it part of one of the earlier secure storage PRs? Honestly I find it pretty confusing.

I think we are relatively close to introducing some more backends and I think we may need a different sort of refactor. Currently this module seems to represent anything that would consume a block device. If we add support for something like ZFS, that would fall into this category, but it would work quite differently from the luks + ext4 approach. 

Anyway, do you think we can keep this PR just to switching to the crate and do the refactoring separately when it's a bit more clear what we will need for other backends?

Xynnn007:
If we want this PR to get in, the CDH will be delivered with glibc than musl because of the static build issue for cryptsetup.

Thus we need to build CDH for different distros for the consumers:
1. [CoCo](https://github.com/kata-containers/kata-containers/blob/main/versions.yaml#L122-L163) uses ubuntu 24.04 
2. [Peerpods](https://github.com/confidential-containers/cloud-api-adaptor/tree/main/src/cloud-api-adaptor/podvm) might need more, like rhel, fedora and alibaba cloud? @stevenhorsman @mkulke 

Also, need to install cryptsetup lib in the rootfs for dynamic linking. Kata has some [prior work](https://github.com/kata-containers/kata-containers/commit/31c0ad207653c15998a3ef41316528231ec44312), while Idk if CAA does.

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
Xynnn007: Yes. Current enums of `Action` are named somehow coupled with `dm-crypt`. My design idea is that `volume-type` field varies for different inputs. Note that the secure mount target is always the plaintext. Based on the output format - device or system and whether the input is an encrypted device, `encryption_type` (type `BlockDeviceEncryptType`) can have four enums in any cases:
- input **encrypted** device and open to target plaintext **device** (hinting there will be decryption) - `DecryptMap`
- input **unencrypted** device and open to target plaintext **device** (hinting there will be encryption) - `EncryptMap`
- input **encrypted** device and mount to target plaintext **filesystem** dir (hinting there will be decryption) - `DecryptMount`
- input **unencrypted** device and mount to target plaintext **filesystem** dir (hinting there will be encryption) - `EncryptFormatMount`
Let's see an example json of secure storage
```json
{
    "volume_type": "block-device",
    "options": {
        "devicePath": "/tmp/test.img",
        "action": "decryptMount",
        "encryptionType": "luks2",
        "decryptionKey": "file:///tmp/encryption_key",
        "filesystemType": "ext4"
    },
    "flags": [],
    "mount_point": "/mnt/luks2ext4-cdh-test"
}
```

The `volume-type` field varies for different inputs. For ZFS, the input is still a block device, so we can still use `block-device`. Concretely, the ZFS shoud add a new enum for `encryption_type` (type `BlockDeviceEncryptType`), and option 1 and option 2 mode will not work.

Let me change the `Action` enum and do some refactoring

manuelh-dev:
> > Hi @manuelh-dev I have update the codebase and hopefully the integrity will work.
> 
> Thank you very much! I will give this a try again on the kata code base with `agent.secure_storage_integrity=true` on Monday and report back.

@Xynnn007 @fitzthum - the integrity option did not work for me on kata-containers, unfortunately. I still observe the prior issue:
```
CDH secure_mount failed: rpc status: Status { code: INTERNAL, message: "[CDH] [ERROR]: Secure Mount failed", details: [], special_fields: SpecialFields { unknown_fields: UnknownFields { fields: None }, cached_size: CachedSize { size: 0 } } }

Stack backtrace:
   0: <unknown>
   1: <unknown>
   2: <unknown>
   3: <unknown>
   4: <unknown>
   5: <unknown>
   6: <unknown>
   7: <unknown>
   8: <unknown>
   9: <unknown>
  10: <unknown>
  11: <unknown>
options: {"deviceId": "8:0", "encryptionType": "luks2", "filesystemType": "ext4", "action": "encryptFormatMount", "dataIntegrity": "true"}
```

The CDH logs say:
```
[2026-02-02T23:45:52Z INFO  ttrpc_cdh] [ttRPC] Confidential Data Hub starts to listen to request: unix:///run/confidential-containers/cdh.sock
[2026-02-02T23:45:53Z DEBUG ttrpc_cdh::ttrpc_server] [ttRPC CDH] get new secure mount request
[2026-02-02T23:45:53Z INFO  confidential_data_hub::hub] secure mount called
[2026-02-02T23:45:53Z DEBUG confidential_data_hub::storage::volume_type::blockdevice] generate a random key. All data on the device will be overwritten.
[2026-02-02T23:45:53Z WARN  confidential_data_hub::storage::volume_type::blockdevice] encrypting the device. This will wipe original data on the disk.
[2026-02-02T23:46:10Z DEBUG confidential_data_hub::storage::volume_type::blockdevice] No mapper name provided, generating a random one
[2026-02-02T23:46:10Z DEBUG confidential_data_hub::storage::volume_type::blockdevice] opening device: /dev/sda
[2026-02-02T23:46:10Z DEBUG confidential_data_hub::storage::drivers::luks2] activating device: cd5a7ed1-1b31-415e-b3cd-3cd68b88b06f
[2026-02-02T23:46:12Z DEBUG confidential_data_hub::storage::drivers::luks2] device activated: cd5a7ed1-1b31-415e-b3cd-3cd68b88b06f
[2026-02-02T23:46:12Z DEBUG confidential_data_hub::storage::volume_type::blockdevice] formatting device: /dev/mapper/cd5a7ed1-1b31-415e-b3cd-3cd68b88b06f
[2026-02-02T23:46:12Z ERROR ttrpc_cdh::ttrpc_server] [ttRPC CDH] Secure Mount :
    Secure Mount failed
    Caused by: Error when mounting Block device: Failed to make filesystem Ext4 of device /dev/mapper/cd5a7ed1-1b31-415e-b3cd-3cd68b88b06f: Failed to format device /dev/mapper/cd5a7ed1-1b31-415e-b3cd-3cd68b88b06f with options:
     FsFormatter {
        fs_type: Ext4,
        force: true,
        args: [],
    }
    stdout: Creating filesystem with 12680463 4k blocks and 3170304 inodes
        Filesystem UUID: 3876b1f7-e50e-493a-8701-949d42ffedfb
        Superblock backups stored on blocks:
                32768, 98304, 163840, 229376, 294912, 819200, 884736, 1605632, 2654208,
                4096000, 7962624, 11239424

        Allocating group tables:   0/387       done
        Writing inode tables:   0/387       done
        Creating journal (65536 blocks): done
        Writing superblocks and filesystem accounting information:   0/387
    stderr: mke2fs 1.47.0 (5-Feb-2023)
        Warning: could not erase sector 2: Input/output error
        Warning: could not read block 0: Input/output error
        Warning: could not erase sector 0: Input/output error
        mkfs.ext4: Input/output error while writing out and closing file system

    Caused by: Failed to make filesystem Ext4 of device /dev/mapper/cd5a7ed1-1b31-415e-b3cd-3cd68b88b06f: Failed to format device /dev/mapper/cd5a7ed1-1b31-415e-b3cd-3cd68b88b06f with options:
     FsFormatter {
        fs_type: Ext4,
        force: true,
        args: [],
    }
    stdout: Creating filesystem with 12680463 4k blocks and 3170304 inodes
        Filesystem UUID: 3876b1f7-e50e-493a-8701-949d42ffedfb
        Superblock backups stored on blocks:
                32768, 98304, 163840, 229376, 294912, 819200, 884736, 1605632, 2654208,
                4096000, 7962624, 11239424

        Allocating group tables:   0/387       done
        Writing inode tables:   0/387       done
        Creating journal (65536 blocks): done
        Writing superblocks and filesystem accounting information:   0/387
    stderr: mke2fs 1.47.0 (5-Feb-2023)
        Warning: could not erase sector 2: Input/output error
        Warning: could not read block 0: Input/output error
        Warning: could not erase sector 0: Input/output error
        mkfs.ext4: Input/output error while writing out and closing file system

    Caused by: Failed to format device /dev/mapper/cd5a7ed1-1b31-415e-b3cd-3cd68b88b06f with options:
     FsFormatter {
        fs_type: Ext4,
        force: true,
        args: [],
    }
    stdout: Creating filesystem with 12680463 4k blocks and 3170304 inodes
        Filesystem UUID: 3876b1f7-e50e-493a-8701-949d42ffedfb
        Superblock backups stored on blocks:
                32768, 98304, 163840, 229376, 294912, 819200, 884736, 1605632, 2654208,
                4096000, 7962624, 11239424

        Allocating group tables:   0/387       done
        Writing inode tables:   0/387       done
        Creating journal (65536 blocks): done
        Writing superblocks and filesystem accounting information:   0/387
    stderr: mke2fs 1.47.0 (5-Feb-2023)
        Warning: could not erase sector 2: Input/output error
        Warning: could not read block 0: Input/output error
        Warning: could not erase sector 0: Input/output error
        mkfs.ext4: Input/output error while writing out and closing file system
```

UVM kernel logs:
```
[   24.211759] trusted_key: device-mapper: crypt: dm-0: INTEGRITY AEAD ERROR, sector 0
[   24.212041] trusted_key: device-mapper: crypt: dm-0: INTEGRITY AEAD ERROR, sector 0
[   24.212180] Buffer I/O error on dev dm-1, logical block 0, async page read
[   24.212410] trusted_key: device-mapper: crypt: dm-0: INTEGRITY AEAD ERROR, sector 0
[   24.212551] Buffer I/O error on dev dm-1, logical block 0, async page read
[   24.212994] trusted_key: device-mapper: crypt: dm-0: INTEGRITY AEAD ERROR, sector 0
[   24.213249] trusted_key: device-mapper: crypt: dm-0: INTEGRITY AEAD ERROR, sector 0
[   24.213841] trusted_key: device-mapper: crypt: dm-0: INTEGRITY AEAD ERROR, sector 0
[   24.213981] Buffer I/O error on dev dm-1, logical block 0, async page read
[   24.214216] trusted_key: device-mapper: crypt: dm-0: INTEGRITY AEAD ERROR, sector 0
[   24.214344] Buffer I/O error on dev dm-1, logical block 0, async page read
[   24.214595] trusted_key: device-mapper: crypt: dm-0: INTEGRITY AEAD ERROR, sector 0
[   24.214852] trusted_key: device-mapper: crypt: dm-0: INTEGRITY AEAD ERROR, sector 0
[   24.467351] trusted_key: device-mapper: crypt: dm-0: INTEGRITY AEAD ERROR, sector 262144
```

I tried a few things on host and guest side, but I don't think I can adjust things properly outside of CDH. The point is that, looking at the workaround in https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/hub/src/storage/scripts/luks-encrypt-storage, the workaround happens after `luksOpen`, but before `mkfs.ext4` and before `mount`. So, this means, one would need to address this in CDH when `/dev/mapper/<uuid>` is present.

On the added test case in this PR: this works because we only exercise `cryptSetup open`, not the subsequent steps.

I am new to the confidential-containers code base, but I can also volunteer to take a look into this issue. Before e.g. bumping the CDH version in kata-containers and e.g. before using this for https://github.com/kata-containers/kata-containers/pull/10559/changes, we should ensure we can provide integrity-protection.

Side note: Please let me know if you want to chat. There is always slight chances I am doing something wrong in the integration of your work. On the bright side, this works well with your latest rebased branch against the kata code base without integrity.

Xynnn007:
@manuelh-dev Sorry that I did not do integrity test in the previous version. It's caused by the ext4 formatter will read the un-initialized blocks thus the checking will fail. I added the logic to fix this together with a unit test for both with integrity and without integrity

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
Xynnn007: Applied the refactoring. Two vertical options: `target_type` (filesystem, device) and `source_type` (encrypted, empty) are introduced to cover all cases.

For ZFS, you can implement only `target_type == filesystem` cases.

fitzthum: Thanks, this looks a lot better.

confidential-data-hub/docs/use-cases/secure-mount-with-block-device.md
```diff
@@ -11,10 +11,10 @@ This guide helps an user to use confidential data hub to secure mount inside TEE
 ### Create a loop device
 
 ```shell
-$ loop_file="/tmp/test.img"
```
manuelh-dev, line 14: I see the latest changes in the PR have updated the API. Would be good to catch up with this in the README

manuelh-dev:
> @manuelh-dev Sorry that I did not do integrity test in the previous version. It's caused by the ext4 formatter will read the un-initialized blocks thus the checking will fail. I added the logic to fix this together with a unit test for both with integrity and without integrity

This is looking better now. I have a prototyping environment with Kata in which I see:
```
root@e913fe272656:/# lsblk
NAME                                  MAJ:MIN RM  SIZE RO TYPE  MOUNTPOINTS
sda                                     8:0    0 48.8G  0 disk
`-76f84a55-225d-4b65-9514-16a1f5419e0e_dif
                                      253:0    0 48.4G  0 crypt
  `-76f84a55-225d-4b65-9514-16a1f5419e0e
                                      253:1    0 48.4G  0 crypt /run/kata-containers/image
vda                                   254:0    0  1.8G  1 disk
`-vda1                                254:1    0  1.7G  1 part  /
<le /dev/mapper/76f84a55-225d-4b65-9514-16a1f5419e0e
0 101443704 crypt capi:authenc(hmac(sha256),xts(aes))-plain :96:logon:cryptsetup:fd74f322-dafc-43ab-a062-308af25c67c9-d0 0 253:0 0 2 integrity:32:aead sector_size:4096
```
cryptsetup in that environment tells me:
```
/dev/mapper/76f84a55-225d-4b65-9514-16a1f5419e0e is active and is in use.
  type:    LUKS2
  cipher:  aes-xts-plain
  keysize: 768 bits
  key location: keyring
  integrity: hmac(sha256)
  integrity keysize: 256 bits
  integrity tag size: 32 bytes
  device:  /dev/sda
  sector size:  4096
  offset:  0 sectors
  size:    101443704 sectors
  mode:    read/write
```
So, this is looking good. In my environment though, I still see some concerning dmesg output, but as the output shows, creating dm-1 eventually succeeds:
```
[   19.532513] trusted_key: device-mapper: crypt: dm-0: INTEGRITY AEAD ERROR, sector 0
[   19.532775] trusted_key: device-mapper: crypt: dm-0: INTEGRITY AEAD ERROR, sector 0
[   19.532911] Buffer I/O error on dev dm-1, logical block 0, async page read
[   19.533129] trusted_key: device-mapper: crypt: dm-0: INTEGRITY AEAD ERROR, sector 0
[   19.533271] Buffer I/O error on dev dm-1, logical block 0, async page read
[   19.567837] trusted_key: device-mapper: crypt: dm-0: INTEGRITY AEAD ERROR, sector 8
[   19.927890] EXT4-fs (dm-1): mounted filesystem 7038ad8f-334e-488e-8951-59221b565db0 r/w with ordered data mode. Quota mode: disabled.
```

Thanks for revising this code. The errors we are seeing may (or may not) be related to the specific UVM environment in which I am testing  these changes.

Xynnn007:
@stevenhorsman Thanks for the updates!

> We are trying to move from fedora to ubuntu based images in peerpods mkosi builds, but still have alibaba cloud and RHEL pieces in the old packer flow.

For alibaba cloud, the base linux OS image still has some compatibility issue with the upstream thus we are now building the images manually. So it's ok for us to install the cryptsetup dynamic lib. The fix upon the compatibility issue needs some time. 



> We seem to be installing cryptsetup in both our packer and mkosi images if that helps?

I did a search with key word "cryptsetup" upon CAA repo, and found that the installation might be only for `s390x` arch. The code's [here](https://github.com/confidential-containers/cloud-api-adaptor/blob/186104711d607a4ab11d77b35610ac13305bb39a/src/cloud-api-adaptor/podvm/Makefile#L91-L100).

For x86-64, I did a test. The running process works well with `cryptsetup` install. But the building process relies on `cryptsetup-devel` which is not available on `ubi9` by default and we need a free source like the following
```shell
cat >/etc/yum.repos.d/rocky-crb.repo <<'EOF'
[rocky-baseos]
name=Rocky Linux 9 - BaseOS
baseurl=https://dl.rockylinux.org/pub/rocky/9/BaseOS/x86_64/os/
enabled=1
gpgcheck=0

[rocky-appstream]
name=Rocky Linux 9 - AppStream
baseurl=https://dl.rockylinux.org/pub/rocky/9/AppStream/x86_64/os/
enabled=1
gpgcheck=0

[rocky-crb]
name=Rocky Linux 9 - CRB
baseurl=https://dl.rockylinux.org/pub/rocky/9/CRB/x86_64/os/
enabled=1
gpgcheck=0
EOF
```

confidential-data-hub/docs/use-cases/secure-mount-with-block-device.md
Xynnn007: I have fixed the document. Thanks for the note!

PR #1054, opened 2025-07-15
ci: fix image-rs and ocicrypt-rs build dependency installation

mythi:
Make sure apt package db is up-to-date before install. Otherwise we might get 404's for some packages.


stevenhorsman (approved):
LGTM. Thanks!

fitzthum (approved):
LGTM

PR #1064, opened 2025-07-23
hygon_dcu attester: fix detect_platform

uril:
When detect_platform is called for attester hygon_dcu, it only checks /sys/devices/virtual/kfd/kfd/topology/nodes, but get_evidence tries to open /dev/mkfd and may fail with file-not-found.

Fix it by verifying /dev/mkfd exists in detect_platform().


fitzthum:
So this is going to cause a relatively catastrophic failure on any guest that does not have that dir?

Xynnn007 (approved):
Thanks. There is surely a reading upon `/dev/mkfd`. https://github.com/openanolis/csv-rs/blob/master/src/api/dcu/mod.rs#L56

PR #1074, opened 2025-08-01
attester: add eventlog mechanism for CSV

Xynnn007:
This patch mainly changes two things:

1. Original AAEL needs a configuration item in AA's launch config to specify the hash algorithm. Now AAEL is embedded inside CCEL, so the hash algorithm should be determined by OVMF. hw vendors usually selects a hash algorithm inside their OVMF builds (sha384 for tdx, sm3 for csv), thus we add a new `ccel_hash_algorithm` to `Attester` trait and get rid of hash algorithm setting logic in AA.
2. Add eventlog for CSV support.

The whole workflow has been test on csv machine. with EAR token like
```json
{
    "eat_profile": "tag:github.com,2024:confidential-containers/Trustee",
    "iat": 1754034551,
    "ear.verifier-id": {
        "developer": "https://confidentialcontainers.org",
        "build": "attestation-service 0.1.0"
    },
    "submods": {
        "cpu0": {
            "ear.status": "warning",
            "ear.trustworthiness-vector": {
                "configuration": 3,
                "executables": 33,
                "hardware": 3
            },
            "ear.appraisal-policy-id": "default",
            "ear.veraison.annotated-evidence": {
                "csv": {
                    "build": 2344,
...
{
                            "details": {
                                "data": {
                                    "content": "just for test",
                                    "domain": "coco-tenant",
                                    "operation": "test"
                                },
                                "string": "coco-tenant test just for test",
                                "unicode_name": "AAEL"
                            },
                            "digests": [
                                {
                                    "alg": "SM3",
                                    "digest": "11cd4f1fc33e79ac244e8b862a33eee46d1589e86f9384ad9d1568b894033b6e"
                                }
                            ],
                            "event": "TEVBQR4AAABjb2NvLXRlbmFudCB0ZXN0IGp1c3QgZm9yIHRlc3Q=",
                            "index": 4,
                            "type_name": "EV_EVENT_TAG"
                        }
                    ],
                    "user_pubkey_digest": "0000000000000000000000000000000000000000000000000000000000000000",
                    "version": "2",
                    "vm_id": "00000000000000000000000000000000",
                    "vm_version": "00000000000000000000000000000000"
                },
                "report_data": "00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
                "runtime_data_claims": null
            }
        }
    },
    "exp": 1754034851
}
```


Cargo.toml
```diff
@@ -38,7 +38,7 @@ hmac = "0.12.1"
 jwt-simple = { version = "0.12", default-features = false, features = [
     "pure-rust",
 ] }
-kbs-types = { "git" = "https://github.com/virtee/kbs-types.git", rev = "e3cc706" }
+kbs-types = { "git" = "https://github.com/Xynnn007/kbs-types.git", rev = "718a83a" }
```
Xynnn007: I will change this once https://github.com/virtee/kbs-types/pull/69 gets merged.

attestation-agent/attestation-agent/config.example.json
```diff
@@ -9,8 +9,7 @@
         }
     },
     "eventlog_config": {
-        "eventlog_algorithm": "sha384",
```
mythi, line 12: while I'm in favor of this in general, I think it goes against what we decided with initdata that the TEE native hash algorithm should not matter. maybe we need to revisit that.

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -211,6 +212,10 @@ impl Attester for TdxAttester {
             _ => 4,
         }
     }
+
+    fn ccel_hash_algorithm(&self) -> HashAlgorithm {
```
mythi, line 216: `let extend_data: [u8; 48] = pad(&event_digest);` in this file becomes redundant I think

attestation-agent/attester/src/csv/mod.rs
```diff
@@ -159,6 +160,10 @@ impl Attester for CsvAttester {
             _ => 4,
         }
     }
+
+    fn ccel_hash_algorithm(&self) -> HashAlgorithm {
+        HashAlgorithm::Sm3
```
mythi, line 174: > "This is defined by the platform's OVMF build."

nit: I believe this should just say it's defined by the platform. OVMF just starts the (BIOS) measurements but won't be able to define what the platform hash algo is.

attestation-agent/attestation-agent/config.example.json
Xynnn007: I think initdata is somehow different from the eventlog hash algorithm here. How to set initdata is fully decided by the VMM software and not enforced by hardware. But for eventlog hash, it is determined by hardware/ovmf I think and has less flexibility.

fitzthum (approved):
LGTM. kbs-types PR was just merged

attestation-agent/attester/src/tdx/mod.rs
Xynnn007: Yes. Let me delete this.

Xynnn007: Oh. I realized that the API is exposed and public. Thus we should still keep this, or other caller than AA crate will still need to check the input length explicitly.

attestation-agent/attester/src/csv/mod.rs
Xynnn007: Fixed.

attestation-agent/attestation-agent/config.example.json
mythi: It's just inconsistent user experience. With initdata the desire was to unify the reference values between TEEs. The cost is that CoCo init and runtime measurements withing a single platform are not unified. Even if the algorithm user selection is easier for initdata, it does not mean we should keep doing it.

Xynnn007: While this may seem intuitive, there are two different hash calculation methods behind the scenes.

For initdata, a simple sha256sum <initdata.toml> is sufficient for users. However, for eventlog, integrity calculation and register binding verification are automated. The specific entries in the policy are almost always compared to the actual contents of the eventlog, eliminating the need for manual hash calculation by the user. Therefore, this does not involve any user experience at all imo.

attestation-agent/attester/src/tdx/mod.rs
mythi: It's probably fine to keep it since it's used in another place in this file too. The API definition is a bit unclear since it talks about "TEE specific register". Therefore, doing a `len != 48` check instead of `pad()` does not sound an API break

Xynnn007: This will bring  TEE specific implementations to upper caller. Currently upper caller does not need to care about this.

mythi: But the question is: is changing user data without user knowing it better than doing some input validation to let user to check their inputs. how csv attester handles this?

Xynnn007: CSV will truncate/pad it in underlying crate. Probably a good way is to add more notes upon the trait function.

Xynnn007: patch added.

Xynnn007:
Hi @mythi are there any blockers before getting this merged?

mythi:
> Hi @mythi are there any blockers before getting this merged?

no blockers but I'm not an approver here. I still disagree with the initdata topic but that's not relevant for this PR :-)

PR #1075, opened 2025-08-01
Add RTMR extending (AAEL event) to ASR

Xynnn007:
Almost the same as https://github.com/confidential-containers/guest-components/pull/635/files

This patch adds ASR support for AAEL. There are two open questions:
1. Should the encoding format be GET + query (base64-encoded parameters) or POST + body with JSON parameters?
2. This API can call AA to make additions to the underlying AAEL, modifying the state of TEE remote attestation. Does this require an additional enabled feature, requiring a higher level of support? Or is the current `attestation` control sufficient?

cc @fitzthum @mkulke

### DIFF
--- a/api-server-rest/protos/attestation_agent.proto
+++ b/api-server-rest/protos/attestation_agent.proto
@@ -31,6 +31,7 @@ message ExtendRuntimeMeasurementRequest {
     string Content = 3;
 
     // Which PCR will be extended with the hash of this entry.
+    // If not specified, the PCR set by AA will be used.
     optional uint64 RegisterIndex = 4;
 }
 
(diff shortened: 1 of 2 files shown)


api-server-rest/protos/attestation_agent.proto
```diff
@@ -18,7 +18,26 @@ message GetTokenResponse {
...
+    // The domain to which this event entry belongs. This domain is used to distinguish the semantics of log entries in different contexts.
+    string Domain = 1;
+
+    // Concrete operation type that this event entry records.
+    string Operation = 2;
+
+    // Detailed content of the operation that this event entry records.
+    string Content = 3;
+
+    // Which PCR will be extended with the hash of this entry.
```
fitzthum, line 33: maybe add a note here about what will happen if this is not specified

fitzthum (commented):
I think this should probably be a post request since it changes the state of the AA.

It probably is not necessary to add an additional feature beyond enabling attestation in the ASR.

fitzthum (approved):
LGTM

mythi:
@Xynnn007 #635 can be closed now?

Xynnn007:
> @Xynnn007 #635 can be closed now?

yes

PR #1090, opened 2025-08-22
protos: bring all proto generation logic to a separate crate

Xynnn007:
Before this commit, we have ttrpc/grpc proto file generation logic in many different places, like kbs_protocol, attestation-agent, confidential-data-hub, ocicrypt-rs, images-rs and asr. This would lead to too many copies of a same file.

This commit deletes all the copies and just maintain one in a separate crate. All the codes that needs to use the ttrpc/grpc proto files can import the crate. This will help decrease the burden of maintainance.


mythi:
Looks like a nice improvement! Perhaps the CI workflows could be updated too? `protoc` is no longer needed for most workflows and the newly created crate is not built.

README.md
```diff
@@ -70,5 +70,12 @@ default except for only builtin `offline-fs-kbc`, you can build with `NO_RESOURC
 make build TEE_PLATFORM=$(TEE_PLATFORM) NO_RESOURCE_PROVIDER=true
 ```
 
+### Optional build parameters
+
+The ttRPC and gRPC protos can be updated via run
```
mythi, line 75: maybe not directly related to this PR but should we also make dependabot to ignore the proto gen related dependencies or add a comment when `make build-protos` should be run after those dependencies are updated.

Xynnn007: Good question. I have also thought about this problem. Protos related dependencies need to be updated, but I don't know if there is a way to let dependantbot automatically run the make protos command when updating dependencies such as tonic-build and ttrpc-gen. Thus I originally expected that every time we encounter such an updation, we need to put a PR manually.

mythi: I was also thinking fully manual flow and let dependabot know it does not need to worry about proto gen related dependencies. Anyway, I don't think we need to make those changes in this PR.

PR #1093, opened 2025-08-26
Add TPM as additional device and fix TPM report data size

iroykaufman:
This pull request introduces two changes:

1. Adds TPM as an additional Attestable Device: Allows a TPM to be configured as an additional device, rather than only as the primary TEE.
2. Fixes TPM Report Data Size: Corrects the report data (nonce) size to 32 bytes to align with the SHA256 digest size, as required by TPM 2.0 library standards and FIPS PUB 180-4.

### DIFF
--- a/attestation-agent/attester/src/lib.rs
+++ b/attestation-agent/attester/src/lib.rs
@@ -226,13 +226,13 @@ pub fn detect_attestable_devices() -> Vec<Tee> {
         additional_devices.push(Tee::HygonDcu);
     }
 
+    // This relies on the TPM check being last in detect_tee_type()
+    // so it only adds the TPM as an additional device if it is not the primary one.
     #[cfg(feature = "tpm-attester")]
-    if detect_tee_type() != Tee::Tpm {
-        
-        if tpm::detect_platform() {
-            additional_devices.push(Tee::Tpm);
-        }
+    if detect_tee_type() != Tee::Tpm && tpm::detect_platform() {
+        log::warn!("The TPM device was detected as an additional device, but please note that it is not bound to the TEE, so there may be security risks.");
+        additional_devices.push(Tee::Tpm);
     }
-    
+
     additional_devices
 }


attestation-agent/attester/src/lib.rs
```diff
@@ -211,5 +211,13 @@ pub fn detect_attestable_devices() -> Vec<Tee> {
         additional_devices.push(Tee::HygonDcu);
     }
 
+    #[cfg(feature = "tpm-attester")]
+    if detect_tee_type() != Tee::Tpm {
```
fitzthum: You may want to add a comment here. It looks like this will trigger only if TPM is enabled as well as some other cpu platform (besides the sample one) and it depends on the TPM check being last in `tee_type_check` which could change in the future. Maybe add a comment there too about how it should be last.

Keep in mind that there is currently no binding between the TPM and the CPU platform so whatever your working on here probably is not secure. I guess that is potentially ok, but there is some risk to adding features that people might not realize are experimental.

fitzthum (commented):
Seems fine. See minor comment

attestation-agent/attester/src/tpm/mod.rs
```diff
@@ -22,6 +22,7 @@ const TPM_HASH_ALGORITHM: &str = "SHA256";
 /// Evidence structure for the TPM Attester.
 #[derive(Serialize, Deserialize, Debug)]
 pub struct Evidence {
+    pub tpm_svn: String,
```
Xynnn007: Why do we need this svn? btw the commit message might be a typo with `snv`

attestation-agent/attester/src/tpm/mod.rs
```diff
@@ -75,7 +76,8 @@ impl Attester for TpmAttester {
         )?;
 
         let evidence = Evidence {
-            tpm_svn: "1".to_string(),
+            svn: "1".to_string(),
+            report_data: base64::engine::general_purpose::STANDARD.encode(&report_data),
```
Xynnn007: The report data is already included inside the TpmQuote and we have [verifier logic](https://github.com/confidential-containers/trustee/blob/main/deps/verifier/src/tpm/mod.rs#L206). What would be the reason to add it here?

Xynnn007 (commented):
Hi @iroykaufman I left some questions. Btw, if we want to add TPM as a device, we do not have any binding between TEE hardware and TPM as @fitzthum points out. We should explicitly record this somewhere in the doc to tell users "TPM (except Azure vTPM series) device is not bound to TEE endorsement thus please ensure that you trust the TPM". Would you please to put a PR, probably [here](https://github.com/confidential-containers/trustee/blob/main/attestation-service/README.md#attestation-service)?

attestation-agent/attester/src/tpm/mod.rs
iroykaufman: I used an old version of [PR#851](https://github.com/confidential-containers/trustee/pull/851)  in that version, the evidence parser was different and expected to get those two fields. I'll make the necessary changes and create a PR for the document.

dcmiddle:
> It would be great if it were bound to a TEE, but for now it's mostly for development purposes.

Can you say the future use case?

attestation-agent/attester/src/lib.rs
```diff
@@ -211,5 +211,12 @@ pub fn detect_attestable_devices() -> Vec<Tee> {
         additional_devices.push(Tee::HygonDcu);
     }
 
+    // This relies on the TPM check being last in detect_tee_type()
+    // so it only adds the TPM as an additional device if it is not the primary one.
+    #[cfg(feature = "tpm-attester")]
+    if detect_tee_type() != Tee::Tpm && tpm::detect_platform(){
+            additional_devices.push(Tee::Tpm);
```
Xynnn007: Agree with https://github.com/confidential-containers/guest-components/pull/1093#discussion_r2301562731 from @fitzthum . Could we add a `warn!("The TPM device was detected as an additional device, but please note that it is not bound to the TEE, so there may be security risks.")`

Xynnn007:
Please run `cargo fmt` and take rebase to push

PR #1112, opened 2025-09-23
gha: Remove offline_sev_kbc workflow

BbolroC:
As SEV support is deprecated and its test constantly fails, the CI WG have decided to remove the `offline_sev_kbc` workflow from CI.


mythi:
> Just let me know if you want to remove all the relevant codes and the workflow together. Thanks.

+1 for doing a PR that includes both. Then we can tag the relevant maintainers to approve before merging.

PR #1115, opened 2025-09-25
deps: bump kbs-types to 0.14.0

mkulke:
The version has been released and bumped in trustee. Since this includes a breaking change in how tee strings are serialized we also need to bump in guest-components.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -994,6 +994,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "cbor-codec"
+version = "0.7.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e083a023562b37c52837e850131a51b1154cceb9d149f41ee3d386737b140f46"
+dependencies = [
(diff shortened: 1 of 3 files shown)


fitzthum:
> the kbs protocol version did not change?

Interesting point. The root cause of this is https://github.com/confidential-containers/kbs-types/commit/9a3427784cd7cfd34caf33f963c157c776e3c43f. Whether this is a breaking change to the KBS Protocol is a gray area. It's not totally clear that the Tee variant names are part of the spec. We do list the "known HW-TEE platforms" [here](https://github.com/confidential-containers/trustee/blob/main/kbs/docs/kbs_attestation_protocol.md#request), but believe it or not we have the Azure platforms with hyphens. Also below that we link to the kbs-types code for the most recent versions.

Since this is a little fuzzy and bumping the KBS Protocol causes churn (affecting all users rather than just the Azure users), I lean towards not bumping the protocol here. That said, we may want to tweak KBS Types so that it also understands the old format.

Issue #1116, opened 2025-09-25
[guest-components] inconsistencies in built packages arch's across components


nmwael:
For 

`2e2d0ca816f779fc474ae7ce3499fd915a403a78`

agent has a x86_64 variant
https://github.com/confidential-containers/guest-components/pkgs/container/guest-components%2Fattestation-agent

api-server-rest has no x86_64 variant
https://github.com/confidential-containers/guest-components/pkgs/container/guest-components%2Fapi-server-rest

mythi:
@nmwael thanks for reporting this! At first it looked like a harmless glitch but it seems `publish-cdh-and-asr` post merge workflow is failing because it builds using rust 1.90 (!). That should not happen at all.

`publish-artifacts.yml` is buggy but I cannot see how `RUST_TARGET` gives us stable/1.90 override.

cc @BbolroC

PR #1121, opened 2025-10-01
workflows: publish artifacts: checkout code before toolchain install

mythi:
We use rust-toolchain.toml to pin toolchain version and actions-rust-lang/setup-rust-toolchain knows to set an override accordingly.

However, the repository must be checked out for rust-toolchain.toml to be available. Without it, the latest Rust stable is used.

Building CDH using rust stable fails currently so the publish to ORAS post-merge job fails to run.

Fixes: #1116

### DIFF
--- a/.github/workflows/publish-artifacts.yml
+++ b/.github/workflows/publish-artifacts.yml
@@ -131,6 +131,8 @@ jobs:
         version: 1.2.0
       if: matrix.arch != 's390x'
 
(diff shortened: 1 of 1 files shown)


fitzthum (approved):
LGTM. Good catch

mythi:
would be good to get this merged because any code merge fails to get published

mkulke (approved):

PR #1123, opened 2025-10-02
grpc-cdh: Return detailed error in get_resource

beraldoleal:
The get_resource endpoint currently returns only the top-level error message (e.g., "Get Resource failed") while logging the full error chain with root causes.

The detailed error is already computed using format_error!() but only used for logging. This changes the response to include the full chain, which would help with debugging without the need to see server logs.

### DIFF
--- a/confidential-data-hub/hub/src/bin/grpc_server/mod.rs
+++ b/confidential-data-hub/hub/src/bin/grpc_server/mod.rs
@@ -80,7 +80,7 @@ impl GetResourceService for Cdh {
             .map_err(|e| {
                 let detailed_error = format_error!(e);
                 error!("[gRPC CDH] Call CDH to get resource failed:\n{detailed_error}");
-                Status::internal(format!("[CDH] [ERROR]: {e}"))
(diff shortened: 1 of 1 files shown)


mkulke:
I don't think we have to be concerned with specific confidentiality aspects here, it's just a question of good old API design. You want to give the user errors that are informative/useful/actionable to without leaking internals of a service which would be potentially harmful to the service owner. 

E.g. "400. Error: can't write /opt/kbs/policy/bla: no space of left on device" is not actionable for the user, and leaks insights about the folder structure of the service host. an opaque 500 error would be sufficient it that case. It would hopefully be picked up by a monitoring system, and fixed the by the operator. Or a "404: secret foo not found error" and a "403: unauthorized, not allowed to access secret bar" will give an unauthorized user hints about the presence of certain secret keys on the system, etc...

Having very rich error messages might be convenient for us who glue those things together (because we don't have to look at service logs to see actual the error) but for a security-related API we probably want to err on the side of being overly cautious rather than unintentionally leaking info.

Xynnn007:
Hi @beraldoleal I wonder if this PR can be closed - or if there are more considerations?

PR #1125, opened 2025-10-05
cdh: check for base64 padding in sealed secrets

beraldoleal:
Add validation to catch base64 payloads with padding ('=') before decode.

CDH expects URL_SAFE_NO_PAD format, in case of a secret with pad arrives, the generic error doesn't explain why.

This might help if someone manually encodes sealed secrets (though that's probably not a common workflow).

Not sure if this is worth the extra check, but figured a clearer error message doesn't hurt.


confidential-data-hub/hub/src/secret/mod.rs
```diff
@@ -58,6 +58,12 @@ impl Secret {
...
 
+        if sections[2].ends_with('=') {
+            return Err(SecretError::ParseFailed(
+                "base64 payload contains padding character '='. Sealed secrets must use URL_SAFE_NO_PAD encoding without padding",
+            ));
+        }
+
         let secret_json = b64
             .decode(sections[2])
             .map_err(|_| SecretError::ParseFailed("base64 decode Secret body"))?;
```
Xynnn007: Thanks for this patch @beraldoleal for user.

I noticed that we have an error message here. If the error message you'd got is this, is it ok to make here "base64 url safe(without '=' paddings) decode secret body failed" or something similar?

beraldoleal: Makes sense. pushed.

fitzthum (approved):
LGTM

Issue #1148, opened 2025-10-22
unpack fails on hardlinks with absolute paths due to canonicalization error


Park-Jiyeonn:
## Description:

In the guest-pull mode of Kata Containers, we encountered an issue when unpacking images containing hard links with absolute paths. The unpack operation fails with an error during canonicalization, such as:

`
No such file or directory (os error 2) while canonicalizing /usr/bin/unzip
`
This happens when using `image_client.pull_image`, which internally relies on the coco component for unpacking.

## Error Log:
```log
level=error msg="createContainer failed" error="rpc error: code = Internal desc = failed to handle layer: hasher sha256: failed to unpack `/run/kata-containers/image/layers/sha256_0d2c63be5ab65296c909dca41311bd7f01b82d8121842b7422166c9bc8ee5ee4/usr/bin/zipinfo`\n\nCaused by:\n    No such file or directory (os error 2) while canonicalizing /usr/bin/unzip ......
```

## Details:
In coco, the core unpack logic looks like this:

```rust
use tokio_tar::ArchiveBuilder;

pub async fn unpack<R: AsyncRead + Unpin>(input: R, destination: &Path) -> Result<()> {
    while let Some(file) = entries.next().await {
        ...
        file.unpack_in(destination).await?;
        ...
    }
}
```

In `astral-tokio-tar` (currently at version 0.5.6), the logic for hard links in `unpack_in` is:

```rust
if kind.is_hard_link() {
    let link_src = match target_base {
        Some(p) => {
            let link_src = p.join(src); // src is the linkname
            self.validate_inside_dst(p, &link_src).await?;
            link_src
        }
        None => src.into_owned(),
    };
}
```

If `src` is an absolute path, `Path::join` replaces the base path entirely. This causes `validate_inside_dst()` to canonicalize the original absolute path (e.g., `/usr/bin/unzip`), which may not exist in the guest VM image, resulting in the error above.

## Suggested Fix:

We propose to **locally override** the hardlink resolution in coco, without modifying astral-tokio-tar, to avoid affecting upstream behavior or other consumers.

How about this, this approach is similar to the behavior in [containerd/archive/tar.go](https://github.com/containerd/containerd/blob/0034ae955d6befb86026c7126bba0e1f3af4fa09/pkg/archive/tar.go#L373), which handles hardlinks.

---

**I'd be happy to work on a fix for this issue.**  
[...]

PR #1152, opened 2025-10-24
fix(image_rs): resolve hardlinks with absolute paths during unpack

Park-Jiyeonn:
### Problem
[#1148](https://github.com/confidential-containers/guest-components/issues/1148)

When unpacking container images using `coco` in guest-pull mode, hardlinks with absolute `Linkname`s (e.g., `/usr/bin/unzip`) cause unpack failures due to canonicalization errors.

### Solution
If `file.unpack_in` fails, it attempts a fallback using `try_hardlink_fallback`, which creates hard links for entries with absolute link paths.

The fallback logic resolves the destination path, normalizes the link path to ensure safety (e.g., removes root prefix and prevents escaping above root). If the fallback succeeds, the unpack continues; otherwise, it returns an error.

### Notes
If you add `println!("{:?}", e)` inside the `match file.unpack_in(destination).await {Err(e) => {} }` block while running the `test_unpack` test case, you can see the output which similar to the issue:
```log
Custom { kind: NotFound, error: TarError { desc: "failed to unpack `/tmp/image_test_dir/usr/bin/test_binary_hardlink`", io: Custom { kind: NotFound, error: "No such file or directory (os error 2) while canonicalizing /usr/bin/test_binary" } } }
```

### DIFF
--- a/image-rs/src/stream/unpack.rs
+++ b/image-rs/src/stream/unpack.rs
@@ -223,6 +223,7 @@ pub async fn unpack<R: AsyncRead + Unpin>(input: R, destination: &Path) -> Unpac
                 source: anyhow!("GID too large"),
             })?;
         let mode = file.header().mode().ok();
+        let kind = file.header().entry_type();
 
         if attr_available && is_whiteout(entry_name) {
             convert_whiteout(entry_name, &entry_path, uid, gid, mode, destination)
@@ -231,9 +232,21 @@ pub async fn unpack<R: AsyncRead + Unpin>(input: R, destination: &Path) -> Unpac
             continue;
         }
 
-        file.unpack_in(destination)
-            .await
-            .map_err(|source| UnpackError::UnpackFailed { source })?;
+        match file.unpack_in(destination).await {
+            Ok(_) => {}
+            Err(e) => match try_hardlink_fallback(&kind, &mut file, destination).await {
+                Ok(true) => {}
+                Ok(false) => return Err(UnpackError::UnpackFailed { source: e }),
+                Err(f) => {
+                    return Err(UnpackError::UnpackFailed {
+                        source: io::Error::new(
+                            io::ErrorKind::Other,
+                            format!("Fallback failed, {f:?}, after original unpack error: {e:?}"),
+                        ),
+                    })
+                }
+            },
+        }
 
         let path = format!(
             "{}/{}",
@@ -247,7 +260,6 @@ pub async fn unpack<R: AsyncRead + Unpin>(input: R, destination: &Path) -> Unpac
 
         let file_path = path_cstring.to_str().expect("must be utf8");
 
-        let kind = file.header().entry_type();
         let mtime = file
             .header()
             .mtime()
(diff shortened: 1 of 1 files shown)


image-rs/src/stream/unpack.rs
```diff
@@ -364,16 +369,85 @@ async fn set_perms_ownerships(
...
+    let mut out = PathBuf::new();
+    for comp in p.components() {
+        match comp {
+            Component::CurDir => {}
+            Component::Normal(s) => out.push(s),
+            Component::ParentDir => {
+                if !out.pop() {
+                    return None;
+                }
+            }
```
Xynnn007: Sometimes e.g. `a/b/../c` is not same to `a/c` if `b` is a softlink. I am afraid that `normalize_rel` will result in an unexpected directory. Do you think this will happen?

image-rs/src/stream/unpack.rs
```diff
@@ -231,9 +232,14 @@ pub async fn unpack<R: AsyncRead + Unpin>(input: R, destination: &Path) -> Unpac
...
-            .await
-            .map_err(|source| UnpackError::UnpackFailed { source })?;
+        match file.unpack_in(destination).await {
+            Ok(_) => {}
+            Err(e) => {
+                if !try_hardlink_fallback(&kind, &mut file, destination).await? {
+                    return Err(UnpackError::UnpackFailed { source: e });
+                }
+            }
+        }
```
Xynnn007, line 249: Something about the error message. If an error happens because of hardlink but not because of absolute issue, the new error in `try_hardlink_fallback` would overwrite the original one. What about to concat the error message into one?

Xynnn007 (commented):
Thanks for the fix! There are some comments inside the code.

image-rs/src/stream/unpack.rs
Park-Jiyeonn: Thx for pointing out, I agree with your suggestion, let me update it accordingly.

image-rs/src/stream/unpack.rs
Park-Jiyeonn: Yep, it could happen in edge cases, handling this correctly is a bit complexly, should I implement a RootPath function like containerd? 

In [unpack | astral-tokio-tar](https://github.com/astral-sh/tokio-tar/blob/ae1fc1d484f5dd8aecdca7ec7643b43b86faddf1/src/entry.rs#L566), it seems like it didn't handle this case.

* [RootPath function | Containerd](https://github.com/containerd/continuity/blob/5bea7844a31990fd39404b353857b54d25aaec43/fs/path.go#L225) 
* [handle hardlink | Containerd](https://github.com/containerd/containerd/blob/4f0b20fe855b9f9aeb6c180e65dad01d3d8a1e66/pkg/archive/tar.go#L373)

Xynnn007: oh this seems a lot of work. before getting into it, a question is what if we do not normalize path? is it acceptable?

Park-Jiyeonn: It's not necessarily to normalize, how about this:

only remove the prefix '/' from the absolute linkname and spell it into `destination`, then use canonization to resolve the real path and apply `starts_with (destination)` to prevent escape.

Xynnn007: This is a practical to me.

image-rs/src/stream/unpack.rs
```diff
@@ -364,16 +376,78 @@ async fn set_perms_ownerships(
...
+    }
+
+    // Drop the leading root from linkname
+    let stripped = linkname
+        .strip_prefix(Path::new("/"))
+        .unwrap_or(linkname.as_path());
+    let anchored_src = destination.join(stripped);
+
+    // Resolve symlinks according to the actual FS
+    let dst_canon = fs::canonicalize(destination)
```
Xynnn007, line 419: One thing to note that the `destination` should not contain any `..` like paths, or the target could still escape. But in the context we do not have such risk as the function `unpack` receives `destination` provided by https://github.com/confidential-containers/guest-components/blob/main/image-rs/src/layer_store.rs#L53-L56

Park-Jiyeonn:
> you may want to add a negative test case to make sure the anti-escape code works

Thx! I've added a negative test case to cover the anti-escape logic, 
```rust
+    #[tokio::test]
+    async fn test_unpack_rejects_escaping_absolute_hardlink() {
+        let td = tempfile::tempdir().unwrap();
......
```

and updated the error message to:
```rust
-                            format!("Fallback failed {f} after original unpack error {e}"),
+                            format!("Fallback failed, {f:?}, after original unpack error: {e:?}"),
```

PR #1163, opened 2025-10-31
trustee-attester: Allow passing initdata

Jakob-Naucke:
Add `--initdata` flag to trustee-attester, taking initdata as a string.

Preliminary change: some manpage fixes

cc @uril


fitzthum (approved):
LGTM 

It might be good to note somewhere that this is the plaintext init-data and that Trustee will expect the hash of the init-data to have been already measured somehow.

Xynnn007 (approved):

PR #1165, opened 2025-11-03
misc updates and cleanups

mythi:
Minor cleanups here and there:

- image-rs: use std::io::Error::other 
- confidential-data-hub: drop unused structs 
- attester: drop sgx-attester from all-attesters 
- image-rs: snapshots: allow derivable impls for SnapshotType


Xynnn007 (approved):

bpradipt (approved):
/lgtm
Thanks @mythi

PR #1172, opened 2025-11-07
github: fix architecture mapping while publishing images for ppc64le

Amulyam24:
Rust uses `powerpc64le` as the target triple, but OCI/Docker images conventionally use `ppc64le`. This change modifies the flow to set arch appropriately while publishing the components  to ensure images are tagged correctly, while keeping build targets unchanged.

### DIFF
--- a/.github/workflows/publish-artifacts.yml
+++ b/.github/workflows/publish-artifacts.yml
@@ -86,7 +86,10 @@ jobs:
         cd oras
         cp ../target/${{ env.RUST_TARGET }}/release/attestation-agent .
         tar cJf attestation-agent.tar.xz attestation-agent
-        arch_tag="${{ github.sha }}-${{ matrix.platform.tee }}_${{ matrix.platform.arch }}"
+        arch="${{ matrix.platform.arch }}"
+        # After building for target powerpc64le, tag and push the image as ppc64le to match standard arch naming.
+        [ "$arch" = "powerpc64le" ] && arch="ppc64le"
+        arch_tag="${{ github.sha }}-${{ matrix.platform.tee }}_${arch}"
         image="${REGISTRY}/${IMAGE_NAME}/attestation-agent"
         tag="${{ github.sha }}-${{ matrix.platform.tee }}"
         oras push "${image}:${arch_tag}" attestation-agent.tar.xz
@@ -126,7 +129,7 @@ jobs:
           libc: gnu
         - arch: aarch64
           libc: gnu
-        - arch: ppc64le
+        - arch: powerpc64le
           libc: gnu
     runs-on: ${{ matrix.arch == 's390x' && 's390x' || matrix.arch == 'ppc64le' && 'ubuntu-24.04-ppc64le' || 'ubuntu-24.04' }}
     env:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-artifacts.yml
```diff
@@ -86,7 +86,7 @@ jobs:
         cd oras
         cp ../target/${{ env.RUST_TARGET }}/release/attestation-agent .
         tar cJf attestation-agent.tar.xz attestation-agent
-        arch_tag="${{ github.sha }}-${{ matrix.platform.tee }}_${{ matrix.platform.arch }}"
+        arch_tag="${{ github.sha }}-${{ matrix.platform.tee }}_${OCI_ARCH}"
```
Amulyam24: @Xynnn007, Would this change have any implications for the images of the x86_64 and arm64 archs being consumed?

Xynnn007: Yes. PeerPods are now [fetching](https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/podvm/Makefile.inc#L204C62-L204C66) the binaries that ORAS pushes here and the env `ARCH` is set as form [x86_64](https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/Makefile.defaults#L27). If we change here there will be issue on that side.

Amulyam24: Modified it, PTAL!

.github/workflows/publish-artifacts.yml
```diff
@@ -86,7 +86,9 @@ jobs:
         cd oras
         cp ../target/${{ env.RUST_TARGET }}/release/attestation-agent .
         tar cJf attestation-agent.tar.xz attestation-agent
-        arch_tag="${{ github.sha }}-${{ matrix.platform.tee }}_${{ matrix.platform.arch }}"
+        arch="${{ matrix.platform.arch }}"
```
Xynnn007, line 89: Can we add a comment here to tell why we do this? I thin currently it's clear, but in later time we could forget that

.github/workflows/publish-artifacts.yml
```diff
@@ -176,8 +178,10 @@ jobs:
 
     - name: Publish CDH + ASR with ORAS
       id: publish
+      env:
+       OCI_ARCH: ${{ matrix.arch == 'powerpc64le' && 'ppc64le' || matrix.arch }}
```
Xynnn007: The logic here is sound. Is it possible to express it in the same way as AA? s.t.
no `env`, but set `arch_tag`, e.g.
```
arch="${{ matrix.arch }}"
[ "$arch" = "powerpc64le" ] && arch="ppc64le"
tag="${{ github.sha }}-${arch}"
```

Amulyam24: It makes sense, modified it, Thanks @Xynnn007! PTAL

.github/workflows/publish-artifacts.yml
Amulyam24: Yes, adding a comment will be helpful.

Xynnn007:
https://github.com/confidential-containers/guest-components/actions/runs/19257574542

PR #1179, opened 2025-11-11
github: avoid using short circuit evaluation in workflow

Amulyam24:
Short circuit evaluation returns an error code when condition evaluates to false. Change the format of if condition to avoid exiting the workflow on error.

### DIFF
--- a/.github/workflows/publish-artifacts.yml
+++ b/.github/workflows/publish-artifacts.yml
@@ -166,7 +166,9 @@ jobs:
         sudo apt-get update
         sudo apt-get install -y --no-install-recommends \
           libdevmapper-dev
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-artifacts.yml
```diff
@@ -166,7 +166,9 @@ jobs:
         sudo apt-get update
         sudo apt-get install -y --no-install-recommends \
           libdevmapper-dev
-        [ "${{ matrix.arch }}" = "powerpc64le" ] && sudo apt-get install -y libclang-dev cmake protobuf-compiler
```
Amulyam24, line 169: We can add `|| true` at the end of the statement, modified it as it would be a better and safer approach.

PR #1180, opened 2025-11-13
chore(deps): protobuf suites (prost/tonic things) Manual for #1160

Xynnn007:
tonic-build starts to be splited into tonic-prost-build crate to support prost abilities. Thus we bring in the new crates to build protobuf stuff. Also, this crate updates the relative protobuf built files.


protos/src/grpc/cdh/api.rs
```diff
@@ -694,11 +344,128 @@ pub mod sealed_secret_service_server {
...
-        const NAME: &'static str = SERVICE_NAME;
+    }
+    /// Generated gRPC service name
+    pub const SERVICE_NAME: &str = "api.SealedSecretService";
+    impl<T> tonic::server::NamedService for SealedSecretServiceServer<T> {
+        const NAME: &'static str = SERVICE_NAME;
+    }
+}
+/// Generated client implementations.
+pub mod get_resource_service_client {
```
mythi, line 355: any idea why this move happens? It makes the diff pretty big.

Xynnn007: It should be related to the updation of the underlying crates. Previously the order is "all client" -> "all server"

but new version seems "client-server" one-by-one.

The code structure looks complex and I gave up after digging into for quite a while

mythi: ok let's keep it as it is

mythi (approved):
LGTM

PR #1189, opened 2025-11-21
image-rs: add image pull benchmark

Xynnn007:
- Add criterion benchmark for image pull performance
- Configure benchmark harness in Cargo.toml
- Add image-pull benchmark that measures pull time and throughput

The terminal log can be
```
image-pull/pull_image   time:   [2.8439 s 4.1208 s 5.6908 s]
                        thrpt:  [1.5182 MiB/s 2.0966 MiB/s 3.0379 MiB/s]
                 change:
                        time:   [−55.955% −7.4571% +97.347%] (p = 0.88 > 0.05)
                        thrpt:  [−49.328% +8.0580% +127.04%]
                        No change in performance detected.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild
```

The generated webpage looks like

image

### DIFF
--- a/attestation-agent/kbs_protocol/test/policy.rego
+++ b/attestation-agent/kbs_protocol/test/policy.rego
@@ -2,6 +2,6 @@ package policy
 
 default allow = false
 
-allow {
+allow if {
 	input["submods"]["cpu0"]["ear.veraison.annotated-evidence"]["sample"]
 }


image-rs/benches/image-pull.rs
fitzthum, line 1: Can we call this `benchmarks` rather than `benches`. Ik it's `cargo bench`, but `benchmark` seems a little better here.

image-rs/benches/image-pull.rs
```diff
@@ -0,0 +1,80 @@
...
+        return entries
+            .map(|entry| get_total_file_size(&entry.unwrap().path()))
+            .sum();
+    }
+
+    0
+}
+
+fn pull_image() -> u64 {
+    let image_url = "ghcr.io/confidential-containers/test-container:unencrypted";
```
fitzthum: Maybe we should have a way to select the image? Would it be good to have multiple images to test? Also, I wonder if we should have a dedicated image for benchmarking? People might get rate-limited here (and then have problems running the other tests). This is probably fine to start, but let's think if there is another way to handle the images. Maybe it would actually make sense to have a local registry for this so we don't have network jitters. That's a bit annoying to implement tho.

Xynnn007: Good point. Currently dockerhub has rate limit so I gave up using images like `ubuntu`. However `ghcr.io` does not have such limit and that's the reason that we had replaced images on image-rs from dockerhub to ghcr.io.

Let me make the image name a variable slice on top of the file, thus users can easily find the way to change code and do the test. Usually people do not do benchmark on CI but only provide the benchmark code to users to help them to test in their own environments.

image-rs/benches/image-pull.rs
Xynnn007: Yes I also think that might not be a good name as I once typo-ed to `benchmark` to, but it's the official [cargo layout](https://doc.rust-lang.org/cargo/guide/project-layout.html) name here, or it will not be detected by Cargo.toml here

Xynnn007:
> Is there an easy way to keep track of memory usage here? Peak memory consumption is a key metric for image pulling because it affects the guest memory size. I'm not sure the best way to measure that tho.

Let me find a way to resolve this in the PR.

PR #1198, opened 2025-12-01
CDH | Extend AAEL for PullImage Event

### LINKED ISSUE #1176: [CDH] Extend AAEL for PullImage Event
After we pull the image using CDH, CDH needs to attempt to call AA's extend runtime measurement interface to extend a PullImage AAEL project.

Xynnn007:
Close #1176 

cc @mythi

### DIFF
--- a/confidential-data-hub/hub/Cargo.toml
+++ b/confidential-data-hub/hub/Cargo.toml
@@ -33,13 +33,13 @@ path = "src/bin/secret_cli.rs"
 required-features = ["cli"]
 
 [dependencies]
-anyhow = { workspace = true, optional = true }
+anyhow.workspace = true
 async-trait.workspace = true
-attestation-agent = { path = "../../attestation-agent/attestation-agent", default-features = false, optional = true }
+attestation-agent = { path = "../../attestation-agent/attestation-agent", default-features = false }
 base64.workspace = true
 cfg-if.workspace = true
 clap = { workspace = true, features = ["derive"], optional = true }
-config = { workspace = true, optional = true }
+config = { workspace = true }
 crypto.path = "../../attestation-agent/deps/crypto"
 env_logger = { workspace = true, optional = true }
 image-rs = { path = "../../image-rs", default-features = false, features = [
@@ -51,7 +51,7 @@ prost = { workspace = true, optional = true }
 protos = { path = "../../protos", default-features = false, optional = true }
 rand.workspace = true
 resource_uri.path = "../../attestation-agent/deps/resource_uri"
-serde = { workspace = true, optional = true }
+serde.workspace = true
 serde_json.workspace = true
 strum = { workspace = true, features = ["derive"] }
 tempfile = { workspace = true, optional = true }
@@ -96,7 +96,7 @@ sev = ["kms/sev"]
 ehsm = []
 
 # Binary RPC type
-bin = ["anyhow", "attestation-agent", "clap", "config", "env_logger", "serde"]
+bin = ["clap", "env_logger"]
 ttrpc = ["dep:ttrpc", "tokio/signal", "protos/ttrpc"]
 grpc = ["prost", "tonic", "tokio/signal", "protos/grpc"]
 


Xynnn007:
Have test this patch on alibaba cloud peerpod. The trustee side claims of eventlog
image

attestation-agent/attester/src/az_snp_vtpm/mod.rs
```diff
@@ -59,6 +64,17 @@ impl Attester for AzSnpVtpmAttester {
...
+    fn pcr_to_ccmr(&self, pcr_index: u64) -> u64 {
+        pcr_index
+    }
+
+    fn ccel_hash_algorithm(&self) -> HashAlgorithm {
+        HashAlgorithm::Sha256
+    }
+
+    // TODO: add get_runtime_measurement function
+    // See https://github.com/confidential-containers/guest-components/issues/1201
```
Xynnn007, line 77: Currently, the Azure suite lacks this function to enable AAEL recording functionality, so I've added TODO notes and issue link to the code.

cc @mkulke

attestation-agent/attester/src/lib.rs
```diff
@@ -107,7 +112,7 @@ pub trait Attester {
         _event_digest: Vec<u8>,
         _register_index: u64,
     ) -> Result<()> {
-        bail!("Unimplemented")
+        panic!("The Attester does not support extending runtime measurement")
```
fitzthum: do we want to panic here? Seems like we could bail with the new message.

protos/protos/attestation-agent/attestation-agent.proto
```diff
@@ -39,7 +39,15 @@ message ExtendRuntimeMeasurementRequest {
     optional uint64 RegisterIndex = 4;
 }
 
-message ExtendRuntimeMeasurementResponse {}
+enum RuntimeMeasurementResult {
+    NOT_SUPPORTED = 0;
+    NOT_ENABLED = 1;
+    OK = 2;
```
fitzthum: nit: maybe `OK` should be 0.

image-rs/src/image.rs
```diff
@@ -153,6 +153,17 @@ pub struct ImageMeta {
     pub layer_metas: Vec<LayerMeta>,
 }
 
+/// The information of the pulled image.
+#[derive(Debug, Clone, PartialEq)]
+pub struct ImageInfo {
+    /// The digest of the image configuration.
```
fitzthum, line 159: Is the id a digest? I thought it was more like the container name?

confidential-data-hub/hub/src/hub.rs
```diff
@@ -86,12 +97,64 @@ impl DataHub for Hub {
...
+                Domain: "github.com/confidential-containers".to_string(),
+                Operation: "PullImage".to_string(),
+                Content: format!(
+                    r#"{{"image":"{image_url}", "digest":"{}"}}"#,
+                    image_info.digest
+                ),
+                ..Default::default()
+            };
+            let res = aa_client
+                .extend_runtime_measurement(with_timeout(10 * 1000 * 1000 * 1000), &req)
```
fitzthum: Do we have a constant we can use for this timeout?

confidential-data-hub/hub/src/hub.rs
```diff
@@ -86,12 +97,64 @@ impl DataHub for Hub {
...
-        let manifest_digest = client
+
+        let image_info = client
             .lock()
             .await
             .pull_image(image_url, Path::new(bundle_path), &None, &None)
             .await?;
-        Ok(manifest_digest)
+
+        #[cfg(feature = "ttrpc")]
```
fitzthum, line 112: Remind me does the AA not have grpc support? Should we be giving people some kind of warning if the right feature is not enabled?

fitzthum (commented):
Nice. A few comments

attestation-agent/attester/src/lib.rs
Xynnn007: Ok. Currently, only AA's `extend runtime measurement` calls this function, and the function won't be called if`supports_runtime_measurement` is not true, so both `panic` and `bail` are acceptable behaviors. I'll change it to `bail` to maintain the previous behavior.

protos/protos/attestation-agent/attestation-agent.proto
Xynnn007: Yes, let me fix this

image-rs/src/image.rs
Xynnn007: It's the digest of `config` of [manifest](https://github.com/opencontainers/image-spec/blob/main/manifest.md), usually format like `sha256:1807eadf4c1740ebb315fc39e587e96ffda165917aada5de292b5368307d19fa`, which is the same format as manifest digest

Xynnn007: Let add more notes upon the field and rename the field

confidential-data-hub/hub/src/hub.rs
Xynnn007: Let me add one

confidential-data-hub/hub/src/hub.rs
Xynnn007: AA supports grpc. 

There's actually a slight hack here. When building CDH and enabling the ttrpc feature, it automatically detects whether there's a corresponding ttrpc listener (not gRPC). Considering that CoCo currently uses ttrpc, there's no additional option to configure whether it's a gRPC/ttrpc.

Does this make sense for now?

confidential-data-hub/hub/src/hub.rs
```diff
@@ -86,12 +97,75 @@ impl DataHub for Hub {
...
+                .aa_client
+                .get_or_try_init(|| async move { initialize_aa_client().await })
+                .await?;
+
+            let Some(aa_client) = aa_client else {
+                warn!("Attestation Agent socket file not found, so all runtime measurement extension will be skipped.");
+                return Ok(image_info.manifest_digest);
+            };
+
+            info!("Extend image pull event via AA's runtime measurement API...");
```
mythi, line 130: instead of adding `debug!()` to image-rs, how about just adding the same or full `image_info` content in this?

Xynnn007: Yes, I added another debug! right after the info! here. As this content might be long and suitable to use debug

PR #1206, opened 2025-12-02
image-rs: drop Occlum UnionFS snapshotter

mythi:
Given enclave-cc deprecation, the Occlum libOS specific snapshotter used by enclave-cc is no longer needed.

Remove it along with the cargo features used to toggle between the two snapshotter implementations.

The Snapshotter trait object handling is kept but this could also be removed in the future. default_snapshotter setting in Image Config is kept but ignored for compatibility.

The Display impl for Snapshotter is also removed since it can be replaced with OverlayFS internals.


image-rs/src/snapshots/overlay.rs
```diff
@@ -19,14 +19,16 @@ pub struct OverlayFs {
 
 impl OverlayFs {
     /// Create a new instance of [OverlayFs].
-    pub fn new(data_dir: PathBuf) -> Self {
+    pub fn new(work_dir: PathBuf) -> Self {
+        let data_dir = work_dir.join(OVERLAYFS_FS_TYPE);
```
Xynnn007, line 23: This might not be needed? e.g. original the data will be under `xxx/` but after this will be `xxx/overlay`, which is not the same, right?

mythi: It was like this originally:
```
let data_dir = work_dir.join(SnapshotType::Overlay.to_string());
```
(L202 which I moved here).

Xynnn007: Oh, I didn't notice that before. Because the `work_dir` parameter input to `init_snapshot` is the overall `work_dir` of `ImageClient`, not the snapshotter's. I was confused here, thinking that the caller of `OverlayFs::new` processed this sub-level before the call.

mythi: are you ok as it is now?

Xynnn007: Yes.

mythi:
ok to merge?

fitzthum (approved):
LGTM

Issue #1210, opened 2025-12-03
image-rs: Whiteout conversion skipped entirely when xattr check fails, causing deleted files to reappear


ajaypvictor:
### Describe the bug

Container image layers that delete files (e.g., `RUN rm -rf /etc/apache2`) create whiteout files that must be converted to character devices. However, image-rs skips ALL whiteout conversion when the `user.* xattr` namespace is unavailable, even though regular deletion whiteouts only require `mknod()` and don't need `xattrs` at all.

Impact: Deleted files reappear in running containers on systems where tmpfs doesn't support the `user.* xattr` namespace.

Affected Systems: Any environment where `/run/kata-containers/image/layers/` is on tmpfs without `user_xattr` mount option.

The current code in `image-rs/src/stream/unpack.rs` at `line 228` treats both types in same way:
```
if attr_available && is_whiteout(entry_name) {
    convert_whiteout(...)
}
```
When `attr_available` returns false, both types are skipped, even though regular deletion whiteouts doesn't need `xattrs`.

Function `is_attr_available()` tests `xattr` support using the `user.* namespace`:
```
fn is_attr_available(path: &Path) -> bool {
    // Tests: set_xattr("user.overlay.origin", b"")
    // ...
}
```

Many systems mount tmpfs without the `user_xattr` option:
```
tmpfs on /run type tmpfs (rw,nosuid,nodev,size=...,mode=755,inode64)
```
where the whiteouts aren't:
```
root@9a848a9b0552:/# ls -la /run/kata-containers/image/layers/1/etc/apache2
crw------- 1 root root 0, 0 Dec  3 14:41 /run/kata-containers/image/layers/1/etc/apache2
root@9a848a9b0552:/#
```

### How to reproduce

Step 1: Create Test Image
```
FROM debian:stable-slim
RUN apt-get update && apt-get install -qy apache2
RUN rm -rf /etc/apache2
CMD ["find", "/etc/apache2"]
```
Step 2: Deploy on System with `tmpfs` without `user_xattr` and see the `whiteout` layer
Having `/run/kata-containers/image/layers/` on `tmpfs` mounted without `user_xattr`: 

In kata guest VM, verify mount:
`mount | grep "/run "` have: `tmpfs on /run type tmpfs (rw,nosuid,nodev,...)` without `user_xattr` option

and check the whiteout layer to be:
```
# ls -l /run/kata-containers/image/layers/2/etc/.wh.apache2
[...]

PR #1211, opened 2025-12-03
image-rs: Fix whiteout conversion when xattr check fails

ajaypvictor:
This patch fixes a bug where ALL whiteout conversion was skipped when the user.* xattr namespace was unavailable, even though regular deletion whiteouts only require mknod() and don't need xattr support.

The fix separates the requirements:
Regular whiteouts (.wh.*): Only need mknod(), don't need xattrs Opaque directories (.wh..wh..opq): Need xattrs to set trusted.overlay.opaque

##### Current flow - without the PR changes
1. Check the whiteout file created by image-rs:
```
ls -l /run/kata-containers/image/layers/2/etc/.wh.apache2
---------- 1 root root 0 .wh.apache2  
```

2. With tmpfs mount options:

```
mount | grep "^tmpfs on /run "
tmpfs on /run type tmpfs (rw,nosuid,nodev,...) 
```

3. Verify `mknod()` works:
```
cd /tmp
mknod test_char c 0 0
ls -l test_char
crw-r--r-- 1 root root 0, 0 test_char  
```

4. Manual fix to prove it works:
```
rm /run/kata-containers/image/layers/2/etc/.wh.apache2
mknod /run/kata-containers/image/layers/2/etc/.wh.apache2 c 0 0
ls -l /run/kata-containers/image/layers/2/etc/.wh.apache2
crw-r--r-- 1 root root 0, 0 .wh.apache2 
```

5. Test overlay mount with corrected whiteout:
```
mount -t overlay overlay -o lowerdir=layer2:layer1 /tmp/test
ls /tmp/test/etc/apache2
ls: cannot access: No such file or directory 
```

##### With PR changes
```
ls -l /run/kata-containers/image/layers/2/etc/apache2
c--------- 1 root root 0, 0 apache2  
```

```
kubectl logs test-whiteout
find: '/etc/apache2': No such file or directory  
```

Fixes: #1210

### DIFF
--- a/image-rs/src/stream/unpack.rs
+++ b/image-rs/src/stream/unpack.rs
@@ -122,7 +122,12 @@ fn is_whiteout(name: &str) -> bool {
     name.starts_with(WHITEOUT_PREFIX)
 }
 
-/// Converts a whiteout file or opaque directory. See OverlayFS and Aufs documentation for details
+/// Converts a whiteout file or opaque directory.
+///
+/// Regular deletion whiteouts only need mknod() and work without xattr.
+/// Opaque directory whiteouts require xattr to set trusted.overlay.opaque.
+///
+/// See OverlayFS and Aufs documentation for details:
 /// https://www.kernel.org/doc/Documentation/filesystems/overlayfs.txt
 /// https://aufs.sourceforge.net/aufs.html
 async fn convert_whiteout(
@@ -132,6 +137,7 @@ async fn convert_whiteout(
     gid: u32,
     mode: Option<u32>,
     destination: &Path,
+    attr_available: bool,
 ) -> Result<()> {
     let parent = path
         .parent()
@@ -139,6 +145,15 @@ async fn convert_whiteout(
 
     // Handle opaque directories
     if name == WHITEOUT_OPAQUE_DIR {
+        // Opaque directory whiteout requires xattr support
+        if !attr_available {
+            debug!(
+                "Skipping opaque directory whiteout (xattr unavailable) for: {:?}",
+                destination.join(parent)
+            );
+            return Ok(());
+        }
+
         let destination_parent = destination.join(parent);
         xattr::set(destination_parent, "trusted.overlay.opaque", b"y")?;
         return Ok(());
@@ -225,10 +240,18 @@ pub async fn unpack<R: AsyncRead + Unpin>(input: R, destination: &Path) -> Unpac
         let mode = file.header().mode().ok();
         let kind = file.header().entry_type();
 
-        if attr_available && is_whiteout(entry_name) {
-            convert_whiteout(entry_name, &entry_path, uid, gid, mode, destination)
-                .await
-                .map_err(|source| UnpackError::ConvertWhiteoutFailed { source })?;
+        if is_whiteout(entry_name) {
+            convert_whiteout(
+                entry_name,
+                &entry_path,
+                uid,
+                gid,
+                mode,
+                destination,
+                attr_available,
+            )
+            .await
+            .map_err(|source| UnpackError::ConvertWhiteoutFailed { source })?;
             continue;
         }
 


image-rs/src/stream/unpack.rs
```diff
@@ -225,7 +225,8 @@ pub async fn unpack<R: AsyncRead + Unpin>(input: R, destination: &Path) -> Unpac
         let mode = file.header().mode().ok();
         let kind = file.header().entry_type();
 
-        if attr_available && is_whiteout(entry_name) {
+        // Regular whiteouts only need mknod(), not xattrs; opaque dirs need xattrs
+        if is_whiteout(entry_name) && (entry_name != WHITEOUT_OPAQUE_DIR || attr_available) {
```
c3d: Shouldn't that logic be inside `convert_whiteout`? It seems like you are breaking the abstraction barrier a little, exposing the internal logic of whiteouts outside of whiteout-specific functions. This is made visible by:
1. Having to deal with a `WHITEOUT_OPAQUE_DIR` name at the call site
2. Having the caller need to know how internally `convert_whiteout` uses `xattrs` or `mknod()`

This is likely to cause problem if we ever need to call `convert_whiteout` elsewhere.

Xynnn007:
@stevenhorsman Yes. Previously, we encountered errors when pulling images. After fixing this bug, I need to slightly refactor the test module to support post-pull check.

image-rs/src/stream/unpack.rs
Xynnn007: Hi @c3d @ajaypvictor There might be some trade-offs here. Currently, whiteout is handled in two ways:

1. Files: Do not depend on xattr

2. Directories: Require the current file system to support xattr

Meanwhile, there isn't a particularly good way to detect xattr support. `is_attr_available` directly adds the xattr to the current directory to check if it's successful; this method is very heavy and should ideally only be executed once.

Therefore, I suspect that probably three design approaches are possible:

1. The `convert_whiteout` function checks each target, i.e., calls `is_attr_available` (needs a little change upon the function) for each file. If it's a directory, whiteout is performed directly. This is heavy.

2. `convert_whiteout` accepts an additional parameter `is_attr_available`, which determines whether to handle directory-level whiteout. (or rename to `skip_directory`)

3. Like the design in the PR, expose some internal details.

Not sure the division is right, but I am ok with either 2 or 3. wdyt?

ajaypvictor: Thanks alot for the suggestion @Xynnn007 and @c3d 

Could you please check if the latest changes looks better?

ajaypvictor: Apologies for the ping again, @Xynnn007 @c3d could you please re-review when you get a chance?

Xynnn007 (approved):

image-rs/src/stream/unpack.rs
Xynnn007: yes thanks. lgtm!

image-rs/src/stream/unpack.rs
```diff
@@ -123,6 +123,11 @@ fn is_whiteout(name: &str) -> bool {
 }
 
 /// Converts a whiteout file or opaque directory. See OverlayFS and Aufs documentation for details
+///
+/// Regular deletion whiteouts only need mknod() and work without xattr.
+/// Opaque directory whiteouts require xattr to set trusted.overlay.opaque.
+///
+/// See OverlayFS and Aufs documentation for details:
```
mythi, line 130: Drop the first occurrence of this in the same comment block

ajaypvictor: Thanks for the spotting this, fixed it..

ajaypvictor: > I think CoCo sets CONFIG_TMPFS_XATTR=y a [hard requirement](https://github.com/kata-containers/kata-containers/commit/42531cf6c4cada273385ce127aa496781e86a396). Is this PR fixing a subset of images to avoid that requirement?

I think we are not avoiding this requirement, sorry if the way I wrote the description created any confusions..
This fix was for ensuring that the regular whiteouts don't need xattr at all, but requires to proceed with `mknod()`..

mythi (approved):
LGTM

PR #1224, opened 2025-12-10
image-rs: allow signature configs in config file

fitzthum:
Today the image signature configurations are referenced via a URI in the image config or the kernel command line. The URI can point to a remote resource or a local file inside the guest. In addition, let's allow the configuration to be specified directly in the image config. This will allow it to easily be set via the init-data.

Generally, it seems like a nicer flow to set the image signature configuration via init-data rather than provisioning the config files to Trustee. Also, retrieving unsigned configuration files over the network does have some subtleties.

Rather than overloading the existing URI fields, let's introduce a new field where the config can be set directly. This is similar to what we already do with the registry config. The URI will take priority if both fields are set and won't break any existing use cases (such as when the URI is set via the kernel command line).

Ideally these new fields would have Struct types allowing the configurations to be integrated directly into the image config and the cdh config, but to keep things simple we store them as strings.

### DIFF
--- a/image-rs/src/config.rs
+++ b/image-rs/src/config.rs
@@ -98,6 +98,12 @@ pub struct ImageConfig {
     #[serde(default = "Option::default")]
     pub image_security_policy_uri: Option<String>,
 
+    /// The image security policy (see above) can also be specified directly
+    /// as a string. If the policy URI and this field are both set,
+    /// the policy URI will be used.
+    #[serde(default = "Option::default")]
+    pub image_security_policy: Option<String>,
+
     /// Sigstore config file URI for simple signing scheme.
     ///
     /// When `image_security_policy_uri` is set and `SimpleSigning` (signedBy) is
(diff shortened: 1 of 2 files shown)


image-rs/src/config.rs
```diff
@@ -98,6 +98,11 @@ pub struct ImageConfig {
     #[serde(default = "Option::default")]
     pub image_security_policy_uri: Option<String>,
 
+    /// The image security policy (see above) can also be specified directly
+    /// as a string.
```
Xynnn007: Let's mention the case if both are given

Xynnn007 (commented):
Good. In terms of priority, the URI currently takes precedence over the configured settings. I don't have a strong preference here, but I'm just mentioning it to ensure that this meets your expectations.

image-rs/src/config.rs
fitzthum: Yeah, i made URI take priority since it is what is currently supported. Also, I think stuff set via the kernel command line should generally have precedence. Let me add to this comment.

PR #1250, opened 2025-12-29
ci: update apt repos in weekly Rust stable tests

mythi:
Workflow runs get 404 errors when installing packages if the runner apt metadata is outdated. Always refresh the metadata before installing anything.


fitzthum (approved):
LGTM. Thanks

PR #1259, opened 2026-01-07
Add a compatibility matrix

bpradipt:
I took help from AI agent to pull up the details. Keeping it in draft for the following reasons:
- Get early feedback on whether something like this will be useful to make it easier for users and adopters to know which version of guest-components will work with Trustee
- Correctness of the data

cc @confidential-containers/guest-components-maintainers @confidential-containers/trustee-maintainers

### DIFF
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -1,0 +1,122 @@
+# Compatibility Matrix
+
+This document tracks compatibility between guest-components and [Trustee](https://github.com/confidential-containers/trustee) (the server-side components including KBS, Attestation Service, and RVPS).
+
+## KBS Protocol Version
+
+Guest-components and Trustee communicate using the KBS (Key Broker Service) protocol. Both sides must use compatible protocol versions.
+
+| Protocol Version | Status     | Notes                                      |
+|------------------|------------|--------------------------------------------|
+| v0.4.0           | Current    | Used in CoCo v0.15.0+                      |
+| v0.3.0           | Deprecated | Legacy, not recommended                    |
+
+## Release Compatibility
+
+The following table shows tested version combinations. Versions are coordinated through [CoCo umbrella releases](https://github.com/confidential-containers/confidential-containers/tree/main/releases).
+
+| CoCo Release | guest-components | Trustee  | KBS Protocol | Release Date |
+|--------------|------------------|----------|--------------|--------------|
+| v0.17.0      | v0.16.0          | v0.16.0  | v0.4.0       | Nov 2025     |
+| v0.16.0      | v0.15.0          | v0.15.0  | v0.4.0       | Oct 2025     |
+| v0.15.0      | v0.14.0          | v0.14.0  | v0.4.0       | Sep 2025     |
+| v0.14.0      | v0.13.0          | v0.13.0  | v0.4.0       | May 2025     |
+| v0.13.0      | v0.12.0          | v0.12.0  | v0.4.0       | Mar 2025     |
+
+**Note:** Using mismatched versions may work but is not tested. For production deployments, use versions from the same CoCo release.
+
+## TEE Platform Support
+
+### Attestation Support Matrix
+
+| TEE Platform       | guest-components | Trustee  | Architecture | Status       |
+|--------------------|------------------|----------|--------------|--------------|
+| Intel TDX          | v0.8.0+          | v0.8.0+  | x86_64       | Stable       |
+| Intel SGX          | v0.7.0+          | v0.7.0+  | x86_64       | Stable       |
+| AMD SEV-SNP        | v0.7.0+          | v0.7.0+  | x86_64       | Stable       |
+| AMD SEV(-ES)       | v0.7.0+          | v0.7.0+  | x86_64       | Stable       |
+| Azure TDX vTPM     | v0.9.0+          | v0.9.0+  | x86_64       | Stable       |
+| Azure SNP vTPM     | v0.9.0+          | v0.9.0+  | x86_64       | Stable       |
+| ARM CCA            | v0.10.0+         | v0.10.0+ | aarch64      | Experimental |
+| IBM Secure Exec    | v0.10.0+         | v0.10.0+ | s390x        | Stable       |
+| Hygon CSV          | v0.12.0+         | v0.12.0+ | x86_64       | Experimental |
+| NVIDIA GPU         | v0.15.0+         | v0.15.0+ | x86_64       | Experimental |
+
+### Feature Flags
+
+guest-components uses feature flags to enable TEE-specific code. The following table maps TEE platforms to the required build configuration:
+
+| TEE Platform       | `TEE_PLATFORM` | Attester Feature Flag |
+|--------------------|----------------|-----------------------|
+| Intel TDX          | `tdx`          | `tdx-attester`        |
+| Intel SGX          | (manual)       | `sgx-attester`        |
+| AMD SEV-SNP        | `snp`          | `snp-attester`        |
+| AMD SEV(-ES) + SNP | `amd`          | `snp-attester`        |
+| Azure TDX vTPM     | `az-tdx-vtpm`  | `az-tdx-vtpm-attester`|
+| Azure SNP vTPM     | `az-snp-vtpm`  | `az-snp-vtpm-attester`|
+| Azure CVM vTPM     | `az-cvm-vtpm`  | Both Azure attesters  |
+| ARM CCA            | `cca`          | `cca-attester`        |
+| IBM Secure Exec    | `se`           | `se-attester`         |
+
+## Component Dependencies
+
+Trustee embeds certain guest-components crates as dependencies:
+
+| Trustee Component     | guest-components Crate | Purpose                          |
+|-----------------------|------------------------|----------------------------------|
+| KBS                   | `kbs_protocol`         | Protocol types and client        |
+| Attestation Service   | `kms`                  | KMS provider implementations     |
+
+These dependencies are pinned to specific git revisions in Trustee's `Cargo.toml`. When upgrading, ensure the pinned revision is compatible with your guest-components version.
+
+## Breaking Changes
+
+### v0.15.0+
+- KBS configuration format updated
+- Token validation changes
+
+### v0.10.0+
+- Added support for multiple concurrent attesters
+- Changed attestation evidence format for some platforms
+
+### v0.8.0+
+- Initial stable KBS protocol v0.4.0
+
+## Testing Compatibility
+
+To verify compatibility between guest-components and Trustee:
+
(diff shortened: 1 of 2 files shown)


COMPATIBILITY.md
```diff
@@ -0,0 +1,122 @@
...
+| NVIDIA GPU         | v0.15.0+         | v0.15.0+ | x86_64       | Experimental |
+
+### Feature Flags
+
+guest-components uses feature flags to enable TEE-specific code. The following table maps TEE platforms to the required build configuration:
+
+| TEE Platform       | `TEE_PLATFORM` | Attester Feature Flag |
+|--------------------|----------------|-----------------------|
+| Intel TDX          | `tdx`          | `tdx-attester`        |
+| Intel SGX          | (manual)       | `sgx-attester`        |
```
fitzthum, line 52: not sure what `manual` is

COMPATIBILITY.md
```diff
@@ -0,0 +1,122 @@
...
+| Intel TDX          | v0.8.0+          | v0.8.0+  | x86_64       | Stable       |
+| Intel SGX          | v0.7.0+          | v0.7.0+  | x86_64       | Stable       |
+| AMD SEV-SNP        | v0.7.0+          | v0.7.0+  | x86_64       | Stable       |
+| AMD SEV(-ES)       | v0.7.0+          | v0.7.0+  | x86_64       | Stable       |
+| Azure TDX vTPM     | v0.9.0+          | v0.9.0+  | x86_64       | Stable       |
+| Azure SNP vTPM     | v0.9.0+          | v0.9.0+  | x86_64       | Stable       |
+| ARM CCA            | v0.10.0+         | v0.10.0+ | aarch64      | Experimental |
+| IBM Secure Exec    | v0.10.0+         | v0.10.0+ | s390x        | Stable       |
+| Hygon CSV          | v0.12.0+         | v0.12.0+ | x86_64       | Experimental |
+| NVIDIA GPU         | v0.15.0+         | v0.15.0+ | x86_64       | Experimental |
```
fitzthum, line 43: NVIDIA should be stable now.

Are we missing TPM here?

COMPATIBILITY.md
```diff
@@ -0,0 +1,122 @@
...
+## TEE Platform Support
+
+### Attestation Support Matrix
+
+| TEE Platform       | guest-components | Trustee  | Architecture | Status       |
+|--------------------|------------------|----------|--------------|--------------|
+| Intel TDX          | v0.8.0+          | v0.8.0+  | x86_64       | Stable       |
+| Intel SGX          | v0.7.0+          | v0.7.0+  | x86_64       | Stable       |
+| AMD SEV-SNP        | v0.7.0+          | v0.7.0+  | x86_64       | Stable       |
+| AMD SEV(-ES)       | v0.7.0+          | v0.7.0+  | x86_64       | Stable       |
```
fitzthum, line 37: Not supported anymore, although I'm not sure if we've ripped the code out of GC yet.

COMPATIBILITY.md
```diff
@@ -0,0 +1,122 @@
...
+### v0.15.0+
+- KBS configuration format updated
+- Token validation changes
+
+### v0.10.0+
+- Added support for multiple concurrent attesters
+- Changed attestation evidence format for some platforms
+
+### v0.8.0+
+- Initial stable KBS protocol v0.4.0
```
fitzthum, line 83: This doesn't seem right. v0.4.0 would have been after multi-device attestation (v0.3.0) and init-data plaintext support (v0.4.0)

COMPATIBILITY.md
```diff
@@ -0,0 +1,122 @@
...
+   ```
+
+2. **Integration Tests**: Use the same CoCo release versions
+   ```shell
+   # Start Trustee KBS
+   docker run -p 8080:8080 ghcr.io/confidential-containers/staged-images/kbs:v0.16.0
+
+   # Test with guest-components client
+   cd attestation-agent
+   cargo test --features kbs
```
fitzthum, line 103: Not sure if either of these tests will actually work. I think the unit test will be skipped if the version doesn't match. At least someone should try running this.

fitzthum (commented):
Something like this could make sense. I wonder if we will be able to keep it up-do-date and if this is the right place for it. Maybe in coco/coco or the release notes or the website would make more sense?

There are a number of errors in this doc as well I think.

COMPATIBILITY.md
bpradipt: Yeah.
Wondering do we need to add this level of details or just leave at GC and Trustee version ? Might become hard to maintain going forward. Wdyt ?

COMPATIBILITY.md
```diff
@@ -0,0 +1,122 @@
...
+
+| CoCo Release | guest-components | Trustee  | KBS Protocol | Release Date |
+|--------------|------------------|----------|--------------|--------------|
+| v0.17.0      | v0.16.0          | v0.16.0  | v0.4.0       | Nov 2025     |
+| v0.16.0      | v0.15.0          | v0.15.0  | v0.4.0       | Oct 2025     |
+| v0.15.0      | v0.14.0          | v0.14.0  | v0.4.0       | Sep 2025     |
+| v0.14.0      | v0.13.0          | v0.13.0  | v0.4.0       | May 2025     |
+| v0.13.0      | v0.12.0          | v0.12.0  | v0.4.0       | Mar 2025     |
+
+**Note:** Using mismatched versions may work but is not tested. For production deployments, use versions from the same CoCo release.
```
Xynnn007, line 26: This will not work. There is version check logic https://github.com/confidential-containers/trustee/blob/main/kbs/src/attestation/backend.rs#L220-L227 to assert the versions are the same?

COMPATIBILITY.md
```diff
@@ -0,0 +1,122 @@
...
+|--------------------|------------------|----------|--------------|--------------|
+| Intel TDX          | v0.8.0+          | v0.8.0+  | x86_64       | Stable       |
+| Intel SGX          | v0.7.0+          | v0.7.0+  | x86_64       | Stable       |
+| AMD SEV-SNP        | v0.7.0+          | v0.7.0+  | x86_64       | Stable       |
+| AMD SEV(-ES)       | v0.7.0+          | v0.7.0+  | x86_64       | Stable       |
+| Azure TDX vTPM     | v0.9.0+          | v0.9.0+  | x86_64       | Stable       |
+| Azure SNP vTPM     | v0.9.0+          | v0.9.0+  | x86_64       | Stable       |
+| ARM CCA            | v0.10.0+         | v0.10.0+ | aarch64      | Experimental |
+| IBM Secure Exec    | v0.10.0+         | v0.10.0+ | s390x        | Stable       |
+| Hygon CSV          | v0.12.0+         | v0.12.0+ | x86_64       | Experimental |
```
Xynnn007, line 42: We also have `Hygon DCU` from v0.15.0

COMPATIBILITY.md
mythi: protocol history details is in this issue https://github.com/confidential-containers/trustee/issues/907

COMPATIBILITY.md
```diff
@@ -0,0 +1,122 @@
+# Compatibility Matrix
+
+This document tracks compatibility between guest-components and [Trustee](https://github.com/confidential-containers/trustee) (the server-side components including KBS, Attestation Service, and RVPS).
+
+## KBS Protocol Version
+
+Guest-components and Trustee communicate using the KBS (Key Broker Service) protocol. Both sides must use compatible protocol versions.
```
mkulke, line 7: Currently we have no enforced schema for the evidence _payload_, so if some encoding changes (e.g. because of a bincode update/replacement in GC or in a dependency). a change on the payload would break Trustee interop. Until we have this, it's an aspirational goal, but we cannot really guarantee this.

COMPATIBILITY.md
wainersm: so add a table of retired TEEs?

COMPATIBILITY.md
```diff
@@ -0,0 +1,122 @@
...
+| AMD SEV-SNP        | v0.7.0+          | v0.7.0+  | x86_64       | Stable       |
+| AMD SEV(-ES)       | v0.7.0+          | v0.7.0+  | x86_64       | Stable       |
+| Azure TDX vTPM     | v0.9.0+          | v0.9.0+  | x86_64       | Stable       |
+| Azure SNP vTPM     | v0.9.0+          | v0.9.0+  | x86_64       | Stable       |
+| ARM CCA            | v0.10.0+         | v0.10.0+ | aarch64      | Experimental |
+| IBM Secure Exec    | v0.10.0+         | v0.10.0+ | s390x        | Stable       |
+| Hygon CSV          | v0.12.0+         | v0.12.0+ | x86_64       | Experimental |
+| NVIDIA GPU         | v0.15.0+         | v0.15.0+ | x86_64       | Experimental |
+
+### Feature Flags
```
wainersm, line 45: This Feature Flags seems more for the guest-component's main documentation than for this compatibility document. I would drop it.

COMPATIBILITY.md
```diff
@@ -0,0 +1,122 @@
...
+| Intel SGX          | (manual)       | `sgx-attester`        |
+| AMD SEV-SNP        | `snp`          | `snp-attester`        |
+| AMD SEV(-ES) + SNP | `amd`          | `snp-attester`        |
+| Azure TDX vTPM     | `az-tdx-vtpm`  | `az-tdx-vtpm-attester`|
+| Azure SNP vTPM     | `az-snp-vtpm`  | `az-snp-vtpm-attester`|
+| Azure CVM vTPM     | `az-cvm-vtpm`  | Both Azure attesters  |
+| ARM CCA            | `cca`          | `cca-attester`        |
+| IBM Secure Exec    | `se`           | `se-attester`         |
+
+## Component Dependencies
```
wainersm, line 61: Likewise. That section is more like an internal implementation detail of guest-components.

COMPATIBILITY.md
wainersm: I would drop that section from this doc as well.

wainersm:
> > > Something like this could make sense. I wonder if we will be able to keep it up-do-date and if this is the right place for it. Maybe in coco/coco or the release notes or the website would make more sense?
> > 
> > 
> > These would be my concerns and questions too. The release notes would force us to keep it regularly updated (or the release checklist could have an item to update this doc in its own file at least)
> 
> IMO that document should be moved to https://github.com/confidential-containers/confidential-containers/ , have a link on release notes and a item in the release checklist to remember us of getting it up-to-date.

Ah, I was about to forget: amazing document! thanks @bpradipt !

COMPATIBILITY.md
mkulke: see #1267 we probably want to have something similar at least for the TEEs that have some envelope over binary data.

PR #1267, opened 2026-01-14
attester: add versioned Evidence structs for azure

mkulke:
This is the introduction of versioned evidences payloads. This is crucial for making statements about compatibility between guest-components and trustee. We will probably have to have a mirror of those structs in trustee and perform deserialization by taking the version into account. Absence of a version string would imply v0, which is the schema that we had before this change.

- Bump az-*-vtpm crates to 0.8.0 (bincode is dropped, sev bumped to v7)
- Use URL-safe base64 encoding for all binary fields (report, vcek, quotes) in the json represenation. The ootb rendering defaults to a line for every entry in Vec<u8>, which is unusable
- Convert VCEK from PEM to DER format in az-snp-vtpm
- Create local TpmQuote wrapper with adjusted serialization

Affects az-snp-vtpm and az-tdx-vtpm attesters.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -350,6 +350,7 @@ dependencies = [
  "num-traits",
  "nvml-wrapper",
  "occlum_dcap",
+ "pem-rfc7468 1.0.0",
  "picky-asn1 0.10.1",
  "picky-asn1-der 0.5.4",
  "picky-asn1-x509 0.15.2",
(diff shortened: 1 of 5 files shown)


mkulke:
> I guess technically we should merge the Trustee side first?

yeah, this probably make sense.

PR #1271, opened 2026-01-16
Enable WASM Workloads with Encrypted Layer Support

rodneyosodo:
## What does the PR do?

This PR adds support for encrypted WebAssembly (WASM) workloads by implementing the encrypted WASM media type (`application/vnd.wasm.content.layer.v1+wasm+encrypted`) and enhancing the keyprovider service with full decryption capabilities. The implementation includes a new cryptographic module supporting AES-256-GCM and AES-256-CTR encryption algorithms, and completes the previously unimplemented UnWrapKey gRPC method that retrieves Key Encryption Keys (KEK) from KBS and performs layer key unwrapping.

The changes enable pulling encrypted WASM layers from OCI registries with proper stream processing that writes WASM binary modules directly to disk (as `module.wasm`) rather than unpacking them as tar archives. The image-rs components have been updated to recognize and handle encrypted WASM media types throughout the decryption, decoding, and pulling workflows, with support for SHA-256/SHA-512 digest verification.

## Reference
- https://github.com/absmach/propeller/pull/118

### DIFF
--- a/attestation-agent/coco_keyprovider/src/enc_mods/kbs.rs
+++ b/attestation-agent/coco_keyprovider/src/enc_mods/kbs.rs
@@ -4,11 +4,57 @@
 //
 
 use anyhow::*;
+use base64::Engine;
 use jwt_simple::prelude::{Claims, Duration, Ed25519KeyPair, EdDSAKeyPairLike};
+use kbs_protocol::{
+    evidence_provider::NativeEvidenceProvider, KbsClientBuilder, KbsClientCapabilities, ResourceUri,
+};
+use log::debug;
 use reqwest::Url;
 use tracing::debug;
 
 const KBS_URL_PATH_PREFIX: &str = "kbs/v0/resource";
+
+pub(crate) async fn get_kek(kbs_addr: &Url, kid: &str) -> Result<Vec<u8>> {
+    let kid = kid.strip_prefix('/').unwrap_or(kid);
+
+    // Construct the resource URI in the format: kbs:///<repository>/<type>/<tag>
+    let resource_uri_str = format!("kbs:///{}", kid);
+    let resource_uri: ResourceUri = resource_uri_str
+        .as_str()
+        .try_into()
+        .map_err(|e| anyhow!("Failed to parse resource URI: {}", e))?;
+
+    let evidence_provider = NativeEvidenceProvider::new()
+        .context("Failed to create evidence provider for attestation")?;
+    let mut kbs_client =
+        KbsClientBuilder::with_evidence_provider(Box::new(evidence_provider), kbs_addr.as_str())
+            .build()
+            .context("Failed to build KBS client")?;
+
+    let mut key = kbs_client.get_resource(resource_uri).await?;
+
+    // If the key is not 32 bytes (256-bit AES key), try base64 decoding
+    // Some KBS implementations return base64-encoded keys or text with newlines
+    if key.len() != 32 {
+        // First, try to interpret as UTF-8 string and trim whitespace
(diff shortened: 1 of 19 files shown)


fitzthum (commented):
Looks cool. 

Can you expand the commit message a little bit.

Also, are there any testing images available for this? I guess it might be tricky since it is encrypted, but ideally we would have coverage here.

Anyway, I will spin the CI. Welcome

rodneyosodo:
Hello @fitzthum 

I found out the issue might also be related to https://github.com/containers/skopeo

I have submitted these changes https://github.com/containers/container-libs/pull/596 thereafter I will upstream them to https://github.com/containers/skopeo/pull/2791. 

I noticed we are using a specific version of `skopeo` -> https://github.com/containers/skopeo/commit/f64a376. How would we adopt the new changes?

fitzthum:
Yeah as @Xynnn007 mentions, skopeo is not a direct dependency for us. It shows up a few places in the repo because we use it to make test images, but we don't have a particular version requirement.

To make sure that image-rs can unpack an image that corresponds to a particular version of skopeo (and whatever else), it's best to provide an image that we can try unpacking in a unit test.

image-rs/src/stream/mod.rs
```diff
@@ -93,6 +93,41 @@ pub async fn stream_processing(
...
+        .await
+        .map_err(|source| StreamError::FailedToRollBack { source })?;
+
+    match tokio::io::copy(&mut hash_reader, &mut wasm_file).await {
+        Ok(_) => Ok(hash_reader.hasher.digest_finalize()),
+        Err(e) => {
+            tokio::fs::remove_dir_all(destination).await.ok();
+            Err(StreamError::FailedToRollBack { source: e })
+        }
+    }
```
mkulke:
```suggestion
    let Ok(_) = tokio::io::copy(&mut hash_reader, &mut wasm_file).await else {
        tokio::fs::remove_dir_all(destination).await.ok();
        return Err(StreamError::FailedToRollBack { source: e });
    }
    Ok(hash_reader.hasher.digest_finalize())
```

attestation-agent/coco_keyprovider/src/enc_mods/kbs.rs
```diff
@@ -4,12 +4,57 @@
...
+    let mut key = kbs_client.get_resource(resource_uri).await?;
+
+    // If the key is not 32 bytes (256-bit AES key), try base64 decoding
+    // Some KBS implementations return base64-encoded keys or text with newlines
+    if key.len() != 32 {
+        // First, try to interpret as UTF-8 string and trim whitespace
+        let key_str = String::from_utf8_lossy(&key);
+        let trimmed = key_str.trim();
+        let engine = base64::engine::general_purpose::STANDARD;
+        let decoded = engine.decode(trimmed.as_bytes())?;
```
mkulke, line 44: this kind of ambiguity is suprising. which KBS implementations are we talking about? do we perform similar heuristics for other layer decryptions?

mkulke (commented):
thanks!

some documentation would be good, e.g. instructions on how to create an image with encrypted wasm layers.

there are also some semantics that would need to be explained (e.g. `scheme.starts_with(".provider")`) if those are newly introduced it would be good to provide a rationale in the PR description and cover the new behaviour in unit tests

This crate is a port of the ocicrypt golang project, is there similar logic/prior art in the golang ocicrypt library for encrypted wasm layer. We probably want to keep close to that.

rodneyosodo:
Close this in favour of https://github.com/confidential-containers/guest-components/pull/1629

PR #1272, opened 2026-01-16
Add signature support to sealed secrets

fitzthum:
See individual commit messages for information such as how we deal existing sealed secrets with fake signatures.

Will add docs to the PR next week. We will also need to update the website.

### DIFF
--- a/confidential-data-hub/docs/SEALED_SECRET.md
+++ b/confidential-data-hub/docs/SEALED_SECRET.md
@@ -10,7 +10,7 @@ in conjunction with an attestation.
 The Confidential Data Hub provides an API for unsealing secrets inside
 of a confidential guest.
 
-You can also use the secret cli tool to generate a sealed secret:
+You can use the secret cli tool to generate a sealed secret:
 
 ```bash
 cargo run -p confidential-data-hub --bin secret
@@ -140,57 +140,47 @@ BASE64URL(UTF8(JWS Protected Header)) || '.
 We can leverage the ["kid"](https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.4)
 field to specify the public key used to verify this signature.
 
-Signatures for secrets are not yet implemented, but sealed secrets are
-required to have a header and signature. A sealed secret should be of the form
-
-`sealed`.`JWS header`.`JWS body (secret content)`.`signature`
-
-Since signature verification is not yet supported, dummy values
-can be used for the header and signature. A sealed secret could look like this.
-
-```
-secret=sealed.fakejwsheader.ewogICAgInZlcnNpb24iOiAiMC4xLjAiLAogICAgInR5cGUiOiAidmF1bHQiLAogICAgIm5hbWUiOiAia2JzOi8vL2RlZmF1bHQvc2VhbGVkLXNlY3JldC90ZXN0IiwKICAgICJwcm92aWRlciI6ICJrYnMiLAogICAgInByb3ZpZGVyX3NldHRpbmdzIjoge30sCiAgICAiYW5ub3RhdGlvbnMiOiB7fQp9Cg.fakesignature
-```
+A public key, identified by a `kid`, must be provided to verify the secret.
+The public key can either be provisioned as a CDH credential or retrieved
+directly from Trustee. In the latter case the `kid` should be a resource URI.
 
 ## Usage in CoCo
 
-You can create a Kubernetes secret from a sealed secret
-and Kata will automatically expose it to your workload.
+The easiest way to create a sealed secret is with the secret CLI tool
 
-Start with a sealed secret such as
-```json
-{
-	"version": "0.1.0",
-	"type": "envelope",
-	"provider": "xxx",
-	"key_id": "xxx",
-	"encrypted_key": "ab27dc=",
-	"encrypted_data": "xxx",
-	"wrap_type": "A256GCM",
-	"iv": "xxx",
-	"provider_settings": {
-		...
-	},
-	"annotations": {
-		...
-	}
-}
-```
-
-Encode the payload in BASE64URL
-```
-ewoJInZlcnNpb24iOiAiMC4xLjAiLAoJInR5cGUiOiAiZW52ZWxvcGUiLAoJInByb3ZpZGVyIjogInh4eCIsCgkia2V5X2lkIjogInh4eCIsCgkiZW5jcnlwdGVkX2tleSI6ICJhYjI3ZGM9IiwgCgkiZW5jcnlwdGVkX2RhdGEiOiAieHh4IiwKCSJ3cmFwX3R5cGUiOiAiQTI1NkdDTSIsCgkiaXYiOiAieHh4IiwKCSJhbm5vdGF0aW9ucyI6IHsKCQkiY3J5cHRvX2NvbnRleHQiOiB7CgkJCSJhbGdvcml0aG0iOiAiQTI1NkdDTSIKCQl9LAoJCSJwcm92aWRlcl9zZXR0aW5nIjogewoJCQkia21zX2luc3RhbmNlX2lkIjogInh4eCIKCQl9Cgl9Cn0
-```
-Then add a prefix `sealed.` and JWS header and signature.
-
-```
-sealed.fakejwsheader.ewoJInZlcnNpb24iOiAiMC4xLjAiLAoJInR5cGUiOiAiZW52ZWxvcGUiLAoJInByb3ZpZGVyIjogInh4eCIsCgkia2V5X2lkIjogInh4eCIsCgkiZW5jcnlwdGVkX2tleSI6ICJhYjI3ZGM9IiwgCgkiZW5jcnlwdGVkX2RhdGEiOiAieHh4IiwKCSJ3cmFwX3R5cGUiOiAiQTI1NkdDTSIsCgkiaXYiOiAieHh4IiwKCSJhbm5vdGF0aW9ucyI6IHsKCQkiY3J5cHRvX2NvbnRleHQiOiB7CgkJCSJhbGdvcml0aG0iOiAiQTI1NkdDTSIKCQl9LAoJCSJwcm92aWRlcl9zZXR0aW5nIjogewoJCQkia21zX2luc3RhbmNlX2lkIjogInh4eCIKCQl9Cgl9Cn0.fakesignature
-```
-
-Create a Kubernetes secret
+You can create a vault secret with the following command.
 
 ```bash
-kubectl create secret generic sealed-secret --from-literal='secret=sealed.fakejwsheader.ewoJInZlcnNpb24i...'
+cargo run -p confidential-data-hub --bin secret seal \
+    --signing-kid kbs://signing/key/uri --signing-jwk-path ./path/to/jwk \
+    vault \
+    --resource-uri kbs://secret/key/uri --provider kbs
+```
+
+This will create a sealed secret that points to the resource `secret/key/uri` in Trustee.
+The secret will be validated with a public key stored in Trustee.
+The public/private keypair should be a JWK with a P256 EC key.
+
+This key should look like this.
+```json
+{
+    "kty": "EC",
+    "d": "_z0AOMG12p1lt...",
+    "use": "sig",
+    "crv": "P-256",
+    "kid": "test",
+    "x": "wjCZnuv_tLKiCt...",
+    "y": "gsSc2YE_O2kmHx...",
+    "alg": "ES256"
+}
+```
+This includes the public and private keypair. Only the public component
+needs to be provisioned to Trustee.
+
+### Create a Kubernetes secret
+
+```bash
+kubectl create secret generic sealed-secret --from-literal='secret=<sealed-secret>'
 ```
 
 Use this secret in a workload


confidential-data-hub/hub/src/config.rs
```diff
@@ -169,6 +179,10 @@ impl CdhConfig {
         if let Some(kbs_cert) = &self.kbc.kbs_cert {
             env::set_var("KBS_CERT", kbs_cert);
         }
+
+        if self.skip_sealed_secret_verification {
+            env::set_var("SKIP_SEALED_SECRET_VERIFICATION", "true");
```
mkulke, line 184: set_var is a bit [iffy](https://doc.rust-lang.org/std/env/fn.set_var.html#safety) on posix systems, do we need to set a global variable like this? I understand we did that elsewhere because some value had to propagated to child processes that we didn't want to modify, but if it's merely in other crates and we had to adjust them to respect that env, could this just be explicitly handled application state in Rust code? (idk "config::sealed_secret_verification()" or something)

confidential-data-hub/hub/src/config.rs
```diff
@@ -93,6 +102,7 @@ impl CdhConfig {
                     credentials: Vec::new(),
                     socket: DEFAULT_CDH_SOCKET_ADDR.into(),
                     image: ImageConfig::from_kernel_cmdline(),
+                    skip_sealed_secret_verification: false,
```
Xynnn007, line 105: If we by default set this to false, there would be some breaks for current existing sealed secrets. We should either claim this severe change, or set this to true.

confidential-data-hub/hub/src/config.rs
Xynnn007: we could extend this part to have another parameter

https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/hub/src/secret/mod.rs#L33

confidential-data-hub/hub/src/secret/mod.rs
```diff
@@ -73,31 +134,116 @@ impl Secret {
...
+            return Err(SecretError::ParseFailed("Invalid KID Key Path"));
+        }
+
+        let verification_key = tokio::fs::read(kid_path).await?;
+
+        Ok(verification_key)
+    }
+
+    fn validate_ps256(secret: String, verification_key: Vec<u8>, signature: Vec<u8>) -> Result<()> {
+        println!("validating");
```
Xynnn007: Seems like leftover. let's mark this as debug! or trace!. Or delete it

confidential-data-hub/hub/src/secret/mod.rs
```diff
@@ -73,31 +134,116 @@ impl Secret {
...
+        let jwk: Jwk = serde_json::from_slice(&verification_key)?;
+
+        println!("after parse key");
```
Xynnn007: ditto

confidential-data-hub/hub/src/secret/mod.rs
```diff
@@ -73,31 +134,116 @@ impl Secret {
...
+        let public_key: p256::PublicKey = match &jwk.key {
+            jose_jwk::Key::Ec(ec) if ec.crv == EcCurves::P256 => ec
+                .try_into()
+                .map_err(|_| SecretError::BadSigningKey("Could not parse verification key."))?,
+            _ => return Err(SecretError::BadSigningKey("Key must be P256")),
+        };
+
+        let verifying_key = VerifyingKey::from(&public_key);
+
+        println!("after making key");
```
Xynnn007: ditto

confidential-data-hub/hub/src/secret/mod.rs
```diff
@@ -73,31 +134,116 @@ impl Secret {
...
+        let sections: Vec<&str> = secret.split('.').collect();
+        let signing_input = format!("{}.{}", sections[0], sections[1]);
+
+        let sig = Signature::from_slice(&signature)?;
+        println!("after making sig");
+        println!("{}", signing_input);
```
Xynnn007: ditto

confidential-data-hub/hub/src/secret/mod.rs
```diff
@@ -73,31 +134,116 @@ impl Secret {
...
+        verifying_key.verify(signing_input.as_bytes(), &sig)?;
+
+        println!("after verifying sig");
```
Xynnn007: ditto

confidential-data-hub/hub/src/secret/mod.rs
```diff
@@ -50,21 +63,69 @@ impl Secret {
...
+                let alg = jws
+                    .signature
+                    .protected
+                    .as_ref()
+                    .and_then(|p| p.oth.alg)
+                    .ok_or(SecretError::ParseFailed("Could not get algorithm"))?;
+
+                let sig = jws.signature.signature.to_vec();
+
+                println!("before get key");
```
Xynnn007: ditto

confidential-data-hub/hub/src/secret/mod.rs
```diff
@@ -73,31 +134,116 @@ impl Secret {
...
     }
 
-    // TODO: add real signature generation
-    pub fn to_signed_base64_string(&self) -> Result<String> {
+    /// Given a key id, get the key.
+    /// If the key id is in the form of a resource URI, use the KMS.
+    /// Otherwise, look for a local credential (provisioned to the fs
+    /// at CDH startup).
+    async fn get_kid(kid: &String) -> Result<Vec<u8>> {
+        if kid.starts_with("kbs:/") {
```
Xynnn007: `kbs://`

confidential-data-hub/hub/src/secret/mod.rs
```diff
@@ -73,31 +134,116 @@ impl Secret {
...
+            let verification_key = kms::new_getter("kbs", ProviderSettings::default())
+                .await?
+                .get_secret(kid, &Annotations::default())
+                .await?;
+
+            return Ok(verification_key);
+        }
+
+        // Get key as local credential.
+        let base = fs::canonicalize(SIGNING_CREDENTIALS_PATH)?;
```
Xynnn007: let's add a more concrete error message here. The case is kid is neither a KBS URI nor a valid path

confidential-data-hub/hub/src/secret/mod.rs
```diff
@@ -73,31 +134,116 @@ impl Secret {
...
+        if kid.starts_with("kbs:/") {
+            let verification_key = kms::new_getter("kbs", ProviderSettings::default())
+                .await?
+                .get_secret(kid, &Annotations::default())
+                .await?;
+
+            return Ok(verification_key);
+        }
+
+        // Get key as local credential.
```
Xynnn007, line 150: btw this behavior is somehow not that safe. It can allow a malicious host to do read operation upon any place in the guest. Just points out this as I did not think of an successful attaking routine.

confidential-data-hub/hub/src/secret/mod.rs
Xynnn007: btw there is some overlapping with the logic https://github.com/confidential-containers/guest-components/blob/main/image-rs/src/resource/mod.rs#L73

probably we can bring them together into one place.

confidential-data-hub/hub/src/secret/mod.rs
```diff
@@ -73,31 +134,116 @@ impl Secret {
...
+            ..Default::default()
+        };
+
+        let header_json = serde_json::to_vec(&header)?;
+        let header_b64 = b64.encode(&header_json);
+        let payload_b64 = b64.encode(secret_json.as_bytes());
 
-        let secret_string = format!("sealed.fakejwsheader.{secret_base64}.fakesignature");
+        let signing_input = format!("{}.{}", header_b64, payload_b64);
+        //println!("{}", signing_input);
```
Xynnn007: ditto

confidential-data-hub/hub/src/secret/mod.rs
```diff
@@ -131,19 +277,31 @@ mod tests {
...
 
         let parsed: Secret = serde_json::from_str(secret_json).expect("deserialize failed");
         assert_eq!(parsed, secret_object);
 
+        let jwk = include_str!("./tests/test-key.json");
+        let jwk: Jwk = serde_json::from_str(&jwk).expect("Could not parse signing JWK");
+
+        let kid = "test-key".to_string();
+        let kid_cred_path = format!("{SIGNING_CREDENTIALS_PATH}/{}", &kid);
+        println!("{}", kid_cred_path);
```
Xynnn007: ditto

confidential-data-hub/hub/src/secret/mod.rs
```diff
@@ -131,19 +277,31 @@ mod tests {
...
+
+        std::fs::create_dir_all(SIGNING_CREDENTIALS_PATH).unwrap();
+        std::fs::write(kid_cred_path, serde_json::to_string(&jwk).unwrap()).unwrap();
```
Xynnn007, line 288: here is a writing operation. is this expected?

confidential-data-hub/hub/src/config.rs
fitzthum: I've never been a big fan of the env var approach, but this is sort of orthogonal to this PR. Maybe we can do a follow-up were we rework the configuration logic.

It is easy to extend the parameter one level up the call tree, but that's where things start to get a bit complicated. `secret::unseal_secret` is called in a few far-flung places. Instead, I kept the flag at a lower level (so we can use if for testing and the CLI), but had the `unseal_secret` method implement the config logic so the caller doesn't need to know about the configuration or change at all.

confidential-data-hub/hub/src/config.rs
fitzthum: Let's ask at the meeting on thursday. I would prefer to have a more secure default, but obviously it is a breaking change.

confidential-data-hub/hub/src/secret/mod.rs
fitzthum: whoops. well now you know it wasn't AI

confidential-data-hub/hub/src/secret/mod.rs
fitzthum: Yes, we have to provide the signing key for the test to work. This is not ideal, but I couldn't think of another way to do it without a lot of mock code.

confidential-data-hub/hub/src/secret/mod.rs
fitzthum: Yeah this is a very interesting point. I was thinking through this a bit. There are a few potential issues.

First, is there any way for an attacker to craft a signing key and payload that would leak information about the private key? Typically the public key and the payload are both public anyway, so I don't think this is a problem.

Second, can an attacker leak information about some other guest data by setting the kid to a different path. The code does block directory traversal, but you could still snoop on another credential. Maybe if you craft a certain payload you can leak the secret. In practice I think this would be very hard, but it's worth thinking about.

Probably we have a very similar concern about other code that uses credentials. Personally, I have never liked storing the credentials in the filesystem. I think we might consider re-working that as a follow-up.

confidential-data-hub/hub/src/secret/mod.rs
fitzthum: I was thinking about using that interface, but constructing a `ResourceProvider` requires you to know some config information. The KMS abstraction handles this (via the env var ofc). tbh I find all these abstractions to be a little unclear. We may want to do some general cleanup at some point.

Xynnn007: yes. we can do it later

confidential-data-hub/hub/src/secret/mod.rs
Xynnn007: ok. 
The files in the global path `/run` have been modified. Is it possible to implement this functionality using the `tempfile` crate, while minimizing the impact of testing on the environment?

confidential-data-hub/hub/src/secret/mod.rs
```diff
@@ -73,31 +133,110 @@ impl Secret {
         Ok(secret)
     }
 
-    // TODO: add real signature generation
-    pub fn to_signed_base64_string(&self) -> Result<String> {
+    /// Given a key id, get the key.
+    /// If the key id is in the form of a resource URI, use the KMS.
```
dcmiddle, line 137: /KMS/KBS

fitzthum: In guest components we have a KMS abstraction that wraps around the KBS. You can sort of see this on line 142. So I think either is correct here. Maybe can put both in the comment.

dcmiddle: I missed that. Yeah it's good either way.

confidential-data-hub/hub/src/bin/secret_cli.rs
```diff
@@ -41,6 +42,16 @@ struct SealArgs {
     /// Type of the Secret, i.e. `vault` or `envelope`
     #[command(subcommand)]
     r#type: TypeArgs,
+
+    /// The kid of the signing key. This will be used when unsealing
+    /// to find the key.
+    #[arg(short, long)]
+    signing_kid: String,
```
Xynnn007, line 49: Is this required? I mean, if the sealed secret is not to be signed, could we still generate the old life-style sealed secret?

confidential-data-hub/hub/src/bin/secret_cli.rs
```diff
@@ -41,6 +42,16 @@ struct SealArgs {
...
+
+    /// A path to a file containing a JWK (with a private component)
+    /// for signing the sealed secret.
+    #[arg(short, long)]
+    signing_jwk_path: String,
```
Xynnn007, line 54: The same question. Btw I must have missed the final conclusion about the compability of sealed secret. Do people agree on totally switch to signed sealed secret?

confidential-data-hub/hub/src/secret/tests/test-key.json
```diff
@@ -0,0 +1,10 @@
+{
+    "kty": "EC",
+    "d": "7baJLiTk68NcLgYPv05jJvElqoAhbzi1ZWx4OyILUtA",
+    "use": "sig",
+    "crv": "P-256",
+    "kid": "test-key",
```
Xynnn007, line 6: Could we have a kid that uses KBS URI for example?

confidential-data-hub/hub/src/bin/secret_cli.rs
fitzthum: We talked about it on the Trustee call and people said to go with the more secure option even if it breaks compatibility. I am planning to ask on this Thursday as well.

confidential-data-hub/hub/src/bin/secret_cli.rs
fitzthum: In this PR we only support unsigned secrets in unsealing. If you are sealing a secret, you must provide the signing info. We could relax this, but I think it's probably good to force people to use signatures when making any new secrets.

confidential-data-hub/hub/src/secret/tests/test-key.json
fitzthum: Unsealing that will require having some kind of KBS we can connect to. Do we have any tooling for a fake KMS backend?

Xynnn007: offline-fs-kbc can work, but needs some change upon /etc/ files, thus probably not good in unit test.

btw, looks like this file is only used for serde/de-serde test, so we might not need to care about the unsealing?

confidential-data-hub/hub/src/bin/secret_cli.rs
dcmiddle: Yep move on to more secure option.

confidential-data-hub/hub/src/secret/tests/test-key.json
fitzthum: Ok let me see about adding another case.

confidential-data-hub/hub/src/secret/mod.rs
```diff
@@ -73,31 +133,110 @@ impl Secret {
...
+                    "Key must be EC P256 private key",
+                ))
+            }
+        };
+
+        let signing_key = SigningKey::from(&signing_key);
+
+        let header = Protected {
+            oth: Unprotected {
+                alg: Some(Signing::Ps256),
```
dcmiddle: Should this be `Es256` for elliptic curve P-256?
https://docs.rs/jose-jwa/0.1.2/jose_jwa/enum.Signing.html#variant.Es256
https://docs.rs/jose-jwa/0.1.2/jose_jwa/enum.Signing.html#variant.Ps256

confidential-data-hub/hub/src/secret/mod.rs
```diff
@@ -73,31 +133,110 @@ impl Secret {
...
+        if !kid_path.starts_with(&base) {
+            return Err(SecretError::ParseFailed("Invalid KID Key Path"));
+        }
+
+        let verification_key = tokio::fs::read(kid_path).await?;
+
+        Ok(verification_key)
+    }
+
+    fn validate_ps256(secret: String, verification_key: Vec<u8>, signature: Vec<u8>) -> Result<()> {
```
dcmiddle: Assuming you want elliptic curve and not RSA 
```suggestion
    fn validate_es256(secret: String, verification_key: Vec<u8>, signature: Vec<u8>) -> Result<()> {
```

confidential-data-hub/hub/src/secret/mod.rs
```diff
@@ -50,21 +63,68 @@ impl Secret {
...
+                    .ok_or(SecretError::ParseFailed("Could not get algorithm"))?;
+
+                let sig = jws.signature.signature.to_vec();
+
+                let verification_key = Secret::get_kid(kid).await?;
+
+                // Try validating the secret with the key using whatever
+                // algorithm is specified in the JWS header.
+                match alg {
+                    Signing::Ps256 => Secret::validate_ps256(secret, verification_key, sig)?,
```
dcmiddle: Do you want Es256?

confidential-data-hub/hub/src/secret/mod.rs
fitzthum: Oh yeah switched this around halfway through. Names are fixed up now.

confidential-data-hub/hub/src/secret/mod.rs
fitzthum: The only issue there is that the filename must match what is in the secret header. There are some possible workarounds, but not sure if it is worth the complexity.

confidential-data-hub/hub/src/secret/tests/test-key.json
fitzthum: Ok, this test actually does both the serialization to/from json and to/from string. When we serialize to string we trigger the signature code. We could split this into two tests, but imo the serialization to/from json isn't particularly interesting anyway.

confidential-data-hub/hub/src/secret/mod.rs
fitzthum: I looked into this a bit more. I think the core problem here is that we have some fixed paths all across the repo. Besides leaving files behind, arguably the more significant problem is that you have to run a number of the tests as root. I think we should make these paths more configurable, then we can use a tempdir for the entire test scope. That's a bit beyond this PR, but I am looking into it.

PR #1293, opened 2026-01-29
Add basic Trustee plugin support to guest-components

fitzthum:
Following from #1208 and https://github.com/confidential-containers/trustee/issues/1158, let's make the handling of trustee plugins/endpoints in the guest-components.

Specifically, add a plugin parameter to `get_resource`. Use this to create the request to Trustee. This way it is explicit which plugin we are using.

Also, add support for plugins to the trustee-attester. Add an optional plugin argument that can specify a plugin.

This PR does not change the behavior of the CDH. Instead, we hardcode the CDH to always use the resource plugin. Again, this is much more explicit than what we do now. In the future we can extend the CDH to support other plugins via sealed-secret, ASR, and other interfaces.

### DIFF
--- a/attestation-agent/deps/resource_uri/src/lib.rs
+++ b/attestation-agent/deps/resource_uri/src/lib.rs
@@ -160,10 +160,10 @@ mod tests {
     #[rstest]
     #[case("kbs:///alice/cosign-key/213", "alice", "cosign-key", "213", None)]
     #[case(
-        "kbs:///plugin/plugname/resourcename?param1=value1&param2=value2",
-        "plugin",
-        "plugname",
-        "resourcename",
+        "kbs:///a/b/c?param1=value1&param2=value2",
+        "a",
+        "b",
+        "c",
         Some("param1=value1&param2=value2")
     )]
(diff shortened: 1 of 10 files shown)


Xynnn007:
As I once comment: I prefer a form that the plugin name is contained inside the URI, than another separate parameter.

This requires that the existing URI format remain valid; they are all resources. Note that when I designed KBS URI, I put a [KBS host/port field](https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/docs/KBS_URI.md) in the URI.

However, as we used it more, we gradually realized that the actual KBS connection was defined by a global configuration file, such as the CDH configuration file. Therefore, it's completely unnecessary to include this field in the URI. So I suggest updating the KBS URI document, replacing the host/port part with "plugin," and leaving it blank by default as "resource."

This ensures backward compatibility, while also simplifying user experience and code design, avoiding the use of multiple parameters.


This design style also exists in Alibaba Cloud OSS object storage—a global configuration file specifies the OSS address, and the OSS URI (oss://x/x/x) includes both the bucket and the path.

fitzthum:
Superseded by #1468

PR #1308, opened 2026-02-04
ASR: add /info API

### LINKED ISSUE #1307: Introduce CDH/AA/ASR Info endpoint
It might be handy if we exposed an info endpoint to the workload. This could tell a workload what tee platform was running, the version of coco, etc.

Xynnn007:
The API calling will look like
```
$ curl 127.0.0.1:8006/aa/tee-type
sample
$ curl 127.0.0.1:8006/version
Guest Components Version: v0.17.0-33-g7d9635fb-dirty
```

Close #1307

### DIFF
--- a/api-server-rest/openapi/api.json
+++ b/api-server-rest/openapi/api.json
@@ -168,6 +168,39 @@
           }
         }
       }
+    },
+    "/info": {
+      "get": {
+        "tags": [],
+        "operationId": "_version",
+        "responses": {
+          "200": {
+            "description": "success response",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/VersionInfo"
+                },
+                "example": {
+                  "additional_tees": [],
+                  "feature": "resource",
+                  "tee": "tdx",
+                  "version": "v1.0.0"
+                }
+              }
+            }
+          },
+          "403": {
+            "description": "forbid external access"
+          },
+          "404": {
+            "description": "resource not found"
+          },
+          "405": {
+            "description": "only Get method allowed"
+          }
+        }
+      }
     }
   },
   "components": {
@@ -191,6 +224,38 @@
           "operation": {
             "type": "string",
             "description": "Attestation Agent Event Log Operation"
+          }
+        }
+      },
+      "VersionInfo": {
+        "type": "object",
+        "required": [
+          "version",
+          "additional_tees",
+          "feature"
+        ],
+        "properties": {
+          "additional_tees": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "description": "Additional TEE types"
+          },
+          "feature": {
+            "type": "string",
+            "description": "User provided feature command line argument for the API server"
+          },
+          "tee": {
+            "type": [
+              "string",
+              "null"
+            ],
+            "description": "TEE type (optional)"
+          },
+          "version": {
+            "type": "string",
+            "description": "Version string"
           }
         }
       }
--- a/api-server-rest/src/main.rs
+++ b/api-server-rest/src/main.rs
@@ -34,9 +34,11 @@ const CDH_ADDR: &str = "unix:///run/confidential-containers/cdh.sock";
 const AA_ADDR: &str =
     "unix:///run/confidential-containers/attestation-agent/attestation-agent.sock";
 
+const VERSION: &str = include_str!(concat!(env!("OUT_DIR"), "/guest_components_version"));
+
 /// API Server arguments info.
 #[derive(Parser, Debug)]
-#[command(author, about, long_about = None)]
+#[command(author, version = Some(VERSION), about, long_about = None)]
 struct Args {
     /// Bind address for API Server
     #[arg(default_value_t = DEFAULT_BIND.to_string(), short, long = "bind")]
--- a/api-server-rest/src/router.rs
+++ b/api-server-rest/src/router.rs
@@ -16,20 +16,26 @@ use crate::client::{
     cdh::{CDHClient, CDH_RESOURCE_URL, CDH_ROOT},
 };
 use crate::utils::split_nth_slash;
+use crate::VERSION;
 
 pub struct Router {
     aa_client: Option<AAClient>,
     cdh_client: Option<CDHClient>,
+    version: String,
+    feature: String,
 }
 
 impl Router {
     pub fn new(
         aa_client: Option<AAClient>,
         cdh_client: Option<CDHClient>,
+        feature: String,
     ) -> Self {
         Self {
             aa_client,
             cdh_client,
+            version: VERSION.trim_ascii_end().to_string(),
+            feature,
         }
     }
 
@@ -101,6 +107,40 @@ impl Router {
             .query()
             .map(|v| form_urlencoded::parse(v.as_bytes()).into_owned().collect())
             .unwrap_or_default();
+        // First, handle the version request
+        if path == "/info" {
+            #[derive(Serialize)]
+            struct VersionInfo {
+                version: String,
+                #[serde(skip_serializing_if = "Option::is_none")]
+                tee: Option<String>,
+                #[serde(skip_serializing_if = "Vec::is_empty")]
+                additional_tees: Vec<String>,
+                feature: String,
+            }
+
+            if method != Method::GET {
+                return self.not_allowed();
+            }
+            let (tee, additional_tees) = match &self.aa_client {
+                Some(client) => {
+                    let tee = client.get_tee_type().await?;
+                    let additional_tees = client.get_additional_tees().await?;
+                    (Some(tee), additional_tees)
+                }
+                None => (None, vec![]),
+            };
+            let version_info = VersionInfo {
+                version: self.version.clone(),
+                tee,
+                additional_tees,
+                feature: self.feature.clone(),
+            };
+            let version_info = serde_json::to_string(&version_info)?;
+            return self.json_response(version_info);
+        }
+
+        // Then, handle the other requests
         if let Some((root_path, url_path)) = split_nth_slash(path, 2) {
             debug!("root_path {root_path}, url_path {url_path}");
             match root_path {
--- a/api-server-rest/README.md
+++ b/api-server-rest/README.md
@@ -16,6 +16,9 @@ $ curl http://127.0.0.1:8006/aa/evidence\?runtime_data\=xxxx
 $ curl http://127.0.0.1:8006/aa/token\?token_type\=kbs
 {"token":"eyJhbGciOiJFi...","tee_keypair":"-----BEGIN... "}
 
+$ curl http://127.0.0.1:8006/info
+{"version":"v1.0.0","tee":"tdx","additional_tees":[]}
+
 $ curl -X POST http://127.0.0.1:8006/aa/aael \
      -H "Content-Type: application/json" \
      -d '{"domain":"test","operation":"test","content":"test"}'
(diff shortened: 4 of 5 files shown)


api-server-rest/src/aa.rs
```diff
@@ -170,4 +175,15 @@ impl AAClient {
...
         Ok(())
     }
+
+    pub async fn get_tee_type(&self) -> Result<String> {
+        let req = GetTeeTypeRequest {
+            ..Default::default()
+        };
+        let res = self
+            .client
+            .get_tee_type(ttrpc::context::with_timeout(TTRPC_TIMEOUT), &req)
```
fitzthum: I guess this is always getting the primary tee type. Maybe we should change the name or also try to include if there are any additional tees/devices.

fitzthum (commented):
This is great. One comment about the Tee Type

bpradipt:
> The API calling will look like
> 
> ```
> $ curl 127.0.0.1:8006/aa/tee-type
> sample
> $ curl 127.0.0.1:8006/version
> Guest Components Version: v0.17.0-33-g7d9635fb-dirty
> ```

Has there been some thought on returning metadata using a single api ?
eg `curl 127.0.0.1:8006/metadata` which returns primary and additional TEEs, GC version and other info..

Xynnn007:
@fitzthum @bpradipt adds `/version` api to get both primary tee and additional tees.

attestation-agent/attestation-agent/src/bin/grpc-aa/server.rs
```diff
@@ -180,6 +180,32 @@ impl AttestationAgentService for AA {
 
         Result::Ok(Response::new(reply))
     }
+
+    async fn get_additional_tee(
```
fitzthum: Should probably be `get_additional_tees`.

api-server-rest/src/main.rs
```diff
@@ -61,40 +63,19 @@ async fn main() -> Result<()> {
...
-            );
-        }
-
+    let (aa_client, cdh_client) = match args.features.as_str() {
+        "resource" => (None, Some(CDHClient::new(&args.cdh_addr).await?)),
+        "attestation" => (Some(AAClient::new(&args.aa_addr).await?), None),
+        "all" => (
+            Some(AAClient::new(&args.aa_addr).await?),
+            Some(CDHClient::new(&args.cdh_addr).await?),
+        ),
```
fitzthum, line 84: We should probably make this configuration more granular at some point, but that can be a separate PR. For now, the resource option will need to be enabled to use the `version` endpoint?

api-server-rest/openapi/api.json
```diff
@@ -193,6 +225,33 @@
...
+              "type": "string"
+            },
+            "description": "Additional TEE types"
+          },
+          "tee": {
+            "type": [
+              "string",
+              "null"
+            ],
+            "description": "TEE type (optional)"
```
fitzthum, line 254: is this really optional?

api-server-rest/openapi/api.json
```diff
@@ -168,6 +168,38 @@
           }
         }
       }
+    },
+    "/version": {
```
fitzthum: If we're just going to have one endpoint I think `info` might be slightly better than `version`, but either are fine.

api-server-rest/src/main.rs
Xynnn007: The version can always be used. If attestation option is enabled, the tee and additional tees will be showed. If not, only a GC version will be showed

api-server-rest/openapi/api.json
Xynnn007: info seems better. Let me do a change

api-server-rest/openapi/api.json
Xynnn007: Yes. This will be showed only if `attestation` or `all` feature is enabled.

api-server-rest/src/main.rs
fitzthum: Ok, btw can we add the feature to the info response? i.e. explicitly say whether attestation is enabled.

api-server-rest/openapi/api.json
fitzthum: You know we could also have both `version` and `info` and have them return the same thing. Up to you.

api-server-rest/src/main.rs
Xynnn007: Good idea

api-server-rest/openapi/api.json
Xynnn007: Yup. Only one `info` makes more sense

bpradipt: There is no `/info` path described in the spec, but it's used later in the example and router.rs

api-server-rest/src/router.rs
```diff
@@ -4,114 +4,236 @@
...
+
+        let path = req.uri().path();
+        let method = req.method();
+        let params: HashMap<String, String> = req
+            .uri()
+            .query()
+            .map(|v| form_urlencoded::parse(v.as_bytes()).into_owned().collect())
+            .unwrap_or_default();
+        // First, handle the version request
+        if path == "/info" {
```
bpradipt, line 111: should this be `/version` ?

api-server-rest/openapi/api.json
Xynnn007: Oh. Let me update the api json file. We now have only `/info` api. No `/version`

api-server-rest/src/router.rs
Xynnn007: see https://github.com/confidential-containers/guest-components/pull/1308#discussion_r2867246334

api-server-rest/src/router.rs
```diff
@@ -4,114 +4,236 @@
...
+                }
+
+                CDH_ROOT => {
+                    let Some(client) = &self.cdh_client else {
+                        return Ok(Response::builder()
+                            .status(404)
+                            .body(Body::from("Resource Feature Not Enabled"))?);
+                    };
+                    match (url_path, method) {
+                        (CDH_RESOURCE_URL, &Method::GET) => {
```
bpradipt: For an example cdh resource `/cdh/resource/default/key`, we'll have url_path as `/resource/default/key` and the match will be performed with `/resource/default/key` which will fail.
Am I missing something?

Xynnn007: Oh good catch. Thanks for this and I fixed it.

api-server-rest/src/router.rs
```diff
@@ -4,114 +4,238 @@
...
+                                    std::result::Result::Ok(results) => {
+                                        return self.octet_stream_response(results)
+                                    }
+                                    Err(e) => return self.internal_error(e.to_string()),
+                                }
+                            }
+                            None => return self.bad_request(),
+                        },
+                        (AA_AAEL_URL, &Method::POST) => {
+                            println!("Extend AAEL entry");
```
bpradipt: nit: Do you want to replace println with debug ?

attestation-agent/attestation-agent/src/bin/ttrpc_dep/server.rs
```diff
@@ -169,4 +170,33 @@ impl AttestationAgentService for AA {
...
+    ) -> ::ttrpc::Result<GetAdditionalTeesResponse> {
+        debug!("AA (ttrpc): get additional tees ...");
+
+        let additional_tee = self.inner.get_additional_tees();
+
+        let mut res = GetAdditionalTeesResponse::new();
+        for tee in additional_tee {
+            let tee = serde_json::to_string(&tee)
+                .map_err(|e| {
+                    error!("AA (ttrpc): get tee type failed:\n {e:?}");
```
bpradipt: nit: s/get tee type failed/get additional tee type failed/

api-server-rest/src/router.rs
```diff
@@ -4,114 +4,238 @@
...
+                            },
+                            _ => {
+                                return self.not_found();
+                            }
+                        }
+                    }
+
+                    return self.not_found();
+                }
+                _ => return self.not_allowed(),
```
bpradipt: This should also be not_found (404) right since it's an unknown root ?

api-server-rest/openapi/api.json
```diff
@@ -168,6 +168,39 @@
           }
         }
       }
+    },
+    "/info": {
```
bpradipt, line 172: nit: the PR heading and commit should be changed to reflect the new API (ie. /info)

bpradipt:
@Xynnn007 added some minor comments. Rest LGTM

mythi:
FYI: I re-titled the PR to match with the implementation.

api-server-rest/openapi/api.json
mythi: just saw this. I did this already

api-server-rest/openapi/api.json
```diff
@@ -193,6 +226,38 @@
...
+          "additional_tees": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "description": "Additional TEE types"
+          },
+          "feature": {
+            "type": "string",
+            "description": "Feature string when launching the API server"
```
mythi: what is this feature string?

api-server-rest/src/client/aa.rs
```diff
@@ -0,0 +1,119 @@
...
+}
+
+#[derive(Deserialize)]
+pub struct AaelEvent {
+    pub domain: String,
+    pub operation: String,
+    pub content: String,
+}
+
+impl AAClient {
```
mythi, line 35: my preference would be to split the client refactoring to a commit of its own and then have that `/info` addition added on top.

api-server-rest/src/router.rs
Xynnn007: Let me use 404. Original code uses "not allowed".

api-server-rest/openapi/api.json
Xynnn007: It's the launch config for ASR. Defines whether AA/CDH or the two's APIs are exposed. See https://github.com/confidential-containers/guest-components/blob/main/api-server-rest/src/main.rs#L46-L47

mythi: Are users expected to query `/info` features first to see what other endpoints are available?

Xynnn007: Depends on users. This only provides such ability if the user does know

fitzthum: Typically the user will have set the features via an annotation (kata has some piping for this), but putting this info here is handy. Having the wrong features is a somewhat common failure point.

mythi: I have no strong opinions, just wondering what `features` is all about in the first place. I guess the alternative could have been to return 40x with some error code text in the headers saying the requested API is not available since the feature is not turned on.

fitzthum: Today the error you get from an unsupported endpoint is generic. We should look into making it more clear in another PR. Still, nice to have this here. There are some other refinements we might want with the features as well. They aren't quite granular enough currently. Also, the term feature isn't ideal. It's easy to confuse that with the rust features. This is beyond this PR tho.

Xynnn007:
@mythi please check if the PR is good. Feel free to get it merged : )

api-server-rest/openapi/api.json
mythi: if `feature` is what user knows in advance, what's the case for having it in `/info`? Would it be better to keep this api focused on what needs to be pulled from the environment during runtime.

mythi:
```suggestion
            "description": "User provided feature command line argument for the API server"
```

PR #1313, opened 2026-02-11
cdh: store ephemeral LUKS headers in tmpfs

manuelh-dev:
**Outline:**
When creating ephemeral storages, place the LUKS headers in guest memory rather than onto the storage device itself. For this, create LUKS header files in the CDH storage directory, and call into `CryptInit::init_with_data_device` instead of into `CryptInit::init`. This aims to mitigate [Vulnerabilities in LUKS2 disk encryption for confidential VMs](https://blog.trailofbits.com/2025/10/30/vulnerabilities-in-luks2-disk-encryption-for-confidential-vms/) for ephemeral storages.

The path where existing, encrypted storages are utilized remains untouched. We should evaluate switch to utilizing different encryption schemes, such as using ZFS, or VeraCrypt, instead.

**Testing:**
I tested this end-to-end flow against kata-containers for which I created this draft PR for integration of the new CDH bits: https://github.com/kata-containers/kata-containers/pull/12459
In my tests, I was using the [trusted image store functionality](https://github.com/kata-containers/kata-containers/blob/main/docs/how-to/how-to-pull-images-in-guest-with-kata.md) in combination with the `kata-qemu-gpu-snp` handler.

I have also onboarded the luks2 tests to CI, and CI tests pass.

**Outlook:**
The ongoing work stream to support [ephemeral container storage](https://github.com/kata-containers/kata-containers/pull/10559/files) in Kata Containers should benefit from this PR and from the integration branch as well once the guest components are updated in Kata Containers.

**Output:**
In the guest:
```
root@26a7f28fc681:/# ls -lha /run/confidential-containers/cdh/luks-headers/
total 16M
drwxr-xr-x 2 root root  60 Feb 11 19:17 .
drwxr-xr-x 3 root root  60 Feb 11 19:17 ..
-rw-r--r-- 1 root root 16M Feb 11 19:17 ODow
< 160 /run/confidential-containers/cdh/luks-headers/ODow | head -n 10
000000 4c 55 4b 53 ba be 00 02 00 00 00 00 00 00 40 00
000010 00 00 00 00 00 00 00 03 00 00 00 00 00 00 00 00
000020 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
*
000040 00 00 00 00 00 00 00 00 73 68 61 32 35 36 00 00
000050 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
000060 00 00 00 00 00 00 00 00 3d 7b bb b5 fd eb 88 c6
000070 98 53 9b 77 84 ad a5 98 4e 73 bb fe 77 2e 6a b6
000080 3d e3 88 5a 6d 0d 5d 68 77 a0 5c 51 db bb 88 79
000090 bc 09 89 d6 c6 e3 54 ec 82 73 0b f1 91 d9 e8 40
```
`4c 55 4b 53` = LUKS magic
`ba be 00 02` = LUKS2
sha256 appears in the header metadata

On the host:
```
$ od -Ax -tx1 -N 64 /tmp/trusted-image-storage.img
000000 69 6e 74 65 67 72 74 00 04 0f 20 00 f0 01 00 00
000010 78 67 0c 06 00 00 00 00 08 00 00 00 03 0c 00 00
000020 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
*
000040
```
`69 6e 74 65 67 72 74 00` is the dm-integrity magic, which is ASCII integrt\0.

Log from CDH:
```
[2026-02-11T19:17:16Z INFO  ttrpc_cdh] [ttRPC] Confidential Data Hub starts to listen to request: unix:///run/confidential-containers/cdh.sock
[2026-02-11T19:17:17Z DEBUG ttrpc_cdh::ttrpc_server] [ttRPC CDH] get new secure mount request
[2026-02-11T19:17:17Z INFO  confidential_data_hub::hub] secure mount called
[2026-02-11T19:17:17Z DEBUG confidential_data_hub::storage::volume_type::blockdevice] generate a random key. All data on the device will be overwritten.
[2026-02-11T19:17:17Z WARN  confidential_data_hub::storage::volume_type::blockdevice] encrypting the device. This will wipe original data on the disk.
[2026-02-11T19:17:29Z DEBUG confidential_data_hub::storage::volume_type::blockdevice] No mapper name provided, generating a random one
[2026-02-11T19:17:29Z DEBUG confidential_data_hub::storage::volume_type::blockdevice] luks2 opening device: /dev/sda
[2026-02-11T19:17:29Z DEBUG confidential_data_hub::storage::drivers::luks2] activating device: 628918be-bb49-4f52-8ace-8df7ea05a8db
[2026-02-11T19:17:31Z DEBUG confidential_data_hub::storage::drivers::luks2] device activated: 628918be-bb49-4f52-8ace-8df7ea05a8db
[2026-02-11T19:17:31Z INFO  confidential_data_hub::storage::volume_type::blockdevice] formatting device: /dev/mapper/628918be-bb49-4f52-8ace-8df7ea05a8db and mounting it to mount point: /run/kata-containers/image/
[2026-02-11T19:17:31Z DEBUG confidential_data_hub::storage::volume_type::blockdevice] formatting device: /dev/mapper/628918be-bb49-4f52-8ace-8df7ea05a8db
[2026-02-11T19:17:31Z DEBUG confidential_data_hub::storage::volume_type::blockdevice] mounting device: /dev/mapper/628918be-bb49-4f52-8ace-8df7ea05a8db
[2026-02-11T19:17:31Z INFO  confidential_data_hub::storage::volume_type::blockdevice] Target path /run/kata-containers/image/ mounted successfully
[2026-02-11T19:17:31Z DEBUG ttrpc_cdh::ttrpc_server] [ttRPC CDH] secure mount succeeded.
[2026-02-11T19:17:31Z DEBUG ttrpc_cdh::ttrpc_server] [ttRPC CDH] get new image pull request
[2026-02-11T19:17:31Z DEBUG confidential_data_hub::hub] Image client lazy initializing...
[2026-02-11T19:17:31Z INFO  image_rs::resource::kbs] secure channel uses ttrpc
[2026-02-11T19:17:31Z WARN  image_rs::builder] No `image_security_policy` given, thus all images can be pulled by the image client without filtering.
[2026-02-11T19:17:31Z INFO  image_rs::builder] Initialize new metadata.
[2026-02-11T19:17:31Z INFO  image_rs::builder] Image work directory: "/run/kata-containers/image/"
```

### DIFF
--- a/confidential-data-hub/hub/src/auth/kbs.rs
+++ b/confidential-data-hub/hub/src/auth/kbs.rs
@@ -8,15 +8,16 @@
 
 use std::path::PathBuf;
 
+use const_format::concatcp;
 use kms::{plugins::kbs::KbcClient, Annotations, Getter};
 use tokio::fs;
 use tracing::debug;
 
-use crate::{hub::Hub, Error, Result};
+use crate::{hub::Hub, hub::CDH_BASE_DIR, Error, Result};
 
-/// This directory is used to store all the kbs resources get by CDH's init
-/// function, s.t. `[[Credential]]` sections in the config.toml file.
-pub const KBS_RESOURCE_STORAGE_DIR: &str = "/run/confidential-containers/cdh";
+/// Directory for KBS resources (credentials, etc.) fetched at init, see the `[[credentials]]`
+/// sections in CDH's `config.toml` file, for an example, see `example.config.toml`.
+const KBS_RESOURCE_STORAGE_DIR: &str = concatcp!(CDH_BASE_DIR, "/kbs");
 
 impl Hub {
     pub(crate) async fn init_kbs_resources(&self) -> Result<()> {
@@ -86,11 +87,11 @@ fn is_path_valid(path: &str) -> bool {
 mod tests {
     use rstest::rstest;
 
-    use crate::auth::kbs::{is_path_valid, KBS_RESOURCE_STORAGE_DIR};
+    use super::{is_path_valid, KBS_RESOURCE_STORAGE_DIR};
 
     #[rstest]
     #[case("/etc/config.toml".into(), false)]
-    #[case(format!("{KBS_RESOURCE_STORAGE_DIR}/../../config.toml"), false)]
+    #[case(format!("{KBS_RESOURCE_STORAGE_DIR}/../../config.toml",), false)]
     #[case(format!("{KBS_RESOURCE_STORAGE_DIR}/kms-credential/../../../config.toml"), false)]
     #[case(format!("{KBS_RESOURCE_STORAGE_DIR}/kms-credential/./config.toml"), false)]
     #[case(format!("{KBS_RESOURCE_STORAGE_DIR}/kms-credential/aliyun/config.toml"), true)]
--- a/confidential-data-hub/hub/src/config.rs
+++ b/confidential-data-hub/hub/src/config.rs
@@ -142,7 +142,7 @@ impl CdhConfig {
     /// ```
     ///
     /// It is supposed that all the `target path` should be with prefix
-    /// `/run/confidential-containers/cdh` or it will be treated as dangerous
+    /// `/run/confidential-containers/cdh/kbs` or it will be treated as dangerous
     /// path.
     ///
     /// TODO: delete this way after initdata mechanism could cover CDH's launch config.
--- a/confidential-data-hub/hub/src/storage/drivers/luks2.rs
+++ b/confidential-data-hub/hub/src/storage/drivers/luks2.rs
@@ -13,6 +13,10 @@
 use std::path::Path;
 
 use anyhow::Context;
+use base64::{engine::general_purpose::URL_SAFE_NO_PAD as b64, Engine};
+use const_format::concatcp;
+
+use crate::hub::CDH_BASE_DIR;
 use libcryptsetup_rs::consts::flags::{CryptActivate, CryptDeactivate, CryptVolumeKey};
 use libcryptsetup_rs::consts::vals::EncryptionFormat;
 use libcryptsetup_rs::{CryptInit, CryptParamsLuks2, CryptParamsLuks2Ref};
@@ -30,6 +34,34 @@ const LUKS2_VOLUME_KEY_SIZE_BIT_WITHOUT_INTEGRITY: usize = 256;
 
 const SECTOR_SIZE: u32 = 4096;
 
+pub const LUKS_HEADERS_STORAGE_DIR: &str = concatcp!(CDH_BASE_DIR, "/luks-headers");
+pub const LUKS_HEADER_FILE_SUFFIX: &str = ".header";
+pub const LUKS2_HEADER_MIN_SIZE_BYTES: u64 = 16 * 1024 * 1024;
+
+/// Returns the path where the detached LUKS header for the given device is stored.
+pub fn luks_header_path(device_path: &str) -> String {
+    let name = b64.encode(device_path.as_bytes());
+    format!(
+        "{}/{}{}",
+        LUKS_HEADERS_STORAGE_DIR, name, LUKS_HEADER_FILE_SUFFIX
+    )
+}
+
+/// Creates and sizes the LUKS header file at `header_path`.
+pub fn prepare_luks_header_file(header_path: &str) -> std::io::Result<()> {
+    if let Some(parent) = Path::new(header_path).parent() {
+        std::fs::create_dir_all(parent)?;
+    }
+    // error "LUKS header file not found: <path/to/header>" from libcryptsetup if header file doesn't exist.
+    let file = std::fs::OpenOptions::new()
+        .create_new(true)
+        .write(true)
+        .open(header_path)?;
+    // error "Device ... is too small" / OS error 5" from libcryptsetup if header isn't sized.
+    file.set_len(LUKS2_HEADER_MIN_SIZE_BYTES)?;
+    Ok(())
+}
+
 #[derive(Default)]
 pub struct Luks2Formatter {
     pub integrity: bool,
@@ -44,10 +76,10 @@ impl Luks2Formatter {
     pub fn encrypt_device(
         &self,
         device_path: &str,
+        header_path: Option<&str>,
         passphrase: Zeroizing<Vec<u8>>,
     ) -> anyhow::Result<()> {
-        let path = Path::new(device_path);
-        let mut device = CryptInit::init(path)?;
+        let mut device = init_device(device_path, header_path)?;
         let mut volume_key_length = LUKS2_VOLUME_KEY_SIZE_BIT_WITHOUT_INTEGRITY / 8;
         let mut params = CryptParamsLuks2 {
             pbkdf: None,
@@ -82,11 +114,11 @@ impl Luks2Formatter {
     pub fn open_device(
         &self,
         device_path: &str,
+        header_path: Option<&str>,
         name: &str,
         passphrase: Zeroizing<Vec<u8>>,
     ) -> anyhow::Result<()> {
-        let path = Path::new(device_path);
-        let mut device = CryptInit::init(path)?;
+        let mut device = init_device(device_path, header_path)?;
 
         let mut params = CryptParamsLuks2 {
             pbkdf: None,
@@ -130,6 +162,19 @@ impl Luks2Formatter {
     }
 }
 
+fn init_device(
+    device_path: &str,
+    header_path: Option<&str>,
+) -> anyhow::Result<libcryptsetup_rs::CryptDevice> {
+    let device_path = Path::new(device_path);
+    let device_paths = match header_path {
+        Some(header_path) => libcryptsetup_rs::Either::Right((Path::new(header_path), device_path)),
+        None => libcryptsetup_rs::Either::Left(device_path),
+    };
+
+    Ok(CryptInit::init_with_data_device(device_paths)?)
+}
+
 #[cfg(test)]
 mod tests {
     use std::io::Write;
@@ -137,10 +182,19 @@ mod tests {
     use serial_test::serial;
     use zeroize::Zeroizing;
 
-    use crate::storage::drivers::luks2::Luks2Formatter;
+    use super::{luks_header_path, prepare_luks_header_file, Luks2Formatter};
 
     const TEST_PASSPHRASE: &[u8] = b"test";
     const NAME: &str = "test";
+
+    /// Removes the LUKS header file on drop so tests don't leave files behind on panic.
+    /// TODO: Similarly, clean dm-crypt devices when a panic occurs.
+    struct RemoveHeaderOnDrop(String);
+    impl Drop for RemoveHeaderOnDrop {
+        fn drop(&mut self) {
+            let _ = std::fs::remove_file(&self.0);
+        }
+    }
 
     #[test]
     #[serial]
@@ -156,10 +210,12 @@ mod tests {
         let passphrase = Zeroizing::new(TEST_PASSPHRASE.to_vec());
         let luks2_formatter = Luks2Formatter { integrity: false };
         luks2_formatter
-            .encrypt_device(path, passphrase.clone())
+            .encrypt_device(path, None, passphrase.clone())
             .unwrap();
 
-        luks2_formatter.open_device(path, NAME, passphrase).unwrap();
+        luks2_formatter
+            .open_device(path, None, NAME, passphrase)
+            .unwrap();
 
         luks2_formatter.close_device(NAME).unwrap();
     }
@@ -178,12 +234,135 @@ mod tests {
         let passphrase = Zeroizing::new(TEST_PASSPHRASE.to_vec());
         let luks2_formatter = Luks2Formatter { integrity: true };
         luks2_formatter
-            .encrypt_device(path, passphrase.clone())
+            .encrypt_device(path, None, passphrase.clone())
             .unwrap();
 
-        luks2_formatter.open_device(path, NAME, passphrase).unwrap();
+        luks2_formatter
+            .open_device(path, None, NAME, passphrase)
+            .unwrap();
 
         luks2_formatter.close_device(NAME).unwrap();
+    }
+
+    #[test]
+    #[serial]
+    fn encrypt_open_device_no_integrity_with_header() {
+        let mut bin_file = tempfile::NamedTempFile::new().unwrap();
+        bin_file
+            .as_file_mut()
+            .write_all(&vec![0; 20 * 1024 * 1024])
+            .unwrap();
+        let path = bin_file.path().to_str().unwrap();
+        let header_path = luks_header_path(path);
+        prepare_luks_header_file(&header_path).unwrap();
+        let _guard = RemoveHeaderOnDrop(header_path.clone());
+
+        let passphrase = Zeroizing::new(TEST_PASSPHRASE.to_vec());
+        let luks2_formatter = Luks2Formatter { integrity: false };
+        luks2_formatter
+            .encrypt_device(path, Some(&header_path), passphrase.clone())
+            .unwrap();
+
+        luks2_formatter
+            .open_device(path, Some(&header_path), NAME, passphrase)
+            .unwrap();
+
+        luks2_formatter.close_device(NAME).unwrap();
+    }
+
+    #[test]
+    #[serial]
+    fn encrypt_open_device_integrity_with_header() {
+        let mut bin_file = tempfile::NamedTempFile::new().unwrap();
+        bin_file
+            .as_file_mut()
+            .write_all(&vec![0; 20 * 1024 * 1024])
+            .unwrap();
+        let path = bin_file.path().to_str().unwrap();
+        let header_path = luks_header_path(path);
+        prepare_luks_header_file(&header_path).unwrap();
+        let _guard = RemoveHeaderOnDrop(header_path.clone());
+
+        let passphrase = Zeroizing::new(TEST_PASSPHRASE.to_vec());
+        let luks2_formatter = Luks2Formatter { integrity: true };
+        luks2_formatter
+            .encrypt_device(path, Some(&header_path), passphrase.clone())
+            .unwrap();
+
+        luks2_formatter
+            .open_device(path, Some(&header_path), NAME, passphrase)
+            .unwrap();
+
+        luks2_formatter.close_device(NAME).unwrap();
+    }
+
+    #[test]
+    #[serial]
+    fn encrypt_with_existing_header_file() {
+        let mut bin_file = tempfile::NamedTempFile::new().unwrap();
+        bin_file
+            .as_file_mut()
+            .write_all(&vec![0; 20 * 1024 * 1024])
+            .unwrap();
+        let path = bin_file.path().to_str().unwrap();
+        let header_path = luks_header_path(path);
+        prepare_luks_header_file(&header_path).unwrap();
+        let _guard = RemoveHeaderOnDrop(header_path.clone());
+
+        let passphrase = Zeroizing::new(TEST_PASSPHRASE.to_vec());
+        let luks2_formatter = Luks2Formatter { integrity: false };
+        let result = luks2_formatter.encrypt_device(path, Some(&header_path), passphrase);
+        assert!(result.is_ok());
+    }
+
+    #[test]
+    #[serial]
+    fn open_device_missing_header_file_fails() {
+        let mut bin_file = tempfile::NamedTempFile::new().unwrap();
+        bin_file
+            .as_file_mut()
+            .write_all(&vec![0; 20 * 1024 * 1024])
+            .unwrap();
+        let path = bin_file.path().to_str().unwrap();
+        let header_path = luks_header_path(path);
+        prepare_luks_header_file(&header_path).unwrap();
+        let _guard = RemoveHeaderOnDrop(header_path.clone());
+
+        let passphrase = Zeroizing::new(TEST_PASSPHRASE.to_vec());
+        let luks2_formatter = Luks2Formatter { integrity: false };
+        luks2_formatter
+            .encrypt_device(path, Some(&header_path), passphrase.clone())
+            .unwrap();
+
+        std::fs::remove_file(&header_path).unwrap();
+
+        let result =
+            luks2_formatter.open_device(path, Some(header_path.as_str()), NAME, passphrase);
+        assert!(result.is_err());
+    }
+
+    #[test]
+    fn prepare_luks_header_file_rejects_existing_path() {
+        use rand::{distr::Alphanumeric, rng, Rng};
+
+        let path_str = format!(
+            "/dev/{}",
+            rng()
+                .sample_iter(&Alphanumeric)
+                .take(16)
+                .map(char::from)
+                .collect::<String>()
+        );
+
+        let header_path = luks_header_path(&path_str);
+        prepare_luks_header_file(&header_path).unwrap();
+        let result = prepare_luks_header_file(&header_path);
+
+        match result {
+            Err(err) => assert_eq!(err.kind(), std::io::ErrorKind::AlreadyExists),
+            other => panic!("unexpected result: {other:?}"),
+        }
+        let _ = std::fs::remove_file(&header_path);
     }
 
     /// This test can be used to clean useless devices under /dev/mapper/
--- a/confidential-data-hub/hub/src/storage/drivers/luks2.rs
+++ b/confidential-data-hub/hub/src/storage/drivers/luks2.rs
@@ -188,11 +188,18 @@ mod tests {
     const NAME: &str = "test";
 
     /// Removes the LUKS header file on drop so tests don't leave files behind on panic.
-    /// TODO: Similarly, clean dm-crypt devices when a panic occurs.
     struct RemoveHeaderOnDrop(String);
     impl Drop for RemoveHeaderOnDrop {
         fn drop(&mut self) {
             let _ = std::fs::remove_file(&self.0);
+        }
+    }
+
+    /// Closes the dm-crypt device on drop so tests don't leave mapper devices behind.
+    struct CloseDeviceOnDrop(String);
+    impl Drop for CloseDeviceOnDrop {
+        fn drop(&mut self) {
+            let _ = Luks2Formatter::default().close_device(&self.0);
         }
     }
 
@@ -216,8 +223,7 @@ mod tests {
         luks2_formatter
             .open_device(path, None, NAME, passphrase)
             .unwrap();
-
-        luks2_formatter.close_device(NAME).unwrap();
+        let _device_guard = CloseDeviceOnDrop(NAME.to_string());
     }
 
     #[test]
@@ -240,8 +246,7 @@ mod tests {
         luks2_formatter
             .open_device(path, None, NAME, passphrase)
             .unwrap();
-
-        luks2_formatter.close_device(NAME).unwrap();
+        let _device_guard = CloseDeviceOnDrop(NAME.to_string());
     }
 
     #[test]
@@ -266,8 +271,7 @@ mod tests {
         luks2_formatter
             .open_device(path, Some(&header_path), NAME, passphrase)
             .unwrap();
-
-        luks2_formatter.close_device(NAME).unwrap();
+        let _device_guard = CloseDeviceOnDrop(NAME.to_string());
     }
 
     #[test]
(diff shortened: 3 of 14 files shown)


confidential-data-hub/hub/src/auth/kbs.rs
```diff
@@ -12,11 +12,7 @@ use kms::{plugins::kbs::KbcClient, Annotations, Getter};
 use log::debug;
 use tokio::fs;
 
-use crate::{hub::Hub, Error, Result};
-
-/// This directory is used to store all the kbs resources get by CDH's init
-/// function, s.t. `[[Credential]]` sections in the config.toml file.
-pub const KBS_RESOURCE_STORAGE_DIR: &str = "/run/confidential-containers/cdh";
+use crate::{config::KBS_RESOURCE_STORAGE_DIR, hub::Hub, Error, Result};
```
manuelh-dev: Could rename KBS_RESOURCE_STORAGE_DIR to KBS_STORAGE_DIR, then we can a) store KBS resources in a subfolder of /run/confidential-containers/cdh and b) similarly luks-headers in a subfolder. In this PR, the luks headers are stored under /run/confidential-containers/cdh/luks-headers/…

confidential-data-hub/hub/src/config.rs
```diff
@@ -20,6 +20,10 @@ cfg_if::cfg_if! {
     }
 }
 
+/// This directory is used to store all the kbs resources get by CDH's init
+/// function, s.t. `[[Credential]]` sections in the config.toml file.
```
manuelh-dev: Should update this comment (taken from existing comment): looks like we mean the [[credentials]] section. Could point to the example file example.config.toml

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
```diff
@@ -1,95 +1,445 @@
...
+        mkfs_opts: Option<String>,
+    },
+}
+
+#[cfg(feature = "luks2")]
+pub(crate) fn luks_header_path(device_id: Option<&str>, device_path: &Path) -> std::path::PathBuf {
+    let id_bytes: Vec<u8> = device_id
+        .map(|s| s.as_bytes().to_vec())
+        .unwrap_or_else(|| device_path.to_string_lossy().into_owned().into_bytes());
+    let name = b64.encode(&id_bytes);
```
manuelh-dev: Base64 encoded: This should make the input collision-resistant. We don't want to allow duplicate paths or device IDs, as far as I can understand.
Examples:
Device ID 8:0 => ODow
Device path /dev/vdb => L2Rldi92ZGI

Thought of using tempfile::Builder to generate random names, but doesn't seem necessary.

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
```diff
@@ -187,12 +234,15 @@ impl BlockDevice {
...
+            parameters.device_id.as_deref(),
+            parameters.device_path.as_deref(),
+        ) {
             (Some(device_id), _) => {
-                let (maj, min) = parse_device_id(&device_id)?;
+                let (maj, min) = parse_device_id(device_id)?;
                 get_device_path(maj, min).await?
             }
-            (_, Some(device_path)) => device_path,
+            (_, Some(device_path)) => device_path.to_string(),
```
mythi: what's the reason for these changes?

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
```diff
@@ -116,6 +125,44 @@ pub enum TargetType {
...
+    device_path: &Path,
+) -> Result<std::path::PathBuf> {
+    let header_path = luks_header_path(device_id, device_path);
+    if header_path.exists() {
+        return Err(std::io::Error::new(
+            std::io::ErrorKind::AlreadyExists,
+            format!("LUKS header file already exists: {}", header_path.display()),
+        )
+        .into());
+    }
```
mythi: this is redundant and racy in theory at least. Use `create_new(true)` below to catch if the file already exists.

confidential-data-hub/hub/src/config.rs
```diff
@@ -20,6 +20,10 @@ cfg_if::cfg_if! {
     }
 }
 
+/// This directory is used to store all the kbs resources get by CDH's init
```
mythi: maybe `s/get/initialized/`.

confidential-data-hub/hub/src/storage/drivers/luks2.rs
```diff
@@ -130,18 +130,45 @@ impl Luks2Formatter {
...
 
+fn init_device(
+    device_path: &Path,
+    header_path: Option<&Path>,
+) -> anyhow::Result<libcryptsetup_rs::CryptDevice> {
+    let device_paths = match header_path {
+        Some(header_path) => {
+            if !header_path.exists() {
+                bail!("LUKS header file not found: {}", header_path.display());
+            }
```
mythi: could you rely on `CryptInit::init_with_data_device(device_paths)` failing on this instead?

confidential-data-hub/hub/src/config.rs
manuelh-dev: @mythi - shoot, I haven't published my own comments on this PR. I just did now. I was merely relocating this variable and comment to a new place, not changing it. I will resolve this comment here and tag you in the other comment. There's an additional aspect ...

confidential-data-hub/hub/src/config.rs
manuelh-dev: @mythi this is related to your prior comment. One additional comment I would like to tag you is also related.

confidential-data-hub/hub/src/auth/kbs.rs
manuelh-dev: @mythi thoughts?

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
manuelh-dev: I changed the parameters of open_device and encrypt_device (as well as init_device and prepare_luks_header_file and luke_header_path) to take `&Path` as parameters instead of `&str`, see the changes in luks2.rs. 
With this, I resolve device_id and device_path once into a single String and pass it as Path::new(&device_path). 
To me, this results in a bit of easier to read code (I tried without this as well and had to make various 'hoops').
In the match I used .as_deref() so that we don’t "consume" parameters and can still pass "parameters.device_id" into prepare_luks_header_file (without doing it this way I got a use-after-move compile error).

More holistic view:
device_path is a str coming from the CDH API so we need to convert that to a Path somewhere before feeding the path into the function `CryptInit::init_with_data_device` (called via `init_device` by `open_device` and `encrypt_device`). If you see a more elegant solution or prefer the previous solution, I can show you how it looks.
Previously, we had done this conversion in `open_device` and in `encrypt_device` twice via:
```
let path = Path::new(device_path);
let mut device = CryptInit::init(path)?;
```

confidential-data-hub/hub/src/storage/drivers/luks2.rs
manuelh-dev: Yes, thank, you, addressed; the test `open_device_missing_header_file_fails` still succeeds, printing the following error (`Device /run/confidential-containers/cdh/luks-headers/L3RtcC8udG1wN1NueXNR does not exist or access denied.`).

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
manuelh-dev: removed; as expected, the test `prepare_luks_header_file_rejects_existing_path` still succeeds.

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
```diff
@@ -214,13 +257,25 @@ impl BlockDevice {
                 data_integrity,
                 mapper_name,
             } => {
+                if parameters.source_type == SourceType::Empty && parameters.key.is_some() {
+                    return Err(BlockDeviceError::KeyNotAllowedForEphemeralLuksHeader);
+                }
```
fitzthum: Seems like this check could go below inside the `SourceType::Empty` block.

confidential-data-hub/hub/src/config.rs
```diff
@@ -20,6 +20,10 @@ cfg_if::cfg_if! {
...
+/// function, s.t. `[[Credential]]` sections in the config.toml file.
+pub const KBS_RESOURCE_STORAGE_DIR: &str = "/run/confidential-containers/cdh";
```
fitzthum: not sure `config.rs` is the right place for this. It doesn't really have anything to do with configuration.

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
```diff
@@ -116,6 +125,37 @@ pub enum TargetType {
     },
 }
 
+#[cfg(feature = "luks2")]
+pub(crate) fn luks_header_path(device_id: Option<&str>, device_path: &Path) -> std::path::PathBuf {
```
fitzthum: I don't think we need to take both the `device_id` and `device_path` here. It looks like `real_mount` already resolves a device path from either and id or a path.

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
```diff
@@ -116,6 +125,37 @@ pub enum TargetType {
...
+    let id_bytes: Vec<u8> = device_id
+        .map(|s| s.as_bytes().to_vec())
+        .unwrap_or_else(|| device_path.to_string_lossy().into_owned().into_bytes());
+    let name = b64.encode(&id_bytes);
+    std::path::PathBuf::from(format!(
+        "{}/{}/{}",
+        KBS_RESOURCE_STORAGE_DIR, LUKS_HEADERS_DIR, name
```
fitzthum: I wonder if we want these files to have something like `.header` at the end.

fitzthum (commented):
Nice. This seems legit. It's still a little surprising that we have to provide a file to libcryptsetup, but so be it.

Left a few comments.

Let's see what @Xynnn007 says when he is back.

confidential-data-hub/hub/src/auth/kbs.rs
manuelh-dev: @fitzthum - this is somewhat related to your other comment on KBS_RESOURCE_STORAGE_DIR. Thoughts?

confidential-data-hub/hub/src/config.rs
manuelh-dev: @fitzthum - this is somewhat related to your other comment on KBS_RESOURCE_STORAGE_DIR. Thoughts?

confidential-data-hub/hub/src/config.rs
manuelh-dev: I had moved it there from kbs.rs. I will move it back to kbs.rs

manuelh-dev: I remember now - consuming the variable in blockdevice/mod.rs is not possible when moving it back from config.rs to kbs.rs as module `kbs` is private.

@fitzthum - any thoughts on where to move this other than config.rs, or should we use two separate variables?

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
```diff
@@ -19,20 +19,30 @@ use crate::{
...
+use std::path::Path;
 use strum::{Display, EnumString};
 use tokio::{
     fs::{symlink, File},
     io::{AsyncBufReadExt, BufReader},
 };
 use zeroize::Zeroizing;
 
+use crate::config::KBS_RESOURCE_STORAGE_DIR;
+#[cfg(feature = "luks2")]
```
manuelh-dev: In this file, we introduce three new variables: `LUKS_HEADERS_DIR`, `LUKS_HEADER_FILE_SUFFIX`, and `LUKS2_HEADER_MIN_SIZE_BYTES` as well as two new functions `luks_header_path` and `prepare_luks_header_file`.

From a design perspective, is this the right file? The file `storage/drivers/luks2.rs` is a libcryptsetup wrapper. At the moment, it does not seem like this wrapper should make the decision on the luks header path 
stay a pure libcryptsetup wrapper, and since `blockdevice/mod.rs` obviously contains various luks2-specific paths today, I added it there. If you have better suggestions, @mythi or @fitzthum (CC @Xynnn007) please let me know

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
manuelh-dev: Added .header.

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
manuelh-dev: You're right. Adopted.

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
manuelh-dev: Agreed, relocated.

confidential-data-hub/hub/src/auth/kbs.rs
fitzthum: yeah fine to rename

confidential-data-hub/hub/src/config.rs
fitzthum: Hmm since this now pertains to more than just the KBS resource storage, I would just try to put it towards the top of the tree. Maybe in `hub.rs`.

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
fitzthum: I think this is fine. The semantics of our storage stuff could be tightened up, but I would prefer to do that in a future PR when we introduce another backend. The requirements will probably be a bit more clear at that point.

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
```diff
@@ -116,6 +126,32 @@ pub enum TargetType {
...
+    // error "LUKS header file not found: <path/to/header>" from libcryptsetup if header file doesn't exist.
+    let file = std::fs::OpenOptions::new()
+        .create_new(true)
+        .write(true)
+        .open(&header_path)?;
+    // error "Device ... is too small" / OS error 5" from libcryptsetup if header isn't sized.
+    file.set_len(LUKS2_HEADER_MIN_SIZE_BYTES)?;
+    Ok(header_path)
+}
+
```
Xynnn007: Probably we could move the two functions to `drivers/luks2.rs` to avoid the feature flags?

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
```diff
@@ -216,11 +255,20 @@ impl BlockDevice {
             } => {
                 let integrity = data_integrity.map_or(false, |e| e.parse().unwrap_or(false));
                 let formatter = Luks2Formatter::default().with_integrity(integrity);
+                let mut header_path: Option<std::path::PathBuf> = None;
                 // 3.1 if the source type is empty, encrypt the device
                 if parameters.source_type == SourceType::Empty {
+                    if parameters.key.is_some() {
+                        return Err(BlockDeviceError::KeyNotAllowedForEphemeralLuksHeader);
+                    }
```
Xynnn007: What's the reason to add this condition? I am not sure if the code logic changes. The original idea is to support to encrypt an empty device with given key if key is given.

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
```diff
@@ -116,6 +126,32 @@ pub enum TargetType {
     },
 }
 
+#[cfg(feature = "luks2")]
+pub(crate) fn luks_header_path(device_path: &Path) -> std::path::PathBuf {
```
Xynnn007: Looks like this function is used to automatically generate a file path of header file. Could we have the header path as another separate parameter for function `prepare_luks_header_file()`?

Btw, we have `BlockDevice` struct mutable in function `real_mount()` and it has many different members like `symlink_names`, `mount_points`. They are all used to do clean jobs when doing `umount()`. Probably we can rename `symlink_names` to `tempfiles` or something, beause it only records the file paths to be deleted when `umount` is called, and symlink files are just one kind of them. Add the luks header path to this new ``tempfiles` member after generate the file in the body of  `real_mount`. In this way we do not need the struct `RemoveHeaderOnDrop` in tests. Just all `umount` can delete the file.

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
```diff
@@ -596,4 +646,60 @@ mod tests {
 
         bd.umount().await.unwrap();
     }
+
+    #[cfg(feature = "luks2")]
+    #[tokio::test]
+    #[serial]
+    async fn encrypt_empty_device_with_key_is_rejected() {
```
Xynnn007: I have a same question upon this function as the previous one.

Xynnn007 (commented):
Nice to see the PR, thanks @manuelh-dev !

Some design ideas for reference

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
fitzthum: In this case we create a new LUKS device and encrypt it with an existing key. Those operations are possible, but since we don't have any mechanism of propagating the header back to the host, we will never be able to mount the device again. Essentially, we can't currently do persistent storage with detached headers.

Also, this case of creating a new persistent storage, might not be one we even want to support. The expectation may be that for persistent storage the user creates the device (and header) ahead of time.

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
manuelh-dev: We thought about this, in a review comment further above. :) I will resolve this comment and tag you in the comment further above to have this discussion.

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
manuelh-dev: @Xynnn007 - here is the discussion on moving luks2-specific functions (but also variables) into `storage/drivers/luks2.rs` which you suggested in another comment. Could I please ask you to take a look here and provide your opinion? :)

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
manuelh-dev: I suspect this specific comment refers to the comment on `lines +261 to +263`. We can continue the discussion there as there's already a reply in the comment's discussion. Resolving - please re-open if I have misunderstood you.

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
manuelh-dev: Great points! Side note: For `confidential-data-hub\hub\src\storage\drivers\luks2.rs` we will still keep using `RemoveHeaderOnDrop`.

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
manuelh-dev: I have reverted this. The diff is now shorter, resp., there is no more diffs such as highlighted by you. I am pro-actively resolving this comment as I believe this is addressed.

confidential-data-hub/hub/src/config.rs
manuelh-dev: Cleared/moved/rephrased as part of addressing another comment. Resolving this one here to continue the conversation in the other comment.

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
manuelh-dev: Nobody has commented on this so far, assuming this is fine.

confidential-data-hub/hub/src/config.rs
manuelh-dev: I gave it a stab. We now have: `CDH_BASE_DIR` in `hub.rs` and `kbs_resource_storage_dir()` and `luks_headers_storage_dir()` in the relevant files using `CDH_BASE_DIR`

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
fitzthum: Ah yeah I agree with what Ding suggests about moving to the other file.

confidential-data-hub/hub/src/auth/kbs.rs
```diff
@@ -12,11 +12,13 @@ use kms::{plugins::kbs::KbcClient, Annotations, Getter};
...
 
-use crate::{hub::Hub, Error, Result};
+use crate::{hub::Hub, hub::CDH_BASE_DIR, Error, Result};
 
-/// This directory is used to store all the kbs resources get by CDH's init
-/// function, s.t. `[[Credential]]` sections in the config.toml file.
-pub const KBS_RESOURCE_STORAGE_DIR: &str = "/run/confidential-containers/cdh";
+/// Directory for KBS resources (credentials, etc.) fetched at init, see the `[[credentials]]`
+/// sections in CDH's `config.toml` file, for an example, see `example.config.toml`.
+fn kbs_resource_storage_dir() -> String {
```
manuelh-dev: do you prefer using `concatcp` here and in other places? For now I abstained from using it to not add a new crate to hub

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
manuelh-dev: Moved, this looks cleaner now.

confidential-data-hub/hub/src/auth/kbs.rs
manuelh-dev: Discussed with @fitzthum - using `concatcp` now.

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
Xynnn007: imo the ability matrix should provide different usages regardless higher level callers. Overall
1. if key is given. It's suitable for persistent storage for CoCo or CVM.
2. if key is not given. It's suitable for ephemeral storage of CoCo (like write-able layer of container images in guest, which does not need to be persistent or used again) , and the pure CVM case.

For CoCo we could choose which can be used by upper layer caller.

confidential-data-hub/hub/src/storage/drivers/luks2.rs
```diff
@@ -130,18 +162,39 @@ impl Luks2Formatter {
...
+    use super::{luks_header_path, prepare_luks_header_file, Luks2Formatter};
 
     const TEST_PASSPHRASE: &[u8] = b"test";
     const NAME: &str = "test";
 
+    /// Removes the LUKS header file on drop so tests don't leave files behind on panic.
+    struct RemoveHeaderOnDrop(String);
+    impl Drop for RemoveHeaderOnDrop {
+        fn drop(&mut self) {
+            let _ = std::fs::remove_file(&self.0);
```
Xynnn007, line 194: I met a same problem before to clean cryptsetup devices when panic happens. Drop here provides a good way. Could you help to add a TODO note that mentions "clean dm-crypt devices when panic"?

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
```diff
@@ -216,11 +215,18 @@ impl BlockDevice {
...
                 let formatter = Luks2Formatter::default().with_integrity(integrity);
+                let mut header_path: Option<String> = None;
                 // 3.1 if the source type is empty, encrypt the device
                 if parameters.source_type == SourceType::Empty {
+                    if parameters.key.is_some() {
+                        return Err(BlockDeviceError::KeyNotAllowedForEphemeralLuksHeader);
+                    }
                     warn!("encrypting the device. This will wipe original data on the disk.");
+                    header_path = Some(luks2::luks_header_path(&device_path));
+                    luks2::prepare_luks_header_file(header_path.as_ref().unwrap())?;
```
Xynnn007: Let's replace `unwrap` here with error handling

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
```diff
@@ -216,11 +215,18 @@ impl BlockDevice {
...
+                    self.temp_paths.push(header_path.as_ref().unwrap().clone());
```
Xynnn007: ditto for unwrap

Xynnn007 (commented):
Some more ideas.

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
fitzthum: I see where you're coming from, but currently the API does not have any way to communicate the path of the header file, meaning that we really don't support all of the cases. I think adding to the API is beyond the scope of this PR, which is really just about ephemeral support. 

But we could drop this check if you want and have the caller deal with it. Either way is fine for me.

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
manuelh-dev: @Xynnn007 addressed in new revision, ptal

confidential-data-hub/hub/src/storage/drivers/luks2.rs
manuelh-dev: I added a TODO in the initial commit introducing LUKS header support, and added a new commit on top adding this new drop guard. ptal if that looks good, also happy to remove this commit again.

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
manuelh-dev: I am dropping this check for now. We can bring it back in a separate commit or a separate PR if we like. @Xynnn007 - ptal :)

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
```diff
@@ -494,41 +500,57 @@ mod tests {
             .unwrap();
 
         bd.umount().await.unwrap();
-        drop(bd);
+    }
```
Xynnn007, line 503: Noticed that the original create-then-open test is splited into two. I guess the cause is that it's hard to specify the header path on the request.

We can add another optional field e.g. `header`  [here](https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs#L42-L50), and use it when the source is empty and integrity is enabled.

In this way we can do the create-then-open test again

Xynnn007 (commented):
The code looks good to me! One last comment upon the test

fitzthum (approved):
LGTM pending Ding's last comments

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
fitzthum: Can we just keep the scope of this to ephemeral for now. I think there are a few things we need to figure out regarding how to handle the header. For instance, these headers will need to be measured in the persistent case. So maybe we want to provide them via init-data, rather than via the kata-agent. Or maybe we should just use ZFS for persistent storage. Either way, I feel like we should think about this in a separate PR.

Xynnn007: Emmm. This breaks current unit test's target. But we can have a TODO here to get it back if this is the aim.

fitzthum: That's true, but even if we add a header path, the feature won't really be usable until we have some way to measure the header. Anyway, I'm fine with your suggestion but I think it's reasonable to skip this case for now and add it back when we really support it.

Xynnn007 (approved):

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
manuelh-dev: I have a commit here upon we can expand in the future in a subsequent PR: https://github.com/manuelh-dev/guest-components/commit/9082118141b0cebb7f2f10971326cb7366284582

PR #1316, opened 2026-02-13
CI: makefile: propagate LIBC environment variable

manuelh-dev:
Propagating the LIBC variable aims to help with CI build errors such
as an error observed in CI to build CDH in publish-cdh-and-asr.
With the newly introduced dependency to libcryptsetup, building the
target x86_64-unknown-linux-musl fails with "pkg-config has not been
configured to support cross-compilation" when omitting the LIBC
variable.

For the sake of completeness, also add the ARCH variable explicitly
to Makefile.


fitzthum (approved):
LGTM

Let's merge this and see if it's fixes.

PR #1317, opened 2026-02-13
ci: Use gnu libc to build CDH

manuelh-dev:
Align the CI task with recent changes to the CDH build process.


fitzthum (approved):
Ok, looked more closely this time. This makes sure that we build the CDH with gnu, which is required now that we are using the libcryptsetup bindings. This looks like it will solve our problem.

PR #1318, opened 2026-02-13
ci: Install new CDH dependencies

manuelh-dev:
Install the libcryptsetup-dev, pkg-config packages similar to the recent change in `image_rs_build.yml`, see commit
`ba683043f7ca90bb2e7a121119a43a399216bb0b`. This should unblock the still failing "publish-cdh-and-asr (x86_64)" CI task.


fitzthum (approved):
LGTM

Ideally we could work through this without doing one PR at a time, but this seems good.

PR #1332, opened 2026-02-24
CDH: remove golang implementation

### LINKED ISSUE #1256: cdh: remove golang interface
Currently repo contains a golang implementation of cdh client and interface in `confidential-data-hub/golang`. This code is not aligned with the latest developments and I'm not sure if any users are using it. I suggest we can remove this code. cc @Xynnn007 @fitzthum

Xynnn007:
Close #1256


ChengyuZhu6 (commented):
I think we also need remove  https://github.com/confidential-containers/guest-components/blob/2ce8d5f43ce6fefa4dc7a8a2c33dea743324d57d/.github/dependabot.yml#L16-L19?

PR #1340, opened 2026-02-25
cdh: Add test case using mkfsOpts

manuelh-dev:
Add a test case which passes mkfsOpts for the mkfs.ext4 command.

I have a local addition to this where I use a function `verify_ext4_mkfs_opts` to verify these options are considered, and it works for me. As of now, I was hesitant to add that function to the PR as this function uses `tune2fs`.


manuelh-dev:
@fitzthum @Xynnn007 fyi - I cannot add reviewers :)

Xynnn007 (approved):

fitzthum (approved):
LGTM. I guess we could potentially expand this in the future to check the properties of the fs or to try exercising stuff like xattrs.

PR #1344, opened 2026-02-26
CDH | Add ZFS Blockdevice plugin

Xynnn007:
This patch mainly adds the ZFS block device ability using commands than rust-native crates as there is no active maintained lib for ZFS.

See concrete commit message for better review.

cc @fitzthum @manuelh-dev

### DIFF
--- a/confidential-data-hub/hub/src/storage/drivers/filesystem.rs
+++ b/confidential-data-hub/hub/src/storage/drivers/filesystem.rs
@@ -6,8 +6,12 @@
 use anyhow::{anyhow, bail, Result};
 use serde::{Deserialize, Serialize};
 use strum::AsRefStr;
+use tracing::warn;
 
 use crate::storage::drivers::run_command;
+
+const EXT4_COMMAND: &str = "mkfs.ext4";
+const DD_COMMAND: &str = "dd";
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Copy, AsRefStr, Default)]
 pub enum FsType {
@@ -53,8 +57,18 @@ impl FsFormatter {
     /// Thus we need to get the block numbers and write to those blocks before hand.
     pub fn format_integrity_compatible(&self, device_path: &str) -> Result<()> {
         let command = match self.fs_type {
-            FsType::Ext4 => "mkfs.ext4",
+            FsType::Ext4 => {
+                if !is_ext4_installed() {
+                    bail!("ext4 is not installed. Consider installing `e2fsprogs` (ubuntu).");
+                }
+                EXT4_COMMAND
+            }
         };
+
+        if !is_dd_installed() {
+            bail!("dd is not installed. Consider installing `coreutils` (ubuntu).");
+        }
+
         let args = vec!["-F", "-n", device_path];
         let (stdout, stderr) = run_command(command, &args, None)?;
 
@@ -87,7 +101,7 @@ impl FsFormatter {
 
         for num in nums {
             let _ = run_command(
-                "dd",
+                DD_COMMAND,
                 &[
                     "if=/dev/zero",
                     &format!("of={device_path}"),
@@ -108,7 +122,12 @@ impl FsFormatter {
 
     pub fn format(&self, device_path: &str) -> Result<()> {
         let command = match self.fs_type {
-            FsType::Ext4 => "mkfs.ext4",
+            FsType::Ext4 => {
+                if !is_ext4_installed() {
+                    bail!("ext4 is not installed. Consider installing `e2fsprogs` (ubuntu).");
+                }
+                EXT4_COMMAND
+            }
         };
 
         let mut args = vec![device_path];
@@ -123,4 +142,20 @@ impl FsFormatter {
 
         Ok(())
     }
+}
+
+fn is_ext4_installed() -> bool {
+    let installed = run_command(EXT4_COMMAND, &["-V"], None).is_ok();
+    if !installed {
+        warn!("ext4 is not installed. Consider installing `e2fsprogs` (ubuntu).");
+    }
+    installed
+}
+
+fn is_dd_installed() -> bool {
+    let installed = run_command(DD_COMMAND, &["--version"], None).is_ok();
+    if !installed {
+        warn!("dd is not installed. Consider installing `coreutils` (ubuntu).");
+    }
+    installed
 }


confidential-data-hub/hub/src/storage/drivers/zfs.rs
```diff
@@ -0,0 +1,280 @@
...
+
+/// The default encryption algorithm.
+pub const DEFAULT_ENCRYPTION_ALGORITHM: &str = "aes-256-gcm";
+
+/// The default encryption key format.
+pub const DEFAULT_ENCRYPTION_KEY_FORMAT: &str = "passphrase";
+
+/// This function checks if zfs is installed and available.
+pub fn is_zfs_installed() -> bool {
+    run_command("zpool", &["list"], None).is_ok()
```
fitzthum: you might move the log message about zfs not being installed into this function rather than doing it in all the callers

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
```diff
@@ -48,6 +50,18 @@ pub enum BlockDeviceEncryptType {
         #[serde(rename = "mapperName")]
         mapper_name: Option<String>,
     },
+
+    #[strum(serialize = "zfs")]
+    #[serde(rename = "zfs")]
+    Zfs {
+        /// The name of the zpool to use.
+        /// If not set, [`crate::storage::drivers::zfs::DEFAULT_ZPOOL_NAME`] will be used.
```
fitzthum: might be good to mention what the actual default value is here (and for the next option)

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
```diff
@@ -48,6 +50,18 @@ pub enum BlockDeviceEncryptType {
...
+        pool: Option<String>,
+
+        /// The name of the zdataset to use.
+        /// If not set, [`crate::storage::drivers::zfs::DEFAULT_ZDATASET_NAME`] will be used.
+        dataset: Option<String>,
+    },
```
fitzthum: It might be a little bit more modular to have this enum contain a struct that comes from the ZFS file, rather than having all these ZFS-specific fields in `mod.rs`. Not a big deal either way.

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
```diff
@@ -343,6 +371,47 @@ impl BlockDevice {
                     }
                 }
             }
+            BlockDeviceEncryptType::Zfs { pool, dataset } => {
```
fitzthum: This method is getting really big. We may want to split this huge match statement into a few different methods.

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
```diff
@@ -343,6 +371,47 @@ impl BlockDevice {
...
+                }
+
+                // if the source type is empty, create a new zpool and zdataset
+                if parameters.source_type == SourceType::Empty {
+                    warn!("creating a new zpool and zdataset on the device. This will wipe original data on the disk.");
+                    create_zpool(&pool, &device_path)
+                        .map_err(|source| BlockDeviceError::ZfsError { source })?;
+                    self.zfs_pools.push(pool.clone());
+                    create_zdataset(&pool, &dataset, key.clone(), mount_point)
+                        .map_err(|source| BlockDeviceError::ZfsError { source })?;
```
fitzthum: We could introduce a new function in the zfs file called `create_ephemeral_zdataset` or something like that, which has the functionality of this match arm. I think it's better to have more of the ZFS-specific stuff in the ZFS file. (although i see there is some state related to zfs in this struct that we'd need to think about)

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
```diff
@@ -171,7 +186,12 @@ pub struct BlockDevice {
     /// clean up.
     ///
     /// It is (device-path, dev-mapper-name)
+    #[cfg(feature = "luks2")]
```
fitzthum, line 136: As an aside, I guess a downside of the libcryptsetup bindings is that the code won't build without the dependencies. Hence we need a feature. I prefer the flow you have for ZFS where there is no feature and we just check at runtime if we have the binaries.

confidential-data-hub/docs/use-cases/secure-mount-with-block-device.md
```diff
@@ -238,4 +210,140 @@ Then you can use the following `secure_mount` request to mount the encrypted blo
...
+```shell
+$ cat /mnt/luks2ext4-cdh-test/confidential-data-file
+some-data
+```
+
+---
+
+## ZFS usage
+
+Commands, steps, and expected behavior when using **ZFS** secure mount (`encryptionType: "zfs"`).
```
fitzthum, line 239: hmm. To me ZFS is the filesystem type not the encryption type. I guess we don't have a field for the filesystem type currently, so maybe it's ok to do this so we don't have to change the interface again with all these PRs in flight. Still, this doesn't seem quite right to me.

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
Xynnn007: It's set [here](https://github.com/confidential-containers/guest-components/pull/1344/changes#diff-ababdaebf88bb53ba180ae83e6f4c7fa1b458b9ab828cb4aa699c7225f297ee6R381). and it's on line an inline comment and you can see it as a link. By clicking, it can be skipped to the actual `DEFAULT_ZPOOL_NAME` variable. It helps us to only maintain one value `DEFAULT_ZPOOL_NAME` than in two places (the const and the comment here)

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
Xynnn007: Sounds good. Let me add a separate commit to do the refactoring.

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
Xynnn007: Let me do the refactoring

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
Xynnn007: Yes. That's the shortage that I once mentioned for libcryptsetup-rs -- it cannot be built statically thus we need different binaries for different distros.

If it's hard for us to do different releases of CDH for different distros. I can do the change to transfer it from `libcryptsetup-rs` to running commands.

confidential-data-hub/docs/use-cases/secure-mount-with-block-device.md
Xynnn007: I once mentioned somewhere that the "block-device" here means the data source is a block-device. Thus it's natural for us to have different "modes" like luks2 and zfs than mentioning it as fileSystem. The encryption ability of both are used to encrypt the block device. Thus it makes sense to me to use `encryptionType`. But also, if you like, it's also acceptable to change it to `mode` or something else

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
Xynnn007: Let me change the code together with this item

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
fitzthum: Yeah, I could go either way on this one. We can discuss more later.

confidential-data-hub/docs/use-cases/secure-mount-with-block-device.md
fitzthum: Yeah, I guess I mainly think of ZFS as a filesystem. Now we do already specify `ext4` as a filesystem, but that's part of the LUKS options. It seems like maybe filesystem should be a top-level option, but it's a little tricky to model because LUKS+ext4 and ZFS have different properties.

Maybe we should be more explicitly separating everything into two steps. 1) block/volume operations 2) filesystem operations. So we could have LUKS + ext4 or LUKS + some other filesystem. For ZFS the first step would be a no-op.

Anyway, let's not change the API for now, but worth thinking about in the future.

confidential-data-hub/docs/use-cases/secure-mount-with-block-device.md
```diff
@@ -4,126 +4,117 @@ This guide helps an user to use confidential data hub to secure mount inside TEE
 
 ## Preliminaries
 
-- Ensure that [cryptsetup](https://gitlab.com/cryptsetup/cryptsetup/) is installed.
+- Ensure that [cryptsetup](https://gitlab.com/cryptsetup/cryptsetup/) is installed for **LUKS2** mode.
```
manuelh-dev: I am a bit confused here. I thought we're using libcryptsetup. What again is cryptsetup being used for? Also, in SECURE_STORAGE.md we have a few references to the cryptsetup binary.

manuelh-dev: For sanity, should we add that other binaries are needed to, such as dd and mkfs.ext4 (resp., relevant packages)? See the filesystem.rs file

confidential-data-hub/hub/src/storage/drivers/zfs.rs
```diff
@@ -0,0 +1,339 @@
...
+    /// Mount the zfs pool and zdataset to the given mount point.
+    /// Returns a vector of the created zfs pools.
+    pub async fn do_mount(
+        self,
+        device_path: &str,
+        mount_point: &str,
+        key: Zeroizing<Vec<u8>>,
+        source_type: SourceType,
+    ) -> Result<()> {
+        if !is_zfs_installed() {
```
manuelh-dev, line 96: I wonder if we can author a similar for the LUKS path. A few other binaries/libraries need to be present there as well

confidential-data-hub/hub/src/storage/drivers/zfs.rs
```diff
@@ -0,0 +1,339 @@
...
+    /// - clean up
+    #[test]
+    fn test_whole_zfs_flow() {
+        if !is_zfs_installed() {
+            return;
+        }
+
+        let temp_device = TempFileLoopDevice::new(200 * 1024 * 1024).unwrap();
+
+        const TEST_KEY: &str = "12345678";
```
manuelh-dev: just curious, can we reuse the zfs-disk-passphrase file here as well?

confidential-data-hub/hub/src/storage/drivers/mod.rs
```diff
@@ -3,6 +3,92 @@
...
+};
+
+use anyhow::{anyhow, bail, Result};
+use tempfile::NamedTempFile;
+use tracing::debug;
+
 pub mod filesystem;
 #[cfg(feature = "luks2")]
 pub mod luks2;
+pub mod zfs;
```
manuelh-dev, line 18: can you guys remind me again why we don't use `#[cfg(feature = "zfs")]` here? Because, in contrast to luks2 we don't use additional crates like for libcryptsetup?

fitzthum: Yes, ZFS impl only has runtime dependencies (which we check for at runtime). For LUKS we have build dependencies so it's nice to have a feature.

manuelh-dev: Resolving, see discussion above.

confidential-data-hub/docs/use-cases/secure-mount-with-block-device.md
manuelh-dev: Suggest to strive to well-define the API as early as possible so that we avoid breaking changes. 

So, the top-level fields are required for all "modes/types" of encryption? The fields of each encryption type (zfs|luks) are unique to these use cases? Can we think of at least one more use case and see if this pattern applies?

What happens if e.g. `volume_type` may not be `block-device`?

fitzthum:
Regarding licensing, I don't think this PR has any ZFS code in it. Potentially we should add the dependencies to our GHA workflows, but as is, I don't think there are any licensing concerns. That may be a different story for Kata.

confidential-data-hub/docs/use-cases/secure-mount-with-block-device.md
Xynnn007: Let me delete useless parts in the documents. They were promoted by cursor and it mixes the lib and the binary : (

Xynnn007: Yes. Let me add this

confidential-data-hub/hub/src/storage/drivers/zfs.rs
Xynnn007: Yes. Let me do  this

confidential-data-hub/hub/src/storage/drivers/zfs.rs
Xynnn007: I guess you mean the thread https://github.com/confidential-containers/guest-components/pull/1344#discussion_r2862099957.

Now luks2 is dependent on cryptsetup-rs (FFI dynamic library) than a binary. If we want to use binary call (like ZFS in this PR) than FFI binding, we can do this. This is better in another PR.

manuelh-dev: Agree, I was referring to mkfs.ext4 and dd binaries we invoke though (for the case where integrity is requested - format_integrity_compatible)

Xynnn007: Yes. Added.

PR #1347, opened 2026-02-27
fix: update openssl-src to 300.5.5

pavithiran34:
this is vulnerability in previous version of openssl-src, so updated to 300.5.5+3.5.5

CVE-2025-69421 - check this for more info


Xynnn007 (approved):
Thanks. LGTM

PR #1356, opened 2026-03-02
AA/CDH: Add debug level to configuration file

### LINKED ISSUE #1324: Add logging configuration to config files
Today, we rely on a parent process setting the  `RUST_LOG ` env var to configure the debug level. Instead, we should introduce logging configurations directly to the config files, allowing for logging to be enabled alongside any other configurations specified in the init-data.

cc @burgerdev

Xynnn007:
Close #1324

### DIFF
--- a/confidential-data-hub/hub/src/bin/config/mod.rs
+++ b/confidential-data-hub/hub/src/bin/config/mod.rs
@@ -21,17 +21,10 @@ pub fn read_config(config_path: Option<String>) -> Result<(CdhConfig, String)> {
             (config, log)
         }
         None => {
-            if let std::result::Result::Ok(env_path) = env::var("CDH_CONFIG_PATH") {
-                let log = format!("Read CDH's config path from env: {env_path}");
-                let config = CdhConfig::from_file(&env_path[..])
-                    .with_context(|| format!("failed to read config file {env_path}"))?;
-                (config, log)
-            } else {
-                let log = "No CDH config path specified. Using default configuration.".to_string();
-                let config = CdhConfig::default_with_kernel_cmdline()
-                    .with_context(|| "failed to read default configuration".to_string())?;
-                (config, log)
-            }
+            let log = "No CDH config path specified. Using default configuration.".to_string();
+            let config = CdhConfig::default_with_kernel_cmdline()
+                .with_context(|| "failed to read default configuration".to_string())?;
+            (config, log)
         }
     };
 
@@ -51,9 +44,6 @@ pub fn read_config(config_path: Option<String>) -> Result<(CdhConfig, String)> {
 mod tests {
     use std::{env, io::Write};
 
-    use anyhow::anyhow;
-    use confidential_data_hub::{CdhConfig, KbsConfig, LogConfig, DEFAULT_CDH_SOCKET_ADDR};
-    use image_rs::config::ImageConfig;
     use serial_test::serial;
 
     use crate::config::read_config;
@@ -101,41 +91,5 @@ authenticated_registry_credentials_uri = "kbs:///default/auth/1"
             config.image.authenticated_registry_credentials_uri,
             Some("kbs:///default/auth/1".into())
         );
-    }
-
-    #[test]
-    #[serial]
-    fn test_config_path() {
-        // --config takes precedence,
-        // then env.CDH_CONFIG_PATH
-
-        let config = CdhConfig::default_with_kernel_cmdline().expect("Must be successful");
-        let expected = CdhConfig {
-            log: LogConfig::default(),
-            kbc: KbsConfig {
-                name: "offline_fs_kbc".into(),
-                url: "".into(),
-                kbs_cert: None,
-            },
-            credentials: Vec::new(),
-            socket: DEFAULT_CDH_SOCKET_ADDR.into(),
-            image: ImageConfig::from_kernel_cmdline(),
-            skip_sealed_secret_verification: false,
-        };
-        assert_eq!(config, expected);
-
-        let config = CdhConfig::from_file("/thing").unwrap_err();
-        let expected = anyhow!("configuration file \"/thing\" not found");
-        assert_eq!(format!("{config}"), format!("{expected}"));
-
-        env::set_var("CDH_CONFIG_PATH", "/byenv");
-        let result = read_config(None).unwrap_err();
-        let expected = anyhow!("failed to read config file /byenv");
-        assert_eq!(format!("{result}"), format!("{expected}"));
-        env::remove_var("CDH_CONFIG_PATH");
-
-        let config = CdhConfig::from_file("/thing").unwrap_err();
-        let expected = anyhow!("configuration file \"/thing\" not found");
-        assert_eq!(format!("{config}"), format!("{expected}"));
     }
 }


attestation-agent/attestation-agent/src/bin/grpc-aa/main.rs
```diff
@@ -69,9 +69,25 @@ struct Cli {
...
+        }
+        None => {
+            let config = Config::default_with_kernel_cmdline();
+            let log = format!("No AA config file specified. Using default configuration and the kbs address will be read from kernel cmdline.");
+            (config, log)
+        }
+    };
+
     let env_filter = match std::env::var_os("RUST_LOG") {
         Some(_) => EnvFilter::try_from_default_env().expect("RUST_LOG is present but invalid"),
```
fitzthum: Wait, is that an `expect`? We should probably try to fall back to the log in this case. Actually, I would be fine with removing support for setting via env var.

attestation-agent/attestation-agent/src/bin/ttrpc-aa.rs
```diff
@@ -87,9 +87,27 @@ pub fn start_ttrpc_service(aa: AttestationAgent) -> Result<HashMap<String, Servi
...
+
+    let (config, config_log) = match cli.config_file {
+        Some(config_path) => {
+            let config = Config::try_from(&config_path[..])?;
+            let log = format!("Using config file: {config_path}");
+            (config, log)
+        }
+        None => {
+            let config = Config::default_with_kernel_cmdline();
+            let log = format!("No AA config file specified. Using default configuration and the kbs address will be read from kernel cmdline.");
```
fitzthum: But the env var can still be set in this case, right?

Xynnn007: env only works for CDH

attestation-agent/attestation-agent/src/bin/grpc-aa/main.rs
Xynnn007: Yes it's `expect`. It's because up to now the code still does not initialize the logging (`Subscriber`) thus even if we use `info!`-like macros there will be nothing on both stderr or stdout.

RUST_LOG env is somehow a "by-default" behavior for rust bins. Probably we can use `info` if errors happen here

confidential-data-hub/hub/src/bin/cdh-oneshot.rs
```diff
@@ -94,14 +96,26 @@ struct PullImageArgs {
...
-    let config = CdhConfig::new(args.config).expect("failed to initialize cdh config");
-    config.set_configuration_envs();
+    let cli = Cli::parse();
 
+    let (config, config_log) = config::read_config(cli.config).expect("failed to read config");
+
+    let env_filter = match std::env::var_os("RUST_LOG") {
+        Some(_) => EnvFilter::try_from_default_env().expect("RUST_LOG is present but invalid"),
+        None => EnvFilter::try_new(&config.log.level).unwrap_or_else(|e| {
+            eprintln!("Invalid log level '{e}'. Using default log level 'info'.");
```
mkulke: I would panic if the level can't be parsed, it's most likely the result of a typo and a gracious fallback is probably never what the user wants. an alternative is to check this at the config struct using an enum instead of a string for the level

confidential-data-hub/hub/src/bin/grpc-cdh.rs
```diff
@@ -32,9 +33,16 @@ struct Cli {
...
 #[tokio::main]
 async fn main() -> Result<()> {
+    let cli = Cli::parse();
+
+    let (config, config_log) = config::read_config(cli.config).expect("failed to read config");
+
     let env_filter = match std::env::var_os("RUST_LOG") {
         Some(_) => EnvFilter::try_from_default_env().expect("RUST_LOG is present but invalid"),
-        None => EnvFilter::new("info"),
+        None => EnvFilter::try_new(&config.log.level).unwrap_or_else(|e| {
```
mkulke: see comment above about gracios fallbacks

confidential-data-hub/hub/src/bin/ttrpc-cdh.rs
```diff
@@ -47,9 +47,16 @@ struct Cli {
...
 async fn main() -> Result<()> {
+    let cli = Cli::parse();
+
+    let (config, config_log) = config::read_config(cli.config).expect("failed to read config");
+
     let env_filter = match std::env::var_os("RUST_LOG") {
         Some(_) => EnvFilter::try_from_default_env().expect("RUST_LOG is present but invalid"),
-        None => EnvFilter::new("info"),
+        None => EnvFilter::try_new(&config.log.level).unwrap_or_else(|e| {
+            eprintln!("Invalid log level '{e}'. Using default log level 'info'.");
```
mkulke: see comment above about gracious fallbacks

attestation-agent/attestation-agent/src/bin/grpc-aa/main.rs
```diff
@@ -69,9 +69,26 @@ struct Cli {
...
+    let cli = Cli::parse();
+    let (config, config_log) = match cli.config_file {
+        Some(config_path) => {
+            let config = Config::try_from(&config_path[..])?;
+            let log = format!("Using config file: {config_path}");
+            (config, log)
+        }
+        None => {
+            let config = Config::default_with_kernel_cmdline();
+            let log ="No AA config file specified. Using default configuration and the kbs address will be read from kernel cmdline.".to_string();
```
mkulke: somehow rust fmt --check doesn't run on this?
```suggestion
            let log = "No AA config file specified. Using default configuration and the kbs address will be read from kernel cmdline.".to_string();
```

attestation-agent/attestation-agent/src/bin/grpc-aa/main.rs
```diff
@@ -69,9 +69,26 @@ struct Cli {
...
+            (config, log)
+        }
+    };
+
     let env_filter = match std::env::var_os("RUST_LOG") {
         Some(_) => EnvFilter::try_from_default_env().expect("RUST_LOG is present but invalid"),
-        None => EnvFilter::new("info"),
+        None => EnvFilter::try_new(&config.log.level).unwrap_or_else(|e| {
+            eprintln!("Invalid log level '{e}'. Using default log level 'info'.");
```
mkulke: see other comment about falling back on invalid messages

attestation-agent/attestation-agent/src/bin/ttrpc-aa.rs
```diff
@@ -87,9 +87,27 @@ pub fn start_ttrpc_service(aa: AttestationAgent) -> Result<HashMap<String, Servi
...
+            let log ="No AA config file specified. Using default configuration and the kbs address will be read from kernel cmdline.".to_string();
+            (config, log)
+        }
+    };
+
     let env_filter = match std::env::var_os("RUST_LOG") {
         Some(_) => EnvFilter::try_from_default_env().expect("RUST_LOG is present but invalid"),
-        None => EnvFilter::new("info"),
+        None => EnvFilter::try_new(&config.log.level).unwrap_or_else(|e| {
+            eprintln!("Invalid log level '{e}'. Using default log level 'info'.");
```
mkulke: see other comment about falling back on invalid messages

confidential-data-hub/hub/src/bin/config/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+    };
+
+    if let std::result::Result::Ok(env) =
+        env::var(CDH_DEFAULT_IMAGE_AUTHENTICATED_REGISTRY_CREDENTIALS)
+    {
+        info!("Read authenticated registry credentials URI from env: {env}");
+        config.image.authenticated_registry_credentials_uri = Some(env);
+    }
+
+    config.extend_credentials_from_kernel_cmdline()?;
```
mkulke, line 38: does the kernel cmdline override or extend the credentials config specified in configfile or env?

if it extends, does it make sense to do the same with env?

attestation-agent/attestation-agent/src/bin/grpc-aa/main.rs
```diff
@@ -69,9 +69,25 @@ struct Cli {
...
+            let log = format!("Using config file: {config_path}");
+            (config, log)
+        }
+        None => {
+            let config = Config::default_with_kernel_cmdline();
+            let log = format!("No AA config file specified. Using default configuration and the kbs address will be read from kernel cmdline.");
+            (config, log)
+        }
+    };
+
```
mythi: how about having a unified config/log init function between grpc and ttprc builds

confidential-data-hub/hub/src/bin/config/mod.rs
Xynnn007: This field is used to configure the credentials during CDH startup. It's used to retrieve resources from the KBS side at startup and is different from `authenticated_registry_credentials_uri`. However, I'm considering whether this configuration is still needed since I'm already using initdata.

mkulke: hmm, not sure I understand.

`authenticated_registry_credentials_uri` <= looks like a uri for kbs?
`.extend_credentials_from_kernel_cmdline()` <= also a uri for kbs?

Xynnn007: `authenticated_registry_credentials_uri` is a URI for KBS to get the image pull secret
`.extend_credentials_from_kernel_cmdline()` is a URI for KBS to get any resource

so my original question is whether it's acceptable to get rid of reading the things from env?

mkulke: > `.extend_credentials_from_kernel_cmdline()` is a URI for KBS to get any resource

interesting. I wasn't aware, is this resource injected to the workload as file or sealed secret?
 
> so my original question is whether it's acceptable to get rid of reading the things from env?

IMO yes. 2 caveats I can think of:
- RUST_LOG should still work as expected, we shouldn't break that
- the default location for the config file cannot be on a read-only filesystem (/etc) or we need a cli toggle --config-file that can override the default

Xynnn007: > interesting. I wasn't aware, is this resource injected to the workload as file or sealed secret?
> 
> >

It's [rejected as files into the guest rootfs](https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/hub/src/auth/kbs.rs#L23-L73) than workload pod/container.



> IMO yes. 2 caveats I can think of:
> 
> * RUST_LOG should still work as expected, we shouldn't break that
> * the default location for the config file cannot be on a read-only filesystem (/etc) or we need a cli toggle --config-file that can override the default

Now the code logic is as you mentioned. First check env `RUST_LOG`, then check if `--config-file` is given. if not, finally, use the default config.

fitzthum (approved):
LGTM

mythi (approved):
LGTM

PR #1357, opened 2026-03-02
agents: add AGENTS.md file

mkulke:
Equivalent to what we have in trustee's repo, so we can assign copilot to chores and have it create useful pull requests.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -128,7 +128,7 @@ subproject.
 It's unlikely that local development will happen in a TEE that is able to produce genuine
 attestation evidence. Therefore there are "sample" and a "sample-device" attester/verifier
 pairs that can be used as dummy stubs for testing end-to-end flow in local development and
-testing. We should be aware that there is a risk that that these dummy implementations are
+testing. We should be aware that there is a risk that that these dummy implementations and
 their evidence are accidentally used in production. We should make sure that in presence of
 real TEE, we don't default or fallback to sample evidence.
 


AGENTS.md
```diff
@@ -0,0 +1,161 @@
...
+repositories are developed and released in tandem, but they are decoupled and can be used
+independently. guest-components's kbs-protocol crate is consumed by Trustee in the kbs-client
+subproject.
+
+## Sample Verifiers
+
+It's unlikely that local development will happen in a TEE that is able to produce genuine
+attestation evidence. Therefore there are "sample" and a "sample-device" attester/verifier
+pairs that can be used as dummy stubs for testing end-to-end flow in local development and
+testing. We should be aware that there is a risk that that these dummy implementations are
```
dcmiddle:
```suggestion
testing. We should be aware that there is a risk that that these dummy implementations and
```

AGENTS.md
```diff
@@ -0,0 +1,161 @@
...
+strictly necessary for the core use cases.
+
+## API Stability
+
+The project is packaged and distributed downstream. Guest-Components do not provide
+machine-enforcable schemas for all its payloads and protocols yet, but we strive to avoid breakage
+for consumers when iterating. If it's unavoidable, we point it out expclicitly in commit messages,
+so it can be tracked and aggregated in release notes.
+
+## Relationship to Trustee
```
dcmiddle, line 117: I'm not sure what the implications of this are for the agent. Like if it changes a protocol in guest components it knows to make a parallel change in trustee? Or can the agent only act on this repo?

Similar consideration for the website docs vs local docs.

mkulke: at the moment to my knowledge the github copilot agent will not work across repos yet, although I'm pretty sure that it'll be able to work org-wide soon. an agents.md file is not bound to a specific agent technology though, it should be picked up claude code, codex, copilot, gemini etc.

The rationale behind this statement is to instruct an agent to avoid altering schemas. At the moment we rely on implicit contracts e.g. in the data layout of evidence, so it's easy to introduce changes that break parsing on the consumer side (trustee)

mythi (approved):
Thanks

PR #1363, opened 2026-03-03
Replace NVML bindings with NVAT bindings

fitzthum:
Switch from unmaintained NVML bindings to the official NVAT Rust bindings. See first commit message for some more details here. 

I will need to fiddle with the workflows a bit, so leaving this as a draft for now.

### DIFF
--- a/attestation-agent/attestation-agent/Cargo.toml
+++ b/attestation-agent/attestation-agent/Cargo.toml
@@ -76,7 +76,6 @@ all-attesters = [
     "se-attester",
     "cca-attester",
     "tpm-attester",
-    "nvidia-attester",
 ]
 tdx-attester = ["kbs_protocol?/tdx-attester", "attester/tdx-attester"]
 tdx-attester-libtdx = [
--- a/attestation-agent/attester/Cargo.toml
+++ b/attestation-agent/attester/Cargo.toml
@@ -76,7 +76,6 @@ all-attesters = [
     "cca-attester",
     "se-attester",
     "tpm-attester",
-    "nvidia-attester",
 ]
 
 # tsm-report enables a module that helps attesters to use Linux TSM_REPORTS for generating


attestation-agent/attestation-agent/Cargo.toml
```diff
@@ -76,7 +76,6 @@ all-attesters = [
     "se-attester",
     "cca-attester",
     "tpm-attester",
-    "nvidia-attester",
```
Xynnn007, line 79: probably this deletion is not intended?

attestation-agent/attester/src/nvidia/mod.rs
```diff
@@ -4,33 +4,61 @@
...
+
+static INIT: OnceLock<Result<()>> = OnceLock::new();
 
 const NVIDIA_NONCE_SIZE: usize = 32;
 
+fn ensure_sdk_init() -> Result<()> {
+    INIT.get_or_init(|| -> Result<()> {
+        let opts = SdkOptions::new()?;
+        let sdk = NvatSdk::init(opts)?;
+        std::mem::forget(sdk);
```
Xynnn007, line 28: This looks unusual. Just out of curiosity, do you have any document for this?

attestation-agent/attester/src/nvidia/mod.rs
```diff
@@ -4,33 +4,61 @@
...
 
 /// NRAS knows about "switch" and "gpu" but the expected evidence
 /// content is the same. nvidia-attester can compose a list of
 /// all CC enabled nvml/nscq devices using this evidence struct.
-#[derive(Serialize)]
+#[derive(Deserialize, Serialize)]
 struct NvDeviceReportAndCert {
-    arch: DeviceArchitecture,
-    uuid: String,
+    arch: Architecture,
```
Xynnn007, line 67: Note that this will break the current verifier side compatibility. https://github.com/confidential-containers/trustee/blob/main/deps/verifier/src/nvidia/mod.rs#L88-L93

Do you have a plan to update that side?

attestation-agent/attester/src/nvidia/mod.rs
```diff
@@ -110,3 +91,35 @@ impl Attester for NvAttester {
...
     }
 }
+
+/// Internal helper for getting evidence from NVIDIA devices.
+fn get_device_evidence(report_data: Option<[u8; 32]>) -> Result<Vec<NvDeviceReportAndCert>> {
+    ensure_sdk_init()?;
+
+    let nonce = match report_data {
+        Some(data_vec) => Nonce::from_hex(&hex::encode(data_vec))?,
+        None => Nonce::generate(32)?,
```
Xynnn007, line 116: Let's add a debug! or something logging here to tell that no nonce is given. This might be helpful for debugging.

attestation-agent/attester/Cargo.toml
```diff
@@ -22,9 +22,7 @@ crypto.path = "../deps/crypto"
 hex.workspace = true
 iocuddle = { version = "0.1.1", optional = true }
 kbs-types.workspace = true
-nvml-wrapper = { git = "https://github.com/rust-nvml/nvml-wrapper", rev = "7e0752f331", optional = true, default-features = false, features = [
-    "serde",
-] }
+nv-attestation-sdk = { git  = "https://github.com/NVIDIA/attestation-sdk", rev="0c1be386a8fbb8f2766a6a556d10df86f5fed9d3", optional = true }
```
Xynnn007: please take a toml format here.
btw, does nv-attestation-sdk have a plan to cut a release?

We can then use the release version than git+tag. This will not block the PR.

attestation-agent/attestation-agent/Cargo.toml
fitzthum: See last commit message. This would be temporary until a `nv-attestation-devel` package is released.

attestation-agent/attester/src/nvidia/mod.rs
fitzthum: I can add a comment. Basically the SDK is very c-like and expects to be initialized exactly once.

attestation-agent/attester/src/nvidia/mod.rs
fitzthum: Yeah, Trustee PPCIE PR has a default value for UUID. The Arch fields are compatible. I could add the default value here as well if we want them to be compatible.

attestation-agent/attester/src/nvidia/mod.rs
fitzthum: Ok. Note that the none case is just for when you call this function from the detect platform function where there is no nonce.

attestation-agent/attester/Cargo.toml
fitzthum: There is a tag in the repo we can use. The crate will hopefully be added to crates.io in the future.

attestation-agent/attester/src/nvidia/mod.rs
fitzthum: (Added comment)

attestation-agent/attester/src/nvidia/mod.rs
fitzthum: (Added UUID back and put a default value)

attestation-agent/attester/Cargo.toml
fitzthum: (Switched to using the tag)

attestation-agent/attester/src/nvidia/mod.rs
Xynnn007: Yes. What suprises me is the `std::mem::forget(sdk);`.

I get the document [here](https://docs.nvidia.com/attestation/nv-attestation-sdk-cpp/latest/sdk-rust/user_guide.html#sdk-initialization). I am not sure if we need to forget this object manually.

Did you test a case when a same NvidiaAttester struct's get report api be called sereval times? If it works, the writing style will dispel my doubts.

fitzthum: Yes, in fact it's called multiple times in every case, since we use it in the platform detection function and again when we get the evidence. I have also tested plenty of runs where it retries the request and it works fine. I am going to talk to the crate maintainers tho and see if they can implement a better API.

PR #1364, opened 2026-03-04
ci/codeql: add Rust to codeql scanning language list

Xynnn007:


mythi:
> The test can be run after getting this merged. like https://github.com/Xynnn007/guest-components/actions/runs/22656119510

I could not find the details but looks like there are some problems found:

```
Analysis produced the following metric data:

|                            Metric                            | Value |
+--------------------------------------------------------------+-------+
| Total number of Rust files that were extracted with errors   |     5 |
...
```

Issue #1235, opened 2025-12-16
RFC: kbs client: add support for more EC keys and make tee-pubkey selection configurable


mythi:
Triggered by https://github.com/confidential-containers/trustee/pull/1003.

Two things we could do:

1. Add P521 keypair in deps/crypto
2. allow users to select which tee-pubkey keypair  to use

The first one is rather straightforward and we could even switch to it by default (thoughts?) once implemented.  Even without 1., we currently have tee-pubkey implementations that users could select and use. It would also be good to be able to have e2e test coverage added for all of them.

Questions: is 2. useful, and if yes, what would be the best place to have it? Initial idea: CDH config:

```
[kbc]
tee_pubkey_algo: <names TBD> 
```

mythi:
Any feedback on changing the default

PR #1366, opened 2026-03-04
aa: make TeePubkey algorithm configurable and add P521 Curve to EC keys

mythi:
Fixes: #1235

Goals:
- existing behavior does not change
- add the enablers to improve Trustee JWE e2e testing beyond unit tests
- allow users to choose the most suitable algo for their needs


attestation-agent/attestation-agent/src/config/kbs.rs
```diff
@@ -8,13 +8,57 @@ use serde::Deserialize;
...
 #[derive(Clone, Debug, Deserialize, PartialEq)]
 pub struct KbsConfig {
     /// URL Address of KBS.
     pub url: String,
 
     /// Cert of KBS
     pub cert: Option<String>,
+
+    #[serde(default)]
+    pub tee_key_algorithm: TeeKeyAlgorithm,
```
Xynnn007, line 22: Maybe we can directly `pub use kbs_protocol::TeeKeyAlgorithm` without defining it again here. Sometimes we define two different structs with a same name but that's usually for two repos maintained by different orgs and to glue the same-name-different-usage structs. For `attestation-agent` and `kbs_protocol`, they two are in a same org and workspace, thus might be better to declare only once.

Xynnn007 (approved):
Only one nit. LGTM other parts. I am happy to see that there are other people interested in cryptography from CoCo!

attestation-agent/attestation-agent/src/config/kbs.rs
mythi: I thought about this too. No strong opinions here but this approach keeps `kbs_protocol` out of deserialization which is what user-facing AA logic needs to handle.

Xynnn007: imo it's acceptable for the struct of kbs-protocol to derive De-ser maros or so in kbs_protocol.

mythi: updated + rebased

PR #1369, opened 2026-03-05
cosign: strip newlines from image signatures

fitzthum:
Some signing implementations put newlines into the image signatures. The cosign cli seems to remove these automatically while validating, but the crate does not.

Let's remove these characters if they are present. This shouldn't affect valid signatures.

I add a test image with this sort of signature. This case fails without the change.


Xynnn007:
Looks like a cargo fmt is needed.

image-rs/src/signature/policy/cosign/mod.rs
```diff
@@ -151,6 +151,18 @@ impl CosignParameters {
             .trusted_signature_layers(&auth, &source_image_digest, &cosign_image)
             .await?;
 
+        // Some cosign implementations leave newlines in the signature. Strip these out.
+        let signature_layers: Vec<sigstore::cosign::SignatureLayer> = signature_layers
+            .iter()
+            .map(|layer| {
+                let mut cleaned = layer.clone();
+                if let Some(sig) = &cleaned.signature {
+                    cleaned.signature = Some(sig.replace(['\n', '\r'], ""));
```
mythi, line 160: would `.trim_end()` be enough?

fitzthum: The newlines are in the middle of the signature.

fitzthum:
@Xynnn007 @mythi ptal

mythi (approved):

PR #1375, opened 2026-03-09
workflows: Use IBM Actions Z runners for s390x CI jobs

BbolroC:
Switch all self-hosted runners labeled 's390x' to IBM Actions Z runners.

### DIFF
--- a/.github/workflows/aa_basic.yml
+++ b/.github/workflows/aa_basic.yml
@@ -34,13 +34,13 @@ jobs:
       matrix:
         instance:
           - ubuntu-24.04
-          - s390x
+          - ubuntu-24.04-s390x
           - ubuntu-24.04-arm
         include:
           - instance: ubuntu-24.04
             make_args: ""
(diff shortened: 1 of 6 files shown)


confidential-data-hub/hub/src/storage/drivers/luks2.rs
```diff
@@ -204,6 +204,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg_attr(target_arch = "s390x", ignore)]
```
Xynnn007, line 207: not a blocker: why it's not supported for this on ibm Z?

BbolroC: It works on VM-based runners, but IBM Actions runners are implemented using LXD containers, which impose several security restrictions. LUKS2 operations are blocked by AppArmor due to the reduced set of capabilities. The service team appears to be planning to introduce VM-based runners. In the meantime, let's skip the affected tests. Thanks!

Xynnn007: Thanks for the explainations!

PR #1380, opened 2026-03-11
Make CDH statically built-able again with replacing libcryptsetup-rs with binary calling

Xynnn007:
Currently, CDH is not statically compileable, and I'm afraid we'll need to release different binaries for each different distribution, even on the same architecture. This is because the same binary might cause issues running on different distributions' dynamic link libraries.

This PR aims to solve this problem. However, this requires https://github.com/confidential-containers/guest-components/pull/1344, thus I've marked as a WIP for now, for replacing some common functions.

### DIFF
--- a/.github/workflows/cdh_basic.yml
+++ b/.github/workflows/cdh_basic.yml
@@ -57,7 +57,7 @@ jobs:
 
       - name: Install cryptsetup and zfs
         run: |
-          sudo apt-get update && sudo apt-get install -y libcryptsetup-dev pkg-config zfsutils-linux
+          sudo apt-get update && sudo apt-get install -y cryptsetup pkg-config zfsutils-linux
 
       - name: Build and install
         run: |
@@ -72,4 +72,4 @@ jobs:
 
       - name: Run cargo test
         run: |
-          sudo -E PATH=$PATH -s RUST_LOG=warn cargo test --features kbs,aliyun,sev,bin,luks2 -p kms -p confidential-data-hub -- --nocapture
+          sudo -E PATH=$PATH -s RUST_LOG=warn cargo test --features kbs,aliyun,sev,bin -p kms -p confidential-data-hub -- --nocapture
--- a/.github/workflows/image_rs_build.yml
+++ b/.github/workflows/image_rs_build.yml
@@ -70,7 +70,8 @@ jobs:
       
       - name: Install cryptsetup
         run: |
-          sudo apt-get install -y libcryptsetup-dev pkg-config
+          sudo apt-get install -y cryptsetup pkg-config
+        if: matrix.instance == 'ubuntu-24.04'  || matrix.instance == 'ubuntu-24.04-arm'
 
       - name: Install cross-compliation support dependencies
         run: |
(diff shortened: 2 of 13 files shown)


Xynnn007:
@manuelh-dev @fitzthum PTAL.

btw. I found that the Kata side hard code the libc for GC with GNU. https://github.com/kata-containers/kata-containers/blob/main/tools/packaging/static-build/coco-guest-components/Dockerfile#L53

Also, do you think we should also raise an issue to track the statically built-version of nvidia attester?

fitzthum (approved):
LGTM 

do you think it makes sense to have a test for the binary before we try to use it? might be a little cleaner than having the commands fail

Xynnn007:
@fitzthum Yes, I added `which` to judge the existence of the binary.

mythi (commented):
`luks2` feature does not seem to be related to `libcryptsetup-rs` usage. there could be users who don't install `cryptsetup` and so prefer to keep `luks2` feature disabled as well?

Xynnn007:
> `luks2` feature does not seem to be related to `libcryptsetup-rs` usage. there could be users who don't install `cryptsetup` and so prefer to keep `luks2` feature disabled as well?

We do not need the feature, actually. The features are usually used to mark the "building time" dependencies, while in the current version we use command call and do not need any extra dependencies to facilate luks2. Before the command call, we will `which` the command, if no command is found, it will raise error during runtime.

mythi:
> > `luks2` feature does not seem to be related to `libcryptsetup-rs` usage. there could be users who don't install `cryptsetup` and so prefer to keep `luks2` feature disabled as well?
> 
> We do not need the feature, actually. The features are usually used to mark the "building time" dependencies, while in the current version we use command call and do not need any extra dependencies to facilate luks2. Before the command call, we will `which` the command, if no command is found, it will raise error during runtime.

But still, removing the feature is independent of making CDH built static. It's not related to the topic of PR was my comment.

Xynnn007:
> But still, removing the feature is independent of making CDH built static. It's not related to the topic of PR was my comment.

They are a same coin for both sides - if we get rid of `libcryptsetup-rs` dependency, we do not need the feature anymore. Thus they are two different commits.

fitzthum:
I think it makes sense to get rid of the feature. With these changes the LUKS code doesn't add any signficant overhead. If you don't want to use LUKS, just don't use it at runtime. Fewer features = simpler imo.

mythi:
I'm OK with removal. Could be isolated into one commit if it needs to be done in this PR.

Xynnn007:
> I'm OK with removal. Could be isolated into one commit if it needs to be done in this PR.

Already in a separate commit.

mythi:
> > I'm OK with removal. Could be isolated into one commit if it needs to be done in this PR.
> 
> Already in a separate commit.

It was scattered in multiple commits and my feedback was to get it in one.

Xynnn007:
> > > I'm OK with removal. Could be isolated into one commit if it needs to be done in this PR.
> > 
> > 
> > Already in a separate commit.
> 
> It was scattered in multiple commits and my feedback was to get it in one.

Oops I mis-read the comment. But I think it would not cause much trouble?

PR #1406, opened 2026-04-08
AA: bring initdata validation logic to Initdata validator

Xynnn007:
1. delete initdata checking logic from AA init logic
2. Add initdata-validator
3. Add initdata-validator publishment CI pipeline

### DIFF
--- a/.github/actions/publish-oras-artifact/action.yml
+++ b/.github/actions/publish-oras-artifact/action.yml
@@ -1,0 +1,65 @@
+name: Publish ORAS artifact
+description: Package a binary and publish as ORAS artifact with manifest
+
+inputs:
+  registry:
+    description: OCI registry hostname
+    required: true
+  image_name:
+    description: Repository image namespace
+    required: true
+  artifact_name:
+    description: Binary/artifact name
+    required: true
+  rust_target:
+    description: Rust target triple
+    required: true
+  git_sha:
+    description: Git SHA for tag
+    required: true
+  tee:
+    description: TEE platform suffix for tag
+    required: true
+  arch:
+    description: Build architecture (x86_64/aarch64/powerpc64le/...)
+    required: true
+  oci_arch:
+    description: OCI architecture for manifest annotation
+    required: true
+
+outputs:
+  image:
+    description: Published image name
+    value: ${{ steps.publish.outputs.image }}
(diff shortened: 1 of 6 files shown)


fitzthum:
Seems like this is similar to what @burgerdev and @Apokleos have worked on for decoupling

Xynnn007:
> Seems like this is similar to what @burgerdev and @Apokleos have worked on for decoupling

Yes. Here's the newest thread - if you're interested (I'm sure you must be : ) ) ptl

https://github.com/kata-containers/kata-containers/issues/11532#issuecomment-4152533560

fitzthum (commented):
This looks good. My one question is whether we should drop the first commit for now so that we have a smoother transition. Otherwise, as soon as we merge this, we will need to get everything else in place before we can bump gc. Maybe we keep existing functionality for now and remove it once we are actually using the new flow.

Xynnn007:
> This looks good. My one question is whether we should drop the first commit for now so that we have a smoother transition. Otherwise, as soon as we merge this, we will need to get everything else in place before we can bump gc. Maybe we keep existing functionality for now and remove it once we are actually using the new flow.

Good point. Let me make this as draft and I will take a test with the whole stack. If it works, I will turn it back to ready

burgerdev (commented):
Thanks for working on this! I think we will need to have both AA as validator and the additional validator binary for a transition period.

Xynnn007:
> Thanks for working on this! I think we will need to have both AA as validator and the additional validator binary for a transition period.

in my original plan, we can do the whole change togetherly without any mid-state. To achieve this aim, I will take the whole stack with also @Apokleos 's PR. This way we can avoid some clean up jobs - actually we will not use attestation-agent for initdata validation anymore. wdyt?

burgerdev:
@Xynnn007, we're following this plan: https://docs.google.com/document/d/1LEE6p0qL61MUCNQw3fh9dhpFeN-Uv4d4AuOBp4ftK34/edit?usp=sharing. We'll need the validator for phase 1, but can only skip processing in AA in phase 3. The reason for the phased introduction is to reduce review burden on Kata side and general safety of these changes.

Xynnn007:
Let's get this in and keep moving the work forward.

PR #1408, opened 2026-04-09
image-rs: fix signature validation of multi-arch images

fitzthum:
There are a number of ways to sign multi-arch images. One relatively common approach is to sign the manifest list digest. The manifest list is the map of manifests for various platforms. It contains the digest of each platform-specific manifest that it contains, so checking the digest of the manifest list digest implicitly checks the digest of whatever platform is used.

Unfortunately, image-rs does not support this mode. Instead, we always check the signature against the manifest digest. In a multi-arch image, the manifest digest is the digest of the platform-specific manifest. In cases where only the manifest list digest is signed, this will fail.

Fortunately, rust-oci-client was recently updated to report both the manifest digest and the manifest list digest when fetching a manifest. This PR adds a secondary check for multi-arch images. If the manifest digest does not match the signature, it will check the manifest list digest against the signature. If either one of these matches, the image is valid.

Tested this locally, but the test image is not public sadly. Can add one if anybody has an example.

Using a SHA for now. i will keep an eye on the `rust-oci-crate` to see when they cut a release.

cc @esposem

### DIFF
--- a/image-rs/Cargo.toml
+++ b/image-rs/Cargo.toml
@@ -28,7 +28,7 @@ hex = { workspace = true, optional = true }
 kbc = { path = "../attestation-agent/kbc", default-features = false, optional = true }
 loopdev = { git = "https://github.com/mdaffin/loopdev", rev = "c9f91e8f0326ce8a3364ac911e81eb32328a5f27" }
 nix = { workspace = true, features = ["mount", "fs"] }
-oci-client = { version = "0.15", default-features = false, optional = true }
+oci-client = { git = "https://github.com/oras-project/rust-oci-client", rev = "83092a67ebcaeed832eae090895a1c178d867344", default-features = false, optional = true }
 oci-spec = "0.9.0"
 ocicrypt-rs = { path = "../ocicrypt-rs", default-features = false, features = [
     "async-io",
@@ -154,8 +154,7 @@ signature-cosign-rustls = [
 signature-cosign-native = ["signature-cosign", "sigstore/native-tls"]
 
 oci-client-rustls = [
-    "oci-client/rustls-tls",
-    "oci-client/rustls-tls-native-roots",
+    "oci-client/rustls-tls"
 ]
 oci-client-native = ["oci-client/native-tls"]
 
--- a/image-rs/src/pull.rs
+++ b/image-rs/src/pull.rs
@@ -106,9 +106,11 @@ impl<'a> PullClient<'a> {
     }
 
     /// pull_manifest pulls an image manifest and config data.
-    pub async fn pull_manifest(&mut self) -> PullLayerResult<(OciImageManifest, String, String)> {
+    pub async fn pull_manifest(
+        &mut self,
+    ) -> PullLayerResult<(OciImageManifest, String, String, Option<String>)> {
         self.client
-            .pull_manifest_and_config(&self.reference, self.auth)
+            .pull_manifest_and_config_and_list_digest(&self.reference, self.auth)
             .await
             .map_err(|source| PullLayerError::PullManifestError { source })
     }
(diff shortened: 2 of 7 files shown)


image-rs/src/signature/image/mod.rs
```diff
@@ -46,6 +48,11 @@ impl Image {
         self.manifest_digest = Digest::try_from(digest)?;
         Ok(())
     }
+
+    pub fn set_manifest_list_digest(&mut self, digest: Option<&str>) -> Result<()> {
```
mkulke: looks a bit odd at first glance. should we decide at the caller site whether we want to set the manifest or not, depending on the whether we have a digest or not?

image-rs/src/signature/mod.rs
```diff
@@ -132,6 +133,12 @@ impl SignatureValidator {
             .set_manifest_digest(image_digest)
             .map_err(|_| SignatureError::IllegalImageDigest(image_digest.to_string()))?;
 
+        image
+            .set_manifest_list_digest(manifest_list_digest)
+            .map_err(|_| {
+                SignatureError::IllegalImageDigest(manifest_list_digest.unwrap().to_string())
```
mkulke: see comment above, if we let Some() the option here, we don't have to use unwrap()

image-rs/Cargo.toml
```diff
@@ -28,7 +28,7 @@ hex = { workspace = true, optional = true }
 kbc = { path = "../attestation-agent/kbc", default-features = false, optional = true }
 loopdev = { git = "https://github.com/mdaffin/loopdev", rev = "c9f91e8f0326ce8a3364ac911e81eb32328a5f27" }
 nix = { workspace = true, features = ["mount", "fs"] }
-oci-client = { version = "0.15", default-features = false, optional = true }
+oci-client = { git = "https://github.com/oras-project/rust-oci-client", rev = "83092a67ebcaeed832eae090895a1c178d867344", default-features = false, optional = true }
```
mkulke, line 31: it's a bit unfortunate that with those sha pins we will miss out on dependabot updates. maybe create an issue to bump to the new version once available (or wait for the next release?)

image-rs/src/signature/image/mod.rs
fitzthum: Ok, tweaked this a bit. We still need an option in the struct to capture if it is a multi-arch image, but now the setter just takes a `&str` rather than `Option<&str>`

image-rs/Cargo.toml
fitzthum: Yeah I would like to sneak this into the upcoming kata release. Not sure what the release cadence of the crate is. Interestingly we are currently two releases behind in our lockfile.

mkulke (approved):

image-rs/src/pull.rs
```diff
@@ -106,9 +106,11 @@ impl<'a> PullClient<'a> {
     }
 
     /// pull_manifest pulls an image manifest and config data.
-    pub async fn pull_manifest(&mut self) -> PullLayerResult<(OciImageManifest, String, String)> {
+    pub async fn pull_manifest(
```
Xynnn007, line 109: In some way the manifest and the oci index are not the same. So this function name is not accurate but the cause is not from image-rs. It's from the oci-client lib. Let's get this in and I will take a look if the underlying crate can have a more descriptive API definition for both manifest and index.

Xynnn007 (approved):

esposem:
A bit late but I can confirm this works for me also (tested locally). Thank you Tobin!

image-rs/src/pull.rs
fitzthum: oh the underlying crate has `pull_manifest_and_config_and_list_digest`. that name is so long though i didn't want to use it here too. `pull_manifest` already returns both manifest and config so i just kept that. but if you have any other ideas we can update

PR #1416, opened 2026-04-15
deps: switch from canon_json

fitzthum:
canon-json has some compatibility with corim-rs (which we want to use in Trustee). It's not particularly mature, so let's switch to this version which seems slightly better.


fitzthum:
Some of the image-pulling tests are failing. I suspect this is because of our recent changes to how we generate the test images.

fitzthum:
Well this took a lot of messing around to adjust for the changes to the test images, but I think it should be working now.

attestation-agent/kbc/src/offline_fs_kbc/aa-offline_fs_kbc-keys.json
```diff
@@ -1,5 +1,5 @@
 {
-  "default/key/key_id1": "cGFzc3BocmFzZXdoaWNobmVlZHN0b2JlMzJieXRlcyE=",
+  "default/key/key_id1": "REXwTCcCFObQdE9ZfL0a/iZtucCwBIo36/L0aL5l7UA=",
```
Xynnn007, line 2: Why we need to change the keys? Seeing that this PR only cares about canon json, which is only used in RCAR composite attestation handshake?

fitzthum: Last couple commits [here](https://github.com/confidential-containers/infra) overwrote all our test images with new ones. Not really related to this PR. This just happens to be the first one that hits it. I can split into two PRs if preferred.

Xynnn007: Oh ok. Ok to include in this PR and no need another

Xynnn007 (approved):
I am not sure whether we will meet other issues if we change content of `default/key/key_id1`, but this is the right direction for having infra transparent. Let's fix them when we met!

PR #1417, opened 2026-04-17
se-attester: Bind runtime data to IBM SEL attestation evidence

BbolroC:
This PR makes the following changes:
- Deserialize SeAttestationRequest from the nonce in AttestationRequest
- Inject runtime_data_digest directly into the request
- Serialize SeAttestationRequest again as primary_runtime_data
- Construct user_data from runtime_data_digest

This binds user_data to the measurement in the attestation evidence.

Depends-on: https://github.com/confidential-containers/trustee/pull/1284

Assisted-by: IBM Bob

### DIFF
--- a/attestation-agent/attester/src/se/mod.rs
+++ b/attestation-agent/attester/src/se/mod.rs
@@ -12,10 +12,26 @@ use pv::{
 };
 use serde::{Deserialize, Serialize};
 use serde_with::{base64::Base64, serde_as};
-use std::fs;
 use tracing::debug;
 
-const DIGEST_FILE: &str = "/run/peerpod/initdata.digest";
+const RUNTIME_DIGEST_SIZE: usize = 48; // SHA-384 digest size in bytes
+
+/// Structured user data for IBM SEL attestation
+/// Currently, only contains runtime data digest bound to the attestation measurement
+/// This could be extended in the future to include additional fields
+#[serde_as]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct UserData {
+    #[serde_as(as = "Base64")]
+    pub runtime_data_digest: Vec<u8>,
+}
+
+impl UserData {
+    /// Return runtime_data_digest as raw bytes for attestation command
+    pub fn to_bytes(&self) -> Vec<u8> {
+        self.runtime_data_digest.clone()
+    }
+}
 
 pub fn detect_platform() -> bool {
     misc::pv_guest_bit_set()
@@ -26,15 +42,20 @@ pub fn detect_platform() -> bool {
 #[derive(Debug, Serialize, Deserialize)]
 pub struct SeAttestationRequest {
     #[serde_as(as = "Base64")]
-    request_blob: Vec<u8>,
-    measurement_size: u32,
-    additional_size: u32,
+    pub request_blob: Vec<u8>,
+    pub measurement_size: u32,
+    pub additional_size: u32,
     #[serde_as(as = "Base64")]
-    encr_measurement_key: Vec<u8>,
+    pub encr_measurement_key: Vec<u8>,
     #[serde_as(as = "Base64")]
-    encr_request_nonce: Vec<u8>,
+    pub encr_request_nonce: Vec<u8>,
     #[serde_as(as = "Base64")]
-    image_hdr_tags: BootHdrTags,
+    pub image_hdr_tags: BootHdrTags,
+    /// Runtime data digest computed from the full JSON structure:
+    /// (tee-pubkey, nonce, additional-evidence)
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde_as(as = "Option<Base64>")]
+    pub runtime_data_digest: Option<Vec<u8>>,
 }
 
 #[repr(C)]
@@ -71,14 +92,51 @@ impl Attester for SeAttester {
             encr_measurement_key,
             encr_request_nonce,
             image_hdr_tags,
+            runtime_data_digest,
         } = request;
-        let mut user_data = vec![0];
-        if fs::metadata(DIGEST_FILE).is_ok() {
-            user_data = fs::read(DIGEST_FILE)?;
-        }
+
+        // Construct user_data with runtime_data_digest
+        let runtime_digest =
+            runtime_data_digest.ok_or_else(|| anyhow!("runtime_data_digest not provided"))?;
+
+        debug!(
+            "Using runtime_data_digest, length: {}, content: {:?}",
+            runtime_digest.len(),
+            runtime_digest
+        );
+
+        let runtime_digest = match runtime_digest.len() {
+            len if len > RUNTIME_DIGEST_SIZE => {
+                bail!(
+                    "Invalid runtime_data_digest length: expected {} bytes (SHA-384), got {} (too large)",
+                    RUNTIME_DIGEST_SIZE,
+                    len
+                );
+            }
+            len if len < RUNTIME_DIGEST_SIZE => {
+                debug!(
+                    "Padding runtime_data_digest from {} to {} bytes with zeros",
+                    len, RUNTIME_DIGEST_SIZE
+                );
+                let mut padded = runtime_digest;
+                padded.resize(RUNTIME_DIGEST_SIZE, 0);
+                padded
+            }
+            _ => runtime_digest, // Exact match, use as-is
+        };
+
+        let user_data = UserData {
+            runtime_data_digest: runtime_digest,
+        };
+
+        let user_data_bytes = user_data.to_bytes();
+
+        // user_data is used to produce the measurement (HMAC),
+        // meaning if the data is different, the measurement will be different.
+        // See: https://www.ibm.com/docs/en/linux-on-systems?topic=commands-pvattest
         let mut uvc: AttestationCmd = AttestationCmd::new_request(
             request_blob.into(),
-            Some(user_data.to_vec()),
+            Some(user_data_bytes.clone()),
             measurement_size,
             additional_size,
         )?;
@@ -91,7 +149,7 @@ impl Attester for SeAttester {
         let response: SeAttestationResponse = SeAttestationResponse {
             measurement: uvc.measurement().to_vec(),
             additional_data,
-            user_data,
+            user_data: user_data_bytes,
             cuid: *cuid,
             encr_measurement_key,
             encr_request_nonce,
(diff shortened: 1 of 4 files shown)


attestation-agent/attester/src/se/mod.rs
```diff
@@ -71,14 +110,55 @@ impl Attester for SeAttester {
...
+            "Using runtime_data_digest, length: {}, content: {:?}",
+            runtime_digest.len(),
+            runtime_digest
+        );
+
+        // Determine initdata_digest source:
+        // 1. Try reading from DIGEST_FILE first (peer-pods)
+        // 2. If initdata_digest is given in request, use it (baremetal)
+        // 3. Otherwise, None
+        let initdata = if fs::metadata(DIGEST_FILE).is_ok() {
```
Xynnn007, line 138: For other architectures, the initdata digest is written to the "hardware" (whether it's the MRCONFIGID register during VM initialization or the PCR of vtpm). For SE, we use a file inside guest for this, there would be a security risk, as the file could be tampered with during the VM's lifecycle, not like "hardware". Maybe we should highlight this part somewhere.

BbolroC: Good catch! Up until z16, yes, it is true. But from z17, a new feature called `retrievable secrets` is introduced, in which secrets are safely stored in ultravisor. So the initdata will be stored in add-secret-request (https://www.ibm.com/docs/en/linux-on-systems?topic=tasks-submit-secret) in a trusted area and fed into as an initdata annotation. In the end, it will be safely stored in ultravisor and retrieved<del>/generated to the file</del> in the attester's `bind_init_data()`. I will implement this in the near future. 😉 But, yes. I will create an issue to mention your comment. Thanks!

Xynnn007: Wow cool! It's excited to see some hardware support for this because it's more reliable.

mkulke: I think the digest needs to be signed by the TEE, is that the case for IBM SE? (like HOSTDATA, MRCONFIG or a TPM PCR)

BbolroC:
For reviewers, the change is verified by the current Kata nightly test for IBM SEL.

attestation-agent/attester/src/se/mod.rs
BbolroC: > I think the digest needs to be signed by the TEE, is that the case for IBM SE? (like HOSTDATA, MRCONFIG or a TPM PCR)

yes, via add-secret-request. 😉 Unless it is signed by a correct key, the data storing will be rejected by ultravisor. But, as I said, from z17 (latest one).

attestation-agent/kbs_protocol/src/client/rcar_client.rs
```diff
@@ -164,25 +168,66 @@ impl KbsClient<Box<dyn EvidenceProvider>> {
...
+                        debug!("Injected initdata_digest into SE attestation request");
+                    }
+
+                    // Serialize the modified request to pass to the attester
+                    serde_json::to_vec(&request)
+                        .context("Failed to serialize modified SeAttestationRequest")?
+                }
+                #[cfg(not(feature = "se-attester"))]
+                {
+                    runtime_digest
```
Xynnn007: This path seems un-accessible. Maybe we can panic here, or do some thing in the code, for maintainance?

Xynnn007 (approved):
Overall, LGTM. One question is that

the reason why not change `userdata: Vec<u8>` to `userdata: UserData` in the request body, is to keep compability?

attestation-agent/kbs_protocol/src/client/rcar_client.rs
BbolroC: Yeah, panicking looks reasonable because it shouldn't be reachable. I will update the PR. Thanks a lot!

BbolroC:
I've switched this PR to a draft again to prevent it from getting merged before I get a thumbs-up from @stevenhorsman. Thanks @Xynnn007 for the review!

attestation-agent/kbs_protocol/src/client/rcar_client.rs
```diff
@@ -164,25 +168,66 @@ impl KbsClient<Box<dyn EvidenceProvider>> {
...
+                        debug!("Injected initdata_digest into SE attestation request");
+                    }
+
+                    // Serialize the modified request to pass to the attester
+                    serde_json::to_vec(&request)
+                        .context("Failed to serialize modified SeAttestationRequest")?
+                }
+                #[cfg(not(feature = "se-attester"))]
+                {
+                    panic!("Tee::Se requires se-attester feature to be enabled")
```
stevenhorsman: I'm not a rust expert, but would `bail!` be more appropriate than `panic!` which crashes the process?

attestation-agent/attester/src/se/mod.rs
```diff
@@ -71,14 +110,55 @@ impl Attester for SeAttester {
...
-            user_data = fs::read(DIGEST_FILE)?;
-        }
+
+        // Construct user_data with runtime_data_digest (required) and initdata_digest (optional)
+        let runtime_digest =
+            runtime_data_digest.ok_or_else(|| anyhow!("runtime_data_digest not provided"))?;
+
+        debug!(
+            "Using runtime_data_digest, length: {}, content: {:?}",
+            runtime_digest.len(),
```
stevenhorsman: Should the runtime digest only ever be from sha-384? Is it worth adding a len == 48 check here if so?

stevenhorsman (approved):
I'm definitely not an expert on attestation, but this seems okay to me. Just left a couple of more general comments/questions

attestation-agent/kbs_protocol/src/client/rcar_client.rs
BbolroC: yeah, I will change it to bail!(). Thanks!

attestation-agent/attester/src/se/mod.rs
BbolroC: Yep, I will add the check. Thanks!

attestation-agent/attester/src/se/mod.rs
BbolroC: > For other architectures, the initdata digest is written to the "hardware" (whether it's the MRCONFIGID register during VM initialization or the PCR of vtpm). For SE, we use a file inside guest for this, there would be a security risk, as the file could be tampered with during the VM's lifecycle, not like "hardware". Maybe we should highlight this part somewhere.

Created https://github.com/confidential-containers/guest-components/issues/1418 Thanks!

BbolroC:
I would like to postpone the merge until https://github.com/confidential-containers/trustee/pull/1284 gets a final thumbs-up. It is because there could be a chance to update this PR based on the change from Trustee. Thanks!

attestation-agent/attester/src/se/mod.rs
```diff
@@ -71,14 +110,62 @@ impl Attester for SeAttester {
...
+
+        if runtime_digest.len() != 48 {
+            bail!(
+                "Invalid runtime_data_digest length: expected 48 bytes (SHA-384), got {}",
+                runtime_digest.len()
+            );
         }
+
+        // Determine initdata_digest source:
+        // 1. Try reading from DIGEST_FILE first (peer-pods)
```
fitzthum, line 135: I don't think we want any peer-pods specific logic in an attester.

Last I checked the peer pods flow for binding the init-data was ad-hoc and only supported on some platforms. The solution to that is to make sure that all peer pods images are correctly calling the attestation agent (or the upcoming init-data validator component) to bind the init-data.

attestation-agent/attester/src/se/mod.rs
fitzthum: Hmmm. Yeah at first glance combining the runtime-data and the init-data into user-data is a red flag. runtime-data / user-data have different properties than init-data.

afaik there is no way to support init-data on z16 which is why we haven't had proper init-data support.

so z17 introduces secret sealing. this PR doesn't seem to be interacting with those secrets directly. i guess they are automatically provisioned to the guest fs or some other component is putting them in place. either way, I'm not sure that this helps us with init-data.

a key property of init-data is that it cannot be modified from inside the guest. for instance, if a workload escapes from a container into the podvm, the init-data will have already been set and can't be modified. thus, any future attestation requests coming from that podvm will have the same init-data. 

i don't think this holds even with securely injected secrets. sure, we can inject a secret, but if the guest is compromised, it doesn't have to use it, and it doesn't seem like there is any way for the verifier to check if the correct init-data is being used.

attestation-agent/attester/src/se/mod.rs
```diff
@@ -91,7 +178,7 @@ impl Attester for SeAttester {
         let response: SeAttestationResponse = SeAttestationResponse {
             measurement: uvc.measurement().to_vec(),
```
Xynnn007, line 154: Could you add a link here about how `uvc.measurement()` is derived? Concretely I wonder if the `user_data_bytes` influences the final shape, s.t. different `user_data_bytes` gets different `uvc.measurement()`?

attestation-agent/attester/src/se/mod.rs
```diff
@@ -71,14 +110,62 @@ impl Attester for SeAttester {
...
+                "Invalid runtime_data_digest length: expected 48 bytes (SHA-384), got {}",
+                runtime_digest.len()
+            );
         }
+
+        // Determine initdata_digest source:
+        // 1. Try reading from DIGEST_FILE first (peer-pods)
+        // 2. If initdata_digest is given in request, use it (baremetal)
+        // 3. Otherwise, None
+        let initdata = if fs::metadata(DIGEST_FILE).is_ok() {
```
Xynnn007: Let's also add a note here to the issue link, i.e.
```rust
// See https://github.com/confidential-containers/guest-components/issues/1418
```

and also if the initdata file reading is ok, raise a `warning` to simply tell something like "initdata is not hardware binded and may be compromised. Be careful to use it" or if you like, add the link issue also.

attestation-agent/attester/src/se/mod.rs
BbolroC: > Yeah at first glance combining the runtime-data and the init-data into user-data is a red flag. runtime-data / user-data have different properties than init-data.

> i don't think this holds even with securely injected secrets. sure, we can inject a secret, but if the guest is compromised, it doesn't have to use it, and it doesn't seem like there is any way for the verifier to check if the correct init-data is being used.

Let me convince you why the retrievable secret works like other TEEs on Slack first, and leave a comment we both agree with here. Thanks

attestation-agent/attester/src/se/mod.rs
BbolroC: This means we don't need the 1st arm (I keep this because of backword-compatibility) to check the init data any more? Could you point me where your remark below is implemented/discussed on the CAA side?

> The solution to that is to make sure that all peer pods images are correctly calling the attestation agent (or the upcoming init-data validator component) to bind the init-data.

attestation-agent/attester/src/se/mod.rs
BbolroC: Yeah, the user data is used to create a measurement (HMAC) by ultravisor. So if a different user data is used, a different measurement is created. I will add a link or a comment for this. Thanks!

attestation-agent/attester/src/se/mod.rs
BbolroC: Sure! Thanks!

attestation-agent/attester/src/se/mod.rs
```diff
@@ -71,14 +110,62 @@ impl Attester for SeAttester {
...
+            runtime_digest.len(),
+            runtime_digest
+        );
+
+        if runtime_digest.len() != 48 {
+            bail!(
+                "Invalid runtime_data_digest length: expected 48 bytes (SHA-384), got {}",
+                runtime_digest.len()
+            );
         }
```
Xynnn007: Usually we will do padding here to the target length of the platform. Raise warning if the length is not expected. See [TDX](https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/attester/src/tdx/mod.rs#L123-L127) for example

attestation-agent/attester/src/se/mod.rs
mkulke: > yes, via add-secret-request. 😉 Unless it is signed by a correct key, the data storing will be rejected by ultravisor. But, as I said, from z17 (latest one).

ok, and we would verify that signature on Trustee? (so it would be like SNP where hostdata is part of the signed attestation report)

BbolroC: After discussing with Tobin, I’ve decided to drop the initdata part for now, since there is no safety mechanism for initdata on z16 (which the current implementation targets), unlike other TEEs. Initdata will be handled in a future implementation targeting z17.

> ok, and we would verify that signature on Trustee? (so it would be like SNP where hostdata is part of the signed attestation report)

I will raise an issue for the implementation plan for z17 and cover your questions there. Thanks!

BbolroC:
The verifier [side](https://github.com/confidential-containers/trustee/pull/1284) will be updated soon. Thanks!

BbolroC:
@fitzthum  @Xynnn007 Could you review this again? The initdata part is dropped. Thanks a lot!

attestation-agent/kbs_protocol/src/client/rcar_client.rs
```diff
@@ -164,25 +168,48 @@ impl KbsClient<Box<dyn EvidenceProvider>> {
...
         // Calculate the runtime data for the primary attester, which includes
         // the device evidence retrieved above.
+        let runtime_data_json = json!({
+            "tee-pubkey": runtime_data.tee_pubkey,
+            "nonce": runtime_data.nonce,
+            "additional-evidence": additional_evidence,
+        });
+        let runtime_data_serialized = serialize_json_canonically(&runtime_data_json)
+            .context("serialize runtime data failed")?;
+        let runtime_digest = hash_algorithm.digest(&runtime_data_serialized);
```
fitzthum: I think I might prefer keeping this inside the match, but if you move it here can you add `primary` to these variable names to be clear.

fitzthum (commented):
Ok, good to decouple this from the z17 stuff. One comment on the style, but otherwise seems fine.

attestation-agent/kbs_protocol/src/client/rcar_client.rs
BbolroC: Yep, I will update the PR. Thanks for the feedback!

attestation-agent/attester/src/se/mod.rs
```diff
@@ -5,17 +5,37 @@
...
+    pub runtime_data_digest: Vec<u8>,
+}
+
+impl UserData {
+    /// Serialize to Base64-encoded bytes for attestation command
+    pub fn to_bytes(&self) -> Vec<u8> {
+        // Encode to Base64 to ensure UTF-8 compatibility
+        general_purpose::STANDARD
+            .encode(&self.runtime_data_digest)
+            .into_bytes()
```
Xynnn007: a naive question:

Looks like the `to_vec()`-processed UserData is only used as inputs to `AttestationCmd::new_request`. And we base64 encode it here. So does `AttestationCmd::new_request` requires a base64-encoded byte vector - or it does not? What I mean is that if it does not require, we can directly use `runtime_data_digest` (type `Vec<u8>`) here.

Xynnn007 (approved):
LGTM. only one design question for optimization

attestation-agent/attester/src/se/mod.rs
BbolroC: It is just for backward-compatibility for the following https://github.com/confidential-containers/trustee/blob/21ce8d9bb2c3e4f0ba2df5eb5e003f97bca95713/deps/verifier/src/se/ibmse.rs#L220

Otherwise, an error will return on the trustee side.

BbolroC: Regarding the error I mentioned above, it occurred when concatenating runtime_data_digest and init_data_digest. Now we use runtime_data_digest only. No error would be expected. Let me update the PR and test it with the current Trustee. Thanks!

BbolroC:
The Trustee currently used in Kata Containers returns the following error like:

```
ERROR kbs::error: AttestationError(RcarAttestFailed { source: verify TEE evidence failed

Caused by:
    Verifier evaluate failed: invalid utf-8 sequence of 1 bytes from index 0 })
```

which is expected (check out https://github.com/confidential-containers/guest-components/pull/1417#discussion_r3122005564). For the next update in Kata, https://github.com/confidential-containers/trustee/pull/1284 should be included along with this. Thanks!

fitzthum (approved):
LGTM

This looks ready to merge

PR #1425, opened 2026-04-27
ci: release AA on TDX with musl

Xynnn007:
Currently the TDX attester can be built with musl. Thus we change the published AA binary with musl to run on different distros.


fitzthum (approved):
LGTM

mkulke (approved):

mythi (approved):

mythi:
> Currently the TDX attester can be built with musl. Thus we change the published AA binary with musl to run on different distros.

offtopic, are we getting closer to drop `tdx-attest-dcap-ioctls`?

PR #1423, opened 2026-04-23
chore(deps): update openssl to 0.10.78 to fix vulnerabilities

Xynnn007:

PR #1431, opened 2026-04-29
chore(deps): update openssl to 0.10.78 to fix vulnerabilities

supritai:
Pin workspace [`openssl`](Cargo.toml:43) to [`0.10.78`](Cargo.toml:43) to avoid resolving vulnerable `rust-openssl` versions.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ jwt-simple = { version = "0.12", default-features = false, features = [
 ] }
 kbs-types = "0.15.0"
 nix = "0.31"
-openssl = "0.10"
+openssl = "0.10.78"
 prost = "0.14"
 protobuf = "3.7.2"
 rand = "0.10.0"


fitzthum (commented):
I think this won't let us bump past 0.11.0, which could be prolbmeatic in the future. You may want to use `>= 0.10.78`

Cargo.toml
```diff
@@ -40,7 +40,7 @@ jwt-simple = { version = "0.12", default-features = false, features = [
 ] }
 kbs-types = "0.15.0"
 nix = "0.31"
-openssl = "0.10"
+openssl = "0.10.78"
```
stevenhorsman, line 43: You are missing the actual cargo update here, please run `cargo update -p openssl`. Plus do you need to update the Cargo.toml at all?

fitzthum: I think @Xynnn007 already bumped this in the lockfile. I guess you might want to specify this if you are building without the lockfile, but cargo should pick the newest version anyway

stevenhorsman: oh ok - I'm not sure I really understand the need for this change then

Xynnn007:
We have updated `Cargo.lock` due to https://github.com/confidential-containers/guest-components/pull/1423. I think it's the same as you required. Due  https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html, `0.10.78` and `0.10` can be same when we have `Cargo.lock`

supritai:
> We have updated `Cargo.lock` due to #1423. I think it's the same as you required. Due https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html, `0.10.78` and `0.10` can be same when we have `Cargo.lock`

Yes, right. Some packages in cargo.toml had exact versions so thought to add it.

supritai:
Closing it as https://github.com/confidential-containers/guest-components/pull/1441 addresses upcoming vulnerability as well

PR #1432, opened 2026-04-29
protos: Add missing license to crate

stevenhorsman:
protos is missing a license, so causes cargo deny to fail on it's license check if image-rs is in a dependency.


fitzthum (approved):
LGTM. Thanks

PR #1436, opened 2026-05-01
deps: Remediate two CVEs by updating astral-tokio-tar to 0.6.1

stevenhorsman:
Another security vuln, which dependabot doesn't track, but the OpenSSF tool does:

Update astral-tokio-tar from 0.6.0 to 0.6.1, remediating
- RUSTSEC-2026-0112
- RUSTSEC-2026-0113

Assisted-by: IBM Bob


fitzthum (approved):
LGTM

PR #1440, opened 2026-05-05
image-rs: introduce pathrs (incrementally)

fitzthum:
Pathrs is a handy crate that allows you to perform fs operations atomically and constrain them to a particular directory. We should probably use this crate in many places in image-rs, but let's start small.

The hardlink fallback method is a great fit for pathrs. Previously we were anchoring and canonicalizing the paths, but pathrs will do both of those things automatically. Since pathrs behaves like chroot, it handles absolute paths elegantly without needing any additional anchoring logic.

We are constraining hardlinks to the layer dir. This means that hard links cannot be created between layers. The image spec is slightly ambiguous about this (it says that this kind of link "may" fail). This should be fine.

### DIFF
--- a/image-rs/src/stream/unpack.rs
+++ b/image-rs/src/stream/unpack.rs
@@ -7,6 +7,7 @@ use filetime::FileTime;
 use futures::StreamExt;
 use nix::libc::timeval;
 use nix::sys::stat::{mknod, Mode, SFlag};
+use pathrs::InodeType;
 use thiserror::Error;
 use tracing::{debug, warn};
 
@@ -93,6 +94,12 @@ pub enum UnpackError {
         #[source]
         source: anyhow::Error,
         path: String,
+    },
+
+    #[error("pathrs error: {source}")]
+    PathRsFailed {
+        #[source]
+        source: pathrs::error::Error,
     },
 }
 
@@ -185,7 +192,7 @@ async fn convert_whiteout(
 pub async fn unpack<R: AsyncRead + Unpin>(input: R, destination: &Path) -> UnpackResult<()> {
     if destination.exists() {
         warn!(
-            "unpack destination {destination:?} already exists, will delete and rerwrite the layer",
+            "unpack destination {destination:?} already exists, will delete and rewrite the layer",
         );
         fs::remove_dir_all(destination)
             .await
@@ -255,9 +262,8 @@ pub async fn unpack<R: AsyncRead + Unpin>(input: R, destination: &Path) -> Unpac
             continue;
         }
 
-        match file.unpack_in(destination).await {
-            Ok(_) => {}
-            Err(e) => match try_hardlink_fallback(&kind, &mut file, destination).await {
+        if let Err(e) = file.unpack_in(destination).await {
+            match try_hardlink_fallback(&kind, &mut file, destination).await {
                 Ok(true) => {}
                 Ok(false) => return Err(UnpackError::UnpackFailed { source: e }),
                 Err(f) => {
@@ -267,7 +273,7 @@ pub async fn unpack<R: AsyncRead + Unpin>(input: R, destination: &Path) -> Unpac
                         )),
                     })
                 }
-            },
+            }
         }
 
         let path = format!(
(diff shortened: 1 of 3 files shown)


image-rs/src/stream/unpack.rs
```diff
@@ -94,6 +95,12 @@ pub enum UnpackError {
         source: anyhow::Error,
         path: String,
     },
+
+    #[error("pathrs error")]
```
Xynnn007:
```suggestion
    #[error("pathrs error: {source}")]
```

Xynnn007 (commented):
1 st review

Let's change the unit test to cover both escape and absolute path case
```
diff --git a/image-rs/src/stream/unpack.rs b/image-rs/src/stream/unpack.rs
index 69a5aa80..ac5fd302 100644
--- a/image-rs/src/stream/unpack.rs
+++ b/image-rs/src/stream/unpack.rs
@@ -96,7 +96,7 @@ pub enum UnpackError {
         path: String,
     },
 
-    #[error("pathrs error")]
+    #[error("pathrs error: {source}")]
     PathRsFailed {
         #[source]
         source: pathrs::error::Error,
@@ -467,6 +467,8 @@ mod tests {
     };
     use tokio_tar::{Builder, EntryType, Header};
 
+    use rstest::rstest;
+
     use super::*;
 
     #[tokio::test]
@@ -613,21 +615,33 @@ mod tests {
         assert!(unpack(data.as_slice(), destination).await.is_ok());
     }
 
+    #[rstest]
+    #[case::absolute_link_target("/etc/os-release", true)]
+    #[case::relative_escape_link_target("../../etc/os-release", false)]
     #[tokio::test]
-    async fn test_unpack_rejects_escaping_absolute_hardlink() {
+    async fn test_unpack_hardlink_tar_cases(#[case] link_name: &str, #[case] expect_ok: bool) {
         let td = tempfile::tempdir().unwrap();
-        let outside_file = td.path().join("outside_file");
-        let mut f = File::create(&outside_file).await.unwrap();
-        f.write_all(b"outside-data").await.unwrap();
+        let mut ar = Builder::new(Vec::new());
+
+        // 1. Create a file, write content, add it to the archive as the link target.
+        let source_file = td.path().join("os-release");
+        let mut f = File::create(&source_file).await.unwrap();
+        f.write_all(b"test-data\n").await.unwrap();
         f.flush().await.unwrap();
+        ar.append_file(
+            "etc/os-release",
+            &mut File::open(&source_file).await.unwrap(),
+        )
+        .await
+        .unwrap();
+
+        // 2. Append a hardlink entry with case-provided link target.
+        let link_entry_path = "etc/os-release.hardlink";
 
-        // linkname = "/../outside_file" will be anchored in fallback mode as
-        // destination/../outside_file, which points to the parent directory of the destination.
-        let mut ar = Builder::new(Vec::new());
         let mut hdr = Header::new_gnu();
         hdr.set_entry_type(EntryType::Link);
-        hdr.set_link_name("/../outside_file").unwrap();
-        hdr.set_path("subdir/evil_hardlink").unwrap();
+        hdr.set_link_name(link_name).unwrap();
+        hdr.set_path(link_entry_path).unwrap();
         hdr.set_size(0);
         hdr.set_mode(0o644);
         hdr.set_uid(0);
@@ -637,24 +651,27 @@ mod tests {
         ar.append(&hdr, empty()).await.unwrap();
 
         let data = ar.into_inner().await.unwrap();
-
         let destination = td.path().join("dest");
         let res = unpack(data.as_slice(), &destination).await;
 
-        // Expectation: unpacking should fail due to anti-escape check
-        match res {
-            Err(UnpackError::UnpackFailed { source }) => {
-                assert!(
-                    source.to_string().contains("pathrs create"),
-                    "unexpected error: {source}"
-                );
-            }
-            Ok(_) => panic!("unpack unexpectedly succeeded; anti-escape check failed"),
-            Err(e) => panic!("unexpected error variant: {e:?}"),
+        // 3. Assert unpack outcome and resulting paths under destination.
+        let hardlink_path = destination.join(link_entry_path);
+        if expect_ok {
+            res.unwrap();
+            let target = destination.join("etc/os-release");
+            let target_meta = fs::metadata(&target).await.unwrap();
+            let hardlink_meta = fs::metadata(&hardlink_path).await.unwrap();
+            assert!(target_meta.is_file() && hardlink_meta.is_file());
+            assert_eq!(target_meta.ino(), hardlink_meta.ino());
+        } else {
+            assert!(
+                res.is_err(),
+                "unpack should fail when hardlink target resolves outside the layer root: {res:?}"
+            );
+            assert!(
+                fs::metadata(&hardlink_path).await.is_err(),
+                "hardlink entry must not be created under destination"
+            );
         }
-
-        // Confirm that the escaping hardlink was not created
-        let evil_path = destination.join("subdir/evil_hardlink");
-        assert!(fs::metadata(&evil_path).await.is_err());
     }
 }

```

fitzthum:
> Let's change the unit test to cover both escape and absolute path case

Added

Xynnn007 (approved):

PR #1445, opened 2026-05-06
attestation-agent: improve build info

fitzthum:
The attestation-agent reports supported features when it starts up, but our logic for calculating these features is out of date.

Add the additional attester features and add another list of attesters that are always supported (e.g. the sample attester).

### DIFF
--- a/attestation-agent/attestation-agent/build.rs
+++ b/attestation-agent/attestation-agent/build.rs
@@ -14,24 +14,37 @@ fn main() -> std::io::Result<()> {
         use std::path::Path;
 
         // generate an `intro` file that includes the feature information of the build
-        fn feature_list(features: Vec<&str>) -> String {
-            let enabled_features: Vec<&str> = features
+        fn feature_list(always_enabled: Vec<&str>, features: Vec<&str>) -> String {
+            let enabled_features: Vec<&str> = always_enabled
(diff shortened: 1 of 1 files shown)


attestation-agent/attestation-agent/build.rs
```diff
@@ -14,24 +14,36 @@ fn main() -> std::io::Result<()> {
...
+                "AZ_SNP_VTPM_ATTESTER",
+                "AZ_TDX_VTPM_ATTESTER",
+                "SNP_ATTESTER",
+                "SE_ATTESTER",
+                "CCA_ATTESTER",
+                "CSV_ATTESTER",
+                "HYGON_DCU_ATTESTER",
+                "TPM_ATTESTER",
+                "NVIDIA_ATTESTER",
+            ],
```
Xynnn007, line 46: just btw and I was wondering if we could make the prompt appear as "tdx,csv" instead of "TDX_ATTESTER,CSV_ATTESTER".

Xynnn007 (approved):

attestation-agent/attestation-agent/build.rs
fitzthum: Ok, I stripped off `_ATTESTER`.

PR #1447, opened 2026-05-07
deps/kms: replace p12, yasna and ring with openssl

### LINKED ISSUE #1446: [CDH] Replace `p12` and `yasna` with other pkcs12 crates
Now alibaba cloud KMS plugin uses `p12` and `yasna` to facilate pkcs12 abilities, but `p12` is too old and we'd better to find a new crate to do this.

Xynnn007:
In history, we use ring dependency for static building, but now openssl supports static building. Thus we replace ring, yasna and p12 crate with openssl.

Another benefit is that p12 is a very old crate that lacks maintaince. Switching to openssl would provide security and fixes.

close #1446

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -32,13 +32,13 @@ dependencies = [
 
 [[package]]
 name = "aes"
-version = "0.9.0-rc.4"
+version = "0.9.0"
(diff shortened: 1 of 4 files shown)


fitzthum (approved):
LGTM

Issue #1433, opened 2026-04-29
[CDH] discussion on suspected issues with integrity-protected block device encryption


manuelh-dev:
I am opening this issue as part of my ongoing investigation on the CDH ephemeral storage feature, leveraged by kata-containers. When using this feature in kata-containers with integrity-protection enabled, we observe AEAD errors in scenarios with parallel filesystem activity. I have previously opened an issue for the kata-containers repository: https://github.com/kata-containers/kata-containers/issues/12721.

After investigating a bit closer, I realized that I am able to reproduce this issue without the kata specifics within the guest-components repository and made some more specific findings for the integrity-enabled encryption path. I would like to discuss these findings and utilize this input for potential changes to CDH if we find alignment.

To illustrate this issue, I created the following branch in my guest-components fork: https://github.com/manuelh-dev/guest-components/commits/mahuber/luks-parallel-aead/. There are 4 commits in the current intentional order, please see the relevant commit messages for more details.

1. commit: add a test case which exercises a test similar to existing tests in `blockdevice/mod.rs` but *using a loop device*. This test failed for me - for this case I authored commit 2.
2. commit: change `format_integrity_compatible` in `storage/drivers/filesystem.rs‎` to use a fallback when direct I/O fails. Direct I/O works well for scenarios like kata-containers where CDH runs in a VM and where we attach block devices appearing as real disks. However, when loop devices backed by host files are being used, this won't work. This unblocks the test from commit 1 and lets us author a new test with parallel filesystem activity, see the next commit.
3. commit: add a test case which writes to the integrity protected ephemeral storage via parallel workers (and another test which writes with a single worker). This (multi-worker) test fails. *This is the important test which reproduces the integrity errors reported in Kata's issue 12721*. We can now exercise this test thanks to the second commit. This test would *not* fail using the simple `tempfile::NamedTempFile` approach, but since we use a loop device exposed as a block device, the errors observed in the kata scenario reproduce. The failures in dmesg are found at the bottom of this post [1].
4. commit: Disable lazy inode initialization in the ext4 integrity format path. This unblocks the failing test. To show that this change would work on kata-containers side, I have just opened a draft PR for which the adapted, targeted test cases successfully pass CI: https://github.com/kata-containers/kata-containers/pull/12953.

What is your take on this? 

[...]

PR #1451, opened 2026-05-07
cdh: fix ext4 formatting for integrity-protected LUKS2 devices

manuelh-dev:
CDH formats LUKS2 devices with dm-integrity enabled and no full integrity wipe. In that mode, blocks only receive valid integrity tags after they have been written.

The existing tests covered the LUKS2 integrity formatting path through a temporary-file backed device, but not through the loop/block-device path that is closer to CDH secure_mount usage. This PR adds loop-device coverage and ext4 stress tests. The parallel stress test reproduces I/O errors from ext4 inode table metadata activity on a freshly formatted integrity-protected device.

The fix defaults ext4 formatting to `lazy_itable_init=0`, so inode table blocks are initialized during `mkfs.ext4`. Explicit caller-provided `lazy_itable_init` settings are preserved.

This follows the investigation from kata-containers/kata-containers#12721 and aligns with the temporary Kata-side validation in kata-containers/kata-containers#12953.

Solves: #1433

### DIFF
--- a/confidential-data-hub/hub/src/storage/drivers/filesystem.rs
+++ b/confidential-data-hub/hub/src/storage/drivers/filesystem.rs
@@ -114,13 +114,20 @@ impl FsFormatter {
             )?;
         }
 
-        // then do original format command
-        self.format(device_path)?;
+        // then do original format command with integrity-compatible defaults
+        let format_args = match self.fs_type {
+            FsType::Ext4 => disable_lazy_itable_init(&self.args),
+        };
+        self.format_with_args(device_path, &format_args)?;
 
         Ok(())
     }
 
     pub fn format(&self, device_path: &str) -> Result<()> {
+        self.format_with_args(device_path, &self.args)
+    }
+
+    fn format_with_args(&self, device_path: &str, format_args: &[String]) -> Result<()> {
         let command = match self.fs_type {
             FsType::Ext4 => {
                 if !is_ext4_installed() {
@@ -134,7 +141,7 @@ impl FsFormatter {
         if self.force {
             args.push("-F");
         }
-        for arg in &self.args {
+        for arg in format_args {
             args.push(arg);
         }
 
@@ -142,6 +149,37 @@ impl FsFormatter {
 
         Ok(())
     }
+}
+
+// Keep explicit caller choices, but default ext4 to eager inode table initialization.
+fn disable_lazy_itable_init(args: &[String]) -> Vec<String> {
+    let mut args = args.to_vec();
+    if args.iter().any(|arg| arg.contains("lazy_itable_init")) {
+        return args;
+    }
+
+    warn!(
+        "defaulting ext4 lazy_itable_init=0 for LUKS2 dm-integrity; set it explicitly in mkfsOpts to override"
+    );
+
+    if let Some(index) = args.iter().rposition(|arg| arg == "-E") {
+        if let Some(options) = args.get_mut(index + 1) {
+            options.push_str(",lazy_itable_init=0");
+            return args;
+        }
+    }
+
+    if let Some(options) = args
+        .iter_mut()
+        .rfind(|arg| arg.starts_with("-E") && arg.len() > 2)
+    {
+        options.push_str(",lazy_itable_init=0");
+        return args;
+    }
+
+    args.push("-E".to_string());
+    args.push("lazy_itable_init=0".to_string());
+    args
 }
 
 fn is_ext4_installed() -> bool {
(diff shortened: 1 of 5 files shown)


Xynnn007 (approved):

confidential-data-hub/hub/src/storage/drivers/filesystem.rs
```diff
@@ -144,6 +148,33 @@ impl FsFormatter {
     }
 }
 
+// Keep explicit caller choices, but default ext4 to eager inode table initialization.
+fn ext4_args_with_lazy_itable_init_disabled(args: &[String]) -> Vec<String> {
```
fitzthum: nit: maybe just call this `disable_lazy_itable_init`. I don't really love parsing the args like this but I guess it's ok.

fitzthum (approved):
LGTM. one nit

confidential-data-hub/hub/src/storage/drivers/filesystem.rs
manuelh-dev: thanks, modified and updated @fitzthum

confidential-data-hub/hub/src/storage/drivers/filesystem.rs
```diff
@@ -144,6 +148,33 @@ impl FsFormatter {
     }
 }
 
+// Keep explicit caller choices, but default ext4 to eager inode table initialization.
+fn disable_lazy_itable_init(args: &[String]) -> Vec<String> {
```
zvonkok, line 155: Is Clap here overkill? @manuelh-dev @Xynnn007 @fitzthum ?

zvonkok: Do we have other places for args parsing where it could help or are we implementing it ad-hoc for each of the moduels where needed?

fitzthum: It makes me wonder if we don't quite have the right api here. The caller provides the args but we might modify them in some cases. That's not the cleanest. I am fine with merging this tho and we can try to refine as we work on more storage stuff.

Xynnn007: On one hand we'd expect CDH to make everything inside as a black box. With this idea, I don't want CDH callers to need to worry about which parameters they haven't set, causing issues like luks+ext4 failing (e.g., lazy initial inode not being disabled). 
At the same time, users can freely set some parameters when formatting the ext4 file system.


My idea is to merge this PR first, at least this PR guarantees this point. I will make some optimizations later.

Xynnn007: But I must admit that there are two ways to achieve this
1. If user does not provide lazy initial related parameters and at the same time uses luks2, the call will fail before actual execute luks+ext4 mount. This requires the user to have some prior knowledge about secure mount RPC.
2. If user does not provide lazy initial related parameters and at the same time uses luks2, the logic will automatically add the parameter and raise a `warn` to CDH log (but cannot tell the RPC caller), then the mount succeeds. This could lead some users to want Luks2 + ext4 to execute intentially without a lazy initial node—even though they are unaware that the internal logic has already added one, and their original intention is impossible to achieve.

After some thoughts, let me raise this two options out and get some ideas on which choice. This PR is almost about way ~~1~~ 2.

manuelh-dev: @Xynnn007 - for clarification: Are you saying the PR is almost about way _2_ not about way _1_? Except for the `warn`, that's what's happening (if `lazy_itable_init` is not provided by the user, CDH will set it. If it is provided by the user, CDH will accept the user setting).

On your first point: If we follow this suggestion, this would keep things 'less complex' in CDH and the contract would be more explicit - but it's less of a black box and in turn the contract would be 'more complex': If the user does not set `lazy_itable_init=0` we fail early in CDH. We can probably add explicit documentation and notes in the relevant API code about this. Minor point: The caller will no longer be able to overwrite `lazy_itable_init=0`. If the caller is confident that they are *not making parallel writes* and thus does not want `lazy_itable_init=0`, the caller will not be able to influence this setting.
I have a prior draft PR in kata which would actually set the parameter explicitly (and start exercising paralle writes): https://github.com/kata-containers/kata-containers/pull/12953

By the way: whatever choice we make, we should probably only mandate this when integrity is requested. Without integrity, this is not a problem. This current PR _always_ exercises the `disable_lazy_itable_init` function. What do you think about this?

Xynnn007: > whatever choice we make, we should probably only mandate this when integrity is requested. Without integrity, this is not a problem. This current PR _always_ exercises the `disable_lazy_itable_init` function. What do you think about this?

We need some detect logic in the upper level `luks2` module to determine whether we need to do this step, instead of always doing that

Xynnn007: > @Xynnn007 - for clarification: Are you saying the PR is almost about way _2_ not about way _1_? Except for the `warn`, that's what's happening (if `lazy_itable_init` is not provided by the user, CDH will set it. If it is provided by the user, CDH will accept the user setting).

Yes thanks for pointing out, I fixed my comment

manuelh-dev: @Xynnn007 - thank you. I have revised the top commit - new commit: https://github.com/confidential-containers/guest-components/pull/1451/changes/8de14aadbbc5715c2a9ea85725e67669a4c10124. Changes compared to prior revision (I am still sticking with the initially proposed logic for handling the options parameter):
- added comments for new behavior
- exercise `format_integrity_compatible` when `data_integrity` is requested. Only adjust the `mkfsOpts` in `format_integrity_compatible`

This should not provide a decision for this discussion, but regardless of the decision, adding comments in these places and making the logic more selective should be an improvment.

confidential-data-hub/hub/src/storage/drivers/filesystem.rs
```diff
@@ -144,6 +151,37 @@ impl FsFormatter {
...
```
sprt, line 155: See https://github.com/kata-containers/kata-containers/pull/12953#discussion_r3175635016 - do we need that here too?

manuelh-dev: I don't know if that answers the question - but without `lazy_journal_init=0` the tests added in this PR work and also the Kata storage tests as modified in https://github.com/kata-containers/kata-containers/pull/12953 do work. With this in mind, the answer to your question is 'no'.

Are you looking at anything specific such as more specific output from the logs?

sprt: Don't you agree the principle of both options is the same (lazy -> eager), so if journal init remains lazy you risk running into the same issue?

sprt: If you're OK with that risk that's fine by me! Just wanted to point it out in case you missed this.

fitzthum:
Ready to merge?

confidential-data-hub/hub/src/storage/drivers/filesystem.rs
manuelh-dev: It's a good point - for now my stance would be: to keep an eye on this - but since there was no issue with regards to this setting, let's keep things simple for this PR. For `lazy_itable_init` we have to make this surgical operation (`disable_lazy_itable_init`), but since we're not aware of issues with regards to `lazy_journal_init` I would personally not extend this PR for now.

If we get more test coverage from consumers of this functionality or if we have more ideas for test coverage in the rust tests (and can reproduce issues), we may want to revise this in the future.

I'm very happy to adapt though if there's good arguments.

PR #1456, opened 2026-05-08
CDH/luks2: add `--integrity-no-journal` flag when integrity is enabled

Xynnn007:
Before #1380 we use CryptActivate::NO_JOURNAL to promote the performance when integrity protection is enabled. But #1380 ignored this flag. This patch brings back the flag to cryptsetup cmdline.

Let's get everything equal back.

cc @manuelh-dev


manuelh-dev (approved):

fitzthum (approved):
LGTM

PR #1457, opened 2026-05-08
image-rs: use pathrs when unpacking

fitzthum:
This one is a little fiddly. Please take a close look. I think there are a lot of things we can improve about image-rs, but I am taking it slow to avoid breaking stuff.

Let's use pathrs in more places.

In particular, let's update the convert_whiteout code to use pathrs.

We could also use pathrs at the end of the unpack code, but since the path has already been validated at that point, let's leave that for later.

This also adjusts the set_perms_ownerships. First, let's just call it set_permissions. Also, let's change it to take an fd. This way we don't have to worry about the difference between fchown and lchown. I'm not sure we really needed to distinguish between these anyway.

Also, I am continuing my quest to replace destination with layer_dir, which is clearer.

### DIFF
--- a/image-rs/src/stream/unpack.rs
+++ b/image-rs/src/stream/unpack.rs
@@ -5,8 +5,9 @@
 use anyhow::{anyhow, bail, Context, Result};
 use filetime::FileTime;
 use futures::StreamExt;
+use nix::libc;
 use nix::libc::timeval;
-use nix::sys::stat::{mknod, Mode, SFlag};
+use pathrs::flags::OpenFlags;
 use pathrs::InodeType;
 use thiserror::Error;
 use tracing::{debug, warn};
@@ -18,10 +19,10 @@ use std::{
     fs::Permissions,
     io,
     os::{
-        fd::{AsFd, AsRawFd},
+        fd::{AsFd, AsRawFd, BorrowedFd},
         unix::fs::PermissionsExt,
     },
-    path::{Path, PathBuf},
+    path::Path,
 };
 use tokio::{fs, io::AsyncRead};
 use tokio_tar::ArchiveBuilder;
@@ -140,29 +141,31 @@ fn is_whiteout(name: &str) -> bool {
 async fn convert_whiteout(
     name: &str,
     path: &Path,
-    uid: u32,
-    gid: u32,
-    mode: Option<u32>,
-    destination: &Path,
+    layer_dir: &Path,
     attr_available: bool,
 ) -> Result<()> {
     let parent = path
         .parent()
         .ok_or(anyhow!("Invalid whiteout parent for path: {:?}", path))?;
 
+    let layer_dir = pathrs::Root::open(layer_dir).context("Failed initialize layer dir")?;
+
     // Handle opaque directories
     if name == WHITEOUT_OPAQUE_DIR {
+        let opaque_dir = layer_dir.resolve(parent)?;
+
         // Opaque directory whiteout requires xattr support
         if !attr_available {
             debug!(
                 "Skipping opaque directory whiteout (xattr unavailable) for: {:?}",
-                destination.join(parent)
+                opaque_dir
             );
             return Ok(());
         }
 
-        let destination_parent = destination.join(parent);
-        xattr::set(destination_parent, "trusted.overlay.opaque", b"y")?;
+        let opaque_dir = opaque_dir.reopen(OpenFlags::O_RDONLY)?;
+        opaque_dir.set_xattr("trusted.overlay.opaque", b"y")?;
+
         return Ok(());
     }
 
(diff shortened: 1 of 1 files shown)


image-rs/src/stream/unpack.rs
```diff
@@ -171,39 +174,33 @@ async fn convert_whiteout(
...
+        layer_dir.mkdir_all(parent, &Permissions::from_mode(0o755))?;
     }
 
-    mknod(path.as_c_str(), SFlag::S_IFCHR, Mode::empty(), 0)?;
+    layer_dir.create(
+        &original_path,
+        &InodeType::CharacterDevice(Permissions::from_mode(0o000), 0),
+    )?;
 
-    set_perms_ownerships(&path, ChownType::LChown, uid, gid, mode).await
```
Xynnn007, line 188: Noticed that the `set_perms_ownerships` is removed here. Does this have any influence?

image-rs/src/stream/unpack.rs
```diff
@@ -567,6 +545,9 @@ mod tests {
             fs::remove_dir_all(destination).await.unwrap();
         }
 
+        if let Err(e) = unpack(data.as_slice(), destination).await {
+            println!("error {}", e);
+        }
```
Xynnn007: Is this something left when debugging?

image-rs/src/stream/unpack.rs
```diff
@@ -307,8 +298,13 @@ pub async fn unpack<R: AsyncRead + Unpin>(input: R, destination: &Path) -> Unpac
                     .map_err(|source| UnpackError::ReadTarEntriesFailed { source })?
                     .ends_with(b"/")
         {
-            set_perms_ownerships(&path_cstring, ChownType::LChown, uid, gid, mode)
-                .await
+            let f =
+                std::fs::File::open(file_path).map_err(|e| UnpackError::SetOwnershipsFailed {
```
Xynnn007, line 302: Note that the new `set_permissions` function uses `libc::AT_SYMLINK_NOFOLLOW` flag that means if the fd is a link file, change the link file itself than the linked target. However we need to ensure that the fd is the link file itself, but the default `std::fs::File::open` here does not include `O_NOFOLLOW` flag by default thus always the file that the link file points to.

image-rs/src/stream/unpack.rs
fitzthum: I don't think there is any need for this. Notice when we create a whiteout file just above, we set the permissions directly.

For opaque dirs, I don't think the permissions matter, since the xattr is set.

There is an interesting corner case when we call `mkdir_all` on the parents of the opaque dir. We hardcode the permissions to 755 in this case. It's worth noting that we did this before this PR as well, it's just a bit hidden by the crate we call. Either way, I think this is ok. If these intermediate dirs are specified in another layer, their permissions should be updated. If not, I guess the container image itself must not container the perms info for those dirs.

image-rs/src/stream/unpack.rs
fitzthum: Yeah let's think through this. It's a bit confusing.

We have three cases. 1) is_dir 2) is regular file 3) is symlink or hardlink

in case 1 and 2 we follow symlinks when we open the file. this is safe because we've already checked that the symlinks don't leave the layer dir. the fd that we get here will not point to a symlink. in both cases we call `set_permissions`. This sets `SYMLINK_NOFOLLOW`, but that's sort of irrelevant because we have an fd pointing to a real file/dir at this point.

Case 3 isn't explicitly enumerated, but it's the fall through from the other two. In this case we do have a symlink or a hardlink but we don't call `set_permissions`. 

So I think that the current implementation is fine, but maybe it would be clearer if we remove `AT_SYMLINK_NOFOLLOW` from `set_permissions`. On the other hand maybe we want that to be set if people use the method for other stuff in the future. wdyt?

fitzthum: btw originally case 1 and 2 used lchown vs fchown, but I'm not sure if that was necessary.

Xynnn007: The original logic refers to `tar` logic https://github.com/alexcrichton/tar-rs/blob/2349b494827807791473f3c6ca9b2fc5ef274cdb/src/entry.rs#L499-L509 and https://github.com/alexcrichton/tar-rs/blob/2349b494827807791473f3c6ca9b2fc5ef274cdb/src/entry.rs#L622-L633. Note that when the 2nd parameter is None, the underlying logic will use lchown instead of fchown. probably it's caused by programming style that the fd of the dir is not given, thus using path. 

Yes probably we do not need the flag. Due to https://man.freebsd.org/cgi/man.cgi?query=fchownat&sektion=2&n=1 it says
```
AT_EMPTY_PATH
	       If the path argument is an empty	string,	operate	on the file or
	       directory referenced by the descriptor fd.  If fd is  equal  to
	       AT_FDCWD, operate on the	current	working	directory.
```

and we are always giving fd.


Also, you are right that we do not care about case 3 up to now. Maybe we could do it in this PR also? Specifically, if it is a symlink, do lchown to the link file itself

fitzthum: Ok, I got rid of the extra arg and added some code to set symlink ownership.

I'd like to get the pathrs stuff in for the upcoming release, but I do see a few more things to fix. For instance, I think one of these cases might actually be redundant. The comment says that `this crate also does not cover symlink files/dir mtime`.

Yet we are setting the permissions and ownerships of dirs and the permissions, ownerships, and mtime of files. Seems like that might be redundant, but I haven't had time to check the crate.

Before I dive any deeper into the image unpacking, I want to create some more sophisticated image tests where we can directly compare our pulled image to one pulled with a more mainstream tool.

Xynnn007 (approved):
LGTM

PR #1468, opened 2026-05-12
docs: update resource URI and add plugin support

fitzthum:
Update the resource URI doc and add support for plugins.

Specifically, let's allow the scheme to be something like `kbs+plugin://`. This seems pretty reasonable and is backwards compatible. 

If this looks good, I will add some commits to actually implement the change.


attestation-agent/docs/RESOURCE_URI.md
```diff
@@ -0,0 +1,76 @@
...
+
+Since these fields are ignored, most resource URIs leave them out.
+This results in three slashes in a row.
+For example, `kbs:///repository/type/tag`.
+ 
+### Repository, Type, and Tag
+
+Currently resource URIs have three levels of identifiers/scope.
+The terms `repository`, `type`, and `tag` are somewhat arbitrary. 
+These identifiers can be used in any way.
```
Xynnn007, line 58: It's `resource` plugin-specific and no requirements on other plugins

attestation-agent/docs/RESOURCE_URI.md
```diff
@@ -0,0 +1,76 @@
...
+relates to Trustee.
+Some legacy code, such as the SEV KBC or the FS KBC may fulfill resource URIs
+in different ways.
+
+## Specification
+
+A Resource URI must comply with the following format:
+
+```plaintext
+kbs+<plugin>://<kbs_host>:<kbs_port>/<repository>/<type>/<tag>
```
Xynnn007, line 27: I like this design

mythi: The asymmetry here is that with `kbs+<plugin>://` callers need to know the resource is provided by `<plugin>` but with the `resource` plugin, they have no way to say the resources comes from vault.

How the plugin URIs can be supported with api-server-rest queries?

Xynnn007: vault is not a plugin level conception, but a resource plugin level. Thus it's expected that the caller does not know that.

for ASR, currently is `127.0.0.1:8006/cdh/resource/...`. we could generize it as `127.0.0.1:8006/cdh/<plugin-in-name>/...` for `kbs+<plugin-in-name>:///...`

mkulke: I would also that the caller doesn't get to request a concrete backing store for their resource is (say, vault or filesystem), this is up to the KBS, no?

I also like this URI scheme 👍

attestation-agent/docs/RESOURCE_URI.md
mkulke: yeah, I was also thinking whether it is a good idea to require the same path segment for every plugin, for some this hierarchy might just not make any sense. but it keeps implementation complexity at bay for now and that requirement can be relaxed later once it becomes a nuisance

mkulke (approved):

attestation-agent/docs/RESOURCE_URI.md
mythi: > vault is not a plugin level conception, but a resource plugin level.

yup, my comment was just that the implementation detail is visible to the user and they don't necessarily understand: why `kbs+pcks11://` is needed but `kbs+vault://` is not.

attestation-agent/docs/RESOURCE_URI.md
Xynnn007: Yes - and there is already some implementations that do not follow this, e.g. [pkcs11](https://github.com/confidential-containers/trustee/blob/main/kbs/src/plugins/implementations/pkcs11.rs#L135)

attestation-agent/docs/RESOURCE_URI.md
Xynnn007: the pkcs11 apis are different from the vaults. see https://github.com/confidential-containers/trustee/blob/main/kbs/src/plugins/implementations/pkcs11.rs#L135 for example `wrap-key`

attestation-agent/docs/RESOURCE_URI.md
fitzthum: Yes, we need to clarify this. Currently you cannot access the pkcs11 plugin via the resource uri, which is essentially a bug imo. I am open to allowing arbitrary length requests, like we do with rv uris, but that is not how the resource uri works today.

attestation-agent/docs/RESOURCE_URI.md
fitzthum: yeah, also worth noting that the resource plugin only supports one storage backend anyway.

attestation-agent/docs/RESOURCE_URI.md
```diff
@@ -0,0 +1,76 @@
...
+
+### Query Strings
+
+The resource URI also supports query strings.
+
+## Examples
+
+* `kbs://example.cckbs.org:8081/alice/decryption-key/1`
+* `kbs:///a/b/c`
+* `kbs+pkcs11:///a/b/c`
```
Xynnn007, line 75: btw the original resource URI could also support query string, like `kbs+plugin1://a/b/c?d=e`

Xynnn007:
btw, https://github.com/confidential-containers/trustee/pull/1305 needs `POST` and `DELETE` method. So the concrete implementation should care about this.

fitzthum:
Added implementation. PTAL

note that I am not touching the ASR in this PR. We will still have a gap there. Currently you can make a request via the ASR like this

```
curl http://127.0.0.1:8006/cdh/resource/default/key/1
```

I don't think we can make that work with plugins so probably we should add a new endpoint like 
```
curl http://127.0.0.1:8006/cdh/plugin/<plugin-name>/a/b/c
```

But that is still a bit limited. Open to suggestions here. Probably will address in a follow-up rather than this one.

Xynnn007:
> curl http://127.0.0.1:8006/cdh/plugin/<plugin-name>/a/b/c

A naive question without a deep dive and implementation experience: what's the blocker to use `curl http://127.0.0.1:8006/cdh/<plugin-name>/a/b/c` ?

attestation-agent/deps/resource_uri/src/lib.rs
```diff
@@ -47,9 +49,17 @@ impl TryFrom<url::Url> for ResourceUri {
...
+        let plugin = if scheme == SCHEME {
+            None
+        } else if let Some(plugin_name) = scheme.strip_prefix("kbs+") {
+            if plugin_name.is_empty() {
+                return Err("scheme kbs+ requires a plugin name, e.g. kbs+pkcs11");
+            }
+            Some(plugin_name.to_string())
+        } else {
+            return Err("scheme must be kbs or kbs+<plugin>");
+        };
```
Xynnn007, line 64: Just a nit
```rust
let plugin = match scheme.as_str() {
    SCHEME => None,
    s if s.starts_with("kbs+") => {
        let plugin_name = s.trim_start_matches("kbs+");
        if plugin_name.is_empty() {
            return Err("scheme kbs+ requires a plugin name, e.g. kbs+pkcs11");
        }
        Some(plugin_name.to_string())
    }
    _ => return Err("scheme must be kbs or kbs+<plugin>"),
};
```

attestation-agent/deps/resource_uri/src/lib.rs
```diff
@@ -14,15 +14,17 @@ const RESOURCE_ID_ERROR_INFO: &str =
...
-/// Resource Id document <https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/docs/KBS_URI.md>
+/// Resource Id document <https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/docs/RESOURCE_URI.md>
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct ResourceUri {
     pub kbs_addr: String,
     pub repository: String,
     pub r#type: String,
     pub tag: String,
     pub query: Option<String>,
+    pub plugin: Option<String>,
```
Xynnn007, line 27: Maybe we can directly set this as `String`, and give a default `resource` if none is given

attestation-agent/deps/resource_uri/src/lib.rs
```diff
@@ -201,4 +254,32 @@ mod tests {
             ResourceUri::try_from(url_from_string).expect("failed to try from url");
         assert_eq!(resource_from_url, resource);
     }
+
+    #[test]
```
Xynnn007: Can we combine them into a single test with `rstest`? which will look dry

confidential-data-hub/hub/src/secret/mod.rs
```diff
@@ -142,7 +142,7 @@ impl Secret {
     /// Otherwise, look for a local credential (provisioned to the fs
     /// at CDH startup).
     async fn get_kid(kid: &String) -> Result<Vec<u8>> {
-        if kid.starts_with("kbs://") {
+        if kid.starts_with("kbs://") || kid.starts_with("kbs+") {
```
Xynnn007: I once gave the suggestion to make the `plugin` a `String` in `ResourceUri` struct in previous commit. If it's adopted, here we can simply use `kid.starts_with("kbs+")`. 

btw, `kid.starts_with("kbs+")` might not be enough, we might still need to judge `://`. so probably a simple `.contains()` or a regex can be used. wdyt?

confidential-data-hub/hub/src/storage/volume_type/blockdevice/mod.rs
```diff
@@ -48,7 +48,7 @@ async fn get_plaintext_key(key_uri: &str) -> Result<Zeroizing<Vec<u8>>> {
             .map_err(|source| BlockDeviceError::GetKeyFailed {
                 source: source.into(),
             })?
-    } else if key_uri.starts_with("kbs://") {
+    } else if key_uri.starts_with("kbs://") || key_uri.starts_with("kbs+") {
```
Xynnn007: The same question

attestation-agent/coco_keyprovider/src/enc_mods/mod.rs
```diff
@@ -150,9 +148,23 @@ async fn generate_key_parameters(input_params: &InputParams) -> Result<(Vec<u8>,
...
+    } else if let Some(pos) = keyid.find("://") {
+        let scheme = &keyid[..pos];
+        if scheme.starts_with("kbs+") {
+            &keyid[pos + 3..]
+        } else {
+            keyid
+        }
+    } else {
+        keyid
+    };
```
Xynnn007: Maybe we can import `ResourceUri` and parse, then extract `repository`, .. etc

attestation-agent/deps/resource_uri/src/lib.rs
```diff
@@ -14,15 +14,17 @@ const RESOURCE_ID_ERROR_INFO: &str =
...
+const DEFAULT_PLUGIN: &str = "resource";
 
-/// Resource Id document <https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/docs/KBS_URI.md>
+/// Resource Id document <https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/docs/RESOURCE_URI.md>
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct ResourceUri {
     pub kbs_addr: String,
     pub repository: String,
     pub r#type: String,
     pub tag: String,
```
Xynnn007, line 25: Note that due to https://github.com/confidential-containers/guest-components/pull/1468#discussion_r3232038785, we could combine these fields into a `Vec<String>` for any length paths? wdyt?

Xynnn007 (commented):
great work for extension! First round review

fitzthum:
> A naive question without a deep dive and implementation experience: what's the blocker to use curl http://127.0.0.1:8006/cdh/<plugin-name>/a/b/c ?

Risk of collision if we introduce any other CDH endpoints in the future, which we might.

attestation-agent/deps/resource_uri/src/lib.rs
fitzthum: That makes it a little tricky to reconstruct the URI in the `whole_uri` method. wdyt?

Xynnn007:
sounds reasonable. what about to multiplex current resource field to work as the plugin name, s.t. `127.0.0.1:8006/cdh/<plugin name>/...`?

Xynnn007:
well. it is also the same issue

Xynnn007:
so probably cdh/kbs-plugin/xxx or so is the only choice. for legacy, keep legacy `resource` without `kbs-plugin` part. note that imo the prefix `kbs-` is needed because plugin in CDH can be more than kbs.

I have an impression that we once talked about this, but cannot find the thread.

attestation-agent/coco_keyprovider/src/enc_mods/mod.rs
fitzthum: This function has slightly different behavior than the resource Uri parsing. It allows other schemes and even kids without any scheme. So I don't think it gives us much to use the resource uri crate here.

fitzthum:
Yeah, that seems fine, although it doesn't help use with plugins that have query strings or more arguments. I wonder if there is a way to pass the full resource uri.

attestation-agent/deps/resource_uri/src/lib.rs
fitzthum: I am open to this but I think we should do another PR for handling the variable length stuff.

confidential-data-hub/hub/src/secret/mod.rs
fitzthum: Ok, instead let me import `ResourceUri` here and try parsing that way.

MatiasVara:
I am not sure if this PR has to address this but, after compiling trustee-attester, I am getting:
```bash
trustee-attester failed for http://10.44.34.142:8080 resource kbs+provisioner:///default/ad67b12f-86c4-4653-bb85-14ea3447624e/root: Error: Resource path /kbs+provisioner:///default/ad67b12f-86c4-4653-bb85-14ea3447624e/root must follow the format '/<repository>/<type>/<tag>'
```

fitzthum:
> I am not sure if this PR has to address this but, after compiling trustee-attester, I am getting:

Yeah, looks like the Trustee attester expects a "resource path" as input rather than a resource URI. May want to change that or to add another command that takes the URI. I'll probably leave that to another PR.

btw @Xynnn007 I think I addressed most of your comments. I sort of forgot about this PR, but let's get it rolling again. I am also thinking about ASR changes in another PR.

attestation-agent/deps/resource_uri/src/lib.rs
Xynnn007: do you mean when user uses `kbs:///default/1/1` but `whole_uri` gets `kbs+resource:///default/1/1`?

This can be acceptable imo

attestation-agent/docs/RESOURCE_URI.md
Xynnn007: We can access the pkcs11 plugin.

My suggestion is to remove the three-section restriction here. This is because it's specific to the `resource` plugin; for other plugins, there's no need to limit the number of sections.

attestation-agent/deps/resource_uri/src/lib.rs
Xynnn007: It's ok to include into another PR - it should include at least two parts
1. change document to say that the 3-section is for `resource` only, and no limit for others.
2. here we use Vec to allow any length.

attestation-agent/coco_keyprovider/src/enc_mods/mod.rs
Xynnn007: yes I only see that line 158-163 shares some common codes, not replace whole 155-166 with that crate

Xynnn007 (commented):
to make it clear to push work, what blocks from my side
1. consider reuse ResourceUri parsing logic in coco_keyprovider change if possible
2. The three section is for resource only, and open to all other plugins. This could be done in a following PR, up to you

attestation-agent/docs/RESOURCE_URI.md
mythi: Based on https://github.com/confidential-containers/trustee/pull/735 it seems pkcs11 became a plugin of its own only because the API got extended and it did not meet the three-section resource URI rule anymore. Is that the case?

MatiasVara:
> > I am not sure if this PR has to address this but, after compiling trustee-attester, I am getting:
> 
> Yeah, looks like the Trustee attester expects a "resource path" as input rather than a resource URI. May want to change that or to add another command that takes the URI. I'll probably leave that to another PR.
> 

OK. I can submit a PR for that.

attestation-agent/docs/RESOURCE_URI.md
fitzthum: I think 735 could have been a little more clear.

I updated the spec a bit. I think we should support arbitrary-length URIs, but that's logically separate from this change. Let's make another PR. 

> We can access the pkcs11 plugin.

You mean we can access it currently or that we could access it if we support longer URIs?

attestation-agent/docs/RESOURCE_URI.md
fitzthum: Yeah, I remember when we added support for that for the nebula plugin.

attestation-agent/deps/resource_uri/src/lib.rs
fitzthum: Yes. I think that is potentially confusing, but not too different either way.

attestation-agent/coco_keyprovider/src/enc_mods/mod.rs
fitzthum: Ok, changed this. Cleaned up a few things. The keyprovider code is a little messy overall. We might want to do a follow-up with some more fixes although maybe we should transition to another tool entirely.

Xynnn007 (approved):

PR #1469, opened 2026-05-13
se-attester: use SHA-512 for report data in IBM SEL

BbolroC:
This PR uses more secure SHA-512 algorithm for report data in IBM SEL instead of the fallback algorithm SHA-384.
As small refactor, it also extracts the process_runtime_digest() helper function from the SE attester into a general-purpose utility function.

Depends-on: https://github.com/confidential-containers/trustee/pull/1338

### DIFF
--- a/attestation-agent/attester/src/se/mod.rs
+++ b/attestation-agent/attester/src/se/mod.rs
@@ -14,7 +14,10 @@ use serde::{Deserialize, Serialize};
 use serde_with::{base64::Base64, serde_as};
 use tracing::debug;
 
-const RUNTIME_DIGEST_SIZE: usize = 48; // SHA-384 digest size in bytes
+/// The size for SE attestation report data (SHA-512 digest size in bytes).
+/// Only the first `SE_REPORT_DATA_SIZE` bytes of the `runtime_data_digest` field
+/// from `SeAttestationRequest` are used in the attestation command.
+const SE_REPORT_DATA_SIZE: usize = 64;
 
 /// Structured user data for IBM SEL attestation
 /// Currently, only contains runtime data digest bound to the attestation measurement
@@ -105,7 +108,7 @@ impl Attester for SeAttester {
             runtime_digest
         );
 
-        let runtime_digest = validate_and_pad_data(runtime_digest, RUNTIME_DIGEST_SIZE)?;
+        let runtime_digest = validate_and_pad_data(runtime_digest, SE_REPORT_DATA_SIZE)?;
 
         let user_data = UserData {
             runtime_data_digest: runtime_digest,
@@ -147,9 +150,9 @@ impl Attester for SeAttester {
 mod tests {
     use super::*;
 
-    // Mock runtime digest (SHA-384 size: 48 bytes) for tests
+    // Mock runtime digest (SHA-512 size: 64 bytes) for tests
     // This simulates the digest computed from runtime data in production
-    const MOCK_RUNTIME_DIGEST: [u8; 48] = [0xBB; 48];
+    const MOCK_RUNTIME_DIGEST: [u8; 64] = [0xBB; 64];
 
     // Helper to create a dummy BootHdrTags for testing
     fn create_dummy_boot_hdr_tags() -> BootHdrTags {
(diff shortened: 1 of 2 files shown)


mythi:
> Depends-on: [confidential-containers/trustee#1338](https://github.com/confidential-containers/trustee/pull/1338)

The Trustee PR is also set to depend on this one. To avoid the circular dependency, can this be mode so that se-attester uses sha384 hash algo with old Trustee that does not give you `selected-hash-algorithm`. Or is it like this already? 

could `process_runtime_digest()` be made such that it can also be used in `get_evidence()` to deduplicate the logic.

Xynnn007 (approved):

BbolroC:
> > Depends-on: [confidential-containers/trustee#1338](https://github.com/confidential-containers/trustee/pull/1338)
> 
> The Trustee PR is also set to depend on this one. To avoid the circular dependency, can this be mode so that se-attester uses sha384 hash algo with old Trustee that does not give you `selected-hash-algorithm`. Or is it like this already?

Yeah, I am aware of it, unfortunately. For this, the attestation service should keep memorizing which hash algorithm is selected for the extra_params for each attestation challenge and pass the info to a verifier for validation of the report data, which leads to non-marginal changes. I think we can deal with it separately. Or does it make sense to make a fallback for the validation in the verifier side? What do you think @Xynnn007 about this?

> could `process_runtime_digest()` be made such that it can also be used in `get_evidence()` to deduplicate the logic.

Good idea! This also makes the unit tests look nice and clear. I will update the PR. Thanks!

Xynnn007:
> Yeah, I am aware of it, unfortunately. For this, the attestation service should keep memorizing which hash algorithm is selected for the extra_params for each attestation challenge and pass the info to a verifier for validation of the report data, which leads to non-marginal changes. I think we can deal with it separately. Or does it make sense to make a fallback for the validation in the verifier side? What do you think @Xynnn007 about this?

Let's first distinguish a few concepts before looking at what the current PR is doing.

First, the current Attester abstracts the TEE's report retrieval capability into a black box. For architectures like TDX/SNP, there's a challenge parameter called `report_data`, which is a short data value, typically a few tens of bytes long. IBM SE's black box uses the `UserData` field for this.

As for `runtime_data`, it's a large JSON structure, a concept at the KBS protocol level. We bind the entire `runtime_data` to specific evidence when retrieving evidence. However, since `report_data` is short, we hash the entire `runtime_data` and use the hash value as `report_data` to call `get_report`. The hash algorithm can be chosen during kbs protocol handshake.

So, in this PR, the change is for low-level black box to implement IBM SE's `report_data` functionality - we can treat it as what the Intel CPU circuits inside is doing - changing report field from 48 bytes to 64 bytes - thus a breaking change but I think it's ok. It's not a good idea to support both modes in `report_data` ability level as it's low level black box, but we need to ensure that the compatibility for attester/verifier in a same version.

attestation-agent/attester/src/se/mod.rs
```diff
@@ -14,7 +14,7 @@ use serde::{Deserialize, Serialize};
 use serde_with::{base64::Base64, serde_as};
 use tracing::debug;
 
-const RUNTIME_DIGEST_SIZE: usize = 48; // SHA-384 digest size in bytes
+const SE_REPORT_DATA_SIZE: usize = 64; // SHA-512 digest size in bytes
```
Xynnn007: Do you think it would be valuable to add a comment here? like
```
/// we use the first bytes of `runtime_data_digest` field in `SeAttestationRequest`.
/// This size is the part length.
```

Xynnn007 (approved):
LGTM! Thanks @BbolroC 

Note that this is a breaking change for IBM SE architecture.
Also, with a deep diving to code, I realize now SE could now support composite attestation. The `SeAttestationRequest.runtime_data_digest` can be set in the `kbs_protocol` for IBM SE. This work can be done a separate PR.

attestation-agent/attester/src/se/mod.rs
BbolroC: Yeah, reasonable. Thanks for the suggestion! I will update the PR.

Xynnn007:
btw let's merge this when https://github.com/confidential-containers/trustee/pull/1338 is ready also, for there are breaking changes and we need the compability for both side.

mythi:
@BbolroC I think b9f82d4890aa2a0054abbd65bd93de01a27aaadc needs to be the first commit in this PR.

Xynnn007:
Note that https://github.com/confidential-containers/trustee/pull/1338 gets merged

attestation-agent/attester/src/utils.rs
```diff
@@ -20,6 +21,48 @@ pub fn pad<const T: usize>(input: &[u8]) -> [u8; T] {
...
+/// Validates and pads data to a specified size.
+///
+/// This function ensures that the input data does not exceed the expected size,
+/// and pads it with zeros if it's smaller than the expected size.
+///
+/// # Arguments
+///
+/// * `data` - The input data to validate and pad
+/// * `expected_size` - The target size for the data
+/// * `data_name` - A descriptive name for the data (used in error messages)
```
mythi: Is this useful/needed?

mythi (commented):
Looks OK. Have you tested this with existing Trustee releases?

attestation-agent/attester/src/utils.rs
BbolroC: It is useful for me, but I'm okay to remove it if you don't like.

BbolroC:
> Looks OK. Have you tested this with existing Trustee releases?

Yes, I tested it with the latest push for both. It works. But not working with the current Trustee (the version before https://github.com/confidential-containers/trustee/pull/1338 is merged) as @Xynnn007  explained above.

attestation-agent/attester/src/utils.rs
BbolroC: The parameter is removed from the function prototype. Thanks!

mythi (approved):

PR #1470, opened 2026-05-13
Attester: remove libtdx feature for TDX attester

Xynnn007:


mythi:
Thanks! Just one quick comment: I know at least Openshift does not have RTMRs available so `TDX_CMD_GET_REPORT0` ioctl is still needed to do MRCONFIGID parsing. 

Could this PR be about dropping `tdx-attest-dcap-ioctls` featured code only?

Xynnn007:
@mythi Thanks for the information. Do you mean that 
1. read/write rtmrs via ~~configfs~~ sysfs
2. read `mrconfigid` via configfs
are not supported by Openshift?

mythi:
Yes. mrconfigid for initdata is read from report0 ioctl which is available without the DCAP wrapper in the default configuration.

Another clarification: the rtmr abi is not related to configfs reports abi

Xynnn007:
Intel's DCAP internally has a compiler switch called V3_DRIVER that controls the type of ioctl writes for RTMR. I've implemented support for this in my PR. For IOR scenarios, the iocuddle crate can only make read requests with the req parameter using `unsafe`. Maybe there is better ways I did not know to avoid `unsafe`.

Xynnn007:
Test with all configfs/sysfs/ioctl scenarios.

mythi:
There are no kernels with ioctl RTMR extend so I don't think we need to add that here. RTMR extend is only available via sysfs. Upstream kernels have only report0 ioctl (no get quote nor rtmr extend).

Xynnn007:
> There are no kernels with ioctl RTMR extend so I don't think we need to add that here. RTMR extend is only available via sysfs. Upstream kernels have only report0 ioctl (no get quote nor rtmr extend).

We alibaba cloud kernel is using a version with ioctl RTMR, which might be an out-of-tree code. This is what the [DCAP](https://github.com/intel/confidential-computing.tee.dcap/blob/3aa24f6df004f092556aef10e0ab422cb0395e38/QuoteGeneration/quote_wrapper/tdx_attest/tdx_attest.c#L46-L55) lib is doing such and with this PR we can get rid of DCAP but still, keep the functionality work.

I am not sure how other CSPs support the RTMR extending if they does not have 6.16 sysfs patch. Let me make a comment upon the codes to show that this is for out-of-tree code and TODO is to use sysfs when all projects are ready.

mythi:
> > There are no kernels with ioctl RTMR extend so I don't think we need to add that here. RTMR extend is only available via sysfs. Upstream kernels have only report0 ioctl (no get quote nor rtmr extend).
> 
> We alibaba cloud kernel is using a version with ioctl RTMR, which might be an out-of-tree code. This is what the [DCAP](https://github.com/intel/confidential-computing.tee.dcap/blob/3aa24f6df004f092556aef10e0ab422cb0395e38/QuoteGeneration/quote_wrapper/tdx_attest/tdx_attest.c#L46-L55) lib is doing such and with this PR we can get rid of DCAP but still, keep the functionality work.

are you saying you moved to configfs-tsm reports for Qutoes but kept ioctl RTMR?

Xynnn007:
> > > There are no kernels with ioctl RTMR extend so I don't think we need to add that here. RTMR extend is only available via sysfs. Upstream kernels have only report0 ioctl (no get quote nor rtmr extend).
> > 
> > 
> > We alibaba cloud kernel is using a version with ioctl RTMR, which might be an out-of-tree code. This is what the [DCAP](https://github.com/intel/confidential-computing.tee.dcap/blob/3aa24f6df004f092556aef10e0ab422cb0395e38/QuoteGeneration/quote_wrapper/tdx_attest/tdx_attest.c#L46-L55) lib is doing such and with this PR we can get rid of DCAP but still, keep the functionality work.
> 
> are you saying you moved to configfs-tsm reports for Qutoes but kept ioctl RTMR?

Yes. configfs for quotes; ioctl/sysfs for RTMR

mythi:
>; ioctl/sysfs for RTMR

If both are available, we can drop ioctl RTMR extend. I'm not OK to support out-of-tree functionality unconditionally and/or enabled by default.

Xynnn007:
@mythi Ok. Another choice is to have a feature flag to cover the ioctl code and have an explicit roadmap that delete them once main CSPs/distros to support sysfs. But this seems the same as we keep DCAP here also.

So, this work can be pending as now the upstream kernel 6.16 is not widely used and we can wait for main CSPs/distros to support sysfs. At the same time, the DCAP dep can be fully removed. Before that, as the runtime measurement is still under use, DCAP is still required (if we do not allow ioctl MVP codes to get into this repo).

mythi:
One optimization we could do already today is to try to get `mrconfigid` for initdata via sysfs if it's avaiable and only to report0 ioctl as the fallback (similar to how get quote works but so that the fallback is unconditionally enabled).

Xynnn007:
> One optimization we could do already today is to try to get `mrconfigid` for initdata via sysfs if it's avaiable and only to report0 ioctl as the fallback (similar to how get quote works but so that the fallback is unconditionally enabled).

Let me implement it in another PR.

mythi:
> > One optimization we could do already today is to try to get `mrconfigid` for initdata via sysfs if it's avaiable and only to report0 ioctl as the fallback (similar to how get quote works but so that the fallback is unconditionally enabled).
> 
> Let me implement it in another PR.

OK. This PR can be closed?

Xynnn007:
> > > One optimization we could do already today is to try to get `mrconfigid` for initdata via sysfs if it's avaiable and only to report0 ioctl as the fallback (similar to how get quote works but so that the fallback is unconditionally enabled).
> > 
> > 
> > Let me implement it in another PR.
> 
> OK. This PR can be closed?

Let's keep it a draft and then we can track the thread.

PR #1474, opened 2026-05-17
image-rs: refactor the unpack logic and do chown/chmod after unpacking

Xynnn007:
The upstream astral-tokio-tar does not handle correctly about uid/gid,
mode for different kind of files. Thus we are always doing some patching
work in image-rs. Because of the changes will also modify mtime, so we
should also care about the mtime of the files.
    
This patch does a refactoring about the logic, and make it more clear
about what we are doing in image-rs.
    
Also, the unpack logic should also keep the root dir "." with the mtime
it declares in the tar header. This is also fixed by the patch. But note
that the image-rs pulled rootfs is mounted by overlayfs. Overlayfs will
use multiple different extracted directories, thus the target mount
point's "." mtime cannot be determined by any of the tar files, so it's
not the scope of this commit.

### DIFF
--- a/image-rs/src/stream/unpack.rs
+++ b/image-rs/src/stream/unpack.rs
@@ -7,8 +7,7 @@ use filetime::FileTime;
 use futures::StreamExt;
 use nix::libc;
 use nix::libc::timeval;
-use pathrs::flags::OpenFlags;
-use pathrs::InodeType;
+use pathrs::{flags::OpenFlags, InodeType, Root};
 use thiserror::Error;
 use tracing::{debug, warn};
 
@@ -20,12 +19,12 @@ use std::{
     io,
     os::{
         fd::{AsFd, AsRawFd, BorrowedFd},
-        unix::fs::PermissionsExt,
+        unix::{ffi::OsStrExt, fs::PermissionsExt},
     },
     path::Path,
 };
 use tokio::{fs, io::AsyncRead};
-use tokio_tar::ArchiveBuilder;
+use tokio_tar::{Archive, ArchiveBuilder, Entry};
 use xattr::FileExt;
 
 pub type UnpackResult<T> = std::result::Result<T, UnpackError>;
@@ -90,6 +89,13 @@ pub enum UnpackError {
         path: String,
     },
 
+    #[error("Failed to set permissions for path: {path}")]
+    SetPermissionsFailed {
+        #[source]
+        source: anyhow::Error,
+        path: String,
+    },
+
     #[error("Failed to set file mtime: {path}")]
     SetMTimeFailed {
         #[source]
@@ -128,6 +134,40 @@ const WHITEOUT_OPAQUE_DIR: &str = ".wh..wh..opq";
 /// Returns whether the file name is a whiteout file
 fn is_whiteout(name: &str) -> bool {
     name.starts_with(WHITEOUT_PREFIX)
+}
+
+/// Returns whether the tar entry names the layer root directory itself.
+///
+/// OCI image layers may include a `.` entry to carry the root directory metadata.
+/// The layer extraction directory already exists, so this entry must not be unpacked
+/// again; only its ownership and timestamps are applied at the end.
+fn is_layer_root_entry(path: &Path) -> bool {
+    path.as_os_str().is_empty() || path == Path::new(".") || path == Path::new("./")
+}
+
+/// Returns whether the tar entry should be treated as a directory when applying
+/// ownership, mode, and deferred mtime after unpack.
+///
+/// OCI layer tars are usually ustar or PAX. In those formats, directories are
+/// identified by typeflag `5` ([`EntryType::Directory`]), which
+/// [`EntryType::is_dir`] checks.
+///
+/// Pre-ustar (v7 / old BSD) headers do not carry a reliable directory typeflag;
+/// there, a trailing `/` on the path name historically meant "directory". That
+/// heuristic is applied only when [`Header::as_ustar`] is `None`, so ustar and
+/// later headers rely on the typeflag alone. This matches
+/// `tokio_tar::Entry::unpack` ("Old BSD-tar compatibility").
+fn is_dir<R: AsyncRead + Unpin>(
+    kind: &tokio_tar::EntryType,
+    file: &Entry<Archive<R>>,
+) -> UnpackResult<bool> {
+    let is_dir = kind.is_dir()
+        || file.header().as_ustar().is_none()
+            && file
+                .path_bytes()
+                .map_err(|source| UnpackError::ReadTarEntriesFailed { source })?
+                .ends_with(b"/");
+    Ok(is_dir)
 }
 
 /// Converts a whiteout file or opaque directory.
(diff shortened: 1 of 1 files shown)


image-rs/src/stream/unpack.rs
```diff
@@ -243,14 +280,48 @@ pub async fn unpack<R: AsyncRead + Unpin>(input: R, layer_dir: &Path) -> UnpackR
...
+                    .map_err(|source| UnpackError::SetPermissionsFailed { source, path })?;
+                root_times = Some(dir_times(mtime));
+            }
+            continue;
+        }
+
+        // 4. unpack the entry to the layer directory
         // Both of these paths ensure that the entry is actually inside the layer dir
         if let Err(e) = file.unpack_in(layer_dir).await {
             match try_hardlink_fallback(&kind, &mut file, layer_dir).await {
```
fitzthum, line 357: Unrelated, but I don't totally love the fallback approach here. It seems like it might be better to directly detect the case where `unpack_in` is not supported. Not a big deal tho

image-rs/src/stream/unpack.rs
```diff
@@ -243,14 +280,48 @@ pub async fn unpack<R: AsyncRead + Unpin>(input: R, layer_dir: &Path) -> UnpackR
...
                 .map_err(|source| UnpackError::ConvertWhiteoutFailed { source })?;
             continue;
         }
 
+        let is_dir = kind.is_dir()
+            || file.header().as_ustar().is_none()
+                && file
+                    .path_bytes()
+                    .map_err(|source| UnpackError::ReadTarEntriesFailed { source })?
+                    .ends_with(b"/");
```
fitzthum: Maybe could be its own function also the second case seems a little hacky. Maybe add a comment about why it is needed.

image-rs/src/stream/unpack.rs
```diff
@@ -243,14 +280,48 @@ pub async fn unpack<R: AsyncRead + Unpin>(input: R, layer_dir: &Path) -> UnpackR
...
+
+        // 3. layer root ".": do not unpack_in; apply ownership and mode, defer mtime.
+        if is_layer_root_entry(&entry_path) {
+            if is_dir {
```
fitzthum, line 328: hmm if we're checking that the entry is `.` are there really cases where that wouldn't be a directory?

image-rs/src/stream/unpack.rs
```diff
@@ -367,27 +378,181 @@ pub async fn unpack<R: AsyncRead + Unpin>(input: R, layer_dir: &Path) -> UnpackR
...
+enum FollowSymlink {
+    Yes,
+    No,
+}
+
+/// Open file handle or bare path for applying ownership and mode.
+struct EntryHandle<'a> {
+    path: &'a Path,
+    fd: Option<BorrowedFd<'a>>,
+}
```
fitzthum: I think I would prefer using pathrs wherever we can and only dealing with things in terms of fds. Supporting both paths and fds seems like unnecessary complexity.

fitzthum (commented):
Nice. Thanks for taking a look at this.

I think the metadata stuff might be slightly too generic, which ends up creating extra layers and complexity. I would just do everything in terms of fds, and if possible use pathrs basically everywhere (which generates fds).

image-rs/src/stream/unpack.rs
Xynnn007: Yes. We could promote this in a separate PR

image-rs/src/stream/unpack.rs
Xynnn007: Ok. Updated.

image-rs/src/stream/unpack.rs
Xynnn007: Yes. Some malwared tar layer could exist. Let me add a warn if the condition does not match

Xynnn007:
@fitzthum Oh I think this PR does not help. I test again and find `.` is still different. The reason is that for overlayfs, we need to change mtime once the mount is done. In `unpack.rs` we are only handling the layers that are the overlay mount sources than target.

fitzthum:
> Oh I think this PR does not help. I test again and find . is still different. The reason is that for overlayfs, we need to change mtime once the mount is done. In unpack.rs we are only handling the layers that are the overlay mount sources than target.

Still a nice PR tho. I made a comment about mtime on the issue.

image-rs/src/stream/unpack.rs
Xynnn007: Updated.

Xynnn007:
Let me rename the title and change the commit message.

image-rs/src/stream/unpack.rs
```diff
@@ -367,29 +430,139 @@ pub async fn unpack<R: AsyncRead + Unpin>(input: R, layer_dir: &Path) -> UnpackR
...
+        false => filetime::set_file_times(path, ft, ft),
+    }
+    .map_err(|e| UnpackError::SetMTimeFailed {
+        source: e.into(),
+        path: err_path.to_string(),
+    })
+}
+
+#[allow(clippy::too_many_arguments)]
+fn apply_link_metadata(
```
fitzthum: This doesn't seem quite right. Can we do the resolution in the caller so that we don't have so many arguments here. I still feel like we have too many layers of abstraction.

image-rs/src/stream/unpack.rs
```diff
@@ -367,29 +430,139 @@ pub async fn unpack<R: AsyncRead + Unpin>(input: R, layer_dir: &Path) -> UnpackR
...
+    root: &Root,
+    layer_dir: &Path,
+    rel_path: &Path,
+    err_path: &str,
```
fitzthum: No need to pass this. Just construct the error in the caller if you don't have the right info.

image-rs/src/stream/unpack.rs
```diff
@@ -367,29 +430,139 @@ pub async fn unpack<R: AsyncRead + Unpin>(input: R, layer_dir: &Path) -> UnpackR
...
+            libc::AT_EMPTY_PATH,
+        )
+    };
+    if ret != 0 {
+        bail!("failed to fchmodat: {:?}", io::Error::last_os_error());
+    }
     Ok(())
 }
 
+fn fchmod_handle(handle: &Handle, mode: Option<u32>) -> Result<()> {
```
fitzthum: This function seems unnecessary.

image-rs/src/stream/unpack.rs
```diff
@@ -367,29 +430,139 @@ pub async fn unpack<R: AsyncRead + Unpin>(input: R, layer_dir: &Path) -> UnpackR
...
-    if let Some(mode) = mode {
-        let ret = unsafe { libc::fchmod(fd.as_raw_fd(), mode as libc::mode_t) };
-        if ret != 0 {
-            bail!(
-                "failed to set permissions: {:?}",
-                io::Error::last_os_error()
-            );
-        }
-    }
+fn resolve_entry(root: &Root, rel_path: &Path, nofollow: bool) -> UnpackResult<Handle> {
```
fitzthum: We don't need this function. Just use resolve or resolve_nofollow in the caller. That is clearer than using a boolean for follow/nofollow.

fitzthum (commented):
Sorry, I have more nits. I feel like this is more complicated than it needs to be.

In our unpack method we can resolve the entry path using follow or nofollow. Then we have a fuction for setting mode, owernship, and mtime and call them directly from the main conditional as needed.

image-rs/src/stream/unpack.rs
Xynnn007: Yes let me delete this layer

image-rs/src/stream/unpack.rs
Xynnn007: Yes. Like the previous comment, I will resolve together

fitzthum (approved):
LGTM. Thanks for putting up with my nits

PR #1475, opened 2026-05-18
kbs_protocol: support composite attestation for IBM SE

Xynnn007:
Now in the kbs protocol level, the report_data/runtime_data digest could be set, thus the binding between device evidences and CPU evidence can be acieved for IBM SE.

Based on #1469


Xynnn007:
cc @BbolroC I think this can work now. More concretely, the composite attestation relies on the "report_data binding" ability for the attester. It binds a hash digest of all device evidences to the report_data of the CPU attester. Now IBM SE supports "report data" ability, thus supports this ability too.

BbolroC (approved):
Thanks @Xynnn007 for raising this! I love it.

Issue #1414, opened 2026-04-13
[RFC] Standalone binary for image-rs


bpradipt:
Reasons
- Helps to debug and isolate issues with in-guest image pull in the field
- Allows to pre-populate images in VM image for some specific/niche cases

cc @confidential-containers/guest-components-maintainers

fitzthum:
Sure. We have some tests that already do something like that. I think the kata `agent-ctl` might also use `image-rs` for simple container unpacking.

Xynnn007:
I think the implementation would be pretty easy.


A question: which API should be exposed by the binary in which form in your mind?

bpradipt:
For me to debug in-guest image pull issue in the field, enabling the following features for the pull_image API helped
```
"pull-debug-bin", "oci-client-rustls", "signature-cosign-rustls"
```

Here is an example that I used - https://github.com/bpradipt/guest-components/commit/a51f181d00c8f8f02a662933ae1ee4c599a3c4b7

mythi:
Do we think the future is like #1080? Maybe a simple grpc client is sufficient.

bpradipt:
> Do we think the future is like https://github.com/confidential-containers/guest-components/pull/1080? Maybe a simple grpc client is sufficient.

So the flow will be something like grpc-client ->  kata-agent to execute the in-guest image pull ?
Will it be possible on a live system when the Kata VM is up including kata-agent but there are failures in container creation?

mythi:
> > Do we think the future is like [#1080](https://github.com/confidential-containers/guest-components/pull/1080)? Maybe a simple grpc client is sufficient.
> 
> So the flow will be something like grpc-client -> kata-agent to execute the in-guest image pull ? Will it be possible on a live system when the Kata VM is up including kata-agent but there are failures in container creation?

No, I had missed the part that you wanted to debug a live system, sorry.

PR #1480, opened 2026-05-19
image-rs: add conformance test for image pulling

fitzthum:
Add a test that pulls an image with umoci and with image-rs and then compares them gomtree.

To avoid issues with the network, the test spins up a local registry and loads images from test_data/conformance_images into it. This PR does not add any test images (thus the test doesn't do anything). We can add images to the repo (pending some license discussion) or you can just run the test locally with your own test images. This also adds a script to pull a test image and put it in the directory with the right format.

I think we may want to generalize this local registry approach and use it for some of the other tests too. This could be handy for performance tests, for instance, which we should probably add. Anecdotally image-rs seems significantly slower than umoci.

While umoci is a good reference implementation to compare to, it differs from image-rs in that it unpacks an image directly to the filesystem (while image-rs uses overlayfs). This leads to some differences even when the image is unpacked correctly.

For example, overlayfs does not correctly report hardlink counts. Thus, we don't include nlink in our comparison. Overlayfs can also have some discrepancies in terms of size. For now, this is included in the comparison and these differences are reported.

Further, it's not yet clear if the mtime of the root dir will be correct with overlayfs. This is also reported.

This test is not yet ready to be required, but it's pretty handy.

### DIFF
--- a/image-rs/tests/unpack_conformance.rs
+++ b/image-rs/tests/unpack_conformance.rs
@@ -1,0 +1,198 @@
+// Copyright (c) 2026 NVIDIA Corporation
+//
+// SPDX-License-Identifier: Apache-2.0
+
+mod conformance;
+
+use conformance::{conformance_images_dir, image_config_for_registry, LocalRegistry};
+use std::path::PathBuf;
+use tempfile::TempDir;
+
+const REGISTRY_PORT: u16 = 5000;
+
+fn keep_artifacts() -> bool {
+    std::env::var("IMAGE_RS_KEEP_TEST_ARTIFACTS").is_ok()
+}
+
+struct PulledImage {
+    _work_dir: TempDir,
+    bundle_dir: TempDir,
+}
+
+impl PulledImage {
+    fn rootfs(&self) -> PathBuf {
+        self.bundle_dir.path().join("rootfs")
+    }
+
+    fn cleanup(self) {
+        if !keep_artifacts() {
+            let _ = nix::mount::umount(&self.rootfs());
+        }
+    }
+}
+
+async fn pull_image_rs(image_ref: &str, registry_url: &str) -> PulledImage {
+    let work_dir = tempfile::tempdir().unwrap();
+    let bundle_dir = tempfile::tempdir().unwrap();
+
+    let config = image_config_for_registry(work_dir.path().to_path_buf(), registry_url);
+    let mut client = image_rs::builder::ClientBuilder::from(config)
+        .build()
+        .await
+        .expect("Failed to build image client");
+    client
+        .pull_image(image_ref, bundle_dir.path(), &None, &None)
+        .await
+        .expect("image-rs pull failed");
+
+    PulledImage {
+        _work_dir: work_dir,
+        bundle_dir,
+    }
+}
+
+async fn pull_umoci(image_ref: &str) -> PulledImage {
+    use tokio::process::Command;
+
+    let work_dir = tempfile::tempdir().unwrap();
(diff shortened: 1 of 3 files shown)


image-rs/tests/unpack_conformance.rs
```diff
@@ -0,0 +1,198 @@
...
+            "-p",
+        ])
+        .arg(umoci_rootfs)
+        .output()
+        .expect("Failed to execute gomtree");
+    assert!(
+        spec_output.status.success(),
+        "gomtree spec generation failed: {}",
+        String::from_utf8_lossy(&spec_output.stderr)
+    );
```
Xynnn007, line 121: Note that this requires that `umoci` and `gomtree` are both installed on the machine. We'd better do a test before call them.

image-rs/tests/unpack_conformance.rs
```diff
@@ -0,0 +1,198 @@
...
+        .output()
+        .expect("Failed to execute gomtree verify");
+
+    if !verify_output.status.success() {
+        let stderr = String::from_utf8_lossy(&verify_output.stderr);
+        let stdout = String::from_utf8_lossy(&verify_output.stdout);
+        println!(
+            "Rootfs mismatch for {} (image-rs vs umoci):\n{}{}",
+            image_name, stdout, stderr
+        );
```
Xynnn007, line 140: The `println` usually does not work in test and can be ignored. return this as an error to the upper layer, and do panic/assert would be better

image-rs/tests/unpack_conformance.rs
```diff
@@ -0,0 +1,198 @@
...
+        .expect("Failed to read conformance-images directory")
+        .filter_map(|e| e.ok())
+        .map(|e| e.path())
+        .filter(|p| p.extension().is_some_and(|ext| ext == "tar"))
+        .collect();
+
+    if images.is_empty() {
+        eprintln!("No test images found, skipping");
+        return;
+    }
```
Xynnn007, line 162: Up to now, I start to know that the aim of the PR is not to do comparation everytime we do `cargo test`, but provide a way for users to compare with umoci, right?

If this is the case, and also integrate "pull image + comparation" into CI is not urgent, I'd suggest to start from https://github.com/confidential-containers/guest-components/issues/1414. Then we could provide a document to help users to do pull + comparation work using the image-rs tool.

Xynnn007 (commented):
This is a unfinished review, until I realize the aim of the PR. Please correct me then I could better understand.

image-rs/tests/unpack_conformance.rs
fitzthum: Yes, at some point we might want to do explicit checks in every PR, but that requires figuring out a few things (which images do we want; can we add them to the repo; which attributes should we check; etc). For now, this is just for diagnostic purposes. I don't think we even necessarily need to merge, but I figured I would share what I have been using to test since we are working on image-rs. Might as well put it into the repo imo.

I don't think this is related to #1414 tho. It doesn't matter if we use image-rs as a crate or a binary. In fact, it may be better to test it as a crate, since that's what we mostly use.

Xynnn007: Yes. I agree with the part of how umoci pulls images and how we do comparation. 

The issue is that "where we can get an image rootfs pulled by image-rs". you are using an integration test in the PR, while I usually directly change the code of [the unit test](https://github.com/confidential-containers/guest-components/blob/main/image-rs/src/image.rs#L558-L600).

Before this becomes a required component of CI check in a proper way, it might not be proper to get the code in.

Instead, a doc to document these steps can be a merge-able thing imo.

PR #1482, opened 2026-05-20
trustee-attester: support kbs+plugin URI in --path argument

MatiasVara:
Add support for kbs+plugin URI path. When the path is `kbs+plugin:///`, use `try_from()` to get also plugin name. This shall be merged after #1468

### DIFF
--- a/attestation-agent/kbs_protocol/src/bin/trustee-attester/main.rs
+++ b/attestation-agent/kbs_protocol/src/bin/trustee-attester/main.rs
@@ -5,6 +5,7 @@
 
 //! Attest and fetch confidential resources from Trustee
 
+use anyhow::Context;
 use anyhow::Result;
 use base64::engine::general_purpose::STANDARD;
 use base64::Engine;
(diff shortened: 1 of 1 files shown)


attestation-agent/kbs_protocol/src/bin/trustee-attester/main.rs
```diff
@@ -87,18 +87,21 @@ async fn main() -> Result<()> {
...
-
             if let Some(init) = initdata {
                 client_builder = client_builder.add_initdata(init);
             }
             let mut client = client_builder.build()?;
 
-            let resource = ResourceUri::new("", &resource_path)?;
+            let resource = if path.contains("://") {
+                ResourceUri::try_from(path.as_str())
+                    .map_err(|e| anyhow::anyhow!(e))?
```
Xynnn007: We could use
```suggestion
                ResourceUri::try_from(path.as_str()).context("failed to parse ResourceUri")?
```

and `use anyhow::Context`

Xynnn007 (commented):
Only one nit

attestation-agent/kbs_protocol/src/bin/trustee-attester/main.rs
MatiasVara: is it not missing `.map_err(|e| anyhow::anyhow!(e))` before the context?

Xynnn007: It's intend. usually `.context()` would help to automatically convert the original `Error` to `anyhow::Error` with some `context` message thus no need `map_err` to `anyhow::anyhow!`. But not sure if it could alsobe compiled successfully here as `Error` of the `ResourceUri` is `String` if I remembered correctly.

MatiasVara: I could not compile it without `.map_err()`, is the updated version fine?

Xynnn007: Yes. Fine!

fitzthum (approved):
LGTM

PR #1484, opened 2026-05-21
ocicrypt-rs: add warning when OCICRYPT_KEYPROVIDER_CONFIG is not set

Agrek11:
## Summary

  Adds a `tracing::warn!` to `OcicryptConfig::from_env` when the `OCICRYPT_KEYPROVIDER_CONFIG` env var is unset. Today the function returns `Ok(None)` silently — downstream callers then fail with a generic decrypt error that gives no
  hint that the env var was the real cause. Behavior is unchanged: `Ok(None)` is still returned, only an observability log is added.

  `tracing` is already a workspace dependency (used by `image-rs`, `attestation-agent`, `confidential-data-hub`, `api-server-rest`); this PR adds it to `ocicrypt-rs/Cargo.toml` to match.

  Similar in spirit to #630 (don't swallow `pre_unwrap_key()` error) — surfacing a previously-silent failure mode at the boundary where the operator can act, without changing the API contract.

  ## Test plan

  - [x] No change to `from_env` return signature or semantics
  - [x] No change to any other code path
  - [x] Existing `test_ocicrypt_config` continues to exercise the `Ok(filename)` branch
  - [ ] CI: builds across `block-cipher-openssl`, `block-cipher-ring`, `keywrap-keyprovider-*` feature matrices on this PR

### DIFF
--- a/ocicrypt-rs/Cargo.toml
+++ b/ocicrypt-rs/Cargo.toml
@@ -30,6 +30,7 @@ serde_json.workspace = true
 sha2 = { workspace = true, optional = true }
 tokio = { workspace = true, features = ["rt-multi-thread"], optional = true }
 tonic = { workspace = true, optional = true }
+tracing = { workspace = true }
 ttrpc = { workspace = true, features = ["async"], optional = true }
 zeroize = { workspace = true, optional = true }
 
--- a/ocicrypt-rs/src/config.rs
+++ b/ocicrypt-rs/src/config.rs
@@ -287,13 +287,18 @@ impl OcicryptConfig {
             .map_err(|e| anyhow!("Error reading file {:?}", e.to_string()))
     }
 
-    /// from_env tries to read the configuration file at the following locations
-    /// ${OCICRYPT_KEYPROVIDER_CONFIG} == "/etc/ocicrypt_keyprovider.json"
-    /// If no configuration file could be found or read a null pointer is returned
+    /// from_env reads the configuration file pointed at by the given environment
+    /// variable. If the variable is not set, `Ok(None)` is returned and a warning
+    /// is emitted, since no keyproviders will be registered.
     pub fn from_env(env: &str) -> Result<Option<OcicryptConfig>> {
-        // find file name from environment variable, ignore error if environment variable is not set.
         match std::env::var(env) {
-            Err(_e) => Ok(None),
+            Err(_e) => {
+                tracing::warn!(
+                    env_var = env,
+                    "ocicrypt keyprovider config env var is not set; no keyproviders will be registered"
+                );
+                Ok(None)
+            }
             Ok(filename) => OcicryptConfig::from_file(filename.as_str()).map(Some),
         }
     }


ocicrypt-rs/src/config.rs
```diff
@@ -287,13 +288,26 @@ impl OcicryptConfig {
...
         match std::env::var(env) {
-            Err(_e) => Ok(None),
+            Err(_e) => {
+                warn!(
+                    env_var = env,
+                    "ocicrypt keyprovider config environment variable is not set; \
+                     no keyproviders will be registered. Pulls of images with \
+                     encrypted layers (`org.opencontainers.image.enc.keys.provider.*` \
+                     annotations) will fail with a generic decrypt error. \
+                     See https://github.com/containers/ocicrypt/blob/main/docs/keyprovider.md"
```
Xynnn007: This address can be moved as it points to `main` ref. Can we use a fixed tag/commit ref?

ocicrypt-rs/src/config.rs
```diff
@@ -478,4 +492,26 @@ mod tests {
         assert!(p4.ttrpc.is_some());
         assert!(p4.native.is_none());
     }
+
+    /// `from_env` must return `Ok(None)` (not `Err`) when the requested
+    /// environment variable is not set, so callers that intentionally treat
+    /// missing config as "no keyproviders" continue to work. A warning is
+    /// also emitted, but that is observable only via a tracing subscriber
+    /// and is not asserted here -- this test guards the return-value
+    /// contract specifically.
```
Xynnn007: The test name is clear enough, thus we might not need the comments

Xynnn007 (commented):
Thanks @Agrek11  Overall LGTM. two nits

ocicrypt-rs/src/config.rs
```diff
@@ -478,4 +492,26 @@ mod tests {
...
+    #[test]
+    fn test_from_env_returns_none_when_unset() {
+        // Use a name that is guaranteed not to be set in any reasonable
+        // test environment, so this test does not require mutating shared
+        // process env state.
+        let missing = "OCICRYPT_KEYPROVIDER_CONFIG_NOT_SET_FOR_TEST_BB7F2A";
```
fitzthum: What is `BB7F2A`?

fitzthum (commented):
fwiw I would prefer more consice summary of the PR if possible (e.g. "add warning when OCICRYPT_KEYPROVIDER_CONFIG" is not set

I think we copy this behavior from gocrypt-rs. Otherwise I would suggest changing it completely. The env var thing is not very robust.

ocicrypt-rs/src/config.rs
fitzthum: tbh I'm not sure this test is really necessary

ocicrypt-rs/src/config.rs
Agrek11: Done -- pinned to v1.3.0 (latest containers/ocicrypt release tag) in the amended commit 941a6ed

ocicrypt-rs/src/config.rs
Agrek11: right,  dropped the test entirely

ocicrypt-rs/src/config.rs
Agrek11: It was a uniqueness marker to avoid colliding with any real env-var name in shared test process state, but with the whole test gone it's no longer relevant

Agrek11:
Thanks @fitzthum, 
1) trimmed the PR summary.
2) **"env var thing not very robust"** --  agreed  env-var is brittle. keeping this PR scoped to the warn on the
  existing API; rather do the bigger redesign separately
 two redesign paths I see:
      (a) drop env-var entirely, take config path / OcicryptConfig as an explicit arg, deprecate from_env over a release. touches every caller (image-rs, kata-agent peer-pods, etc.) + Go ocicrypt probably wants a parallel change.
     (b) keep from_env, add from_path(&Path) alongside, mark from_env deprecated.
 both are separate PRs with their own design call. happy to open a tracking issue and prototype either

Xynnn007 (approved):

ocicrypt-rs/Cargo.toml
```diff
@@ -30,6 +30,7 @@ serde_json.workspace = true
 sha2 = { workspace = true, optional = true }
 tokio = { workspace = true, features = ["rt-multi-thread"], optional = true }
 tonic = { workspace = true, optional = true }
+tracing.workspace = true
```
mythi: should this be `optional = true` since this is behind `keywrap-keyprovider` feature.

Agrek11: Good catch -- amended. `tracing` is now `optional = true` and pulled in via
  the `keywrap-keyprovider` feature. `from_env` and its test
  `test_ocicrypt_config` are gated to the same feature, since the
  only production caller (lib.rs keyprovider registration block) is
  already cfg'd on `keywrap-keyprovider`. Non-keyprovider builds no longer
  incur the tracing dep.

ocicrypt-rs/src/config.rs
```diff
@@ -287,13 +289,31 @@ impl OcicryptConfig {
...
+    /// Gated behind `keywrap-keyprovider` because that is the only feature
+    /// path that consumes the env-var-loaded `OcicryptConfig`; this keeps
+    /// the `tracing` dependency optional for non-keyprovider builds.
+    #[cfg(feature = "keywrap-keyprovider")]
     pub fn from_env(env: &str) -> Result<Option<OcicryptConfig>> {
-        // find file name from environment variable, ignore error if environment variable is not set.
         match std::env::var(env) {
-            Err(_e) => Ok(None),
+            Err(_e) => {
+                warn!(
```
mythi:
```suggestion
                tracing::warn!(
```

mythi (approved):
Looks OK. lines 11:12 could be easily dropped with the suggestion

ocicrypt-rs/src/config.rs
Agrek11: Applied — tracing::warn! inline now,

fitzthum:
Please rebase.

fitzthum:
Ok, see error in the CI. Possibly caused by adding the feature gates. I"m not sure this needed anyway. Seems fine to include tracing for ocicrypt-rs generally, although maybe @mythi was suggesting something slightly different.

Also, you can shorten the comment and the error message. The functionality is basically self-explanatory.

Agrek11:
Thanks @fitzthum Dropped all the feature gates and made tracing a regular dependency for ocicrypt-rs, so from_env is no longer gated (matches the previous behavior on main) and the unused-import churnis gone. Also shortened the doc comment and the warning message.

Xynnn007:
Let's see how CI works.

mythi:
> Seems fine to include tracing for ocicrypt-rs generally, although maybe @mythi was suggesting something slightly different.

my comment/question was just that if the dependency can be declared as `optional = true`. I believe the change missed `keywrap-keyprovider = [ "tracing" ]`.

fitzthum:
> my comment/question was just that if the dependency can be declared as optional = true. I believe the change missed keywrap-keyprovider = [ "tracing" ].

Ok, are you good with the current version? Seems fine to just add the feature imo

mythi (approved):

mythi:
> > my comment/question was just that if the dependency can be declared as optional = true. I believe the change missed keywrap-keyprovider = [ "tracing" ].
> 
> Ok, are you good with the current version? Seems fine to just add the feature imo

+1

fitzthum (approved):

PR #1486, opened 2026-05-22
resource-uri: generalize ResourceUri for plugin-aware KBS paths

Xynnn007:
The main idea of the patch includes the following:

1. Generalize `ResourceUri` to carry an explicit plugin, optional query, and a variable-length segment path, so URI parsing is no longer tied to a fixed three-segment layout. 
2. Add `ResourcePluginPath` for the Trustee resource plugin convention (repo/type/tag); legacy KBCs (sample, offline_fs, online_sev) convert through it and only accept that plugin with three segments, while cc_kbc forwards the full URI to kbs_protocol. Update kbs_protocol to build KBS URLs from the plugin name and joined segments, and refresh RESOURCE_URI.md. 
3. Also fix online_sev_kbc compilation by using the shared protos crate instead of locally generated keybroker gRPC code, and update bincode 2.x decoding.

cc @salmanyam about https://github.com/confidential-containers/guest-components/pull/1208, when I think about multiple sections about the plugin support, looks like the code cannot be that easy as we also have `kbc` abstraction to be changed. This abstraction will cause a lot of other work different from the plugin support itself. Thus it might be difficult to achieve the goal via a simple code changes. Thus I get this PR. Feel free to take a look if this meet your needs.


attestation-agent/kbc/src/online_sev_kbc/mod.rs
fitzthum, line 1: tbh we may want to remove this entire kbc, but let's worry about that later

attestation-agent/deps/resource_uri/src/lib.rs
```diff
@@ -14,17 +14,21 @@ const RESOURCE_ID_ERROR_INFO: &str =
...
 const SCHEME: &str = "kbs";
-const DEFAULT_PLUGIN: &str = "resource";
+pub const DEFAULT_RESOURCE_PLUGIN: &str = "resource";
 
 /// Resource Id document <https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/docs/RESOURCE_URI.md>
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct ResourceUri {
     pub kbs_addr: String,
-    pub repository: String,
+    pub segment: Vec<String>,
```
fitzthum: can we call this the path. also might be nice to change the order of the struct members to plugin, path, query. Also, maybe we should change `kbs_addr` -> `kbs_address` or `kbs_uri`. We don't use abbreviations like that as much these days.

attestation-agent/deps/resource_uri/src/lib.rs
```diff
@@ -111,31 +141,22 @@ impl ResourceUri {
...
 
     pub fn whole_uri(&self) -> String {
-        let scheme = match &self.plugin {
-            Some(p) => format!("{SCHEME}+{p}"),
-            None => SCHEME.to_string(),
-        };
         let uri = format!(
-            "{scheme}://{}/{}/{}/{}",
-            self.kbs_addr, self.repository, self.r#type, self.tag
+            "kbs+{scheme}://{kbs_addr}/{segment}",
```
fitzthum: nit: `scheme` isnt' the scheme here it's the plugin

attestation-agent/docs/RESOURCE_URI.md
```diff
@@ -24,18 +24,23 @@ in different ways.
...
 ```
 
+The path after the authority consists of one or more slash-separated segments.
+There is no fixed limit on how many segments a URI may contain.
+
 ### Scheme
 
-The scheme always begins with `kbs`. Typically the scheme is simply `kbs://`, but a plus sign
-can be used to specify that the resource should be fulfilled by a particular plugin.
+The scheme identifies which Trustee plugin should fulfill the resource.
```
fitzthum: I think we should soften this a little bit to so that it sounds like the plugin being in a scheme is just an option, not something that we expect everyone to use.

fitzthum (commented):
Nice. A couple small comments. 

In the future I think we should delete all the legacies KBCs. The codebase is a little sprawling and anything we can remove, we should imo. 

Also, I am open to supporting kbs_uris in resource_uri, but let's discuss separately.

fitzthum (approved):
LGTM

fitzthum:
@confidential-containers/guest-components-maintainers PTAL 

let's try to merge this next week

Xynnn007:
I assume the thumb up from @salmanyam means an approval : )

salmanyam:
> I assume the thumb up from @salmanyam means an approval : )

Hi @Xynnn007, I went through the code. Looks good. I am testing your PR with a plugin in trustee. Could you please allow me tonight to do the testing? I need a couple of hours to do that. I will comment here once that is done.

Xynnn007:
> > I assume the thumb up from @salmanyam means an approval : )
> 
> Hi @Xynnn007, I went through the code. Looks good. I am testing your PR with a plugin in trustee. Could you please allow me tonight to do the testing? I need a couple of hours to do that. I will comment here once that is done.

Certainly. Please go with it!

salmanyam:
@Xynnn007 I have tested my plugin. It works fine with the following format: `kbs+<plugin_name>:///<path>`. The default resource plugin also works fine. So, good from my side.

PR #1487, opened 2026-05-22
asr: support base64 URL-safe-nopad runtime_data on evidence API

### LINKED ISSUE #1434: api-server-rest `/aa/evidence` endpoint corrupts binary `runtime_data` due to UTF-8 string parsing
### Describe the bug

The `runtime_data` query parameter in the `/aa/evidence` API cannot correctly handle arbitrary binary data (non-visible characters or bytes in the 128-255 range).
Currently, the query string is parsed into a `HashMap<String, String>`. This forces UTF-8 validation. If raw bytes (e.g., `%FF`) are passed via the URL, they trigger lossy decoding and are replaced by the Unicode replacement character (`U+FFFD`). This corrupts the original byte array and unexpectedly increases the byte length, which may also easily exceed the 64-character limit.

## Passing a non-UTF8 byte (%FF)
curl "http://127.0.0.1:8006/aa/evidence?runtime_data=%FF"

Expected: The backend receives the exact byte `[255]`.
Actual: The backend parses it into a String and `into_bytes()` yields `[239, 191, 189]` (the UTF-8 encoding of ``).
Suggested Fix:
Instead of parsing raw binary data as UTF-8 strings, the API should require `runtime_data` to be Base64 or Hex encoded by the client. The server can then safely parse it as a standard string and decode it back into a `Vec<u8>`.

### CoCo version information

kata v3.28.0

Snp

```shell

```

Xynnn007:
Add an optional `encoding=base64` query parameter to `/aa/evidence`. When set, decode `runtime_data` with URL-safe base64 (no padding) before forwarding to the attestation agent. Preserve the existing behavior when `encoding` is omitted.

Close #1434

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -174,6 +174,7 @@ dependencies = [
  "form_urlencoded",
  "hyper 0.14.32",
  "protos",
+ "rstest",
  "serde",
  "serde_json",
  "shadow-rs",
--- a/api-server-rest/Cargo.toml
+++ b/api-server-rest/Cargo.toml
@@ -21,6 +21,9 @@ tracing-subscriber.workspace = true
 ttrpc = { workspace = true, features = ["async"] }
 protos = { path = "../protos", default-features = false, features = ["ttrpc"] }
 
+[dev-dependencies]
+rstest.workspace = true
+
 [build-dependencies]
 serde_json = { workspace = true }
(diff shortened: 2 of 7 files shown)


fitzthum (approved):
LGTM. Thanks for looking into this @Xynnn007 

(needs rebase)

PR #1489, opened 2026-05-25
attester/tdx: prefer TSM measurements for RTMR and MRCONFIGID

Xynnn007:
Refactor the TDX attester to use upstream kernel interfaces consistently:

- Add TDX sysfs helpers in `tsm_report` for MRCONFIGID read and RTMR read/write under `/sys/devices/virtual/misc/tdx_guest/measurements` (kernel >= 6.16).
- Wire the TDX attester to prefer those sysfs paths for runtime measurement and initdata binding; fall back to `/dev/tdx_guest` ioctl `GET_REPORT0` for RTMR/MRCONFIGID reads when sysfs is unavailable.
- Generate TD quotes only via TSM configfs (`/sys/kernel/config/tsm/report`); remove the libtdx/DCAP quote ioctl fallback.
- Keep optional `tdx-attest-dcap-ioctls` (`tdx-attester-libtdx`) only for RTMR extend when sysfs is absent.
## Why
Upstream Linux 6.16 exposes stable sysfs for TDX guest measurements. Using it avoids parsing raw TD reports for common operations and aligns quote generation with the TSM_REPORT ABI. The ioctl/libtdx paths remain as fallbacks for older or partial kernel setups.

### DIFF
--- a/attestation-agent/attester/src/tdx/mod.rs
+++ b/attestation-agent/attester/src/tdx/mod.rs
@@ -27,30 +27,6 @@ const TDX_GUEST_IOCTL: &str = "/dev/tdx_guest";
 pub fn detect_platform() -> bool {
     TsmReportPath::new(TsmReportProvider::Tdx).is_ok() || Path::new(TDX_GUEST_IOCTL).exists()
 }
-
-#[allow(unused_variables)]
-fn get_quote_ioctl(report_data: &[u8]) -> Result<Vec<u8>> {
-    cfg_if::cfg_if! {
-            if #[cfg(feature = "tdx-attest-dcap-ioctls")] {
-                let tdx_report_data = tdx_attest_rs::tdx_report_data_t {
-                    // report_data.resize() ensures copying report_data to
-                    // tdx_attest_rs::tdx_report_data_t cannot panic.
-                    d: report_data.try_into().unwrap(),
-                };
-
-                match tdx_attest_rs::tdx_att_get_quote(Some(&tdx_report_data), None, None, 0) {
-                    (tdx_attest_rs::tdx_attest_error_t::TDX_ATTEST_SUCCESS, Some(q)) => Ok(q),
-                    (error_code, _) => Err(anyhow!(
-                        "TDX DCAP get_quote: failed with error code: {:?}",
-                        error_code
-                    )),
-                }
-            } else {
-                Err(anyhow!("TDX DCAP ioctls: support not available"))
-        }
-    }
-}
-
 // Return true if the TD environment can extend runtime measurement.
 // The best guess at the moment is that if tdx-attest-dcap-ioctls
 // is enabled, runtime measurement is possible. It's also possible
@@ -126,16 +102,9 @@ impl Attester for TdxAttester {
 
         report_data.resize(TDX_REPORT_DATA_SIZE, 0);
 
-        let quote_bytes = TsmReportPath::new(TsmReportProvider::Tdx).map_or_else(
-            |notsm| {
-                get_quote_ioctl(&report_data)
-                    .context(format!("TDX Attester: quote generation using ioctl() fallback failed after a TSM report error ({notsm})"))
-            },
-            |tsm| {
-                tsm.attestation_report(TsmReportData::Tdx(report_data.clone()))
-                    .context("TDX Attester: quote generation using TSM reports failed")
-            },
-        )?;
+        let quote_bytes = TsmReportPath::new(TsmReportProvider::Tdx)
+            .context("TDX Attester: failed to create TSM Report path")?
+            .attestation_report(TsmReportData::Tdx(report_data))?;
 
         let engine = base64::engine::general_purpose::STANDARD;
         let quote = engine.encode(quote_bytes);
--- a/attestation-agent/attester/src/tdx/mod.rs
+++ b/attestation-agent/attester/src/tdx/mod.rs
@@ -15,11 +15,15 @@ use report::TdReport;
 use scroll::Pread;
 use serde::{Deserialize, Serialize};
 use std::path::Path;
+use tracing::debug;
 
 mod report;
 mod rtmr;
 
 const TDX_REPORT_DATA_SIZE: usize = 64;
+
+/// `TDX_REGISTER_LENGTH` / SHA384 digest size
+const TDX_REGISTER_LENGTH: usize = 48;
 
 const TDX_TSM_SYSFS_PATH: &str = "/sys/devices/virtual/misc/tdx_guest/measurements";
 const TDX_GUEST_IOCTL: &str = "/dev/tdx_guest";
@@ -66,31 +70,24 @@ impl TdxAttester {
 #[repr(C)]
 struct TdxReportReq {
     report_data: [u8; 64],
-
     d: [u8; 1024],
-}
-
-impl Default for TdxReportReq {
-    fn default() -> Self {
-        Self {
-            report_data: [0; 64],
-            d: [0; 1024],
-        }
-    }
 }
 
 const TDX: Group = Group::new(b'T');
 const TDX_CMD_GET_REPORT0: Ioctl<WriteRead, &TdxReportReq> = unsafe { TDX.write_read(0x1) };
 
 impl TdxAttester {
-    fn get_report() -> Result<TdReport> {
-        let mut report = TdxReportReq::default();
-        let mut fd =
-            std::fs::File::open(TDX_GUEST_IOCTL).context("Open TD report ioctl() failed")?;
+    fn ioctl_get_report(&self) -> Result<TdReport> {
+        let mut report = TdxReportReq {
+            report_data: [0; 64],
+            d: [0; 1024],
+        };
+        let mut fd = std::fs::File::open(TDX_GUEST_IOCTL)
+            .with_context(|| format!("open {TDX_GUEST_IOCTL} failed"))?;
 
         TDX_CMD_GET_REPORT0
             .ioctl(&mut fd, &mut report)
-            .context("Get TD report ioctl() failed")?;
+            .context("Get TD report via ioctl() failed")?;
 
         let td_report = report
             .d
@@ -136,9 +133,9 @@ impl Attester for TdxAttester {
         let ccmr_index = self.pcr_to_ccmr(register_index);
         let rtmr_index = ccmr_index - 1;
 
-        let extend_data: [u8; 48] = pad(&event_digest);
+        let extend_data: [u8; TDX_REGISTER_LENGTH] = pad(&event_digest);
 
-        tracing::debug!(
+        debug!(
             "TDX Attester: extend RTMR{rtmr_index}: {}",
             hex::encode(extend_data)
         );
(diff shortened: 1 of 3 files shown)


mythi:
> Refactor the TDX attester to use upstream kernel interfaces consistently:
> 
> * Add TDX sysfs helpers in `tsm_report` for MRCONFIGID read and RTMR read/write under `/sys/devices/virtual/misc/tdx_guest/measurements` (kernel >= 6.16).

`tsm_report` is not for RTMR so let's keep the RTMR changes isolated to TDX attester. In other works, we can drop [ed75d7e](https://github.com/confidential-containers/guest-components/pull/1489/commits/ed75d7eeeac2957f503526db0b451cba05837a69).

> * Generate TD quotes only via TSM configfs (`/sys/kernel/config/tsm/report`); remove the libtdx/DCAP quote ioctl fallback.
> * Keep optional `tdx-attest-dcap-ioctls` (`tdx-attester-libtdx`) only for RTMR extend when sysfs is absent.

This makes sense

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -240,14 +232,17 @@ mod tests {
     #[tokio::test]
     async fn test_tdx_get_evidence() {
         let attester = TdxAttester::default();
-        let report_data: Vec<u8> = vec![0; 48];
+        let report_data: Vec<u8> = vec![0; TDX_REGISTER_LENGTH];
```
mythi: nit: this is not corrrect. 48 comes from kbs protocol but report data itself is 64.

Xynnn007:
@mythi I found the commit https://git.sceen.net/linux/linux-stable.git/commit/include/trace?id=ae5ec8adb8ec9c2aa916f853737c101faa87e5ba that says
"""
    Add a general sysfs scheme for publishing "Measurement" values
    provided by the architecture's TEE Security Manager.
"""

so sysfs for RTMRs is also a sub-capibility of TSM?

attestation-agent/attester/src/tdx/mod.rs
Xynnn007: Yes thanks. Fixed.

Xynnn007:
Or - we need to rename `tsm_report` submodule to `tsm`?

mythi:
> so sysfs for RTMRs is also a sub-capibility of TSM?

True but `tsm_reports` is for the the reports parts of it only. I'd prefer to keep the changes simple and worry about generalizing tsm MR once it gets more users.

mapping a not found error from `std::fs::read(Path::new(TDX_RTMR_PATH).join("mrconfigid"))` to `get_report() + td_report.tdinfo.mrconfigid` fallback seems easy to do in the first phase.

Xynnn007:
> > so sysfs for RTMRs is also a sub-capibility of TSM?
> 
> True but `tsm_reports` is for the the reports parts of it only. I'd prefer to keep the changes simple and worry about generalizing tsm MR once it gets more users.
> 
> mapping a not found error from `std::fs::read(Path::new(TDX_RTMR_PATH).join("mrconfigid"))` to `get_report() + td_report.tdinfo.mrconfigid` fallback seems easy to do in the first phase.

The changes are simple and straightforward, I believe. If upstream has already embraced the sysfs api for tsm, it's stable and why not we use it?

Also, we leave compability design in the PR also for those no `sysfs` is available. If `sysfs` is enabled, we could directly use it instead of ioctl-report0, while ioctl-report0 could serve in scenarios where no `sysfs` is available.

Btw, do you think we should rename the tsm-report to tsm?

mythi:
I did not say we should not use sysfs api, I said all of these changes should be done local to `tdx-attester` so that `tsm_reports` is not changed (and its scope remains as quote generation only). My suggestion is quite close to what `read_mrconfigid()` is with the difference that we read `std::fs::read(Path::new(TDX_RTMR_PATH).join("mrconfigid")` directly.

Xynnn007:
Sysfs is part of TSM ability, so bringing the abstraction to TSM would be better than implement in tdx locally. If we implement it in tdx locally, the whole TSM abstraction and ability will be splited into two different places which will cause the code structure not that elegant.

Once we get a way to read registers for  snp and cca, we can easily extend it in the TSM code also.

Xynnn007:
@mythi Updated the code in only tdx field and leave tsm refactoring work for future. Looks like the CCA rtmr work still needs time. We can do abstraction after it's upstreamed.

mythi:
> @mythi Updated the code in only tdx field and leave tsm refactoring work for future. Looks like the CCA rtmr work still needs time. We can do abstraction after it's upstreamed.

Thanks, I think it makes sense. I have not even seen CCA RTMR patches anywhere so it surely will take time.

> Lines changed: 117 additions & 101 deletions

This looks unnecessarily big to me since there's nothing we need to change for `extend_runtime_measurement()` path. Also, dropping get quote ioctl() fallback is better done in a commit of its own. Do you think the read/write wrappers are helpful to justify the changes?

Xynnn007:
> This looks unnecessarily big to me since there's nothing we need to change for `extend_runtime_measurement()` path. Also, dropping get quote ioctl() fallback is better done in a commit of its own. Do you think the read/write wrappers are helpful to justify the changes?

The refactoring helps to distinguish different RTMR calling routine and in future we can easily remove useless paths. Let me split the logic into pieces of commits.

Xynnn007:
Updated and removed function wrappers for rtmr r/w, etc.

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -138,40 +138,53 @@ impl Attester for TdxAttester {
...
+                }
             }
         }
 
         Ok(())
     }
 
     async fn bind_init_data(&self, init_data_digest: &[u8]) -> Result<InitDataResult> {
-        let report = self.ioctl_get_report()?;
+        let mr_configid = match self.supports_sysfs {
```
mythi: `if let` works better here I think.

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -148,72 +120,89 @@ impl Attester for TdxAttester {
...
 
     async fn get_runtime_measurement(&self, pcr_index: u64) -> Result<Vec<u8>> {
-        let td_report = Self::get_report()?;
-        let ccmr = self.pcr_to_ccmr(pcr_index) as usize;
+        let ccmr = self.pcr_to_ccmr(pcr_index);
+        let index = ccmr - 1;
+        if self.supports_sysfs {
+            let rtmr_path = Path::new(TDX_TSM_SYSFS_PATH).join(format!("rtmr{}:sha384", index));
+            let rtmr = std::fs::read(rtmr_path).context("Failed to read RTMR via sysfs")?;
+            return Ok(rtmr);
```
mythi, line 206:
```suggestion
            std::fs::read(rtmr_path).context("Failed to read RTMR via sysfs")
```

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -148,72 +120,89 @@ impl Attester for TdxAttester {
...
+        self.supports_sysfs || LIBTDX_ENABLED
     }
 
     async fn extend_runtime_measurement(
         &self,
         event_digest: Vec<u8>,
         register_index: u64,
     ) -> Result<()> {
-        if !runtime_measurement_extend_available() {
-            bail!("TDX Attester: runtime measurement extend is not available");
```
mythi, line 160: I think it's important to keep this message until RTMR support becomes more widely available (in known configurations). The check is also early in the function to avoid running unnecessary code.

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -77,36 +46,46 @@ struct TdxEvidence {
 }
 
 #[derive(Debug, Default)]
-pub struct TdxAttester {}
+pub struct TdxAttester {
+    /// Linux TSM sysfs is used to read and write runtime measurement registers,
+    /// and also read the registers like MRCONFIGID. This ability starts with kernel version 6.16.
+    /// If the sysfs is not supported, the attester will use the ioctl to get the report
+    /// and the registers.
+    supports_sysfs: bool,
```
mythi: `supports_tsm_measurements` (?)

mythi (commented):
Looks good, a couple of suggestions

attestation-agent/attester/src/tdx/mod.rs
Xynnn007: This is somehow a bug fix. 

See the commit message of https://github.com/confidential-containers/guest-components/pull/1489/changes/c45a665a69a54ad2607a59d877f18bf2dda1562b

attestation-agent/attester/src/tdx/mod.rs
Xynnn007: I believe it's just a personal preference. 

Let me share some personal experiences of this design:

When I used LLDB to debug with execution line-by-line, a explicit lvalue assignment could help debugger to stop at the time "when the lvalue is assiged, but not return".

If we use the suggested way, it's hard to check what the lvalue is assigned. So I kept the behavior to always assign a lvalue and then return with `Ok(())`.

attestation-agent/attester/src/tdx/mod.rs
Xynnn007: This is also a personal preference I believe. But with some rust reading I'd like to share my **personal** experience. Writing `if` on the right-hand side of the equals sign is no more intuitive to read than `match`. But sure, if there is no lvalue assignment, `match` will be a big burden and `if-else` is preferable.

attestation-agent/attester/src/tdx/mod.rs
mythi: makes sense

attestation-agent/attester/src/tdx/mod.rs
Xynnn007: Yes. The term is always hard to me. I found https://lwn.net/Articles/959887/ thus your suggestion can be a good name. Let me apply the changes.

mythi (approved):
Thanks for this work!

I updated the PR title a bit.

attestation-agent/attester/src/tdx/mod.rs
```diff
@@ -148,72 +122,89 @@ impl Attester for TdxAttester {
...
+            match tdx_attest_rs::tdx_att_extend(&event) {
+                tdx_attest_rs::tdx_attest_error_t::TDX_ATTEST_SUCCESS => {
+                    debug!("TDX extend runtime measurement succeeded.")
+                }
+                error_code => {
+                    bail!(
+                        "TDX Attester: Failed to extend RTMR. Error code: {:?}",
+                        error_code
+                    );
+                }
```
fitzthum, line 170: Previously this function would bail if runtime measurement isn't supported (although I guess we didn't get it quite right). Should we put an error or a warning at the bottom of this function in case we don't have TSM or the feature enabled?

fitzthum (approved):
LGTM

left one question

attestation-agent/attester/src/tdx/mod.rs
Xynnn007: If neither `tdx-attest-dcap-ioctls` nor `tsm` is enabled, this function shall not be called by higher layer. The higher layer will always call `supports_runtime_measurement` to check if runtime measurement is available. Thus we will not face the situation. But let me add a comment here to explicit show this.

PR #1490, opened 2026-05-25
asr: expose AA get_additional_evidence via REST API

Xynnn007:
Add GET /aa/additional-evidence so containers can obtain additional TEE evidence through api-server-rest. Reuse the same runtime_data query handling as /aa/evidence, including optional base64 encoding.

Assisted-by: Cursor


fitzthum (approved):
LGTM. Useful

we may want to add an endpoint for getting the attestation token itself someday

Xynnn007:
@fitzthum Good news is that we already supported this -

```
curl http://127.0.0.1:8006/aa/token\?token_type\=XXX
```

PR #1500, opened 2026-06-01
evidence_getter: TPM report data size to 32 bytes

kimullaa:
Failed to execute `evidence_getter` in Azure CVM + SEV-SNP.(main branch. `499658b307c39040efc009563737987c47a1dd63`)

```
$ printf 'z%.zs' {1..64} | sudo ../../target/release/evidence_getter stdio
2026-06-01T04:51:06.557666Z  INFO tss_esapi::context: Closing context.
2026-06-01T04:51:06.557745Z  INFO tss_esapi::context: Context closed.
2026-06-01T04:51:06.559679Z  INFO tss_esapi::context: Closing context.
2026-06-01T04:51:06.559716Z  INFO tss_esapi::context: Context closed.
WARNING:esys:src/tss2-esys/api/Esys_Quote.c:317:Esys_Quote_Finish() Received TPM Error
ERROR:esys:src/tss2-esys/api/Esys_Quote.c:105:Esys_Quote() Esys Finish ErrorCode (0x000001d5)
2026-06-01T04:51:06.560832Z ERROR tss_esapi::context::tpm_commands::attestation_commands: Error in quoting PCR: structure is the wrong size (associated with parameter number 1)
2026-06-01T04:51:06.560849Z  INFO tss_esapi::context: Closing context.
2026-06-01T04:51:06.560856Z  INFO tss_esapi::context: Closing handle 1075766378
2026-06-01T04:51:06.560862Z  INFO tss_esapi::context: Context closed.

thread 'main' panicked at attestation-agent/attester/src/bin/evidence_getter.rs:65:10:
get evidence failed: tpm error

Caused by:
    0: structure is the wrong size (associated with parameter number 1)
    1: structure is the wrong size (associated with parameter number 1)
    2: Response code value: 0x1d5
```

Truncate to 32 bytes because az-snp-vtpm v0.8.1 only supports SHA-256. <https://github.com/kinvolk/azure-cvm-tooling/blob/v0.8.1/az-cvm-vtpm/src/vtpm/mod.rs#L348>

This patch seems to fix the issue.

```
$ cargo build --no-default-features --features bin,az-snp-vtpm-attester --bin evidence_getter --release
   Compiling attester v0.1.0 (/home/azureuser/guest-components/attestation-agent/attester)
    Finished `release` profile [optimized] target(s) in 5.25s
[azureuser@coco-test attester]$ printf 'z%.zs' {1..64} | sudo ../../target/release/evidence_getter stdio
...
AzSnpVtpm:{"hcl_report":"SENM...
```

See also 79554cf88d1d2e85f480de5aae91d2a26990b91b.

### DIFF
--- a/attestation-agent/attester/src/bin/evidence_getter.rs
+++ b/attestation-agent/attester/src/bin/evidence_getter.rs
@@ -25,6 +25,14 @@ enum Cli {
     File { path: String },
 }
 
+/// Some CPU TEE platforms may impose platform-specific limits on report_data.
+fn adjust_report_data(tee: Tee, report_data: &[u8]) -> Vec<u8> {
+    match tee {
+        Tee::AzSnpVtpm | Tee::AzTdxVtpm => report_data[..report_data.len().min(32)].to_vec(),
+        _ => report_data.to_vec(),
+    }
+}
+
 #[tokio::main(flavor = "current_thread")]
 async fn main() {
     let env_filter = match std::env::var_os("RUST_LOG") {
@@ -60,14 +68,10 @@ async fn main() {
         }
     }
 
-    // Some platforms may impose platform-specific limits on report_data.
-    if let Tee::AzSnpVtpm | Tee:AzTdxVtpm = detect_tee_type() {
-       report_data.truncate(32);
-    }
-
-    let evidence = TryInto::<BoxedAttester>::try_into(detect_tee_type())
+    let tee = detect_tee_type();
+    let evidence = TryInto::<BoxedAttester>::try_into(tee)
         .expect("Failed to initialize attester.")
-        .get_evidence(report_data.clone())
+        .get_evidence(adjust_report_data(tee, &report_data))
         .await
         .expect("get evidence failed");
     println!("{:?}:{evidence}", detect_tee_type());
--- a/attestation-agent/attester/src/bin/evidence_getter.rs
+++ b/attestation-agent/attester/src/bin/evidence_getter.rs
@@ -61,12 +61,9 @@ async fn main() {
     }
 
     // Some platforms may impose platform-specific limits on report_data.
-    report_data = match detect_tee_type() {
-        Tee::AzSnpVtpm => report_data[..32].to_vec(),
-        Tee::AzTdxVtpm => report_data[..32].to_vec(),
-        _ => report_data.clone(),
-    };
-
+    if let Tee::AzSnpVtpm | Tee:AzTdxVtpm = detect_tee_type() {
+       report_data.truncate(32);
+    }
 
     let evidence = TryInto::<BoxedAttester>::try_into(detect_tee_type())
         .expect("Failed to initialize attester.")
--- a/attestation-agent/attester/src/bin/evidence_getter.rs
+++ b/attestation-agent/attester/src/bin/evidence_getter.rs
@@ -5,6 +5,7 @@
 
 use attester::{detect_attestable_devices, detect_tee_type, BoxedAttester};
 use clap::Parser;
+use kbs_types::Tee;
 use std::io::Read;
 use tokio::fs;
 use tracing_subscriber::{fmt, EnvFilter};
@@ -36,7 +37,6 @@ async fn main() {
         .with_writer(std::io::stderr)
         .init();
 
-    // report_data on all platforms is 64 bytes length.
     let mut report_data = vec![0u8; 64];
 
     let cli = Cli::parse();
@@ -57,6 +57,14 @@ async fn main() {
             report_data[..len].copy_from_slice(&content[..len]);
         }
     }
+
+    // Some platforms may impose platform-specific limits on report_data.
+    report_data = match detect_tee_type() {
+        Tee::AzSnpVtpm => report_data[..32].to_vec(),
+        Tee::AzTdxVtpm => report_data[..32].to_vec(),
+        _ => report_data.clone(),
+    };
+
 
     let evidence = TryInto::<BoxedAttester>::try_into(detect_tee_type())
         .expect("Failed to initialize attester.")
--- a/attestation-agent/attester/src/bin/evidence_getter.rs
+++ b/attestation-agent/attester/src/bin/evidence_getter.rs
@@ -42,9 +42,11 @@ async fn main() {
     let cli = Cli::parse();
 
     match cli {
-        Cli::Stdio => std::io::stdin()
-            .read_exact(&mut report_data)
-            .expect("read input failed"),
+        Cli::Stdio => {
+            std::io::stdin()
+                .read(&mut report_data)
+                .expect("read input failed");
+        }
         Cli::Commandline { data } => {
             let len = data.len().min(64);
             report_data[..len].copy_from_slice(&data.as_bytes()[..len]);
--- a/attestation-agent/attester/README.md
+++ b/attestation-agent/attester/README.md
@@ -26,6 +26,7 @@ echo $EVIDENCE_STRING | ../../target/release/evidence_getter stdio
 ```
 
 Here, `$EVIDENCE_STRING` is a string/bytes of up to 64 bytes.
+(Some platforms may impose platform-specific limits on evidence)
 
 ## Adding TPM quote to evidence
 


Xynnn007:
cc @mkulke  do we also need to consider azure-tdx-vtpm?

mkulke:
> cc @mkulke do we also need to consider azure-tdx-vtpm?

yes, it would be the same. I'm a bit puzzled why we need this now, though. What is the size of the nonce usually, have we changed something?

@kimullaa the algo you referred to is for the PCRs not the nonce.

```rust
    let hash_algo = HashingAlgorithm::Sha256;
    let selection_list = PcrSelectionListBuilder::new()
        .with_selection(hash_algo, &VTPM_QUOTE_PCR_SLOTS)
        .build()?;
```

kimullaa:
thank you for a feedback!

> yes, it would be the same. I'm a bit puzzled why we need this now, though. What is the size of the nonce usually, have we changed something?

@mkulke  I ran the `evidence_getter` command.(not `attestation-agent`).

The README.md states:
`Here, $EVIDENCE_STRING is a string/bytes of up to 64 bytes.`
https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/attester/README.md?plain=1#L24-L28

```
$ tr -dc 'A-Za-z0-9' < /dev/urandom | head -c 64 |  sudo ../../target/release/evidence_getter stdio
2026-06-02T00:18:31.903730Z  INFO tss_esapi::context: Closing context.
2026-06-02T00:18:31.903774Z  INFO tss_esapi::context: Context closed.
2026-06-02T00:18:31.905589Z  INFO tss_esapi::context: Closing context.
2026-06-02T00:18:31.905621Z  INFO tss_esapi::context: Context closed.
WARNING:esys:src/tss2-esys/api/Esys_Quote.c:317:Esys_Quote_Finish() Received TPM Error
ERROR:esys:src/tss2-esys/api/Esys_Quote.c:105:Esys_Quote() Esys Finish ErrorCode (0x000001d5)
2026-06-02T00:18:31.906550Z ERROR tss_esapi::context::tpm_commands::attestation_commands: Error in quoting PCR: structure is the wrong size (associated with parameter number 1)
2026-06-02T00:18:31.906620Z  INFO tss_esapi::context: Closing context.
2026-06-02T00:18:31.906655Z  INFO tss_esapi::context: Closing handle 1075766378
2026-06-02T00:18:31.906689Z  INFO tss_esapi::context: Context closed.

thread 'main' panicked at attestation-agent/attester/src/bin/evidence_getter.rs:65:10:
get evidence failed: tpm error

Caused by:
    0: structure is the wrong size (associated with parameter number 1)
    1: structure is the wrong size (associated with parameter number 1)
    2: Response code value: 0x1d5


$ tr -dc 'A-Za-z0-9' < /dev/urandom | head -c 32 |  sudo ../../target/release/evidence_getter stdio

thread 'main' panicked at attestation-agent/attester/src/bin/evidence_getter.rs:47:14:
read input failed: Error { kind: UnexpectedEof, message: "failed to fill whole buffer" }

$ tr -dc 'A-Za-z0-9' < /dev/urandom | head -c 1 |  sudo ../../target/release/evidence_getter stdio

thread 'main' panicked at attestation-agent/attester/src/bin/evidence_getter.rs:47:14:
read input failed: Error { kind: UnexpectedEof, message: "failed to fill whole buffer" }
[azureuser@coco-test attester]$
```

evidence_getter expects 64 bytes(not up to 64bytes)
https://github.com/confidential-containers/guest-components/blob/499658b307c39040efc009563737987c47a1dd63/attestation-agent/attester/src/bin/evidence_getter.rs#L39-L46

Therefore, it seems that `evidence_getter stdio ` is xpecting a 64-byte nonce.

I might be missing something or executing it incorrectly. Could you please let me know if there is a correct way to run this command for az-snp-vtpm?

mkulke:
> evidence_getter expects 64 bytes(not up to 64bytes)
> 
> https://github.com/confidential-containers/guest-components/blob/499658b307c39040efc009563737987c47a1dd63/attestation-agent/attester/src/bin/evidence_getter.rs#L39-L46
> 
> Therefore, it seems that `evidence_getter stdio ` is xpecting a 64-byte nonce.
> 
> I might be missing something or executing it incorrectly. Could you please let me know if there is a correct way to run this command for az-snp-vtpm?

I understand the `evidence_getter` tool has been written for the tpm-attester and hence it hasn't been tested on an az-snp/tdx-tpm TEE. maybe @bpradipt can confirm.

Guest-component produces CoCo evidence that is consumed by Trustee's verifier, so it's not raw evidence but a CoCo-specific data envelope. That also means that we cannot change the behaviour of the producer (attester) unless we have made sure the consumer (verifier) doesn't break. If you want to get raw evidence from those azure TEEs, please follow those instructions:

- https://learn.microsoft.com/en-us/azure/confidential-computing/guest-attestation-confidential-virtual-machines-design
- https://learn.microsoft.com/en-us/azure/confidential-computing/how-to-leverage-virtual-tpms-in-azure-confidential-vms

If you intend to make the tool work, I would see 2 options:
- change the call site, i.e. evidence_getter to work with the az-snp/tdx-vtpm attester
- assert/proof that the change in the attester's behaviour doesn't break the verifier's assumptions, ideally via the type system.

kimullaa:
@mkulke  thank you for a reply!

> I understand the evidence_getter tool has been written for the tpm-attester and hence it hasn't been tested on an az-snp/tdx-tpm TEE. maybe @bpradipt can confirm.

The README.md states: `This crate provides a simple tool to detect the current platform type and get related quote due to given report data.`
https://github.com/confidential-containers/guest-components/blob/1fcebcb66a3c21b62e852819d5b55212c90eba2a/attestation-agent/attester/README.md?plain=1#L11-L13

https://github.com/confidential-containers/guest-components/blob/1fcebcb66a3c21b62e852819d5b55212c90eba2a/attestation-agent/attester/src/bin/evidence_getter.rs#L61-L63

so I think it is better to support az_snp_vtpm attester.(or add a warning/note to README.md.)

To help me consider the way to fix this, please let me ask the following question.

> change the call site, i.e. evidence_getter to work with the az-snp/tdx-vtpm attester

Does this mean we should loosen the validation at the call site, rather than strictly matching exactly 64 bytes?

mkulke:
> Does this mean we should loosen the validation at the call site, rather than strictly matching exactly 64 bytes?

what I am suggesting is to fix the evidence-getter tool to make it work without modifications to the attester code. the reason being:

the guest-components/attester and trustee/verifier crates are inherently coupled, changes to the attester's behavior here will have to be matched on the trustee side.

If we truncate/resize the report-data that is sent by trustee (because it is not compatible with the evidence-getter tool) trustee's verifier will have to change its logic. otherwise, it will choke on a report-data mismatch. that means we have bump evidence versions, etc. we want to avoid that, hence I suggest modifying the evidence-getter to work with az-snp/tdx-vtpm, so it'll produce evidence that can be validated by trustee.

kimullaa:
@mkulke I've revised it—could you please take a look?

attestation-agent/attester/src/bin/evidence_getter.rs
```diff
@@ -58,6 +61,14 @@ async fn main() {
         }
     }
 
+    // Some platforms may impose platform-specific limits on report_data.
+    report_data = match detect_tee_type() {
+        Tee::AzSnpVtpm => report_data[..32].to_vec(),
+        Tee::AzTdxVtpm => report_data[..32].to_vec(),
+        _ => report_data.clone(),
+    };
+
```
mkulke:
```suggestion
    if let Tee::AzSnpVtpm | Tee:AzTdxVtpm = detect_tee_type() {
       report_data.truncate(32);
    }
```

attestation-agent/attester/src/bin/evidence_getter.rs
mkulke, line 79: what about attestable devices, e.g. a GPU, does it need to be 64 byte, so you want to truncate only CPU evidence?

kimullaa: I modified the code because I wanted to truncate only the CPU evidence.

mkulke (approved):

Xynnn007 (approved):

PR #1518, opened 2026-06-17
attester: nvidia: serialize Architecture UPPERCASE

mythi:
WDYT? I had a hiccup with Trustee/ITA since I only used test data that had UPPERCASE Arch. NRAS needs it so maybe use uppercase consistently (the same change could be made on nvidia-verifier side)? 

--- 
Trustee/ITA fails with nvidia-attester originated evidence because device architecture is not UPPERCASE.

While this could be fixed in Trustee/ITA by making everything case insensitive, the observation is that nvidia handles this UPPERCASE end to end. Therefore, keep Archicture UPPERCASE between nvidia-attester and Trustee as well.

### DIFF
--- a/attestation-agent/attester/src/nvidia/mod.rs
+++ b/attestation-agent/attester/src/nvidia/mod.rs
@@ -35,6 +35,7 @@ fn ensure_sdk_init() -> Result<()> {
 }
 
 #[derive(Debug, Deserialize, Display, EnumString, PartialEq, Serialize)]
+#[serde(rename_all(serialize = "UPPERCASE"))]
 #[strum(ascii_case_insensitive)]
 enum Architecture {
     #[serde(alias = "BLACKWELL")]


Xynnn007:
There might be two ways to fix the original issue
1. using the way like this PR. This would make breaking change upon the nvidia evidence, thus versioning is needed for nvidia lazily https://github.com/confidential-containers/trustee/issues/1412
2. Just add a logic around [here](https://github.com/confidential-containers/trustee/blob/main/kbs/src/attestation/intel_trust_authority/mod.rs#L246) to UPPER-lize all `d.arch`. This way no change to verifier/attester pair and just ITA adjustment.

Personally prefer way 2, as it's simpler.

mythi:
> 1. This would make breaking change upon the nvidia evidence

Is it? I checked that uppercase is handled https://github.com/confidential-containers/trustee/blob/03d215a667581b849d0d4899b66565564254bd69/deps/verifier/src/nvidia/mod.rs#L80-L90 and mentioned it in the commit message. 

I'm totally fine with 2. as well since I implemented it first but his just needed only one line change and makes things consistent.

Xynnn007 (approved):

fitzthum (approved):
Should be fine but please keep an eye on the Kata tests when we pick up this version.

mythi:
ok to merge?

PR #1536, opened 2026-07-06
asr: bump hyper minimum version 0.14.27 → 0.14.32

ajaypvictor:
CVE-2021-32714, CVE-2021-32715 and RUSTSEC-2022-0022 were fixed in hyper 0.14.10/0.14.12. Raise the lower bound to reflect the currently resolved version.

Assisted-by: IBM Bob <noreply@ibm.com>

### DIFF
--- a/api-server-rest/Cargo.toml
+++ b/api-server-rest/Cargo.toml
@@ -11,7 +11,7 @@ async-trait.workspace = true
 base64.workspace = true
 clap = { workspace = true, features = ["derive"] }
 form_urlencoded = "1.2.2"
-hyper = { version = "0.14.27", features = ["server", "http1", "runtime"] }
+hyper = { version = "0.14.32", features = ["server", "http1", "runtime"] }
 serde.workspace = true
 serde_json.workspace = true
 shadow-rs.workspace = true


stevenhorsman:
Do you need to run cargo update to update the `Cargo.lock` as well?

ajaypvictor:
> Do you need to run cargo update to update the `Cargo.lock` as well?

Thanks Steve for the suggestion..
Ran `cargo update hyper@0.14.32`, output was \"Locking 0 packages\", which I guess is confirming that the `hyper` was already at `0.14.32` before this change. The `Cargo.lock` entry for `api-server-rest` already references `hyper 0.14.32` so looks like no change to the lock file is needed.

stevenhorsman:
So this is a no-op and doesn't actual resolve https://github.com/advisories/GHSA-5h46-h7hh-c6x9, https://github.com/advisories/GHSA-f3pg-qwvg-p99c and RUSTSEC-2022-0022?

fitzthum:
Btw @ajaypvictor is there somethign that we don't get from dependabot here? That's typically how we handle these bumps.

Xynnn007:
Some of my previous workflow:

Usually for a dependency update, we should search in the `Cargo.lock` and get all the places that use `hyper` old version. Then, check if its parent dependency has new version with newer `hyper` dependency. Update the parent dependency layer by layer, until we could find no old `hyper` versions in `Cargo.lock`.

ajaypvictor:
Thanks for all the inputs, looks like the scanner was flagging the package post scanning `Cargo.toml` under api-server-rest directly, I guess I will close this PR.. Thanks Steve, Tobin and Ding..

@fitzthum I think another downstream tool similar to dependabot reported this, we have seen a pattern that some CVEs are not identified by dependabot on time..

PR #1538, opened 2026-07-10
attester: Add DPU evidence collection

haixinJY:
## Summary

Add DPU (Data Processing Unit) attester implementation for collecting
hardware-rooted attestation evidence from DPU devices (e.g., NVIDIA BlueField-3).

The DPU attester uses the DICE (Device Identifier Composition Engine) certificate
chain to provide hardware-rooted identity and firmware integrity measurements.

## Related PRs

- kbs-types: Add Dpu TEE variant (PR pending)
- trustee: DPU verifier implementation (PR pending)

## Changes

- `attestation-agent/attester/src/dpu/mod.rs`: New module
  - `DpuAttester` implementing the `Attester` trait
  - Platform detection via `/dev/bluefield_attest` and InfiniBand sysfs
  - DICE certificate chain collection (Alias + DeviceID certs)
  - Firmware measurement extraction from DICE layers
  - Unit tests for evidence serialization and platform detection
- `attestation-agent/attester/src/lib.rs`: Register DPU module with feature gate
- `attestation-agent/attester/Cargo.toml`: Add `dpu-attester` feature flag

## Scope

This PR covers evidence **collection** only. Verification logic is in the
corresponding trustee verifier PR. Full integration into the attestation
flow requires the `Tee::Dpu` variant to be added to kbs-types first.

## Testing

```
cargo check -p attester --no-default-features --features dpu-attester
cargo test -p attester --no-default-features --features dpu-attester
```

- All 4 unit tests pass
- No hardware required for basic tests (graceful fallback when device absent)

Signed-off-by: Ying Jiang <jiangying8582@163.com>


attestation-agent/attester/src/dpu/mod.rs
```diff
@@ -0,0 +1,254 @@
+// Copyright (c) 2024 Alibaba Cloud
```
Xynnn007: 2024 -> 2026

attestation-agent/attester/src/dpu/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+//! DPU (Data Processing Unit) Attester
```
Xynnn007: Looks like the heuristic logic here is for `bluefield` only. I'd suggest to use `NvidiaDpu` than the general `Dpu`

attestation-agent/attester/src/dpu/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+#[async_trait::async_trait]
+impl Attester for DpuAttester {
+    /// Collect attestation evidence from the DPU device.
+    ///
+    /// The `report_data` parameter is used as a nonce to bind the evidence
+    /// to a specific attestation session, preventing replay attacks.
+    /// It is truncated or zero-padded to [`DPU_NONCE_SIZE`] bytes.
+    async fn get_evidence(&self, mut report_data: Vec<u8>) -> Result<TeeEvidence> {
+        // Normalize nonce to expected size
+        report_data.resize(DPU_NONCE_SIZE, 0);
```
Xynnn007: Let's add a judgement of the report_data length before this line. like https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/attester/src/tdx/mod.rs#L104-L106

If the data is longer than 64 bytes, there will be information loss

attestation-agent/attester/src/dpu/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+        serde_json::to_value(&evidence).context("Failed to serialize DPU evidence")
+    }
+}
+
+impl DpuAttester {
+    /// Collect evidence by reading from the attestation device node.
+    fn collect_evidence_from_device(&self, report_data: &[u8]) -> Result<DpuEvidence> {
+        // Write nonce to device, then read back evidence
+        // In production: ioctl or write/read on /dev/bluefield_attest
+        let alias_cert = self.read_cert_from_device("alias")?;
```
Xynnn007: Could we make `"alias"` and `"device_id"` as consts?

attestation-agent/attester/src/nvidia_dpu/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+            device_id_cert: vec![0xBE, 0xEF],
+            measurements: vec![],
+            report_data: vec![0u8; 64],
+        };
+
+        let json_str = serde_json::to_string(&evidence).unwrap();
+        let parsed: DpuEvidence = serde_json::from_str(&json_str).unwrap();
+        assert_eq!(parsed.alias_cert, vec![0xDE, 0xAD]);
+        assert_eq!(parsed.report_data.len(), 64);
+    }
```
Xynnn007, line 260: The two unit tests can be combined with string ser/deser. Overall they are somehow overlapped.

attestation-agent/attester/src/nvidia_dpu/mod.rs
```diff
@@ -3,62 +3,62 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-//! DPU (Data Processing Unit) Attester
+//! NVIDIA DPU (Data Processing Unit) Attester
```
Xynnn007, line 6: Ok. I see the changes in the second commit here. Looks like a rebase is needed.

attestation-agent/attester/src/nvidia_dpu/mod.rs
```diff
@@ -77,40 +77,40 @@ pub fn detect_platform() -> bool {
...
     /// to a specific attestation session, preventing replay attacks.
-    /// It is truncated or zero-padded to [`DPU_NONCE_SIZE`] bytes.
+    /// It is truncated or zero-padded to [`NVIDIA_DPU_NONCE_SIZE`] bytes.
     async fn get_evidence(&self, mut report_data: Vec<u8>) -> Result<TeeEvidence> {
         // Normalize nonce to expected size
-        report_data.resize(DPU_NONCE_SIZE, 0);
+        report_data.resize(NVIDIA_DPU_NONCE_SIZE, 0);
 
-        let evidence = if std::path::Path::new(DPU_DEVICE_PATH).exists() {
+        let evidence = if std::path::Path::new(NVIDIA_DPU_DEVICE_PATH).exists() {
```
Xynnn007: It's weird that both `collect_evidence_from_device` and `collect_evidence_from_sysfs` do not have a machanism to bind the `report_data` to the hardware generated evidence. In stead, it's just put as plaintest in the evidence body.

This can face the MITM attach - an attacker can tamper the evidence body with a changed report_data, thus the freshness does not work.

Could you help to check the mechanism/principle behind this?

Xynnn007 (commented):
Hi @56ht86hrd5-ai the code looks much clear! btw, `kbs-protocol`, `attestation-agent` and other crates would also need new feature `nvidia-dpu-attester`.

attestation-agent/attester/src/nvidia_dpu/mod.rs
Xynnn007: btw. two possible ways in my mind
1. Check if the nvidiadpu driver has native support to get signed measurements with nonces. I searched some context with `SPDMGetSignedMeasurements` API, but it's in SPDM context. I am not sure if it's practical.
2. A different way is to use some api to use alias pri key to sign the report_data, and get the signed blob into the evidence

attestation-agent/attester/src/dpu/mod.rs
haixinJY: Good catch, will fix to 2026.

attestation-agent/attester/src/dpu/mod.rs
haixinJY: Agreed — this is BlueField-specific (rshim detection, DICE via sysfs paths). Will rename directory from `dpu/` to `nvidia_dpu/` and update the feature flag to `nvidia-dpu-attester` to be consistent with kbs-types naming.

attestation-agent/attester/src/dpu/mod.rs
haixinJY: Good point, will add the same validation as TDX attester:

```rust
if report_data.len() > 64 {
    bail!("report_data must be no longer than 64 bytes");
}
```

Thanks for the reference.

attestation-agent/attester/src/dpu/mod.rs
haixinJY: Sure, will extract as consts:

```rust
const ALIAS_CERT_KEY: &str = "alias";
const DEVICE_ID_CERT_KEY: &str = "device_id";
```

attestation-agent/attester/src/nvidia_dpu/mod.rs
haixinJY: Agreed, will merge into a single serialize→deserialize round-trip test.

attestation-agent/attester/src/nvidia_dpu/mod.rs
haixinJY: Will rebase and squash into a clean single commit.

attestation-agent/attester/src/nvidia_dpu/mod.rs
haixinJY: Great catch — you're right that the current implementation lacks cryptographic binding between `report_data` and the DICE evidence.

**Current state**: The DICE chain (DeviceID → Alias) proves device identity and firmware integrity, but `report_data` is plaintext metadata alongside the certs. This relies on transport-level protection only.

**Proposed fix (your option 2)**: Use the DICE alias private key to sign `report_data`:

```rust
evidence.report_data_signature = alias_private_key.sign(report_data);
```

The verifier already has the alias public key from the DICE chain, so it can verify this signature — providing equivalent freshness guarantees to TDX/SNP's hardware-signed report_data. The alias private key is accessible at the same sysfs path.

I'll implement this in the next push. This also resolves the corresponding concern on the verifier side (trustee #1479) where `expected_report_data` comparison needs the same cryptographic binding.

(In our full multi-chip protocol design, this binding is achieved by embedding the nonce in the Alias cert's TBS which is DeviceID-signed. For the CoCo single-chain verifier, explicit alias-key signing of report_data is the equivalent mechanism.)

haixinJY: Thanks for the two suggestions!

Re SPDM: BlueField-3 supports SPDM via BMC/rshim management plane, but `SPDMGetSignedMeasurements` isn't accessible from the DPU OS side without a full SPDM requester daemon. Not practical for the in-DPU attester.

Re alias key signing: This is the right fit — alias private key is already at hand, zero extra dependencies. Will implement: sign report_data with alias key, include signature in evidence JSON, and add verification on the trustee verifier side.

Xynnn007: Sounds very good to me!

attestation-agent/attester/src/nvidia_dpu/mod.rs
```diff
@@ -0,0 +1,281 @@
...
+const NVIDIA_DPU_NONCE_SIZE: usize = 64;
+const ALIAS_PRIVATE_KEY_PATH: &str = "/sys/kernel/security/tee/dice/alias_private_key";
+
+const ALIAS_CERT_KEY: &str = "alias";
+const DEVICE_ID_CERT_KEY: &str = "device_id";
+
+/// NVIDIA DPU attestation evidence containing DICE certificate chain
+/// and firmware measurements.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct NvidiaDpuEvidence {
```
fitzthum, line 50: Let's try to make this more similar to how the existing nvidia verifier works. Let's have a top-level struct that containers a vector of DPU evidence where each entry is tagged with an identifier (i..e the architecture). I want to make sure that this is flexible for when we may have other devices or other layouts in the future. 

Also, the top-level evidence struct must have a version field.

attestation-agent/attester/src/nvidia_dpu/mod.rs
```diff
@@ -0,0 +1,281 @@
...
+        }
+    }
+
+    false
+}
+
+/// Sign report_data with DICE alias private key for freshness binding.
+///
+/// Returns a base64-encoded ECDSA P-384 signature, or an empty string
+/// if the alias private key is not available on this device.
```
fitzthum: Can you explain why we are signing the report data in software?

deps/kbs-types/src/error.rs
fitzthum, line 1: No need to put the whole KBS types repo here. Just patch via the cargo file like you did with the other PR

fitzthum (commented):
First round of comments. Seems like this is going in the right direction.

attestation-agent/attester/src/nvidia_dpu/mod.rs
```diff
@@ -0,0 +1,281 @@
...
+const NVIDIA_MLX_VENDOR_ID: &str = "15b3";
+/// PCI device IDs that identify BlueField-3 (as opposed to ConnectX-7 or other NICs).
+/// - a2dc: BF3 integrated NIC function
+/// - a2da: BF3 SoC PCI bridge
+const BF3_DEVICE_IDS: &[&str] = &["a2dc", "a2da"];
+
+const PCI_DEVICES_PATH: &str = "/sys/bus/pci/devices";
+const NVIDIA_DPU_SYSFS_PATH: &str = "/sys/class/infiniband";
+const NVIDIA_DPU_NONCE_SIZE: usize = 64;
+const ALIAS_PRIVATE_KEY_PATH: &str = "/sys/kernel/security/tee/dice/alias_private_key";
```
Xynnn007, line 40: btw, where does this path comes? I mean, if there is any official or public documents/links to show the origin of the key path. Maybe it's good to add a comment here, for better maintaince.

attestation-agent/attester/src/nvidia_dpu/mod.rs
```diff
@@ -0,0 +1,281 @@
...
+#[derive(Debug, Serialize, Deserialize)]
+pub struct NvidiaDpuEvidence {
+    /// DER-encoded Alias certificate (leaf) containing device measurements
+    pub alias_cert: Vec<u8>,
+    /// DER-encoded DeviceID certificate signed by manufacturer root
+    pub device_id_cert: Vec<u8>,
+    /// Firmware measurements from DICE layers
+    pub measurements: Vec<NvidiaDpuMeasurement>,
+    /// Report data (nonce binding for freshness)
+    pub report_data: Vec<u8>,
```
Xynnn007: this field can be removed. As the verifier side already knows what's the expected report data and it only needs to verify the signature. So information entropy is already included in `report_data_signature`. A report data in transition with no protection would meet MITM, and also a useless double check against with expected report data on the verifier side.

attestation-agent/attester/src/nvidia_dpu/mod.rs
```diff
@@ -0,0 +1,281 @@
...
+impl Attester for NvidiaDpuAttester {
+    /// Collect attestation evidence from the DPU device.
+    ///
+    /// The `report_data` parameter is used as a nonce to bind the evidence
+    /// to a specific attestation session, preventing replay attacks.
+    /// It must be no longer than [`NVIDIA_DPU_NONCE_SIZE`] bytes.
+    async fn get_evidence(&self, mut report_data: Vec<u8>) -> Result<TeeEvidence> {
+        if report_data.len() > NVIDIA_DPU_NONCE_SIZE {
+            bail!("report_data must be no longer than {NVIDIA_DPU_NONCE_SIZE} bytes");
+        }
```
Xynnn007, line 154: Now the report data is included in a signature. Thus we might not need the length condition here. That's, any data length is ok

attestation-agent/attester/src/nvidia_dpu/mod.rs
```diff
@@ -0,0 +1,281 @@
...
+        .unwrap_or_default();
+
+        let device_id_cert = std::fs::read(
+            format!("{}/mlx5_0/device/attestation/{}_cert", NVIDIA_DPU_SYSFS_PATH, DEVICE_ID_CERT_KEY),
+        )
+        .unwrap_or_default();
+
+        // Read measurements if available
+        let measurements_path = format!(
+            "{}/mlx5_0/device/attestation/measurements", NVIDIA_DPU_SYSFS_PATH
```
Xynnn007: Can this also be a const?

attestation-agent/attester/src/nvidia_dpu/mod.rs
```diff
@@ -0,0 +1,281 @@
...
+    /// InfiniBand device sysfs path under `device/attestation/`.
+    fn collect_evidence_from_sysfs(&self, report_data: &[u8]) -> Result<NvidiaDpuEvidence> {
+        // Read DICE certificates from sysfs
+        let alias_cert = std::fs::read(
+            format!("{}/mlx5_0/device/attestation/{}_cert", NVIDIA_DPU_SYSFS_PATH, ALIAS_CERT_KEY),
+        )
+        .unwrap_or_default();
+
+        let device_id_cert = std::fs::read(
+            format!("{}/mlx5_0/device/attestation/{}_cert", NVIDIA_DPU_SYSFS_PATH, DEVICE_ID_CERT_KEY),
```
Xynnn007: Can this be a const?

attestation-agent/attester/src/nvidia_dpu/mod.rs
```diff
@@ -0,0 +1,281 @@
...
+
+impl NvidiaDpuAttester {
+    /// Collect evidence from sysfs-exposed DICE attestation attributes.
+    ///
+    /// BlueField-3 exposes DICE certificates and measurements via the
+    /// InfiniBand device sysfs path under `device/attestation/`.
+    fn collect_evidence_from_sysfs(&self, report_data: &[u8]) -> Result<NvidiaDpuEvidence> {
+        // Read DICE certificates from sysfs
+        let alias_cert = std::fs::read(
+            format!("{}/mlx5_0/device/attestation/{}_cert", NVIDIA_DPU_SYSFS_PATH, ALIAS_CERT_KEY),
```
Xynnn007: Can this be a const?

attestation-agent/attester/src/nvidia_dpu/mod.rs
```diff
@@ -0,0 +1,281 @@
...
+                Err(e) => {
+                    warn!("Failed to parse DICE alias private key: {e}");
+                    String::new()
+                }
+            }
+        }
+        Err(_) => {
+            // Device doesn't expose alias private key via sysfs, skip signing
+            String::new()
+        }
```
Xynnn007: Will this happen? Note that if the private key is not used to sign the report data, there will be significant replay attack by simply changing the report field in the report.

Xynnn007 (commented):
Some more comments

attestation-agent/attester/src/nvidia_dpu/mod.rs
haixinJY: Done — restructured to a versioned multi-device format:

```rust
pub struct NvidiaDpuEvidence {
    pub version: String,
    pub devices: Vec<DpuDeviceEvidence>,
}

pub struct DpuDeviceEvidence {
    pub architecture: String,  // e.g. "bluefield3"
    pub alias_cert: Vec<u8>,
    pub device_id_cert: Vec<u8>,
    pub measurements: Vec<NvidiaDpuMeasurement>,
    pub report_data_signature: String,
}
```

This allows future devices (different BlueField generations or other NVIDIA DPU types) to be added without breaking the format.

attestation-agent/attester/src/nvidia_dpu/mod.rs
haixinJY: The DICE certificate chain proves device identity and boot measurements, but does not inherently bind to a specific attestation session. The software signature over `report_data` using the DICE Alias private key creates this session binding — it proves the evidence was freshly generated for *this* challenge nonce, preventing replay of previously-captured valid evidence.

In other words:
- Cert chain = "I am a genuine DPU with these measurements"
- Signature = "and I generated this evidence right now for your nonce"

Without the signature, an attacker who intercepts valid evidence once could replay it indefinitely for any future challenge.

deps/kbs-types/src/error.rs
haixinJY: Done — removed `deps/kbs-types/` entirely and switched to git dependency in workspace Cargo.toml:

```toml
kbs-types = { git = "https://github.com/56ht86hrd5-ai/kbs-types.git", branch = "feat/add-dpu-tee" }
```

Will switch to crates.io version once kbs-types PR is merged.

attestation-agent/attester/src/nvidia_dpu/mod.rs
haixinJY: Added a documentation comment referencing the NVIDIA DOCA DICE Programming Guide ("Key Exposure via sysfs" section). The path comes from the DOCA attestation service deployment layout on BlueField DPUs.

attestation-agent/attester/src/nvidia_dpu/mod.rs
haixinJY: Agreed — removed. The verifier already knows the expected report_data (it generated the nonce), and the signature binds it cryptographically. Transmitting the plaintext separately is redundant.

attestation-agent/attester/src/nvidia_dpu/mod.rs
haixinJY: Agreed — removed the length check. With mandatory signature, the binding constraint is enforced cryptographically.

attestation-agent/attester/src/nvidia_dpu/mod.rs
haixinJY: Done — extracted all sysfs paths to module-level constants (`ATTESTATION_BASE_PATH`, etc.)

attestation-agent/attester/src/nvidia_dpu/mod.rs
haixinJY: Done.

attestation-agent/attester/src/nvidia_dpu/mod.rs
haixinJY: Done.

attestation-agent/attester/src/nvidia_dpu/mod.rs
haixinJY: Good point — changed to a hard requirement. On BlueField-3, the DICE Alias private key is always available when the DOCA attestation service is running (provisioned during secure boot). If the key is unavailable, we now return an error rather than producing unsigned evidence. This closes the replay attack vector.

attestation-agent/attester/src/nvidia_dpu/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+    pub report_data_signature: String,
+}
+
+/// A single DICE layer measurement
+#[derive(Debug, Serialize, Deserialize)]
+pub struct NvidiaDpuMeasurement {
+    /// DICE layer index (0 = ROM, 1 = firmware, 2 = OS, ...)
+    pub layer: u8,
+    /// SHA-384 digest of the layer's code/config
+    pub digest: Vec<u8>,
```
mkulke: is this known to be always sha384? if so, we can use a `&[u8; 48]` to avoid mishandling.

attestation-agent/attester/src/nvidia_dpu/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+    pub measurements: Vec<NvidiaDpuMeasurement>,
+    /// Base64-encoded ECDSA P-384 signature of report_data using DICE alias private key
+    pub report_data_signature: String,
+}
+
+/// A single DICE layer measurement
+#[derive(Debug, Serialize, Deserialize)]
+pub struct NvidiaDpuMeasurement {
+    /// DICE layer index (0 = ROM, 1 = firmware, 2 = OS, ...)
+    pub layer: u8,
```
mkulke, line 84: I guess this is open to extension or are the indices fixes? if it's the latter we could use an enum to represent the layer.

attestation-agent/attester/src/nvidia_dpu/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+}
+
+/// Detect if NVIDIA BlueField-3 DPU hardware is available on the platform.
+///
+/// Scans PCI sysfs for devices matching the BF3 vendor:device IDs.
+/// This distinguishes BlueField-3 from regular ConnectX NICs which have
+/// different PCI device IDs.
+pub fn detect_platform() -> bool {
+    // Scan PCI devices for BF3-specific device IDs
+    if let Ok(entries) = std::fs::read_dir(PCI_DEVICES_PATH) {
```
mkulke: we can reverse this to save an indent:

```
let Ok(entries) = std::fs::read_dir(...) else {
   return false;
}
```

attestation-agent/attester/src/nvidia_dpu/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+    /// Device architecture identifier (e.g. "bluefield3")
+    pub architecture: String,
+    /// DER-encoded DICE Alias certificate (leaf)
+    pub alias_cert: Vec<u8>,
+    /// DER-encoded DICE DeviceID certificate
+    pub device_id_cert: Vec<u8>,
+    /// Firmware measurements from DICE layers
+    pub measurements: Vec<NvidiaDpuMeasurement>,
+    /// Base64-encoded ECDSA P-384 signature of report_data using DICE alias private key
+    pub report_data_signature: String,
```
mkulke, line 72: nit: if you want to keep the raw bytes in the struct you can use something like this

```
    #[serde_as(as = "Base64<UrlSafe>")]
    pub report_data_signature: Vec<u8>,
```

haixinJY: Done — adopted `serde_as` with `Base64` encoding for all binary fields.

attestation-agent/attester/src/nvidia_dpu/mod.rs
haixinJY: Good call — switched to let-else.

haixinJY:
Addressed the automated review feedback in the latest push:

- Binary fields now use base64 encoding via `serde_with`
- Error propagation replaces `unwrap_or_default()` for required cert reads
- Oversize nonce rejected instead of silently truncated
- SigningKey length validated upfront (48 bytes for P-384)
- Measurement buffer length validated before chunking

For the hardcoded sysfs path (`mlx5_0`): acknowledged as an initial limitation — added a TODO for dynamic device discovery in a follow-up PR.

The `[patch.crates-io]` fork reference is temporary pending kbs-types#89 merge.

attestation-agent/attester/src/nvidia_dpu/mod.rs
haixinJY: Yes, always SHA-384. Changed to `[u8; 48]` with base64 serialization — the type system now enforces the invariant.

attestation-agent/attester/src/nvidia_dpu/mod.rs
haixinJY: Open to extension — future firmware versions may define additional DICE layers beyond the current three. Keeping `u8` for forward compatibility, but enhanced the doc comment listing known values.

attestation-agent/attester/src/nvidia_dpu/mod.rs
fitzthum: Nice. Version can be an int btw

haixinJY:
@fitzthum Done — changed `version` to `u32`. Cleaner for what is really just a format revision counter.

fitzthum:
Did you mean to close this? It's coming along well I think.

haixinJY:
@fitzthum No, that was accidental — caused by a force-push that inadvertently created an orphan commit (no common history with main). I have rebased properly and re-submitted as #1575. Sorry for the noise!

haixinJY:
@fitzthum Re: your suggestion on version type -- adopted, version is now u32 in the successor PR #1575. Thanks!

Issue #1537, opened 2026-07-06
Build Extension Images for Kata


fitzthum:
Now that Kata supports [composable VM images](https://github.com/kata-containers/kata-containers/blob/main/docs/design/proposals/composable-vm-images.md), we should think about building these images directly in guest-components.

I think this will lead to improved modularity. Today, we build and setup the guest-components in the Kata repo. Significant changes to this setup require coordination between two repos. Since the e2e tests are in the Kata repo, you can't fully vet your changes until the second step. Ofc the Kata tests can take a while to run so getting through the whole process can be a bit painful.

If we have stable enough APIs (which the composable VM approach in theory provides), we can keep the builder logic alongside the code in this repo. I think this will make our life easier. Crucially, it allows us to do e2e testing earlier and in a simple environment, which can help us to start creating first-class releases of the guest-components. For the time being, I see this as purely additive to the existing tests that we have, but maybe in the longterm we could optimize.

There are a few different ways to think about the testing, but the simplest approach would be to do something similar to the VM-based Trustee tests. Essentially we just use one long-running CVM for all the tests. We unpack the image artifacts inside the CVM and use the binaries directly. There is a question here about handling Trustee versioning. How can we track the versions of both components but avoid circular dependencies?

We could also think about starting a VM for each test, but the composable images aren't full VM images, so we'd need to think about this a bit. It might be cool to also provide generic CVM images. We might also think about how peer pods is connected, but let's start simple.

Of course this has some impact on reference values, but I think it generally makes things more clear. Also, interestingly this might allow us to work on optimizations like reproducible builds without having to worry about the whole Kata rootfs. I'm not sure exactly how Kata will consume the artifact, but we can figure something out.

Xynnn007:
here is a basic question:

Do we have any mechaism to measure extension images for kata when using coco? If not, the practical way is to leverage RTMR to measure the loaded extension images. This needs kata-agent to call ASR or AA's extend runtime measurement API.

Based on this idea, guest-components should somehow be embedded inside the rootfs as "kata-infra" like kata-agent. Or we will fall into the checken-egg issue

fitzthum:
Extension images are measured with dm-verity as part of the kernel command line. We shouldn't need any attestation stuff in the base image. This does put more pressure on calculating the kernel command line as part of a reference value, which is already something we struggle with a bit, but generally I think it is sound.

Xynnn007:
[...]

PR #1539, opened 2026-07-10
coco-extension: build and ship the kata coco-extension image

fidencio:
Kata Containers' composable VM image work splits the CoCo guest components out of the monolithic confidential rootfs into a separate guest extension image (kata's rootfs-image-coco-extension asset). Today that image is assembled in the kata-containers repo from this repo's binaries, which means any change to the extension layout or manifest requires coordinating across two repos and cannot be vetted until the kata side runs.

Build the extension image here instead. From a single assembled rootfs we emit two flavours: a scratch-based OCI container image and an EROFS + dm-verity disk image (with its dm-verity root hash), matching the layout kata expects.

The rootfs mirrors kata's install_image_coco_extension(): the guest component binaries, ocicrypt config, pause bundle, cryptsetup, and the data-driven components.toml manifest (kata-agent contract). On x86_64 it also builds the NVIDIA attester variant (attestation-agent-nv) plus its libnvat closure. The EROFS image is built locally (mkfs.erofs + parted + veritysetup) rather than pulling in kata's osbuilder.

Builds cover x86_64, aarch64, and s390x. Pull requests only build and upload workflow artifacts; pushes to main publish to ghcr.io (tagged with the commit sha and "latest") with build attestations; and published releases publish the images tagged with the release version and attach the per-arch disk images and dm-verity root hashes to the GitHub release so consumers can pin to a stable version.

Refs: https://github.com/confidential-containers/guest-components/issues/1537

### DIFF
--- a/.github/workflows/coco-extension-image.yml
+++ b/.github/workflows/coco-extension-image.yml
@@ -1,0 +1,289 @@
+name: Build and publish the CoCo guest extension image
+
+# Builds the "coco-extension" image (kata-containers-coco-extension.img) that
+# packages the guest components as a kata composable guest extension.
+#
+# Two artifacts are produced per architecture from the same assembled rootfs:
+#   * an OCI container image (scratch-based, rootfs + components.toml manifest)
+#   * an EROFS + dm-verity disk image (+ its dm-verity root hash), using the
+#     same layout as kata's rootfs-image-coco-extension asset.
+#
+# On pull requests everything is built and uploaded as workflow artifacts (no
+# registry push). On pushes to main the artifacts are published to ghcr.io
+# (tagged with the commit sha and "latest") and attested. On a published
+# release the same artifacts are published tagged with the release version, the
+# multi-arch manifests get the version tag, and the per-arch disk images and
+# their dm-verity root hashes are attached to the GitHub release so consumers
+# (e.g. kata-containers) can pin to a stable version.
+on:
+  pull_request:
+  push:
+    branches:
+    - main
+  release:
+    types: [published]
+
+permissions:
+  contents: read
+
+env:
+  REGISTRY: ghcr.io
+  # OCI container image flavour.
+  CONTAINER_IMAGE: ghcr.io/${{ github.repository }}/coco-extension
+  # EROFS + dm-verity disk image flavour.
+  DISK_IMAGE: ghcr.io/${{ github.repository }}/coco-extension-disk
+  # Pause image bundled into the extension (kept in sync with kata versions.yaml).
+  # This is transitional: we still unpack registry.k8s.io/pause into pause_bundle/
+  # because kata's rootfs-image-coco-extension asset and components.toml contract
+  # expect the extension to ship the pause bundle (kata-agent resolves
+  # "pause-bundle" from /run/kata-extensions/coco). It will be dropped once kata
+  # stops baking pause into the coco extension.
+  PAUSE_IMAGE_REPO: docker://registry.k8s.io/pause
+  PAUSE_IMAGE_VERSION: "3.10"
+  NVAT_VERSION: "2026.03.02"
+  UMOCI_VERSION: v0.6.0
+
+jobs:
+  build:
+    name: build (${{ matrix.arch }})
+    permissions:
+      contents: write # needed to upload disk images to the GitHub release
+      packages: write
+      id-token: write
+      attestations: write
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+        - arch: x86_64
+          oci_arch: amd64
+          libc: musl
+          attester: snp-attester,tdx-attester
+          measured_rootfs: "yes"
+          runner: ubuntu-24.04
+        - arch: aarch64
+          oci_arch: arm64
+          libc: gnu
+          attester: cca-attester
+          measured_rootfs: "yes"
+          runner: ubuntu-24.04-arm
+        - arch: s390x
+          oci_arch: s390x
+          libc: gnu
+          attester: se-attester
+          measured_rootfs: "no"
+          runner: s390x
+    runs-on: ${{ matrix.runner }}
+    env:
+      ARCH: ${{ matrix.arch }}
+      LIBC: ${{ matrix.libc }}
+      ATTESTER: ${{ matrix.attester }}
+      # "kbs" enables cc_kbc, the KBS resource provider used on every arch.
+      RESOURCE_PROVIDER: kbs
+      MEASURED_ROOTFS: ${{ matrix.measured_rootfs }}
+      RUST_TARGET: ${{ matrix.arch }}-unknown-linux-${{ matrix.libc }}
+    steps:
+    - name: Harden the runner (Audit all outbound calls)
+      uses: step-security/harden-runner@9af89fc71515a100421586dfdb3dc9c984fbf411 # v2.19.4
+      with:
+        egress-policy: audit
+
+    - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+
+    - uses: actions-rust-lang/setup-rust-toolchain@166cdcfd11aee3cb47222f9ddb555ce30ddb9659 # v1.17.0
+      with:
+        target: ${{ env.RUST_TARGET }}
+        # This action injects RUSTFLAGS="-D warnings" by default, which leaks
+        # into every cargo invocation in the job — including third-party crates
+        # built from source by the NVIDIA Attestation SDK (its vendored regorus
+        # fails on the mismatched_lifetime_syntaxes lint) and our own component
+        # build. We only want to build here, not gate on warnings, so clear it.
+        rustflags: ""
+
+    - name: Install build tooling
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y --no-install-recommends \
+          build-essential clang cmake pkg-config protobuf-compiler \
+          libssl-dev libtss2-dev libdevmapper-dev musl-tools \
+          cryptsetup-bin erofs-utils libclang-dev libxml2-dev parted skopeo zlib1g-dev
+        # umoci is used to unpack the pause image into an OCI runtime bundle.
+        # Prebuilt binaries exist for every architecture in the matrix.
+        sudo curl -fsSL -o /usr/local/bin/umoci \
+          "https://github.com/opencontainers/umoci/releases/download/${UMOCI_VERSION}/umoci.linux.${{ matrix.oci_arch }}"
+        sudo chmod +x /usr/local/bin/umoci
+
+    - name: Install NVIDIA Attestation SDK
+      if: ${{ matrix.arch == 'x86_64' }}
+      run: |
+        tmpdir="$(mktemp -d)"
+        git clone https://github.com/NVIDIA/attestation-sdk "${tmpdir}/attestation-sdk"
+        git -C "${tmpdir}/attestation-sdk" fetch --depth=1 origin "${NVAT_VERSION}"
+        git -C "${tmpdir}/attestation-sdk" checkout FETCH_HEAD
+        cmake -S "${tmpdir}/attestation-sdk/nv-attestation-sdk-cpp" \
+          -B "${tmpdir}/attestation-sdk/nv-attestation-sdk-cpp/build" \
+          -DCMAKE_BUILD_TYPE=Release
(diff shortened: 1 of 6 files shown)


.github/workflows/coco-extension-image.yml
```diff
@@ -0,0 +1,294 @@
...
+      ARCH: ${{ matrix.arch }}
+      LIBC: ${{ matrix.libc }}
+      ATTESTER: ${{ matrix.attester }}
+      # CDH default and kata's coco-guest-components build use kbs,sev on every
+      # arch. "kbs" enables cc_kbc; "sev" only compiles in the online_sev_kbc
+      # plugin (EFI secret pre-attestation for AMD SEV guests). That code is
+      # plain Rust with no arch-specific cfg and links cleanly on all targets
+      # here — including aarch64/s390x, where the plugin is simply unused at
+      # runtime unless configured.
+      RESOURCE_PROVIDER: kbs,sev
```
fitzthum: sev feature is in fact useless these days. we will drop it in a separate PR.

.github/workflows/coco-extension-image.yml
```diff
@@ -0,0 +1,294 @@
...
+
+    - name: Attest OCI container image
+      if: ${{ github.event_name != 'pull_request' }}
+      uses: actions/attest@a1948c3f048ba23858d222213b7c278aabede763 # v4.1.1
+      with:
+        subject-name: ${{ steps.publish-container.outputs.image }}
+        subject-digest: ${{ steps.publish-container.outputs.digest }}
+        push-to-registry: true
+
+    - name: Attest disk image
```
fitzthum: nit: maybe call this something like "Generate Provenance" to avoid any confusion about how we use the term attestation.

tools/coco-extension/components.toml
```diff
@@ -0,0 +1,67 @@
...
+level         = 1
+args          = ["--attestation_sock", "${aa_attestation_uri}"]
+optional_args = [{ when = "initdata_toml_path", args = ["--initdata-toml", "${initdata_toml_path}"] }]
+config        = "${aa_config_path}"
+wait_socket   = "${aa_attestation_socket}"
+timeout_secs  = "${launch_process_timeout}"
+# The extension ships both the stock attestation-agent and the NVIDIA-attester
+# build; the consumer selects one via the "attester_variant" context value
+# (kata-agent uses "default", NVRC uses "nvidia").
+select        = "${attester_variant}"
```
fitzthum, line 29: I'm not sure if the attester variant is necessary in the longterm. It definitely is needed now because the nvidia attester detection heuristic needs the dynamically linked dep. We can add arch gates to this so that we skip this on non-x86 platforms (although ARM support is coming).

Then the question is how to handle x86 builds without nvidia support, but now we always build the nv deps on x86, right? So this may not be a problem at all.

I am also looking into changing the detection heuristic to not use the deps at all. Either way this is fine for now, but I think we can maybe simplify this whole selection thing in the future to follow from the arch.

tools/coco-extension/components.toml
```diff
@@ -0,0 +1,67 @@
...
+
+[paths]
+"ocicrypt-config" = "etc/ocicrypt_config.json"
+"pause-bundle"    = "pause_bundle"
+
+[[process]]
+id            = "attestation-agent"
+level         = 1
+args          = ["--attestation_sock", "${aa_attestation_uri}"]
+optional_args = [{ when = "initdata_toml_path", args = ["--initdata-toml", "${initdata_toml_path}"] }]
```
fitzthum, line 22: cc @Xynnn007 this will intercept https://github.com/confidential-containers/guest-components/pull/1406

btw @fidencio what happens if the process fails to start? Is that a fatal error?

tools/coco-extension/components.toml
```diff
@@ -0,0 +1,67 @@
...
+env          = { OCICRYPT_KEYPROVIDER_CONFIG = "${ocicrypt_config_path}", PATH = "${extension_root}/usr/sbin:/bin:/sbin:/usr/bin:/usr/sbin" }
+wait_socket  = "${cdh_socket}"
+timeout_secs = "${launch_process_timeout}"
+
+[[process]]
+id           = "api-server-rest"
+level        = 3
+path         = "usr/local/bin/api-server-rest"
+args         = ["--features", "${rest_api_features}"]
+timeout_secs = "0"
```
fitzthum, line 67: This is set to 0 because there is no `wait_socket`?

fitzthum (commented):
This looks pretty good. A couple comments; mostly non-blocking about future work.

Let's make sure we hear from @Xynnn007 and maybe @BbolroC

.github/workflows/coco-extension-image.yml
fidencio: I will drop it then, thanks!

.github/workflows/coco-extension-image.yml
fidencio: Ack!

tools/coco-extension/components.toml
fidencio: For now it's only built for x86_64, but I think we'll soon be building for aarch64 s well.

I thought about the arch-gates, but I am not sure if those are needed for now.

tools/coco-extension/components.toml
fidencio: Exactly!

tools/coco-extension/components.toml
fidencio: We don't change Kata Containers behaviour in any way and, AFAIR, it's a fatal error.

tools/coco-extension/components.toml
fitzthum: Yeah dw about it for now but I will make some tweaks to the attester that should mean that we don't need two AA builds. We will just have one per arch.

.github/workflows/coco-extension-image.yml
```diff
@@ -0,0 +1,289 @@
...
+        sudo apt-get update
+        sudo apt-get install -y --no-install-recommends \
+          build-essential clang cmake pkg-config protobuf-compiler \
+          libssl-dev libtss2-dev libdevmapper-dev musl-tools \
+          cryptsetup-bin erofs-utils libclang-dev libxml2-dev parted skopeo zlib1g-dev
+        # umoci is used to unpack the pause image into an OCI runtime bundle.
+        # Prebuilt binaries exist for every architecture in the matrix.
+        sudo curl -fsSL -o /usr/local/bin/umoci \
+          "https://github.com/opencontainers/umoci/releases/download/${UMOCI_VERSION}/umoci.linux.${{ matrix.oci_arch }}"
+        sudo chmod +x /usr/local/bin/umoci
```
Xynnn007, line 114: Note that the scripts will be run on s390x self-hosted runners. `apt` might be ok. What about the `umoci` installations? Maybe not a big deal

.github/workflows/coco-extension-image.yml
```diff
@@ -0,0 +1,289 @@
...
+      working-directory: build/coco-extension-image
+      run: |
+        arch_tag="${{ github.sha }}-${{ matrix.oci_arch }}"
+        files=(kata-containers-coco-extension.img)
+        if [ -f root_hash_coco-extension.txt ]; then
+          files+=(root_hash_coco-extension.txt)
+        fi
+        oras push "${DISK_IMAGE}:${arch_tag}" \
+          --artifact-type application/vnd.confidential-containers.coco-extension.disk \
+          "${files[@]}"
```
Xynnn007, line 211: just wonder. This will push multiple files as different oci layers of a same manifest?

.github/workflows/coco-extension-image.yml
```diff
@@ -0,0 +1,289 @@
...
+        subject-name: ${{ steps.publish-container.outputs.image }}
+        subject-digest: ${{ steps.publish-container.outputs.digest }}
+        push-to-registry: true
+
+    - name: Generate provenance for the disk image
+      if: ${{ github.event_name != 'pull_request' }}
+      uses: actions/attest@a1948c3f048ba23858d222213b7c278aabede763 # v4.1.1
+      with:
+        subject-name: ${{ steps.publish-disk.outputs.image }}
+        subject-digest: ${{ steps.publish-disk.outputs.digest }}
```
Xynnn007, line 229: Is root_hash_coco-extension covered by the provenance?

tools/coco-extension/components.toml
Xynnn007: @fitzthum thanks for the note. I can restart the initdata processor work after this work. As there is no strong push on that thread and needs some more re-design.

.github/workflows/coco-extension-image.yml
fidencio: @Xynnn007, exactly, each file becomes a separate blob/layer referenced by that single manifest.

.github/workflows/coco-extension-image.yml
fidencio: Yes, when it's present (s390x builds with MEASURED_ROOTFS=no, for instance),

mythi:
Just a quick look but one thing that stands out is why this is not using `mkosi`. AFAUI, Peer-pods uses it so we can be aligned here since there's no legacy to support. There's also a `mkosi` action that helps to simplify things.

fidencio:
> Just a quick look but one thing that stands out is why this is not using `mkosi`. AFAUI, Peer-pods uses it so we can be aligned here since there's no legacy to support. There's also a `mkosi` action that helps to simplify things.

What `mkosi` would bring us here, **apart from complexity**, in order to build an image that **only has** the guest components?

I understand using `mkosi` for building a full rootfs image, but this is **definitely not the case here**.

Xynnn007 (approved):
LGTM. Let's get rid of the pause image once kata side work is done.

mkulke:
> > Just a quick look but one thing that stands out is why this is not using `mkosi`. AFAUI, Peer-pods uses it so we can be aligned here since there's no legacy to support. There's also a `mkosi` action that helps to simplify things.
> 
> What `mkosi` would bring us here, **apart from complexity**, in order to build an image that **only has** the guest components?
> 
> I understand using `mkosi` for building a full rootfs image, but this is **definitely not the case here**.

fwiw, mkosi also supports [building sysext's](https://mkosi.systemd.io/sysext.html) afaik. Sysext seems to have some conceptual overlap with the proposed erofs layer scheme. I've never tried this though. I expect it to work nicely if you have a systemd-based rootfs and less great if you don't. 

Do we expect an impact on peerpods image building, or is this only relevant for the kata rootfs builds? peerpods currently consumes oci bundles from kata + guest-component I think.

fidencio:
> fwiw, mkosi also supports [building sysext's](https://mkosi.systemd.io/sysext.html) afaik. Sysext seems to have some conceptual overlap with the proposed erofs layer scheme. I've never tried this though. I expect it to work nicely if you have a systemd-based rootfs and less great if you don't.

systemd-sysext was not exactly the best fit for what we needed, so it doesn't help us here.

> Do we expect an impact on peerpods image building, or is this only relevant for the kata rootfs builds? peerpods currently consumes oci bundles from kata + guest-component I think.

No-one is forced to switch to the guest extensions, it's not mandatory, at all.
My understanding is that peer-pods builds its own images already, so, nothing changes there.

.github/workflows/coco-extension-image.yml
fidencio: Choi confirmed this is fine to do.

BbolroC (approved):
LGTM, thanks @fidencio !

mythi:
(This PR had open threads so not ready for merge yet I think)

> > Just a quick look but one thing that stands out is why this is not using `mkosi`. AFAUI, Peer-pods uses it so we can be aligned here since there's no legacy to support. There's also a `mkosi` action that helps to simplify things.
> 
> What `mkosi` would bring us here, **apart from complexity**, in order to build an image that **only has** the guest components?
> 
> I understand using `mkosi` for building a full rootfs image, but this is **definitely not the case here**.

Complexity in what sense? With just ~30 lines of mkosi recipes I'm able to build an erofs + dm-verity image that adds hello.txt to `/usr/local/share` and nothing else.

fidencio:
> (This PR had open threads so not ready for merge yet I think)

2 maintainers had it approved. Do we need to change the rules so ALL maintainers approve?

> Complexity in what sense? With just ~30 lines of mkosi recipes I'm able to build an erofs + dm-verity image that adds hello.txt to /usr/local/share and nothing else.

Please, feel free to open a PR and we'll evaluate the solution.

PR #1555, opened 2026-07-14
cdh: document and test inline image_security_policy

Xynnn007:
Expose the image-rs image_security_policy option in CDH example configs and cover parsing so init-data can set the policy without a URI.

### DIFF
--- a/confidential-data-hub/example.config.json
+++ b/confidential-data-hub/example.config.json
@@ -22,6 +22,7 @@
         "max_concurrent_layer_downloads_per_image": 3,
         "sigstore_config_uri": "kbs:///default/sigstore-config/test",
         "image_security_policy_uri": "kbs:///default/security-policy/test",
+        "image_security_policy": "{\"default\":[{\"type\": \"reject\"}]}",
         "authenticated_registry_credentials_uri": "kbs:///default/credential/test",
         "registry_configuration_uri": "kbs:///default/registry-configuration/test",
         "registry_config": {
--- a/confidential-data-hub/example.config.toml
+++ b/confidential-data-hub/example.config.toml
@@ -76,9 +76,10 @@ max_concurrent_layer_downloads_per_image = 3
 
 # Sigstore config file URI for simple signing scheme.
 #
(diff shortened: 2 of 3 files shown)


confidential-data-hub/example.config.json
```diff
@@ -19,6 +19,7 @@
         "max_concurrent_layer_downloads_per_image": 3,
         "sigstore_config_uri": "kbs:///default/sigstore-config/test",
         "image_security_policy_uri": "kbs:///default/security-policy/test",
+        "image_security_policy": "{\n    \"default\": [\n        {\n            \"type\": \"reject\"\n        }\n    ]\n}\n",
```
fitzthum: don't really need these newlines here. also might be nice to avoid the double parsing, but that's a out of scope for now

fitzthum (approved):
LGTM. one nit

confidential-data-hub/example.config.json
Xynnn007: it can be a case for understanding that this is original string. btw what do you mean by "double parsing"?

fitzthum: I mean that we have one big json object here that has a field which is a string, but that field happens to be another json object. It might be nice to allow people to just specify one object. That's for later, though.

Regarding the newlines, I like that the test case covers newlines, but in the example config I don't think it makes sense to put in the newlines manually. I would imagine users would either provide an entry that actually spans multiple lines or they would just put it all on one line. I don't see any reason why someone would put in newlines like this. Not a big deal tho

Xynnn007: Ok. One potential way is to use a standard JSON with good ident, and generate "raw string" from it. Then we get `\n`s.

But anyway, let me remove the `\n`s in the example.


Also, about double parsing. For JSON it might be strange, but usually we are using toml as CDH config format, and the content of this field is JSON. That would make more sense. Note that `serde` does not support a outer toml deser + inner json deser, and it only support both outer + inner are same format. So we have to make inner as string.

fitzthum:
Needs rebase. Let's merge this tho.

PR #1556, opened 2026-07-14
attester: introduce az cvm vtpm evidence v2

mkulke:
This PR will make the attester produce v2 evidence, those are structurally equal to v1 evidence, but the hcl_report.var_data.user-data field is now guaranteed to be populated, as we are fetching fresh reports with the kbs provided-nonce mixed into them. This has been the case for TDX already, but we didn't assert it on the verifier side, yet. We also extend this for the SNP case.

Verifiers can assert the freshness not only for the TPM quote, but also for SNP/TDX Attestation Reports.

This has some technical implications. Now evidence retrieval will take a couple of seconds, so there will be more latency for the SNP attester. The TEE platform code is not async and longer running sync blocks that perform FFI need to be put into their own sync context within the attesters async get_evidence() fn's. Finally, to receive a TD quote TDX VMs on azure pass a TD report to IMDS. We switched to the azure-guest-attestation-sdk for this functionality to improve reliability of interactions with this IMDS endpoint. we'll probably migrate further parts to that SDK in the future.

### DIFF
--- a/attestation-agent/attester/Cargo.toml
+++ b/attestation-agent/attester/Cargo.toml
@@ -15,6 +15,7 @@ az-snp-vtpm = { version = "0.8.1", default-features = false, features = [
 az-tdx-vtpm = { version = "0.8.1", default-features = false, features = [
     "attester",
 ], optional = true }
+azure-guest-attestation-sdk = { git = "https://github.com/Azure/azure-guest-attestation-sdk", rev = "e606728446cb66d681913983cbe5563c919dd94b", optional = true }
 base64.workspace = true
 clap = { workspace = true, features = ["derive"], optional = true }
 cfg-if.workspace = true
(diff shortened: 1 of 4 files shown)


attestation-agent/attester/Cargo.toml
```diff
@@ -15,6 +15,7 @@ az-snp-vtpm = { version = "0.8.1", default-features = false, features = [
 az-tdx-vtpm = { version = "0.8.1", default-features = false, features = [
     "attester",
 ], optional = true }
+azure-guest-attestation-sdk = { git = "https://github.com/Azure/azure-guest-attestation-sdk", rev = "e606728446cb66d681913983cbe5563c919dd94b", optional = true }
```
Xynnn007, line 18: just out of curiosity - will you publish the crate?

Xynnn007 (approved):
LGTM. Thanks @mkulke

attestation-agent/attester/Cargo.toml
mkulke: I'm not working on this directly, but yes afaik there should be a crates.io releae soon

fitzthum (approved):
LGTM

PR #1557, opened 2026-07-15
cdh: generate default ocicrypt keyprovider config at startup

### LINKED ISSUE #1552: cdh: make ocicrypt_config.json runtime generated
While reviewing the composable VM images design, there was a comment that we need `/etc/ocicrypt_config.json` so I checked what is it in the first place.

This is the "keyprovider" protocol setting which in case of Kata seems to point back to the CDH socket itself:

```json
{
    "key-providers": {
        "attestation-agent": {
            "ttrpc": "unix:///run/confidential-containers/cdh.sock"
        }
    }
}
```

The system view seems to be:

```
Kata Agent ImagePull RPC -> CDH -> ocicrypt -> UnwrapKey RPC -> CDH
```

This is not only awkward but also problematic because users are allowed to set socket path via initdata cdh config. This can get out-of-sync with the static `/etc/ocicrypt_config.json` since that cannot be configured.

Can we

1. remove `/etc/ocicrypt_config.json` (from the extension image) and maybe auto generate it
~2. break the RPC callback~

Xynnn007:
Deploying CDH with encrypted image support requires shipping a static
ocicrypt_config.json and setting OCICRYPT_KEYPROVIDER_CONFIG, or UnwrapKey
requests are never routed to CDH. The file's only content is wiring that
CDH already knows from its own config.

Now, when the env var is unset, CDH writes the keyprovider config itself
(under /run, a ramfs in CoCo guests) pointing at its own socket. An
explicitly set OCICRYPT_KEYPROVIDER_CONFIG still takes precedence. The
coco-extension payload no longer bundles the static file.

The old config files are kept in-tree since downstream kata-containers
may still reference them.  cc @fidencio 

Close #1552

### DIFF
--- a/confidential-data-hub/hub/src/config/ocicrypt_config.rs
+++ b/confidential-data-hub/hub/src/config/ocicrypt_config.rs
@@ -1,0 +1,160 @@
+// Copyright (c) 2026 Alibaba Cloud
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+//! Default ocicrypt keyprovider config generation.
+//!
+//! This module is only compiled when the `ttrpc` or `grpc` feature is enabled,
+//! since the generated config points ocicrypt-rs at CDH's UnwrapKey RPC
+//! service. Library-only builds (e.g. cdh-oneshot) have no such listener.
+
+use std::{env, fs, path::Path};
+
+use anyhow::{Context, Result};
+
+use crate::hub::CDH_BASE_DIR;
+
+use super::CdhConfig;
+
+/// Historical keyprovider name embedded in encrypted image annotations
+/// (`org.opencontainers.image.enc.keys.provider.attestation-agent`).
+const OCICRYPT_KEYPROVIDER_NAME: &str = "attestation-agent";
+
+/// Environment variable consumed by `ocicrypt-rs` to locate the keyprovider
+/// config file. See <https://github.com/containers/ocicrypt/blob/main/docs/keyprovider.md>.
+pub const OCICRYPT_KEYPROVIDER_CONFIG_ENV: &str = "OCICRYPT_KEYPROVIDER_CONFIG";
+
+impl CdhConfig {
+    /// Ensure `OCICRYPT_KEYPROVIDER_CONFIG` is set so `ocicrypt-rs` can unwrap
+    /// image encryption keys via CDH's KeyProvider service.
+    ///
+    /// If the env var is already set (operator override), it is left alone.
+    /// Otherwise a config file is generated that maps the historical
+    /// `attestation-agent` provider name to this CDH instance's socket.
+    pub(super) fn ensure_ocicrypt_keyprovider_config(&self) -> Result<()> {
+        if env::var(OCICRYPT_KEYPROVIDER_CONFIG_ENV).is_ok() {
+            return Ok(());
+        }
+
+        // In CoCo scenarios /run is a ramfs inside the guest, so it cannot be
+        // tampered with by the host.
+        let parent = Path::new(CDH_BASE_DIR);
+        let path = parent.join("ocicrypt_config.json");
+        let parent = path.parent().expect("path has a parent directory");
+        fs::create_dir_all(parent)
+            .with_context(|| format!("failed to create {}", parent.display()))?;
+        fs::write(&path, self.default_ocicrypt_keyprovider_config()).with_context(|| {
+            format!(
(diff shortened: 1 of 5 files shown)


confidential-data-hub/hub/src/config/ocicrypt_config.rs
```diff
@@ -0,0 +1,157 @@
...
+    /// `attestation-agent` provider name to this CDH instance's socket.
+    pub(super) fn ensure_ocicrypt_keyprovider_config(&self) -> Result<()> {
+        if env::var(OCICRYPT_KEYPROVIDER_CONFIG_ENV).is_ok() {
+            return Ok(());
+        }
+
+        // In CoCo scenarios /run is a ramfs inside the guest, so it cannot be
+        // tampered with by the host.
+        let path = Path::new("/run/confidential-containers/cdh/ocicrypt_config.json");
+        let parent = path.parent().expect("path has a parent directory");
```
mythi:
```suggestion
        let parent = Path::new(CDH_BASE_DIR);
        let path = parent.join("ocicrypt_config.json");
```

mythi (commented):
thanks for doing this. one suggestion

confidential-data-hub/hub/src/config/ocicrypt_config.rs
```diff
@@ -0,0 +1,157 @@
...
+
+use std::{env, fs, path::Path};
+
+use anyhow::{Context, Result};
+
+use super::CdhConfig;
+
+/// Historical keyprovider name embedded in encrypted image annotations
+/// (`org.opencontainers.image.enc.keys.provider.attestation-agent`).
+const OCICRYPT_KEYPROVIDER_NAME: &str = "attestation-agent";
```
fitzthum, line 22: btw if we ever want to transition to a new name in the annotations, we could just create a second entry in the config.

fitzthum (approved):
LGTM

confidential-data-hub/hub/src/config/ocicrypt_config.rs
Xynnn007: emm, imo if we need this, we can directly use the ENV with an extra ocicrypt config file

tools/coco-extension/components.toml
```diff
@@ -55,7 +54,7 @@ config       = "${cdh_config_path}"
 # base-nvidia image's /sbin and /bin. The agent launches CDH with
 # PATH=/bin:/sbin:/usr/bin:/usr/sbin, but setting any env here overrides it
 # wholesale, so prepend the extension's usr/sbin and restore the base dirs.
-env          = { OCICRYPT_KEYPROVIDER_CONFIG = "${ocicrypt_config_path}", PATH = "${extension_root}/usr/sbin:/bin:/sbin:/usr/bin:/usr/sbin" }
+env          = { PATH = "${extension_root}/usr/sbin:/bin:/sbin:/usr/bin:/usr/sbin" }
```
fidencio: While there's no integration of the work here and Kata Containers, I'd avoid drifting from what Kata Containers uses and provides, as it may just make things harder to reconcile later on.

With that said, what I'd do if I were to work on this is one of the two approaches:
1. Get whatever we have as it is used by Kata Containers, and then do improvements on that
2. Change only the code here (not the guest extension), update the code on Kata Containers side, update the extension on Kata Containers side, and then reconcile this back here.

Drifting, IMHO, is a no go, as what this repo has is not yet a source of truth for this integration.

fidencio (dismissed):
Leff one comment about drifting the config with what's in the Kata Containers side.

tools/coco-extension/components.toml
Xynnn007: ok let me drop the guest image commit in this PR. Note that the current functionality is "add-only" for GC, thus will not break kata side.

fidencio: Yeah, I understand it wouldn't break the Kata side. What I don't understand is why GC community is investing on getting the guest extension image being changed without having it consumed by Kata Containers in the first place.

It just makes things way harder to be tested, and the whole process easier to be error prone.

mythi: Going with item 2 makes sense. :+1:

fidencio:
> Leff one comment about drifting the config with what's in the Kata Containers side.

@Xynnn007, thanks for addressing my concern. I've dismissed my review, but I'm not comfortable to approve any PR coming from this community. I'll leave it for the others to do so.

mythi (approved):

PR #1567, opened 2026-07-18
aa/eventlog: fix WAL crash recovery

### LINKED ISSUE #1566: AA eventlog WAL crash recovery is broken
### Describe the bug

If the attestation-agent dies between extending the PCR and finishing the eventlog write, the WAL recovery on the next start doesn't restore a consistent eventlog. The log ends up out of sync with the PCR, so attestation of the AAEL fails afterwards. There are a few reasons:

1. The WAL offset is written big-endian but read back little-endian, so recovery seeks to a garbage position (e.g. offset 512 comes back as 562949953421312):

https://github.com/confidential-containers/guest-components/blob/e0904d9ee692f0ea8e3a67ad84b5fb768a20446a/attestation-agent/attestation-agent/src/eventlog/mod.rs#L149

https://github.com/confidential-containers/guest-components/blob/e0904d9ee692f0ea8e3a67ad84b5fb768a20446a/attestation-agent/attestation-agent/src/eventlog/mod.rs#L170

2. The eventlog file is opened with append(true), so the recovery seek has no effect anyway and writes always go to the end of the file. A half-written entry never gets overwritten, the recovered entry just gets appended after it, creating a problem.
3. stream_position() right after an append-mode open returns 0, so when AA restarts with a non-empty log, pos and the next WAL offset are wrong from the start.

### How to reproduce

1. Run AA and extend at least one AAEL entry, so /run/attestation-agent/eventlog is non-empty.
2. Kill AA in the window between the WAL file being written and removed in extend_entry, leaving /run/attestation-agent/.wal_event_entry behind.
3. Restart AA. Recovery in EventLog::new reads back a byte-swapped offset and appends the entry at the end of the file regardless, instead of rewriting it at the recorded offset.
4. Verify the eventlog against the RTMR/PCR, it fails

### CoCo version information

0.20.0

Tdx

```shell

```

mmahut:
Crash recovery for the eventlog didn't work:

- The WAL offset was written big-endian but read back little-endian,
  so recovery seeked to a wrong position.
- The eventlog file was opened with O_APPEND, so the recovery seek did
  nothing and a half-written entry was never overwritten.
- After reopening an existing log, the writer position started at 0
  instead of the end of the file.

Also use write_all() so short writes are not silently dropped, and
serialize the recovered entry the same way as the normal write path.

Fixes: #1566

### DIFF
--- a/attestation-agent/attestation-agent/src/eventlog/mod.rs
+++ b/attestation-agent/attestation-agent/src/eventlog/mod.rs
@@ -49,9 +49,24 @@ pub struct FileWriter {
     pos: u64,
 }
 
+impl FileWriter {
+    /// Open the log file for writing, positioned at its end. No `O_APPEND`,
+    /// so that recovery can rewrite an entry at its recorded offset.
+    fn open(path: &Path) -> Result<Self> {
+        let mut file = File::options()
+            .write(true)
+            .create(true)
+            .truncate(false)
+            .open(path)
+            .context("open log file")?;
+        let pos = file.seek(SeekFrom::End(0))?;
+        Ok(Self { file, pos })
+    }
+}
+
 impl Writer for FileWriter {
     fn write(&mut self, data: &[u8]) -> Result<()> {
-        self.file.write(data).context("failed to write log")?;
+        self.file.write_all(data).context("failed to write log")?;
         self.file
             .sync_data()
             .context("failed to flush log to I/O media")?;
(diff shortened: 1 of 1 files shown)


Xynnn007 (approved):
LGTM. Thanks! @mmahut

Xynnn007:
Btw, you should finish the commit message and have a sign-off to cover DCO check.

fitzthum (approved):
LGTM. Please update DCO and add some commit message

PR #1568, opened 2026-07-18
ci: publish secret CLI as a released binary artifact

### LINKED ISSUE #1554: Ship the secret CLI as a released binary artifact
## Summary

  Today the secret CLI (the sealed-secret sealing/unsealing tool) can only be obtained by building from source:

  cargo build -p confidential-data-hub --bin secret --features cli,<kms-provider>

  We should distribute it as a prebuilt binary so users don't need a full Rust toolchain and repo checkout just to seal secrets.

  ## Motivation

  Every KMS provider doc instructs users to git clone and cargo build the CLI before they can seal a secret. That's a heavy prerequisite for what is a user-side/operator tool, and it's the main friction point for anyone trying the sealed-secret flow. Shipping a binary makes the "prepare a sealed secret" workflow accessible without a build environment.

  ## Context

  - The secret bin already exists: confidential-data-hub/hub/Cargo.toml → [[bin]] name = "secret", path = "src/bin/secret_cli.rs", required-features = ["cli"].
  - KMS providers are opt-in cargo features (aliyun, aws, ehsm, …), so a shipped binary needs a decision on which providers are compiled in.

  ## Proposed scope

  - [ ] Decide the default feature set for the released CLI (all providers vs. per-provider builds).
  - [ ] Add a release job (GitHub Actions) that builds secret and attaches it as a release asset — and/or publishes it in a container image.
  - [ ] Consider target platforms (at least x86_64-unknown-linux-gnu; possibly musl for static linking and aarch64).
  - [ ] Update the KMS provider docs to point at the released binary instead of cargo build from source.

gaganhr94:
Build the sealed-secret 'secret' CLI with all KMS providers bundled and upload per-platform tarballs (x86_64 gnu/musl, aarch64 gnu) as release assets on publish. Update KMS provider docs to reference the prebuilt binary instead of requiring a build from source.

Fixes #1554

### DIFF
--- a/.github/workflows/secret_cli_release.yml
+++ b/.github/workflows/secret_cli_release.yml
@@ -1,0 +1,119 @@
+name: Release secret CLI binary
+
+on:
+  release:
+    types: [published]
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  build-and-upload:
+    permissions:
+      # Write scopes are only exercised by the release-only upload/publish
+      # steps below (gated on github.event_name == 'release'). On pull_request
+      # the job just builds the binary to catch breakage early; fork PRs get a
+      # read-only token regardless of what is requested here.
+      contents: write
+      packages: write
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+        - arch: x86_64
+          libc: musl
+          runner: ubuntu-24.04
+        - arch: aarch64
+          libc: musl
+          runner: ubuntu-24.04-arm
+    runs-on: ${{ matrix.platform.runner }}
+    env:
+      RUST_TARGET: ${{ matrix.platform.arch }}-unknown-linux-${{ matrix.platform.libc }}
+      # KMS providers bundled into the released binary. Building all of them
+      # into one artifact keeps the released tool provider-agnostic.
+      CLI_FEATURES: cli,aliyun,aws,kbs
+      REGISTRY: ghcr.io
+      IMAGE_NAME: ${{ github.repository }}
+    steps:
+    - name: Harden the runner (Audit all outbound calls)
+      uses: step-security/harden-runner@bf7454d06d71f1098171f2acdf0cd4708d7b5920 # v2.20.0
+      with:
+        egress-policy: audit
+
+    - name: Log in to the Container registry
+      if: github.event_name == 'release'
+      uses: docker/login-action@650006c6eb7dba73a995cc03b0b2d7f5ca915bee # v4.2.0
+      with:
+        registry: ${{ env.REGISTRY }}
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Set up ORAS
+      if: github.event_name == 'release'
+      uses: oras-project/setup-oras@1d808f7d7f6995cc68b7bf507bfe5c5446e1dc9d # v2.0.1
+      with:
+        version: 1.2.0
+
+    - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      with:
+        persist-credentials: false
+
+    - uses: actions-rust-lang/setup-rust-toolchain@166cdcfd11aee3cb47222f9ddb555ce30ddb9659 # v1.17.0
+      with:
+        # The bundled AWS SDK crates require a newer rustc than the repo's
+        # pinned toolchain (rust-toolchain.toml). Build the release binary on
+        # stable so the aws feature compiles.
+        toolchain: stable
+        target: ${{ env.RUST_TARGET }}
+
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y --no-install-recommends \
+          protobuf-compiler libclang-dev cmake pkg-config
+        if [ "${{ matrix.platform.libc }}" = "musl" ]; then
+          sudo apt-get install -y --no-install-recommends musl-tools
+        fi
+
+    - name: Build secret CLI
+      run: |
+        cargo build \
+          -p confidential-data-hub \
+          --bin secret \
+          --release \
+          --target "$RUST_TARGET" \
+          --no-default-features \
+          --features "$CLI_FEATURES"
+
+    - name: Package artifact
+      if: github.event_name == 'release'
+      id: package
+      run: |
+        name="secret-${RUST_TARGET}"
+        mkdir -p "$name"
+        cp "target/${RUST_TARGET}/release/secret" "$name/"
+        tar czf "${name}.tar.gz" "$name"
+        sha256sum "${name}.tar.gz" > "${name}.tar.gz.sha256"
+        echo "asset=${name}.tar.gz" >> "$GITHUB_OUTPUT"
+
+    - name: Upload release assets
+      if: github.event_name == 'release'
+      env:
+        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        TAG: ${{ github.event.release.tag_name }}
+        ASSET: ${{ steps.package.outputs.asset }}
+      run: |
+        gh release upload "$TAG" "$ASSET" "${ASSET}.sha256" \
+          --repo "$GITHUB_REPOSITORY" --clobber
+
+    - name: Publish with ORAS
+      if: github.event_name == 'release'
+      env:
+        TAG: ${{ github.event.release.tag_name }}
+        ASSET: ${{ steps.package.outputs.asset }}
+      run: |
+        image="${REGISTRY}/${IMAGE_NAME}/secret_cli"
+        oras push \
+          "${image}:${TAG}-${RUST_TARGET},latest-${RUST_TARGET}" \
+          "$ASSET"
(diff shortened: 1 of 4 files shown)


.github/workflows/secret_cli_release.yml
```diff
@@ -0,0 +1,92 @@
...
+          runner: ubuntu-24.04
+        - arch: aarch64
+          libc: gnu
+          runner: ubuntu-24.04-arm
+    runs-on: ${{ matrix.platform.runner }}
+    env:
+      RUST_TARGET: ${{ matrix.platform.arch }}-unknown-linux-${{ matrix.platform.libc }}
+      # KMS providers bundled into the released binary. Building all of them
+      # into one artifact keeps the released tool provider-agnostic.
+      CLI_FEATURES: cli,aliyun,aws,kbs,ehsm
```
fitzthum: I think eHSM is not supported anymore so not sure if we need that feature.

fitzthum (commented):
Nice. Thanks for taking a look at this. 

In Trustee we distribute some of our CLI binaries using ORAS. See [here](https://github.com/confidential-containers/trustee/blob/main/.github/workflows/release-clis-to-ghcr.yml). Might make sense to do that here too.

.github/workflows/secret_cli_release.yml
gaganhr94: I'll remove eHSM here

.github/workflows/secret_cli_release.yml
```diff
@@ -0,0 +1,119 @@
...
+        platform:
+        - arch: x86_64
+          libc: gnu
+          runner: ubuntu-24.04
+        - arch: x86_64
+          libc: musl
+          runner: ubuntu-24.04
+        - arch: aarch64
+          libc: gnu
+          runner: ubuntu-24.04-arm
```
Xynnn007: For x86, musl is enough i think. For aarch64, if possible, musl can be good

gaganhr94: done

.github/workflows/secret_cli_release.yml
```diff
@@ -0,0 +1,116 @@
...
+      IMAGE_NAME: ${{ github.repository }}
+    steps:
+    - name: Harden the runner (Audit all outbound calls)
+      uses: step-security/harden-runner@bf7454d06d71f1098171f2acdf0cd4708d7b5920 # v2.20.0
+      with:
+        egress-policy: audit
+
+    - name: Log in to the Container registry
+      if: github.event_name == 'release'
+      uses: docker/login-action@650006c6eb7dba73a995cc03b0b2d7f5ca915bee
```
Xynnn007: The action tag is missed here

.github/workflows/secret_cli_release.yml
```diff
@@ -0,0 +1,116 @@
...
+        gh release upload "$TAG" "$ASSET" "${ASSET}.sha256" \
+          --repo "$GITHUB_REPOSITORY" --clobber
+
+    - name: Publish with ORAS
+      if: github.event_name == 'release'
+      env:
+        TAG: ${{ github.event.release.tag_name }}
+        ASSET: ${{ steps.package.outputs.asset }}
+      run: |
+        image="${REGISTRY}/${IMAGE_NAME}/secret"
```
Xynnn007: `secret_cli` may be better than `secret`?

.github/workflows/secret_cli_release.yml
```diff
@@ -0,0 +1,116 @@
...
+        oras push \
+          "${image}:${TAG}-${RUST_TARGET},latest-${RUST_TARGET}" \
+          "$ASSET"
+
+    - name: Upload workflow artifact
+      if: github.event_name == 'workflow_dispatch'
+      uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+      with:
+        name: secret-${{ env.RUST_TARGET }}
+        path: ${{ steps.package.outputs.asset }}
```
Xynnn007: Why we need this `workflow_dispatch` arm and upload? I guess this is for outer caller, but we can add it when we actually need

.github/workflows/secret_cli_release.yml
```diff
@@ -0,0 +1,116 @@
...
+      packages: write
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+        - arch: x86_64
+          libc: musl
+          runner: ubuntu-24.04
+        - arch: aarch64
+          libc: musl
```
Xynnn007, line 28: Sry, let me make it clear: can we build the cli for aarch64 musl for now? if so, we can use this; if not, gnu is ok

.github/workflows/secret_cli_release.yml
gaganhr94: removed workflow dispatch

.github/workflows/secret_cli_release.yml
gaganhr94: renamed it to secret_cli

.github/workflows/secret_cli_release.yml
gaganhr94: added the tag here

.github/workflows/secret_cli_release.yml
gaganhr94: Tried building it in the pipeline and the build did go through without any issues: https://github.com/confidential-containers/guest-components/actions/runs/29808007146/job/88562530435?pr=1568

So, continuing with musl here

.github/workflows/secret_cli_release.yml
```diff
@@ -0,0 +1,115 @@
+name: Release secret CLI binary
+
+on:
+  release:
+    types: [published]
+  # TEMPORARY: added to verify the aarch64 musl build in CI. Remove before merge.
```
Xynnn007: highlight here

gaganhr94: Removed this

Xynnn007 (approved):
LGTM. Thanks!

.github/workflows/secret_cli_release.yml
```diff
@@ -0,0 +1,108 @@
+name: Release secret CLI binary
+
+on:
+  release:
+    types: [published]
```
fitzthum, line 5: I'm not a huge fan of running fallible flows on the release.

@Xynnn007 do you think we should run this on every merge? That might be too much, but otherwise we'll have to wait two months for this to run.

confidential-data-hub/docs/SEALED_SECRET.md
```diff
@@ -10,7 +10,26 @@ in conjunction with an attestation.
...
 
-You can use the secret cli tool to generate a sealed secret:
+You can use the secret cli tool to generate a sealed secret.
+
+### Prebuilt binary (recommended)
+
+Each [GitHub release](https://github.com/confidential-containers/guest-components/releases)
+ships the `secret` CLI as a prebuilt artifact with all KMS providers bundled,
+so you do not need a Rust toolchain. Download the tarball matching your
+platform, verify it, and extract the binary:
```
fitzthum, line 20: The catch is that we just cut a release without this workflow so these directions are not going to work until the next one.

fitzthum (commented):
Looking good. I just wonder if we should tweak the workflow to run more frequently.

.github/workflows/secret_cli_release.yml
gaganhr94: I'll change this once @Xynnn007 replies back with his inputs here

confidential-data-hub/docs/SEALED_SECRET.md
gaganhr94: What is the change that you are recommending here ? Should I add a line saying this note is not applicable until a certain version of guest-componets is released.

fitzthum: This is kind of a part 2 of my other comment. If we published the artifact more frequently then we wouldn't need to do anything.

.github/workflows/secret_cli_release.yml
Xynnn007: it's ok to run this every merging. See what we are doing https://github.com/confidential-containers/guest-components/blob/main/.github/workflows/publish-artifacts.yml

But I am not sure if the CLI tool needs such storage and computation resource. A intermediate solution would be to only build on every PR CI, but only release on releases.

gaganhr94: Thanks @Xynnn007, will implement the intermediate solution for now.

gaganhr94:
@Xynnn007 @fitzthum this PR is ready for review. I have made changes for the workflow to run on every PR CI and the binary is pushed only during a release event

.github/workflows/secret_cli_release.yml
```diff
@@ -0,0 +1,118 @@
...
+      run: |
+        cargo build \
+          -p confidential-data-hub \
+          --bin secret \
+          --release \
+          --target "$RUST_TARGET" \
+          --no-default-features \
+          --features "$CLI_FEATURES"
+
+    - name: Package artifact
```
Xynnn007, line 89: maybe we can also add `if release` condition this step

Xynnn007 (approved):
One comment. Otherwise LGTM

.github/workflows/secret_cli_release.yml
gaganhr94: Done

fitzthum (approved):
LGTM. Thanks @gaganhr94 

let's see how this release strategy works out. we can update if need be

PR #1569, opened 2026-07-19
certs: fix self-signed cert guidance in CDH example config

nogunix:
## What this does

Corrects the self-signed certificate guidance in
`confidential-data-hub/example.config.toml`. The comment stated that a
self-signed certificate must have the `basicConstraints CA:true` extension,
but rustls (the default TLS backend) does the opposite: it rejects a
`CA:true` certificate when it is presented as a leaf (end-entity), failing
with a `CaUsedAsEndEntity` error.

## Why

A self-signed certificate added via `extra_root_certificates` (or an
`image_pull_proxy` certificate) becomes both the trust anchor and the leaf
presented during the TLS handshake. webpki validates the leaf and rejects
`CA:true` in the end-entity position, so the current guidance leads users to
create certificates that fail the handshake.

## Verification

Reproduced end-to-end with a tokio-rustls server + reqwest (rustls) client,
using the same cert as both trust anchor and leaf, on
rustls 0.23.42 / rustls-webpki 0.103.13 / reqwest 0.12.28:

- `basicConstraints CA:true`  self-signed leaf → **REJECTED** (`CaUsedAsEndEntity`)
- `basicConstraints CA:false` self-signed leaf → **ACCEPTED** (HTTP 200)

## Change

Fixes the `extra_root_certificates` and `image_pull_proxy` comment blocks to
state that a self-signed cert must **not** have `CA:true`, and to point at the
recommended alternative (a `CA:true` CA cert signing a separate `CA:false`
server cert).

### DIFF
--- a/confidential-data-hub/example.config.toml
+++ b/confidential-data-hub/example.config.toml
@@ -173,9 +173,10 @@ location = "123456.mirror.aliyuncs.com"
 #
 # IMPORTANT: This works by adding the cert as a trusted root CA. Therefore:
 # - Supply the CA certificate that signed the registry's (or proxy's) TLS certificate.
-# - If using a self-signed certificate, it must have the CA:true Basic Constraints
-#   extension (i.e., it acts as both CA and leaf). Plain self-signed leaf certificates
-#   (without CA:true) are NOT supported when using rustls (the default TLS backend).
+# - If using a self-signed certificate, it must NOT have the CA:true Basic
+#   Constraints extension: rustls (the default TLS backend) rejects a CA:true cert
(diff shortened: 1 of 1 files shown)


mythi:
@bpradipt PTAL? (ref a3eb154)

bpradipt:
> @bpradipt PTAL? (ref [a3eb154](https://github.com/confidential-containers/guest-components/commit/a3eb1544f3adac92a3b3d830fa0ef8eb0a374704))

Interesting.. @lmilleri do you recall the errors we faced ?

Anyways, the new comment looks correct to me.

bpradipt (approved):
/lgtm

PR #89, opened 2022-12-02
Add feature for image-rs with defaults except cosign

mattarnoatibm:
Add a feature that can be specificed by crates that include the image-rs crate to build image-rs with all default features except cosign.

Fixes confidential-containers#88

Signed-off-by: Matthew Arnold <mattarno@uk.ibm.com>

PR #1575, opened 2026-07-21
attester: Add NVIDIA DPU DICE attester

haixinJY:
Add attestation support for NVIDIA BlueField DPU devices using the DICE trust chain.

This is a re-submission of #1538 which was accidentally closed due to a git history issue (orphan commit with no common ancestor to main).

## Changes
- Add `nvidia_dpu` attester module
- Read DICE certificate chain from device sysfs
- Parse DeviceID and Alias certificates
- Extract hardware measurements and OOB integrity hashes
- Package evidence in structured `DpuEvidence` format

## Design decisions
- `version` uses `u32` for type safety (per @fitzthum suggestion)
- Certificate data uses base64 encoding via `serde_with`
- Error propagation replaces `unwrap_or_default()` for required fields
- Oversize nonce rejected instead of silently truncated
- SigningKey length validated upfront (48 bytes for P-384)
- Measurement buffer length validated before chunking

The `[patch.crates-io]` kbs-types fork reference is temporary pending confidential-containers/kbs-types#89 merge.

Signed-off-by: Ying Jiang <jiangying8582@163.com>
Signed-off-by: Ying Jiang <haixin.jy@alibaba-inc.com>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ hmac = "0.13.0"
 jwt-simple = { version = "0.12", default-features = false, features = [
     "pure-rust",
 ] }
-kbs-types = "0.15.0"
+kbs-types = { git = "https://github.com/confidential-containers/kbs-types", rev = "b26f44e" }
 nix = "0.31"
 openssl = "0.10"
 prost = "0.14"
@@ -74,3 +74,4 @@ ttrpc-codegen = "0.6.0"
 url = "2.5.4"
 uuid = "1"
 zeroize = "1.8.2"
+
--- a/attestation-agent/attester/src/lib.rs
+++ b/attestation-agent/attester/src/lib.rs
@@ -46,6 +46,9 @@ pub mod tpm;
 #[cfg(feature = "nvidia-attester")]
 pub mod nvidia;
 
+#[cfg(feature = "nvidia-dpu-attester")]
+pub mod nvidia_dpu;
+
 pub type BoxedAttester = Box<dyn Attester + Send + Sync>;
 
 impl TryFrom<Tee> for BoxedAttester {
@@ -174,6 +177,17 @@ impl TryFrom<Tee> for BoxedAttester {
                 #[cfg(not(feature = "nvidia-attester"))]
                 {
                     bail!("`nvidia-attester` feature is not enabled!");
+                }
+            }
+            Tee::NvidiaDpu => {
+                #[cfg(feature = "nvidia-dpu-attester")]
+                {
+                    Box::new(nvidia_dpu::NvidiaDpuAttester::default())
+                }
+
+                #[cfg(not(feature = "nvidia-dpu-attester"))]
+                {
+                    bail!("`nvidia-dpu-attester` feature is not enabled!");
                 }
             }
         };
--- a/attestation-agent/attester/src/nvidia_dpu/mod.rs
+++ b/attestation-agent/attester/src/nvidia_dpu/mod.rs
@@ -1,0 +1,242 @@
+// Copyright (c) 2026 Alibaba Cloud
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+//! NVIDIA DPU (Data Processing Unit) Attester
+//!
+//! Collects attestation evidence from NVIDIA DPU devices (e.g., BlueField-3)
+//! using the DICE (Device Identifier Composition Engine) certificate chain.
+//!
+//! The NVIDIA DPU attester reads hardware-rooted attestation evidence including:
+//! - Alias certificate (leaf) containing device measurements
+//! - DeviceID certificate signed by manufacturer root
+//! - Firmware measurements from DICE layers
+//!
+//! Platform detection uses PCI device ID scanning to identify BlueField-3
+//! hardware (vendor `15b3`, device `a2dc` for BF3 NIC or `a2da` for SoC bridge).
+
+use super::{Attester, TeeEvidence};
+use anyhow::{bail, Context, Result};
+use p384::ecdsa::signature::Signer;
+use p384::ecdsa::{Signature, SigningKey};
+use serde::{Deserialize, Serialize};
+use serde_with::{base64::Base64, serde_as};
+
+/// PCI vendor ID for Mellanox/NVIDIA networking devices.
+const NVIDIA_MLX_VENDOR_ID: &str = "15b3";
+/// PCI device IDs that identify BlueField-3 (as opposed to ConnectX-7 or other NICs).
+/// - a2dc: BF3 integrated NIC function
+/// - a2da: BF3 SoC PCI bridge
+const BF3_DEVICE_IDS: &[&str] = &["a2dc", "a2da"];
+
+const PCI_DEVICES_PATH: &str = "/sys/bus/pci/devices";
+const NVIDIA_DPU_NONCE_SIZE: usize = 64;
+
+/// Architecture identifier for NVIDIA BlueField-3 DPU.
+const ARCHITECTURE_BF3: &str = "bluefield3";
+
+/// DICE alias private key exposed by the BlueField platform attestation subsystem.
+///
+/// Requires: `mlnx_bf_attestation` kernel module (part of MLNX_OFED >= 24.10 / DOCA >= 2.8).
+/// This runs on the BF3 ARM SoC Linux (the "DPU OS"), not the x86 host.
+///
+/// Security model: key availability to the attested OS is inherent to DICE.
+/// CDI (Compound Device Identifier, TCG DICE spec): each layer's CDI is derived
+/// from the parent layer's CDI and the measurement of the current firmware layer.
+/// Any firmware change alters the measurement, rotating the CDI and all downstream
+/// DICE-derived keys (including this Alias key) and certificates.
+/// Future: PSC hardware signing (sign-without-extract) for defense-in-depth.
+const ALIAS_PRIVATE_KEY_PATH: &str = "/sys/kernel/security/tee/dice/alias_private_key";
+
+/// Base path for DPU attestation attributes exposed via InfiniBand sysfs.
+///
+/// Requires: `mlx5_core` kernel driver with attestation support
+/// (MLNX_OFED >= 24.10 or DOCA >= 2.8 BFB image with attestation feature enabled).
+/// The standard DOCA production BFB images include this by default since DOCA 2.8.
+///
+/// Reference: <https://docs.nvidia.com/networking/display/dpunicattestation>
+// TODO: dynamically discover InfiniBand device name instead of hardcoding mlx5_0.
+// Multiple devices or different naming (mlx5_1, etc.) require enumeration.
+const ATTESTATION_BASE_PATH: &str = "/sys/class/infiniband/mlx5_0/device/attestation";
+
+/// Top-level NVIDIA DPU attestation evidence.
+/// Versioned and supports multiple device entries for forward-compatibility.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct NvidiaDpuEvidence {
+    /// Evidence format version
+    pub version: u32,
+    /// Per-device attestation evidence entries
+    pub devices: Vec<DpuDeviceEvidence>,
+}
+
+/// Evidence from a single DPU device.
+#[serde_as]
+#[derive(Debug, Serialize, Deserialize)]
+pub struct DpuDeviceEvidence {
+    /// Device architecture identifier (e.g. "bluefield3")
+    pub architecture: String,
+    /// DER-encoded X.509 DICE Alias certificate (leaf, signed by DeviceID key).
+    /// Read as raw bytes from sysfs; the verifier parses the X.509 structure.
+    #[serde_as(as = "Base64")]
+    pub alias_cert: Vec<u8>,
+    /// DER-encoded X.509 DICE DeviceID certificate (signed by manufacturer Root CA).
+    /// Read as raw bytes from sysfs; the verifier parses the X.509 structure.
+    #[serde_as(as = "Base64")]
+    pub device_id_cert: Vec<u8>,
+    /// ECDSA P-384 signature of report_data using DICE alias private key
+    #[serde_as(as = "Base64")]
+    pub report_data_signature: Vec<u8>,
+}
+
+/// Detect if NVIDIA BlueField-3 DPU hardware is available on the platform.
+///
+/// Scans PCI sysfs for devices matching the BF3 vendor:device IDs.
+/// This distinguishes BlueField-3 from regular ConnectX NICs which have
+/// different PCI device IDs.
+pub fn detect_platform() -> bool {
+    // Scan PCI devices for BF3-specific device IDs
+    let Ok(entries) = std::fs::read_dir(PCI_DEVICES_PATH) else {
+        return false;
+    };
+    for entry in entries.flatten() {
+        let vendor_path = entry.path().join("vendor");
+        let device_path = entry.path().join("device");
+
+        let vendor = std::fs::read_to_string(&vendor_path)
+            .unwrap_or_default()
+            .trim()
+            .trim_start_matches("0x")
+            .to_lowercase();
+        let device = std::fs::read_to_string(&device_path)
+            .unwrap_or_default()
+            .trim()
+            .trim_start_matches("0x")
+            .to_lowercase();
+
+        if vendor == NVIDIA_MLX_VENDOR_ID && BF3_DEVICE_IDS.iter().any(|id| device == *id) {
+            return true;
+        }
+    }
+
+    false
+}
+
+/// Sign report_data with DICE alias private key for freshness binding.
+/// Returns the raw signature bytes (DER-encoded ECDSA P-384).
+/// Returns error if the key is unavailable (unsigned evidence is not acceptable).
+fn sign_report_data(report_data: &[u8]) -> Result<Vec<u8>> {
+    // DICE alias private key from NVIDIA attestation subsystem layout.
+    // Reference: https://docs.nvidia.com/networking/display/dpunicattestation
+    let key_bytes = std::fs::read(ALIAS_PRIVATE_KEY_PATH)
+        .context("DICE alias private key not available - cannot produce signed evidence")?;
+    if key_bytes.len() != 48 {
+        bail!(
+            "signing key must be exactly 48 bytes (P-384), got {}",
+            key_bytes.len()
+        );
+    }
+    let signing_key = SigningKey::from_bytes(key_bytes.as_slice().into())
+        .context("Failed to parse DICE alias private key")?;
+    let signature: Signature = signing_key.sign(report_data);
+    Ok(signature.to_bytes().to_vec())
+}
+
+#[derive(Debug, Default)]
+pub struct NvidiaDpuAttester {}
+
+#[async_trait::async_trait]
+impl Attester for NvidiaDpuAttester {
+    /// Collect attestation evidence from the DPU device.
+    ///
+    /// The `report_data` parameter is used as a nonce to bind the evidence
+    /// to a specific attestation session, preventing replay attacks.
+    async fn get_evidence(&self, mut report_data: Vec<u8>) -> Result<TeeEvidence> {
+        // Pad or truncate to fixed nonce register size (ECDSA signs arbitrary length internally)
+        report_data.resize(NVIDIA_DPU_NONCE_SIZE, 0);
+
+        // Collect evidence from sysfs-exposed DICE attestation attributes.
+        // BF3 exposes DICE certs via InfiniBand sysfs, not a chardev.
+        let evidence = self.collect_evidence_from_sysfs(&report_data)?;
+
+        serde_json::to_value(&evidence).context("Failed to serialize NVIDIA DPU evidence")
+    }
+}
+
+impl NvidiaDpuAttester {
+    /// Collect evidence from sysfs-exposed DICE attestation attributes.
+    ///
+    /// BlueField-3 exposes DER-encoded X.509 DICE certificates via the
+    /// InfiniBand device sysfs path under `device/attestation/`.
+    /// The raw DER bytes are passed through as-is; parsing is done by the verifier.
+    fn collect_evidence_from_sysfs(&self, report_data: &[u8]) -> Result<NvidiaDpuEvidence> {
+        // Read DER-encoded X.509 DICE certificates from sysfs
+        let alias_cert = std::fs::read(format!("{}/alias_cert", ATTESTATION_BASE_PATH))
+            .context("failed to read alias cert")?;
+
+        let device_id_cert = std::fs::read(format!("{}/device_id_cert", ATTESTATION_BASE_PATH))
+            .context("failed to read device_id cert")?;
+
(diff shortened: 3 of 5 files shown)


Cargo.toml
```diff
@@ -74,3 +74,6 @@ ttrpc-codegen = "0.6.0"
 url = "2.5.4"
 uuid = "1"
 zeroize = "1.8.2"
+
+[patch.crates-io]
+kbs-types = { git = "https://github.com/56ht86hrd5-ai/kbs-types.git", branch = "feat/add-dpu-tee" }
```
Xynnn007: We can use github rev to point to official kbs-types repo w/o a version and no need this `[patch.crates-io]`

attestation-agent/attester/src/nvidia_dpu/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+
+const PCI_DEVICES_PATH: &str = "/sys/bus/pci/devices";
+const NVIDIA_DPU_NONCE_SIZE: usize = 64;
+
+/// DICE alias private key path exposed by NVIDIA DOCA attestation service via sysfs.
+/// Reference: NVIDIA DOCA DICE Programming Guide - "Key Exposure via sysfs"
+const ALIAS_PRIVATE_KEY_PATH: &str = "/sys/kernel/security/tee/dice/alias_private_key";
+
+/// Base path for DPU attestation attributes exposed via InfiniBand sysfs.
+// TODO: dynamically discover InfiniBand device name instead of hardcoding mlx5_0.
```
Xynnn007, line 59: I  do not have enough context here. @fitzthum can you assert this - if it's ok for long time iteration?

attestation-agent/attester/src/nvidia_dpu/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+// Multiple devices or different naming (mlx5_1, etc.) require enumeration.
+const ATTESTATION_BASE_PATH: &str = "/sys/class/infiniband/mlx5_0/device/attestation";
+
+/// Top-level NVIDIA DPU attestation evidence.
+/// Versioned and supports multiple device entries for forward-compatibility.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct NvidiaDpuEvidence {
+    /// Evidence format version
+    pub version: u32,
```
Xynnn007, line 68: I like the version design here.

attestation-agent/attester/src/nvidia_dpu/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+        } else {
+            Vec::new()
+        };
+
+        let report_data_signature = sign_report_data(report_data)?;
+
+        Ok(NvidiaDpuEvidence {
+            version: 1,
+            devices: vec![DpuDeviceEvidence {
+                architecture: "bluefield3".to_string(),
```
Xynnn007: Let's make it a const for `"bluefield3"`

attestation-agent/attester/src/nvidia_dpu/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+                alias_cert,
+                device_id_cert,
+                measurements,
```
Xynnn007: What's the integrity protection mechanism for `measurements` here? If an attacker modifies this field, how can the verifier side know this?

I am considering whether the `report_data_signature` field can be a more general `signature` one, s.t. `signature := sign(architecture || measurement || reportdata)`

haixinJY: Measurements (FWID) are protected by the DICE certificate chain — firmware layer hashes are embedded as X.509 extensions in the DeviceID certificate, signed by the manufacturer Root CA during measured boot. The verifier extracts FWID from the parsed DICE certificates and validates them against RVPS reference values.

`report_data_signature` serves a different purpose: per-session freshness binding (alias-key signs the RCAR challenge nonce). Keeping them separate means:
- FWID integrity ← boot-time Root CA signature in cert chain (already verified via chain validation)
- Session replay prevention ← per-session alias-key signature (~3ms cost)

The `measurements` field in JSON evidence is informational for logging; authoritative FWID comes from cert extension parsing.

attestation-agent/attester/src/nvidia_dpu/mod.rs
haixinJY: The TODO comment acknowledges this. For the initial implementation, `mlx5_0` covers the common single-DPU-per-host case (which is the shipping BF3 configuration). For multi-DPU scenarios, we plan to enumerate InfiniBand devices dynamically (scan `/sys/class/infiniband/` for devices with attestation capabilities). Happy to add dynamic discovery in a follow-up PR, or keep the TODO for now.

@fitzthum — would appreciate your input on whether this is acceptable for the initial merge.

attestation-agent/attester/src/nvidia_dpu/mod.rs
haixinJY: Thanks! The version field makes it extensible for future DPU generations with different DICE configurations.

attestation-agent/attester/src/nvidia_dpu/mod.rs
haixinJY: Done — added `const ARCHITECTURE_BF3: &str = "bluefield3";` and replaced all usages.

Cargo.toml
haixinJY: Done — removed `[patch.crates-io]` and switched to `kbs-types = { git = "https://github.com/confidential-containers/kbs-types", rev = "b26f44e" }` pointing to the merged official repo.

attestation-agent/attester/src/nvidia_dpu/mod.rs
Xynnn007: If the `measurements` is only a duplication of what we already have in `device_id_cert`, it's pratical to remove it, and only parse it from `device_id_cert` on the verifier side?

attestation-agent/attester/src/nvidia_dpu/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+            if measurement_bytes.len() % 48 != 0 {
+                bail!("measurement buffer length {} is not a multiple of 48", measurement_bytes.len());
+            }
+            let entry_size = 1 + 48; // layer(u8) + SHA-384 digest
+            measurement_bytes.chunks_exact(entry_size)
+                .map(|chunk| NvidiaDpuMeasurement {
+                    layer: chunk[0],
+                    digest: chunk[1..49].try_into().expect("chunk is exactly 49 bytes"),
+                })
+                .collect()
```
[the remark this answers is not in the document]
haixinJY: Good catch — there were indeed two bugs here:

1. **Validation mismatch**: the length check used `% 48` but actual entry size is `1 + 48 = 49` (layer byte + SHA-384 digest). Fixed to `% 49`.
2. **Silent read failure**: `unwrap_or_default()` would silently produce empty measurements if the file exists but cannot be read (e.g. permission denied). Fixed to propagate the error via `.context()?`.

Both fixed.

Cargo.toml
```diff
@@ -40,7 +40,7 @@ hmac = "0.13.0"
 jwt-simple = { version = "0.12", default-features = false, features = [
     "pure-rust",
 ] }
-kbs-types = "0.15.0"
+kbs-types = { git = "https://github.com/confidential-containers/kbs-types", rev = "b26f44e" }
```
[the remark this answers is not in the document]
haixinJY: The `[patch.crates-io]` approach mentioned in the early PR description was replaced with a direct git dependency pinned to the merged commit (`b26f44e` from kbs-types PR #89, merged 2026-07-20). This is intentionally temporary — once a new kbs-types version is published to crates.io that includes the `Dpu` variant, this will be switched to a standard version dependency. The PR description has been updated to reflect this.

attestation-agent/attester/src/nvidia_dpu/mod.rs
haixinJY: Good point. You're right — the verifier already extracts FWID exclusively from the DeviceID certificate's X.509 extensions (which are integrity-protected by the Root CA signature). The standalone `measurements` field is effectively unused on the verifier side.

I'll remove it in a follow-up cleanup PR to keep this one's scope stable:
- Remove `measurements` from `DpuDeviceEvidence` (attester)
- Remove the sysfs read + parsing logic
- Verifier already relies solely on cert-embedded FWID ✓

This simplifies the evidence schema and eliminates an unprotected redundant field. Thanks for the suggestion.

attestation-agent/attester/src/nvidia_dpu/mod.rs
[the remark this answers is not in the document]
haixinJY: Fixed in `c938085` — the modulo check is now `% 49` (1-byte layer + 48-byte SHA-384), consistent with `chunks_exact(49)`. Also replaced `unwrap_or_default()` with `.context()?` for proper error propagation.

Cargo.toml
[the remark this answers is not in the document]
haixinJY: Acknowledged. Will add an inline comment explaining the temporary git rev dependency in the follow-up cleanup PR (once kbs-types publishes to crates.io, this switches to a version specifier).

attestation-agent/attester/src/nvidia_dpu/mod.rs
```diff
@@ -0,0 +1,292 @@
...
+const NVIDIA_DPU_NONCE_SIZE: usize = 64;
+
+/// Architecture identifier for NVIDIA BlueField-3 DPU.
+const ARCHITECTURE_BF3: &str = "bluefield3";
+
+/// DICE alias private key path exposed by the platform attestation subsystem.
+/// Security: within DICE threat model, key availability to the measured OS is
+/// by design — firmware deviation changes CDI, yielding a different key pair.
+// TODO: evaluate hardware-bound signing interface (PSC) when available,
+// avoiding key extraction for defense-in-depth.
```
Xynnn007: what does this mean?

attestation-agent/attester/src/nvidia_dpu/mod.rs
```diff
@@ -0,0 +1,292 @@
...
+    /// Collect attestation evidence from the DPU device.
+    ///
+    /// The `report_data` parameter is used as a nonce to bind the evidence
+    /// to a specific attestation session, preventing replay attacks.
+    async fn get_evidence(&self, mut report_data: Vec<u8>) -> Result<TeeEvidence> {
+        // Reject oversize nonce; pad shorter input with zeros
+        if report_data.len() > NVIDIA_DPU_NONCE_SIZE {
+            bail!("report_data exceeds maximum size of {} bytes", NVIDIA_DPU_NONCE_SIZE);
+        }
+        report_data.resize(NVIDIA_DPU_NONCE_SIZE, 0); // pad with zeros if shorter
```
Xynnn007: we do not need the assertion here as the data will totally be signed

attestation-agent/attester/src/nvidia_dpu/mod.rs
Xynnn007: We can remove it if not needed in this PR, for both attester/verifier side

attestation-agent/attester/src/nvidia_dpu/mod.rs
haixinJY: Reworded for clarity. The DICE alias key is exposed to the measured OS by design (firmware deviation rotates CDI → key pair). PSC is the BlueField-3 hardware signing coprocessor that would allow signing without extracting the key — noted as a future defense-in-depth improvement, not a blocker for this PR.

attestation-agent/attester/src/nvidia_dpu/mod.rs
haixinJY: Fixed — removed the upper-bound assertion. ECDSA internally hashes the input so signing arbitrary-length data is safe. Only the `resize()` padding remains (BF3 nonce register is fixed-size).

attestation-agent/attester/src/nvidia_dpu/mod.rs
haixinJY: Done — removed `measurements` from both the struct and sysfs collection logic. The verifier extracts FWID exclusively from the DeviceID certificate's firmware_layers (integrity-protected by Root CA signature). Corresponding cleanup applied on the verifier side as well.

attestation-agent/attester/src/nvidia_dpu/mod.rs
```diff
@@ -0,0 +1,235 @@
...
+    /// Device architecture identifier (e.g. "bluefield3")
+    pub architecture: String,
+    /// DER-encoded DICE Alias certificate (leaf)
+    #[serde_as(as = "Base64")]
+    pub alias_cert: Vec<u8>,
+    /// DER-encoded DICE DeviceID certificate
+    #[serde_as(as = "Base64")]
+    pub device_id_cert: Vec<u8>,
+    /// Base64-encoded ECDSA P-384 signature of report_data using DICE alias private key
+    pub report_data_signature: String,
```
Xynnn007: Similarly, `#[serde_as(as = "Base64")]` can be used here, and mark the type as `Vec<u8>`

Xynnn007 (commented):
One last comment. Otherwise LGTM!

btw, you need to squash the commits. Also, do not forget to add AI helped sign-offs if AI is used according to https://github.com/confidential-containers/.github/pull/26#pullrequestreview-4741949844

attestation-agent/attester/src/nvidia_dpu/mod.rs
haixinJY: Done — `report_data_signature` now uses `#[serde_as(as = "Base64")]` with `Vec<u8>`, consistent with the cert fields. The `sign_report_data()` function returns raw signature bytes directly and serde handles the base64 encoding/decoding transparently.

Also updated the verifier side (trustee #1479) to match — removed the manual `base64::decode()` call since serde_as handles it.

attestation-agent/attester/src/nvidia_dpu/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+    async fn test_get_evidence_no_hardware() {
+        // Without NVIDIA DPU hardware, get_evidence should fail because
+        // the DICE alias private key is not available (signature is mandatory).
+        let attester = NvidiaDpuAttester::default();
+        let report_data = vec![0u8; 32];
+
+        let result = attester.get_evidence(report_data).await;
+        // On machines without DPU hardware, this should error due to missing key
+        assert!(result.is_err());
+    }
```
Xynnn007: Not sure if this unit test is useful. I prefer to remove it

Xynnn007 (approved):

attestation-agent/attester/src/nvidia_dpu/mod.rs
haixinJY: Agreed — removed. This test only asserts the trivial "no hardware → error" path and doesn't validate any meaningful logic. The remaining `test_detect_platform_no_device` (panic-freedom check) and `test_evidence_serialization` (serde roundtrip) provide better coverage.

Xynnn007 (approved):
LGTM. please squash the commits into one. Also @fitzthum PTAL

attestation-agent/attester/src/nvidia_dpu/mod.rs
fitzthum: Hmm. I think it's fine to start with one device. One thing that we should make more clear is what the requirements for the guest are. Is this path fulfilled by the guest kernel or a driver? Which driver? What version? Basically, can we give a little more detail on the configuration someone would need to actually use this.

attestation-agent/attester/src/nvidia_dpu/mod.rs
```diff
@@ -0,0 +1,227 @@
...
+
+    false
+}
+
+/// Sign report_data with DICE alias private key for freshness binding.
+/// Returns the raw signature bytes (DER-encoded ECDSA P-384).
+/// Returns error if the key is unavailable (unsigned evidence is not acceptable).
+fn sign_report_data(report_data: &[u8]) -> Result<Vec<u8>> {
+    // DICE alias private key path from NVIDIA DOCA attestation service layout.
+    // Reference: NVIDIA DOCA DICE Programming Guide, Section "Key Exposure via sysfs"
```
fitzthum: Can you provide a link to this document

attestation-agent/attester/src/nvidia_dpu/mod.rs
haixinJY: Exactly — removed. The `measurements` field was redundant since FWID is already integrity-protected inside the DeviceID certificate's X.509 extensions (signed by Root CA during measured boot). The verifier now extracts firmware identity exclusively from cert parsing.

haixinJY: Done — removed on both sides. The attester no longer collects raw measurements from sysfs, and the verifier parses FWID solely from the DICE certificate chain.

attestation-agent/attester/src/nvidia_dpu/mod.rs
haixinJY: Good point — added doc comments clarifying the requirements:

1. **Attestation attributes** (`/sys/class/infiniband/mlx5_0/device/attestation/`): provided by the `mlx5_core` kernel driver with attestation support, part of MLNX_OFED >= 24.10 or DOCA >= 2.8 BFB image.
2. **DICE alias private key** (`/sys/kernel/security/tee/dice/`): exposed by the `mlnx_bf_attestation` kernel module, also part of MLNX_OFED/DOCA.

The "guest" here is the BF3 ARM SoC Linux (the DPU OS) — the attester runs there. Standard DOCA production BFB images include attestation support by default since DOCA 2.8.

attestation-agent/attester/src/nvidia_dpu/mod.rs
haixinJY: Added link — the primary public reference is:
https://docs.nvidia.com/networking/display/dpunicattestation

This covers the DICE certificate chain structure and attestation architecture for BlueField-3. The sysfs path specifics (`/sys/kernel/security/tee/dice/`) are part of the `mlnx_bf_attestation` kernel module documented in the DOCA SDK release notes.

haixinJY:
Updated cert field comments — `alias_cert` and `device_id_cert` are DER-encoded X.509 read from sysfs and passed through as-is. Verifier side (trustee#1479) now parses them with `x509_parser`, so both ends are aligned on the wire format.

attestation-agent/attester/src/lib.rs
```diff
@@ -219,6 +224,11 @@ pub fn detect_attestable_devices() -> Vec<Tee> {
         additional_devices.push(Tee::Nvidia);
     }
 
+    #[cfg(feature = "nvidia-dpu-attester")]
+    if nvidia_dpu::detect_platform() {
+        additional_devices.push(Tee::NvidiaDpu);
+    }
```
fitzthum: So this is in the section for additional devices, but when we're running workloads on the DPU itself, what will the primary device be?

attestation-agent/attester/src/nvidia_dpu/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+/// Architecture identifier for NVIDIA BlueField-3 DPU.
+const ARCHITECTURE_BF3: &str = "bluefield3";
+
+/// DICE alias private key exposed by the BlueField platform attestation subsystem.
+///
+/// Requires: `mlnx_bf_attestation` kernel module (part of MLNX_OFED >= 24.10 / DOCA >= 2.8).
+/// This runs on the BF3 ARM SoC Linux (the "DPU OS"), not the x86 host.
+///
+/// Security model: key availability to the attested OS is inherent to DICE —
+/// any firmware change rotates the CDI, producing a different key pair.
```
fitzthum: What does CDI refer to in this context?

fitzthum (commented):
I think this is almost ready. Just two more question from me. Btw needs a rebase

attestation-agent/attester/src/lib.rs
haixinJY: This function only probes secondary devices. When the workload runs on the DPU itself, the AA runs on the DPU's Arm cores and the DPU is the primary TEE, selected via `detect_tee_type()` as `Tee::NvidiaDpu`. Added a doc comment to make this explicit.

attestation-agent/attester/src/nvidia_dpu/mod.rs
haixinJY: CDI is the Compound Device Identifier from the TCG DICE spec: each layer's CDI is derived from the parent CDI and the firmware measurement, so any firmware change rotates the CDI and all downstream keys, including this Alias key. Expanded the doc comment accordingly.

haixinJY:
Thanks @fitzthum. Addressed both comments above, rebased on main and squashed into a single commit.

attestation-agent/attester/src/lib.rs
fitzthum: That comment with DPU-specific stuff should not be added to the top of the function (which is more generic). 

But more fundamentally, if the DPU is the primary TEE, this is in the wrong place.

haixinJY: Fair point. Removed that doc comment and dropped the DPU probing from `detect_attestable_devices()` — this PR now only covers the DPU-as-primary-TEE case (AA running on the DPU's Arm cores, selected via `detect_tee_type()`). Modeling the DPU as an additional device is indeed a different question from GPUs/NICs, so I'd rather defer that path until the model discussion settles. Happy to open a tracking issue for it.

Xynnn007:
@56ht86hrd5-ai Well. I think the DPU model is a little comlicated to integrate with current kbs protocol model. There would be some discussion for whether it's primary or additional.

@fitzthum A practical way is to get attester APIs and verifier APIs in, which part is consensus to be ok I believe. But no kbs protocol parts / calling heuristic func in. Then we can have an issue to talk about this.

wdyt?

haixinJY:
@Xynnn007 That works for me. Once you and @fitzthum agree on the split, I'll trim this PR down to the attester/verifier API implementation only and drop the detection wiring. I can also open the issue for the primary-vs-additional discussion. Thanks!

fitzthum:
Ok. That is fine. PR looks good otherwise.

haixinJY:
Rebased onto main. The PR only carries the attester API implementation (the `nvidia_dpu` module and the `TryFrom<Tee>` mapping) behind the `nvidia-dpu-attester` feature — the detection wiring was already dropped in the previous push. Opened #1607 to track the primary-vs-additional discussion. Thanks @Xynnn007 @fitzthum!

Xynnn007 (approved):
LGTM. Let's see if the CI is happy

fitzthum (approved):
LGTM

Cargo.toml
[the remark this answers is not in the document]
mythi: `kbs-types v0.18.0` is now available.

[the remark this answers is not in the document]
haixinJY: Thanks @mythi! Bumped to `kbs-types = "0.18.0"` — opened a follow-up PR (#1611) in guest-components and updated the trustee PR (#1479) as well.

PR #1577, opened 2026-07-21
Add NVIDIA attester back to all-attesters

fitzthum:
When we added PPCIE support to the NVIDIA attester, we made our life a little more complicated by introducing a dependency on NVAT. Since NVAT has some dependencies which are annoying to build, we removed the nvidia attester from the `all-attesters` feature that we use in the CI. This was we wouldn't have to build NVAT from source every time. Of course, this means that we weren't running our usual checks on the nvidia attester, which is sub-optimal. 

To fix this, introduce a reusable action that installs NVAT on x86. Also, update the nvidia attester so that it only requires NVAT on x86. For now this is the only platform we support. On other platforms `detect_platform` simply returns false.

This should allow us to only ship one AA binary in the extension image btw.


attestation-agent/attester/src/nvidia/mod.rs
```diff
@@ -45,6 +53,7 @@ enum Architecture {
     LS10,
 }
 
+#[cfg(target_arch = "x86_64")]
```
Xynnn007: can we move the arch flag inside the function?

attestation-agent/attester/src/nvidia/mod.rs
```diff
@@ -109,6 +123,7 @@ impl Attester for NvAttester {
 }
 
 /// Internal helper for getting evidence from NVIDIA devices.
+#[cfg(target_arch = "x86_64")]
```
Xynnn007: ditto

attestation-agent/attester/src/lib.rs
```diff
@@ -77,6 +77,7 @@ impl TryFrom<Tee> for BoxedAttester {
             Tee::Tpm => Box::new(tpm::TpmAttester::new()?),
             #[cfg(feature = "nvidia-attester")]
             Tee::Nvidia => Box::<nvidia::NvAttester>::default(),
+            #[allow(unreachable_patterns)]
```
Xynnn007: why we need this macro?

attestation-agent/kbs_protocol/Cargo.toml
```diff
@@ -58,7 +58,7 @@ passport = []
 aa_ttrpc = ["passport", "ttrpc/async", "protos/ttrpc"]
 
 background_check = ["tokio/time", "serde_json_canonicalizer"]
-all-attesters = ["attester/all-attesters"]
+all-attesters = ["attester/all-attesters", "nvidia-attester"]
```
Xynnn007: attester/all-attesters already includes nvidia-attester

attestation-agent/attester/src/nvidia/mod.rs
```diff
@@ -4,13 +4,20 @@
...
+cfg_if::cfg_if! {
+    if #[cfg(target_arch = "x86_64")] {
+        use anyhow::anyhow;
+        use nv_attestation_sdk::{GpuEvidenceSource, Nonce, NvatSdk, SdkOptions, SwitchEvidenceSource};
+        use std::sync::OnceLock;
+        use tracing::warn;
+    }
+}
+
+#[cfg(target_arch = "x86_64")]
```
Xynnn007, line 6: if the arch flag covers many places, we can have different mod files e.g. `x86` and `others`

Xynnn007 (commented):
Looks like only x86-64 can work fine with Nvidia. Two questions:
1. if the target arch is not x86-64, can compilation work fine?
2. if compilation cannot work fine, we can avoid arch flags; if compilations can work fine, can we return error for all member function of Nvidia attester on non-x86 platforms? The returned empty slice would make no sense, but hide true evidence

.github/actions/install-nvat-sdk/action.yml
```diff
@@ -0,0 +1,43 @@
...
+        sudo apt-get update
+        sudo apt-get install -y --no-install-recommends \
+          build-essential clang cmake pkg-config libclang-dev \
+          libcurl4-openssl-dev libssl-dev libxml2-dev \
+          libxmlsec1-dev libxmlsec1-openssl
+
+        tmpdir="$(mktemp -d)"
+        git clone https://github.com/NVIDIA/attestation-sdk "${tmpdir}/attestation-sdk"
+        git -C "${tmpdir}/attestation-sdk" fetch --depth=1 origin "${NVAT_VERSION}"
+        git -C "${tmpdir}/attestation-sdk" checkout FETCH_HEAD
```
mythi:
```suggestion
              git clone --depth=1 --branch "${NVAT_VERSION}" \
                 https://github.com/NVIDIA/attestation-sdk "${tmpdir}/attestation-sdk"
```

.github/actions/install-nvat-sdk/action.yml
```diff
@@ -0,0 +1,43 @@
+name: Install NVIDIA Attestation SDK
+description: Build and install the NVIDIA Attestation SDK from source (x86_64 only)
```
mythi, line 2: `.github/workflows/coco-extension-image.yml` needs an update to use this too.

.github/actions/install-nvat-sdk/action.yml
```diff
@@ -0,0 +1,43 @@
...
+
+inputs:
+  version:
+    description: NVAT SDK version tag to checkout
+    required: false
+    default: "2026.03.02"
```
mythi, line 8: this needs manual updates when dependabot updates the crate. I wonder if we should `yq -r '.dependencies."nv-attestation-sdk".tag' Cargo.toml` to keep the version in sync automatically

fitzthum:
> If the target arch is not x86-64, can compilation work fine?

We aren't supporting nvat on non-x86 yet. We will enable with ARM in the future. This PR adjusts things so we can have the nvidia-attester feature enabled on all platforms.

> if compilation cannot work fine, we can avoid arch flags; if compilations can work fine, can we return error for all member function of Nvidia attester on non-x86 platforms? The returned empty slice would make no sense, but hide true evidence

The `get_evidence` function will never be called on non x86 since `detect_platform` always returns false. We don't want `detect_platform` to return an error since it is always called. This would take us back to where we are now. I can add a warning to non-x86 detect platform that says that the NVIDIA attester is not supported on the platform.

attestation-agent/attester/src/lib.rs
fitzthum: Compiler complains about never reaching the `_` case since we have all enum variants covered. I could put that case inside a platform-specific block, but that seems a little inelegant.

.github/actions/install-nvat-sdk/action.yml
fitzthum: I have been blocking dependabot updates to this crate for now since we haven't been building the nvidia attester in the CI. Versioning needs some work, but maybe in another PR.

attestation-agent/attester/src/nvidia/mod.rs
fitzthum: I did this at first, but it was not as clean as I had hoped.

.github/actions/install-nvat-sdk/action.yml
fitzthum: Yeah, left out any changes to extension stuff to avoid bumping into your PR. This has bigger implications with extension images as well. With this PR, we don't need to build two versions of the attestation agent (one with nvidia stuff and one without). We will simply have nvidia support on x86. I was thinking I would reconcile this in another PR.

attestation-agent/attester/src/nvidia/mod.rs
fitzthum: You want one function with a big cfg_if inside? I kind of like having it above the function.

attestation-agent/attester/src/lib.rs
Xynnn007: Ok, if the enum is exhausted, maybe we can simply remove the last `_ => xxx` line?

attestation-agent/attester/src/nvidia/mod.rs
Xynnn007: Not a big  cfg_if block. See https://github.com/Xynnn007/guest-components/commit/ba3705e0f6470671615c00c9ed885f49ea5f83fc#diff-b1f04b9105fd04870d6322c8e39c9afc048f53a8a08e666acd13b2eec3f5a45f for example

attestation-agent/attester/src/lib.rs
fitzthum: I think that will break if you don't have all the attesters enabled.

attestation-agent/attester/src/nvidia/mod.rs
fitzthum: hmm the only catch is that we will have ARM support soon too. let me think about a better name.

Xynnn007: it's ok to rename `x86_64.rs` to `supported.rs`, and use `if #[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))] {`

attestation-agent/attester/src/lib.rs
Xynnn007: Oh I think I got the cause. It's original design format issue. https://github.com/confidential-containers/guest-components/pull/1588 would help

attestation-agent/attester/src/nvidia/mod.rs
Xynnn007: same with https://github.com/confidential-containers/guest-components/pull/1577#discussion_r3642573340

attestation-agent/attester/src/nvidia/mod.rs
Xynnn007: same with https://github.com/confidential-containers/guest-components/pull/1577#discussion_r3642573340

attestation-agent/attester/src/lib.rs
mythi: `#[allow(unreachable_patterns)]` does look too bad to me. AFAUI most real build targets pick a subset of attesters anyway.

attestation-agent/attester/src/nvidia/mod.rs
fitzthum: Ok, I called it `nvat.rs` since that is the core dependency here. We may split the nvat stuff into a separate crate in the future, but this seems good for now.

attestation-agent/attester/src/lib.rs
fitzthum: Rebased on Ding's PR

attestation-agent/attester/src/nvidia/nvat.rs
```diff
@@ -0,0 +1,150 @@
...
+    LS10,
+}
+
+pub fn detect_platform() -> bool {
+    if ensure_sdk_init().is_err() {
+        warn!("NVIDIA Attestation SDK could not be initialized.");
+        return false;
+    };
+
+    match get_device_evidence(None) {
```
Xynnn007, line 54: Not related to the issue here. I put an issue for the NVRS https://github.com/NVIDIA/attestation-sdk/issues/22 to get a lighter API for detection API here

Xynnn007 (approved):
LGTM

Xynnn007:
cc @mythi

mythi (approved):
LGTM

PR #1581, opened 2026-07-22
image-rs: fix fchmodat EINVAL on kernels older than 6.6

Xynnn007:
The unpack logic sets file modes through fchmodat() with AT_EMPTY_PATH on the O_PATH fd returned by pathrs. The underlying fchmodat(2) syscall has no flags argument, so AT_EMPTY_PATH is only honored via the fchmodat2(2) syscall introduced in Linux 6.6. On older guest kernels the libc falls back and rejects AT_EMPTY_PATH with EINVAL, which breaks image pull on peer-pod guests (cloud-api-adaptor) whose guest kernel predates 6.6, while bare-metal kata guests with newer kernels are unaffected.

Resolve the inode via /proc/self/fd/<N> and call chmod(2) instead, which works on any kernel as long as procfs is mounted (always the case in the guest). chmod is used rather than fchmod because an O_PATH fd cannot be passed to fchmod directly. fchownat is left untouched since that syscall natively supports AT_EMPTY_PATH on all kernels.

Related to https://github.com/confidential-containers/cloud-api-adaptor/pull/3202

### DIFF
--- a/image-rs/src/stream/unpack.rs
+++ b/image-rs/src/stream/unpack.rs
@@ -497,17 +497,20 @@ fn fchown_at(fd: BorrowedFd<'_>, uid: u32, gid: u32) -> Result<()> {
 }
 
 /// Set mode on the inode referenced by an `O_PATH` fd (`Root` / `Handle` from pathrs).
+///
+/// The underlying `fchmodat(2)` syscall has no flags argument, so `AT_EMPTY_PATH`
+/// is only honored through the `fchmodat2(2)` syscall introduced in Linux 6.6.
+/// On older guest kernels the libc falls back and rejects `AT_EMPTY_PATH` with
+/// `EINVAL`. `chmod(2)` cannot be issued on an `O_PATH` fd directly either, so
+/// resolve the inode via `/proc/self/fd/<N>`, which works on any kernel as long
+/// as procfs is mounted (always the case in the guest).
 fn fchmodat(fd: BorrowedFd<'_>, mode: u32) -> Result<()> {
-    let ret = unsafe {
-        libc::fchmodat(
-            fd.as_raw_fd(),
-            c"".as_ptr(),
-            mode as libc::mode_t,
-            libc::AT_EMPTY_PATH,
-        )
-    };
+    let proc_path = format!("/proc/self/fd/{}", fd.as_raw_fd());
+    let proc_path =
+        CString::new(proc_path).map_err(|_| anyhow!("procfs path contains null byte"))?;
+    let ret = unsafe { libc::chmod(proc_path.as_ptr(), mode as libc::mode_t) };
     if ret != 0 {
-        bail!("failed to fchmodat: {:?}", io::Error::last_os_error());
+        bail!("failed to chmod: {:?}", io::Error::last_os_error());
     }
     Ok(())
 }


Xynnn007:
cc @stevenhorsman

stevenhorsman (approved):
Thanks for the update. Let's hope is resolves the problems and isn't a red herring 🤞

mythi:
> Thanks for the update. Let's hope is resolves the problems and isn't a red herring 🤞

Is there a way to get confirmation on this before we merge? For some reason this hasn't triggered on OCP which uses an old kernel too.

Xynnn007:
I got a machine
```
$ uname -r
5.15.0-181-generic

$ cargo test -p image-rs --lib stream::unpack::tests::test_unpack -- --nocapture --test-threads=1
test stream::unpack::tests::test_unpack ... 
thread 'stream::unpack::tests::test_unpack' panicked at image-rs/src/stream/unpack.rs:673:9:
assertion failed: unpack(data.as_slice(), destination).await.is_ok()
FAILED
test stream::unpack::tests::test_unpack_hardlink_tar_cases::case_1_absolute_link_target ... 
thread 'stream::unpack::tests::test_unpack_hardlink_tar_cases::case_1_absolute_link_target' panicked at image-rs/src/stream/unpack.rs:792:17:
called `Result::unwrap()` on an `Err` value: SetPermissionsFailed { source: failed to fchmodat: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }, path: "/tmp/.tmpenkCZU/dest/etc/os-release" }
FAILED
test stream::unpack::tests::test_unpack_hardlink_tar_cases::case_2_relative_escape_link_target ... ok
test stream::unpack::tests::test_unpack_layer_root_mtime ... 
thread 'stream::unpack::tests::test_unpack_layer_root_mtime' panicked at image-rs/src/stream/unpack.rs:743:53:
called `Result::unwrap()` on an `Err` value: SetPermissionsFailed { source: failed to set permissions for layer root

Caused by:
    failed to fchmodat: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }, path: "/tmp/.tmpbaB3GC/layer" }
FAILED

failures:

failures:
    stream::unpack::tests::test_unpack
    stream::unpack::tests::test_unpack_hardlink_tar_cases::case_1_absolute_link_target
    stream::unpack::tests::test_unpack_layer_root_mtime

test result: FAILED. 1 passed; 3 failed; 0 ignored; 0 measured; 74 filtered out; finished in 0.01s
```

while, the CI passes for having a higher version kernel

Xynnn007:
With the PR. 
```
$ cargo test -p image-rs --lib stream::unpack::tests::test_unpack -- --nocapture --test-threads=1

   Compiling image-rs v0.1.0 (/root/guest-components/image-rs)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 2.95s
     Running unittests src/lib.rs (target/debug/deps/image_rs-965aa7eef86f0a0b)

running 4 tests
test stream::unpack::tests::test_unpack ... ok
test stream::unpack::tests::test_unpack_hardlink_tar_cases::case_1_absolute_link_target ... ok
test stream::unpack::tests::test_unpack_hardlink_tar_cases::case_2_relative_escape_link_target ... ok
test stream::unpack::tests::test_unpack_layer_root_mtime ... ok

test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 74 filtered out; finished in 0.01s
```

fitzthum (approved):
LGTM

PR #1586, opened 2026-07-23
image-rs: make mtime restoration best-effort during unpack

### LINKED ISSUE #1585: kata-agent inside the podvm fails to unpack the image layer
### Describe the bug

Tried bringing up a confidential container via peer-pods approach, with an encrypted workload/image and the confidential container failed to start with a ImagePullBackOff error. 

Looks like the changes discussed in #1473 and in PR #1474 is causing this issue.

### How to reproduce

Create a confidential container with the new guest components version v0.21.0 via the peer pod approach with an encrypted workload with TEE=amd

### CoCo version information

guest-components v0.21.0

None

```shell
failed to create containerd task: failed to create shim task: rpc status: 
Status { code: INTERNAL, message: "[CDH] [ERROR]: 
Image Pull error: Failed to pull image <image_name> from all mirror/mapping 
locations or original location: image: <image_name> , error: Errors happened 
when pulling image: Failed to decode layer data stream: Failed to unpack layer: 
Failed to set permissions for path: /run/kata-containers/image/layers/0/etc/group",
details: [], special_fields: SpecialFields { unknown_fields: 
UnknownFields { fields: None }, cached_size: CachedSize { size: 0 } } }
```

gaganhr94:
Fixes #1585

### DIFF
--- a/image-rs/src/stream/unpack.rs
+++ b/image-rs/src/stream/unpack.rs
@@ -395,12 +395,10 @@ pub async fn unpack<R: AsyncRead + Unpin>(input: R, layer_dir: &Path) -> UnpackR
                 }
             })?;
             let ft = FileTime::from_unix_time(mtime, 0);
-            filetime::set_symlink_file_times(layer_dir.join(entry_path), ft, ft).map_err(
-                |source| UnpackError::SetMTimeFailed {
-                    source: source.into(),
-                    path: err_path.to_string(),
-                },
-            )?;
+            set_mtime_best_effort(
+                filetime::set_symlink_file_times(layer_dir.join(entry_path), ft, ft),
+                &err_path,
+            );
         } else if kind.is_hard_link() {
             let handle = layer_root
                 .resolve(&entry_path)
@@ -412,12 +410,10 @@ pub async fn unpack<R: AsyncRead + Unpin>(input: R, layer_dir: &Path) -> UnpackR
                 }
             })?;
             let ft = FileTime::from_unix_time(mtime, 0);
-            filetime::set_file_times(layer_dir.join(entry_path), ft, ft).map_err(|source| {
-                UnpackError::SetMTimeFailed {
-                    source: source.into(),
-                    path: err_path.to_string(),
-                }
-            })?;
+            set_mtime_best_effort(
+                filetime::set_file_times(layer_dir.join(entry_path), ft, ft),
+                &err_path,
+            );
         } else {
             let handle = layer_root
                 .resolve(&entry_path)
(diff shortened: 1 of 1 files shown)


image-rs/src/stream/unpack.rs
```diff
@@ -465,15 +457,26 @@ pub async fn unpack<R: AsyncRead + Unpin>(input: R, layer_dir: &Path) -> UnpackR
...
 
     Ok(())
 }
 
+/// Best-effort application of a file's mtime.
+///
+/// mtime is cosmetic metadata: unlike uid/gid and mode it is neither security- nor
+/// integrity-relevant for a container rootfs. Some layers carry files whose mtime the
+/// kernel or filesystem refuses to set (read-only inodes, timestamps out of range, a
+/// symlink whose target rejects the change). Matching the behaviour of `tar`,
```
Xynnn007: Could you have a link to behavior of tar, containerd and umoci?

I get `tar-rs` and it will raise error like here https://docs.rs/tar/0.4.46/src/tar/entry.rs.html#587-590

gaganhr94: @Xynnn007, got a reference for umoci - https://github.com/opencontainers/umoci/blob/f5d1219acaf67127ebacf6306776d3ff465735ea/oci/layer/unpack.go#L223

gaganhr94: Reference for containerd - https://github.com/containerd/containerd/blob/ffcf715dfa5e2c7ee3c9001bcfc27fa0565d376b/pkg/archive/tar.go#L322

gaganhr94: If I understand it right, should I add these links to the comments ?

Xynnn007: Yes - please do it

Xynnn007: I did not get the logic that shows "best-effort". instead, both will raise an error?

gaganhr94: - mtime → just logs a warning, keeps going (best-effort), treated as non-fatal
- uid/gid, mode, reading the header → use ?, so they abort the pull

Xynnn007: Oh, I meant that both umoci and containerd logic you shared, do not follow the "best-effort" logic. Instead, they directly return error.

gaganhr94: Agh, the comment was probably not very clear. I wanted to convey that umoci and containerd logic handle the cases for mtime not set. This is the reason I wanted to give in order to change the case of mtime not being set to a warning instead of throwing an error. 

I agree that  umoci and containerd have no references to best effort logic. They just handle the case of mtime not being set. That is what I tried to convey through those comments.

Xynnn007: Ok. So a question to me is that 
1. umoci and containerd would bring out error w/o continuing
2. I assume containerd/umoci could successfully pull the image
3. image-rs now follow almost the same logic as containerd/umoci.

why image-rs still fails? So I left [message](https://github.com/confidential-containers/guest-components/issues/1585#issuecomment-5066674669) for a test image. Maybe you can remove the sensitive parts of the image and generate a simple test one. Then we can get better knowledge about what actually happens.

The current PR just ignore all errors that mtime meets

fitzthum: yeah this is a little surprising. i don't think this behavior changed with our recent reshuffle of image-rs, so i'm not sure we've seen this case before. what is your target filesystem?

PR #1587, opened 2026-07-24
cdh: make the AA socket path configurable

kimullaa:
AA allows changing its socket file path via the `--attestation-sock` argument. However, CDH currently connects to the fixed path
`unix:///run/confidential-containers/attestation-agent/attestation-agent.sock`.

This patch allows the socket path to be passed through the configuration file, so that CDH can connect to an arbitrary AA socket path.


Xynnn007 (approved):
LGTM. Thanks! @kimullaa

mythi (approved):
LGTM

fitzthum (approved):
LGTM

PR #1588, opened 2026-07-24
Attester | Fix `try_from<Tee> for BoxedAttester`

Xynnn007:
Original feature flag applies on each enum item. Once we enable all attester features, there will be a lint error to cry on the last `_ => xxx` line.

This patch applies feature flag inside the enum arm, thus no lint error will happen and also a clearer error when a specific attester feature is not enabled.


fitzthum (approved):
LGTM

Issue #1276, opened 2026-01-23
remove online_sev_kbc kbc and ehsm+kbs/sev confidential-data-hub kms plugins


mythi:
- SEV(-ES) support was deprecated in CoCo 0.13.0 and removed in 0.14.0
- `bincode` dependency used by `sev` crate is unmaintained and gives a RUSTSEC finding

With those reason, the issue is to remove remaining sev related content from guest-components.

Summary discussion related to `bincode` in https://github.com/confidential-containers/guest-components/pull/1247#issuecomment-3788852775

PR #1602, opened 2026-07-28
workspace: migrate a subset of crates to Rust edition 2024

### LINKED ISSUE #1599: Move to rust 2024 edition
With the release of Rust 1.85.0,  Rust 2024 was announced. All our crates here follow 2021 but the question is, are we OK to move to 2024? Many of the dependencies we use in our crates here already set MSRV to 1.88.0.

My proposal is we start moving to 2024 in _small_ PRs either by making 2024 available workspace level and then members opting in to that or just crates bumping their own editions. There will be fmt/clippy/fix (suggesting to move to 2024 style too) changes so for example per crate PRs would help to keep changes small.

The biggest impact will be around `env::set|remove_var()` which are declared unsafe. Can we find ways to get rid of some of those.

This issue could be used as the top-level tracker of the progress.

Progress:
- #1651 
- #1602 
- #1687 
- #1689 
- image-rs
- confidential-data-hub

Xynnn007:
Bump `edition` from 2021 to 2024 for all 15 workspace members and adapt the code to the changes that the new edition mandates:

- `std::env::set_var`/`remove_var` are now `unsafe`, so the affected call sites (all in CDH config handling and tests) are wrapped in `unsafe` blocks. These run before any additional threads are spawned, which is the safety condition the new signature makes explicit.
- Nested `if`/`if let` chains are collapsed into let chains, which are stabilized in this edition.
- The whole workspace is reformatted with the edition's rustfmt style, which mainly reorders `use` groups so that types sort before functions. This is required to keep `cargo fmt --check` in CI green.

No functional change intended; the migration is purely mechanical.

Fixes: #1599

In later PR, I will figure out ways to avoid unsafe blocks.

### DIFF
--- a/api-server-rest/Cargo.toml
+++ b/api-server-rest/Cargo.toml
@@ -2,7 +2,7 @@
 name = "api-server-rest"
 version = "0.1.0"
 authors = ["The Confidential Container Authors"]
-edition = "2021"
+edition = "2024"
 license = "Apache-2.0"
 
 [dependencies]
--- a/api-server-rest/src/client/aa.rs
+++ b/api-server-rest/src/client/aa.rs
@@ -11,8 +11,8 @@ use protos::ttrpc::aa::attestation_agent::{
 use protos::ttrpc::aa::attestation_agent_ttrpc::AttestationAgentServiceClient;
 use serde::Deserialize;
 
-use crate::client::ttrpc_client::CachedTtrpcClient;
 use crate::TTRPC_TIMEOUT;
+use crate::client::ttrpc_client::CachedTtrpcClient;
 
 /// ROOT path for Confidential Data Hub API
 pub const AA_ROOT: &str = "/aa";
@@ -123,8 +123,12 @@ impl AAClient {
 
         let res = match res {
             0 => "{\"success\":\"true\"}",
-            1 => "{\"success\":\"false\",\"message\": \"The platform does not support runtime measurement\"}",
-            2 => "{\"success\":\"false\",\"message\": \"Attestation Agent does not enable eventlog recording\"}",
+            1 => {
+                "{\"success\":\"false\",\"message\": \"The platform does not support runtime measurement\"}"
+            }
+            2 => {
+                "{\"success\":\"false\",\"message\": \"Attestation Agent does not enable eventlog recording\"}"
+            }
             _ => "{\"success\":\"false\",\"message\": \"Unknown runtime measurement result\"}",
         };
 
--- a/api-server-rest/src/client/cdh.rs
+++ b/api-server-rest/src/client/cdh.rs
@@ -3,8 +3,8 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-use crate::client::ttrpc_client::CachedTtrpcClient;
 use crate::TTRPC_TIMEOUT;
+use crate::client::ttrpc_client::CachedTtrpcClient;
 use anyhow::*;
 use protos::ttrpc::cdh::api::GetResourceRequest;
 use protos::ttrpc::cdh::api_ttrpc::GetResourceServiceClient;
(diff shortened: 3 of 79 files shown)


mkulke (commented):
this is pretty large, I suppose it was created automatically by clippy and cargo fmt? Can you provide the commands you used for the transformation?

also, I think it would be good to conclude the discussion in the issue about unsafe set_env() before this is merged

Xynnn007:
> this is pretty large, I suppose it was created automatically by clippy and cargo fmt? Can you provide the commands you used for the transformation?

Sure. cargo fmt is a large part. cargo clippy are mostly manual changes.

Xynnn007:
Ok. Let me firstly skip the `env` crates in this PR.

mythi:
Thanks @Xynnn007 for getting this started so quickly.

The RFC suggested to split the work in smaller PRs to help reviews and avoid:

> this is pretty large

With that said, I don't mind keeping these bundled here but two suggestions still:

- could we split the clippy fixes to their own PR similar to how we do with rust stable clippy fixes
- if we have the green light to drop sev/ehsm related (#1276) code, that could be done before this PR because it means at least one crate less to migrate

Xynnn007:
@mythi Thanks for the suggestions. Let me understand

> could we split the clippy fixes to their own PR similar to how we do with rust stable clippy fixes

Do you mean that we have two PRs. One for version updation, `cargo fmt`; and the other for `cargo clippy`?
If so, the first PR's CI will not pass for clippy warns.


> * if we have the green light to drop sev/ehsm related ([remove online_sev_kbc kbc and ehsm+kbs/sev confidential-data-hub kms plugins #1276](https://github.com/confidential-containers/guest-components/issues/1276)) code, that could be done before this PR because it means at least one crate less to migrate

I am ok to get rid of it. But this PR is not a blocker for that. How about getting this merged first and then we can start getting rid of them?

mythi:
> @mythi Thanks for the suggestions. Let me understand
> 
> > could we split the clippy fixes to their own PR similar to how we do with rust stable clippy fixes
> 
> Do you mean that we have two PRs. One for version updation, `cargo fmt`; and the other for `cargo clippy`? If so, the first PR's CI will not pass for clippy warns.

Yes, two PRs to reduce changes in this PR with the clippy part merged first. As we have seen with rust stable clippy fixes, we can take the suggestions in advance. 6c4de3f or https://github.com/confidential-containers/guest-components/pull/1602/changes/f0122d9121dbe999d5e5a391b15e870c9bf2ef63#diff-6df8544f201357fb78cf3bccd57d922b15f556d6eda709e103ab0c987eb570efL179 are not strictly 2024 related. The latter for example is just moved to another clippy category but the linter is already there, AFAUI.

> > * if we have the green light to drop sev/ehsm related ([remove online_sev_kbc kbc and ehsm+kbs/sev confidential-data-hub kms plugins #1276](https://github.com/confidential-containers/guest-components/issues/1276)) code, that could be done before this PR because it means at least one crate less to migrate
> 
> I am ok to get rid of it. But this PR is not a blocker for that. How about getting this merged first and then we can start getting rid of them?

Right, they are independent but the case is same as the clippy work: can we do preparation steps to minimize the changeset in 2024 transition. I guess it's fine to go with this PR first but at least 701cb9e does not seem necessary given the removal plans.

Xynnn007:
https://github.com/confidential-containers/guest-components/pull/1651

mythi (approved):
Thanks 

note that 1d16ed0aac32c91d6aedb8c05801e14a2406a268 commit message is slightly outdated.

Xynnn007:
Updated the commit messages.

mythi (approved):

fitzthum (approved):
LGTM

interesting that there are so many formatting changes between editions. Apparently the alphabet has changed

PR #1604, opened 2026-07-28
aa: let the KBC select attestation policies by id

Xynnn007:
Trustee gained a policy_id_map, with which an administrator maps an id that a client names in the RCAR Request to the Attestation Service policies that evaluate its evidence. Give the KBC a way to name such an id, so that a deployment can have different guests evaluated against different policies without running separate KBS instances.

kbs_protocol carries the id from `KbsClientBuilder::set_policy_selector` into the `extra-params` of the auth Request. AA takes it from the new `attestation_policy_selector ` field of the `[token_configs.kbs]` section, which is empty by default.

An empty id is not sent at all: KBS falls back to a default policy for a Request without an id, but rejects the handshake of a client that sends an id the deployment does not declare. Configuring an id therefore only works against a KBS that maps it, which is why it stays opt-in and why the empty string, rather than an `Option`, expresses "not configured" in the config.

See also https://github.com/confidential-containers/trustee/pull/1521

### DIFF
--- a/attestation-agent/attestation-agent/config.example.toml
+++ b/attestation-agent/attestation-agent/config.example.toml
@@ -6,6 +6,10 @@ url = "http://127.0.0.1:8000"
 [token_configs.kbs]
 url = "https://127.0.0.1:8080"
 tee_key_algorithm = "ECDH-ES+A256KW-P256"
+# The hint that KBS resolves to the attestation policies that evaluate this guest's
+# evidence. Empty by default, which leaves the selection to KBS. The accepted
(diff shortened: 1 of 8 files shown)


Xynnn007:
Related to https://github.com/confidential-containers/trustee/issues/282

Xynnn007:
cc @fitzthum this is the parity one

fitzthum (approved):
Looks ok. Would it be more clear to call it `attestation_policy_selector`?

Xynnn007:
Yes

PR #1603, opened 2026-07-28
attester: implement get_runtime_measurement for az-snp-vtpm

josegarvayo-oc:
The `az-snp-vtpm` attester implemented `extend_runtime_measurement`, `pcr_to_ccmr`,
and `ccel_hash_algorithm`, but not the read counterpart `get_runtime_measurement`
(left as a TODO pointing at this issue). The AAEL record path calls it on every
event write and during crash recovery, so it failed unconditionally on this
platform. This adds it, reading the current SHA-256 PCR bank via `vtpm::get_quote`.

## Testing
- cargo build / test / clippy -D warnings / fmt --check  (attester, az-snp-vtpm-attester feature)

Fixes #1201

Issue #1551, opened 2026-07-13
[CC Eventlog] Use a well-defined struct than String


Xynnn007:
According to @mkulke , The `String` of `cc_eventlog` in TDX/CSV confuses. A more formated struct like `CCEventLog` with members e.g. `Vec<EventEntry>` would make a lot more sense. This could explicitly tell future programmers the eventlogs are well formated.

PR #1605, opened 2026-07-28
Azsnp vtpm eventlog

josegarvayo-oc:
Completes the az-snp-vtpm AAEL path on top of https://github.com/confidential-containers/guest-components/pull/1603

  1. Add `cc_eventlog: Option<String>` to the az-snp-vtpm evidence, populated from
     `read_eventlog()` (mirrors the TDX/CSV attesters). Without it the recorded AAEL
     never reaches the verifier. (using current String representation, which might be change for a typed representation in issue #1551)

  2. Declare SHA-256 in the fixed `EL_HEADER` (used when no CCEL is present, e.g.
     az-snp-vtpm) so the AS eventlog parser can size SHA-256 runtime events instead of
     failing with "Missing digest size for algorithm" on the first one.

  ## Testing
  - cargo build / test / clippy -D warnings / fmt --check (attester, az-snp-vtpm-attester)
  - new unit tests: cc_eventlog serde round-trip; EL_HEADER byte-layout / SHA-256 declaration

### DIFF
--- a/attestation-agent/attester/src/az_snp_vtpm/mod.rs
+++ b/attestation-agent/attester/src/az_snp_vtpm/mod.rs
@@ -4,6 +4,7 @@
 //
 
 use super::{Attester, InitDataResult, TeeEvidence};
+use crate::utils::read_eventlog;
 use anyhow::{bail, Context, Result};
 use az_snp_vtpm::{imds, is_snp_cvm, vtpm};
 use azure_tpm::{Tpm, TpmCommandExt};
@@ -70,6 +71,9 @@ impl From<vtpm::Quote> for TpmQuote {
 /// - `hcl_report` - Hardware Compatibility Layer report containing the
 ///    Hardware attestation report from the AMD SEV-SNP platform
 /// - `vcek` - Versioned Chip Endorsement Key certificate (DER-encoded)
+/// - `cc_eventlog` - Base64-encoded runtime eventlog (AAEL in TCG2 encoding),
+///    or `None` when no runtime eventlog has been recorded. Mirrors the shape
+///    used by the TDX attester so the verifier can consume it uniformly.
 #[serde_as]
 #[derive(Serialize, Deserialize)]
 struct Evidence {
@@ -79,6 +83,7 @@ struct Evidence {
     hcl_report: Vec<u8>,
     #[serde_as(as = "UrlSafeBase64")]
     vcek: Vec<u8>,
+    cc_eventlog: Option<String>,
 }
 
 /// Convert a PEM-encoded certificate to DER format
@@ -92,7 +97,10 @@ fn pem_to_der(pem: &str) -> Result<Vec<u8>> {
     Ok(der)
 }
 
-fn get_evidence_sync(report_data: &[u8]) -> anyhow::Result<TeeEvidence> {
+fn get_evidence_sync(
+    report_data: &[u8],
+    cc_eventlog: Option<String>,
+) -> anyhow::Result<TeeEvidence> {
     let hcl_report = vtpm::get_report_with_report_data(report_data)?;
     let tpm_quote = vtpm::get_quote(report_data)?.into();
     let certs = imds::get_certs()?;
@@ -103,6 +111,7 @@ fn get_evidence_sync(report_data: &[u8]) -> anyhow::Result<TeeEvidence> {
         tpm_quote,
         hcl_report,
         vcek,
+        cc_eventlog,
     };
 
     Ok(serde_json::to_value(&evidence)?)
@@ -111,7 +120,10 @@ fn get_evidence_sync(report_data: &[u8]) -> anyhow::Result<TeeEvidence> {
 #[async_trait::async_trait]
 impl Attester for AzSnpVtpmAttester {
     async fn get_evidence(&self, report_data: Vec<u8>) -> anyhow::Result<TeeEvidence> {
-        spawn_blocking(move || get_evidence_sync(&report_data)).await?
+        // Read the runtime eventlog (AAEL) before entering the blocking section:
+        // `read_eventlog` is async, whereas the vTPM calls below are blocking.
+        let cc_eventlog = read_eventlog().await?;
+        spawn_blocking(move || get_evidence_sync(&report_data, cc_eventlog)).await?
     }
 
     fn supports_runtime_measurement(&self) -> bool {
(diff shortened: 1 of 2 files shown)


mkulke (commented):
thamks, some questions:

- cc_eventlog, is this signed by some entity or merely informational, like a tpm2 eventlog, meant for replaying events and comparing the result with a PCR from a signed quote?
- would we need to bump the version of the Evidence or is this a non-breaking change, as it an Option field?
- It looks like we always populate it, should it be an Option field then?
- Is this orthogonal to a tpm2_eventlog or similar/overlapping/redundant?

attestation-agent/attester/src/az_snp_vtpm/mod.rs
```diff
@@ -258,4 +270,43 @@ MIIB0zCCAXqgAwIBAgIJALg0
...
+            },
+            hcl_report: vec![],
+            vcek: vec![],
+            cc_eventlog: None,
+        };
+        let value_none = serde_json::to_value(&evidence_none).unwrap();
+        assert!(value_none.get("cc_eventlog").unwrap().is_null());
+        let back_none: Evidence = serde_json::from_value(value_none).unwrap();
+        assert_eq!(back_none.cc_eventlog, None);
+    }
```
Xynnn007: Looks like the roundtrip does not make much sense - it's just a test for default behavior for `serde_json` as we do not implement any customized serde/de-ser. Thus ok to remove them

attestation-agent/attester/src/utils.rs
```diff
@@ -63,19 +63,25 @@ pub fn validate_and_pad_data(data: Vec<u8>, expected_size: usize) -> Result<Vec<
...
-pub const EL_HEADER: [u8; 69] = [
+pub const EL_HEADER: [u8; 73] = [
     0x00, 0x00, 0x00, 0x00, 0x03, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
-    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x25, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x29, 0x00, 0x00, 0x00,
     0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
-    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x02, 0x00, 0x00, 0x00, 0x0C, 0x00, 0x30, 0x00,
-    0x12, 0x00, 0x20, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x03, 0x00, 0x00, 0x00, 0x0B, 0x00, 0x20, 0x00,
+    0x0C, 0x00, 0x30, 0x00, 0x12, 0x00, 0x20, 0x00, 0x00,
```
Xynnn007, line 84: Comment: Manually checked and this is ok. Means adds `sha256` and the digest length is `0x20` bytes.

Xynnn007 (commented):
Thanks. Some code nits. There must be some more discussion about the design and integration. Please check with @mkulke

josegarvayo-oc:
Hi @mkulke , thanks for the review; we discussed some of this in this this thread (just for context): https://cloud-native.slack.com/archives/C039JSH0807/p1783419518076759, let me answer what is my understanding inline:

> thamks, some questions:
> 
> * cc_eventlog, is this signed by some entity or merely informational, like a tpm2 eventlog, meant for replaying events and comparing the result with a PCR from a signed quote?

Merely informational, exactly like a TPM2 eventlog; the verifier replays it to reconstruct the expected PCR value and compare against the signed report; this report is the root of trust and the eventlog is the audit trail for the PR value.

> * would we need to bump the version of the Evidence or is this a non-breaking change, as it an Option field?
Since it is an Option string, there is no need for bumping version

> * It looks like we always populate it, should it be an Option field then?
Not always, in a real scenario (on Azure CVM) it is always present, but in tests without Azure firmware or non CVM is not; probably is too strict to force it to be always present. In the slack thread Ding suggested to model it as Option<String>: "actually the only thing is to add another field cc_eventlog: Option<String> in evidence format, and a sentence let cc_eventlog = read_eventlog().await?;"

> * Is this orthogonal to a tpm2_eventlog or similar/overlapping/redundant?
There is some discussion about this in the thread (https://cloud-native.slack.com/archives/C039JSH0807/p1783937190610979?thread_ts=1783419518.076759&cid=C039JSH0807), my understanding is this is similar to a tpm2_eventlog:  the cc_eventlog covers the same functionality (audit trail to replay and verify the PCRs); and the same thing is used in TDX for RTMRs.  But with cc_eventlogs, in the verifier side the replay logic can be the same regardless of PCRs or RTMRs

attestation-agent/attester/src/utils.rs
josegarvayo-oc: thanks

attestation-agent/attester/src/az_snp_vtpm/mod.rs
josegarvayo-oc: Thanks, will remove it; does it make sense to replace it with  a unit test for testing trim_ccel() to check parsing logic with a header including the new SHA-256 with the 0x20 bytes length digest?

Xynnn007: That makes sense!

josegarvayo-oc:
### runtime validation on a real Azure SNP CVM

I took this end-to-end on a live `az-snp-vtpm` peer-pod (Azure `NCC40ads_H100_v5`, SNP + vTPM) to confirm the runtime measurement path actually works, and hit a platform issue worth flagging here :

**Setup.** Enabled AA eventlog recording (`[eventlog_config] enable_eventlog = true`) and drove `extend_runtime_measurement` two ways: (a) via the measured image pull (CDH `PullImage` → AA), and (b) directly via `api-server-rest`'s `POST /aa/aael`.

**Result with the default `init_pcr = 17`:** every extend fails. AA returns the generic `AA extend runtime measurement failed`; because this happens during the measured image pull inside `create_container`, it aborts pod startup entirely:

```
[CDH] [ERROR]: Attestation Agent client error: failed to extend runtime measurement:
  "[ERROR:attestation-agent] AA extend runtime measurement failed"
  at kata_agent::storage::image_pull_handler::ImagePullHandler::create_device
```

**Root cause.** PCR 17 is a DRTM register. On the Azure vTPM a `TPM2_PCR_Extend` on PCR 17 from the guest (locality 0) is rejected — DRTM PCRs are only extendable at locality 4. 

Repointing AA at a locality-0-extendable register — `init_pcr = 23` — makes the whole path succeed:

- `POST /aa/aael` → `{"success":"true"}`;
- the image-pull measurement is recorded automatically — the eventlog contains `github.com/confidential-containers PullImage {"image": "…/alpine:latest", "digest": "sha256:79ff19e9…"}`;
- evidence now carries a non-null `cc_eventlog` that replays cleanly.

Same binary, same config, only the target PCR changed — which isolates the cause to PCR 17's locality restriction, not the read/extend code (both of which are correct).

**Not a blocker.** Since there's an initdata config workaround today — setting `init_pcr = 23` in the AA config (`eventlog_config`) overrides the default without any code change. On the verifier side the target PCR index is carried in the eventlog itself (each TCG2 event records its measurement register), so the verifier can replay each entry and verify against the correct PCR index directly from the log, rather than assuming a fixed register.

josegarvayo-oc:
@mkulke I've replied to your questions above and updated the PR with the requested changes (removed the serde roundtrip test, added trim_ccel unit tests). Would appreciate your review when you get a chance!

fitzthum:
Needs rebase. Maybe we can get this one unstuck.

mkulke (approved):

josegarvayo-oc:
@fitzthum @Xynnn007 , since mkulke approved, and I've just rebased the PR, could you take another look when you have a chance and merge it?

PR #1606, opened 2026-07-28
ci: add zizmor GitHub Actions security scanning workflow

siddu-os:
Why
GitHub Actions workflows are a common attack surface — misconfigurations such as credential persistence, over-broad permissions, and archived action refs can lead to secret exfiltration or supply-chain compromise. zizmor is a static analysis tool purpose-built to audit workflow files for exactly these classes of issues.

What
Adds .github/workflows/zizmor.yaml which runs zizmor in regular persona mode on every push to main, every pull request, and on a weekly schedule (Mondays 03:00 UTC).

Key properties of the new workflow:
permissions: {} — no token permissions granted to the job itself
concurrency group — cancels superseded runs on the same PR/ref
All actions pinned to full commit SHAs (same convention already used across this repo)
advanced-security: false — no GHSA/CodeQL upload required; inline PR annotations are used instead
persona: regular — standard ruleset

Also fixes a pre-existing security finding surfaced by zizmor:
.github/workflows/secret_cli_release.yml — cache-poisoning (high): added cache: false to the actions-rust-lang/setup-rust-toolchain step. This workflow triggers on pull_request and publishes runtime artifacts, making it vulnerable to cache poisoning without this flag.

Local validation
zizmor was run locally against the full repo before this PR:
```
$ zizmor .github/workflows/
No findings to report. Good job! (92 suppressed)
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,8 @@ updates:
     directory: "/"
     schedule:
       interval: daily
+    cooldown:
+      default-days: 7
     open-pull-requests-limit: 3
     allow:
     - dependency-type: direct
@@ -12,8 +14,12 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    cooldown:
+      default-days: 7
 
   - package-ecosystem: docker
     directory: /image-rs/test_data
     schedule:
       interval: daily
+    cooldown:
+      default-days: 7
--- a/.github/workflows/coco-extension-image.yml
+++ b/.github/workflows/coco-extension-image.yml
@@ -178,7 +178,7 @@ jobs:
     # manifests in the publish-manifests job below.
     - name: Log in to the Container registry
       if: ${{ github.event_name != 'pull_request' }}
-      uses: docker/login-action@abd2ef45e78c5afb21d64d4ca52ee8550d9572c7 # v3.6.0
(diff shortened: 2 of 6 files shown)


Xynnn007 (approved):
LGTM.

fitzthum:
Looks like some fixes are needed in the workflows.

siddu-os:
> Looks like some fixes are needed in the workflows.

Thanks for the review.
Fixed — the zizmor CI (running with online audits and v1.26.1) surfaced a few pre-existing findings not caught by my local offline scan:

1. dependabot.yml: added cooldown: default-days: 7 to all three ecosystems (cargo, github-actions, docker) — dependabot-cooldown finding
2. coco-extension-image.yml: corrected docker/login-action SHA from abd2ef45 → 5e57cd11 to match the actual v3.6.0 tag commit — hash-comment-mismatch finding
3. dependency-review.yml: fixed version comment v4 → v5.0.0 to match the pinned SHA — hash-comment-mismatch finding
4. shellcheck.yaml: fixed version comment v2.0.0 → 2.0.0 to match ludeeus/action-shellcheck tag naming — hash-comment-mismatch finding

Xynnn007 (approved):
Thanks for the fix!

fitzthum (approved):
LGTM

PR #1626, opened 2026-08-04
cdh/kms/offline-fs-kbc: add self defined file paths to read keys

Xynnn007:
In some unit tests, we need to test with offline fs kbc but do not want to change current environment. Thus we can use this patch to create a tempfile and use it as key resource. After the test, remove the test file.


esposem (approved):
LGTM thanks!

bpradipt (approved):

PR #1633, opened 2026-08-07
api-server-rest: log internal server errors

kimullaa:
Log internal errors before generating HTTP 500 responses


fitzthum (approved):
LGTM

I'm not sure if the ASR log is actually propagated anywhere

Xynnn007 (approved):

PR #1635, opened 2026-08-07
workflows: allow manually triggering secret-cli workflow

fitzthum:
This will allow us to trigger the release flow manually from GH. This way we don't have to wait for the next release to trigger the workflow.

### DIFF
--- a/.github/workflows/secret_cli_release.yml
+++ b/.github/workflows/secret_cli_release.yml
@@ -4,6 +4,11 @@ on:
   release:
     types: [published]
   pull_request:
+  workflow_dispatch:
+    inputs:
+      tag:
+        required: true
+        type: string
 
 permissions:
   contents: read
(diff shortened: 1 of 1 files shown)


.github/workflows/secret_cli_release.yml
```diff
@@ -99,19 +107,17 @@ jobs:
         echo "asset=${name}.tar.gz" >> "$GITHUB_OUTPUT"
 
     - name: Upload release assets
```
Xynnn007, line 109: I just found that we do not have ` release immutability Loading` for this repo, while for `reference-values` we have. This could be a potential supply chain replacement attack. I guess we should know that

Xynnn007 (approved):

.github/workflows/secret_cli_release.yml
fitzthum: Yeah we may want to enable that

PR #1636, opened 2026-08-08
coco-extension: build in pinned containers and ship 24.04 + 26.04 variants

fidencio:
The extension ships gnu-linked binaries and a cryptsetup binary but no libc of
its own, so its ABI is dictated by whatever userspace it was built on. Until now
that was the GitHub runner image, which tied the payload to Ubuntu 24.04 and made
the guest ABI an implicit side effect of a runner label. Bumping the guest to
26.04 was therefore not expressible without also moving CI infrastructure.

The Ubuntu release is now an explicit build argument of a builder image, and a
matrix axis, so one variant is published per supported guest release. Kata still
pins a 24.04 guest rootfs, so both are built until it moves. Runners keep their
current ubuntu-24.04 label and only need the tooling to turn the assembled rootfs
into the EROFS + dm-verity disk image.

Building the environment costs more than building the payload does, mostly the
NVIDIA SDK compile, so builder images are cached in ghcr.io under a digest of the
recipe and its pinned versions. Pull requests cannot authenticate to pull one and
fall back to building it, as they did before.

The first commit is a prerequisite: a62c7537 bumped the nv-attestation-sdk crate
to the 2026.06.09 tag, but the version CI compiles the C++ SDK from stayed at
2026.03.02, in both the coco-extension workflow and the install-nvat-sdk action.
Since these builds link the system libnvat, the bindings and the library they
bind to have been a release apart ever since. 2026.03.02 additionally fails to
compile against libxml2 2.12+, which 26.04 carries, so the 26.04 variant cannot
be built without this.

**Note: This is a preparation for switching Kata Containers rootfses images to
Ubuntu 26.04.**

## Breaking change

A variant is only usable on a matching guest, so there is no longer an
unqualified image to pull. Every tag and release asset now carries the variant,
and the tags that omitted it are gone:

| before | after |
| --- | --- |
| `<sha>-amd64` | `<sha>-ubuntu24.04-amd64` |
| `latest`, `<version>` | `latest-ubuntu24.04`, `<version>-ubuntu24.04` |
| `kata-containers-coco-extension-<arch>.img` | `kata-containers-coco-extension-ubuntu24.04-<arch>.img` |

Consumers must name the guest release they are targeting. Kata should keep
consuming the ubuntu24.04 variant until its rootfs moves, then switch tags.

## Alternatives considered

Keeping the unqualified tags as an alias for 24.04 would avoid the break, but it
makes the ABI contract invisible again: pulling `latest` onto a 26.04 guest would
silently produce a mismatch rather than a missing tag. Making the variant explicit
trades a one-time update for a failure mode that is obvious.

Moving the runners to ubuntu-26.04 instead was rejected because the runner image
is not the right knob: it would still leave the guest ABI implicit, and it cannot
express two variants at once.

### DIFF
--- a/.github/actions/install-nvat-sdk/action.yml
+++ b/.github/actions/install-nvat-sdk/action.yml
@@ -3,9 +3,12 @@ description: Build and install the NVIDIA Attestation SDK from source (x86_64 on
 
 inputs:
   version:
-    description: NVAT SDK version tag to checkout
+    description: >
+      NVAT SDK version tag to checkout. Defaults to the tag the
+      nv-attestation-sdk crate is pinned to, so the C++ library and the Rust
+      bindings that link it cannot drift apart.
     required: false
-    default: "2026.03.02"
+    default: ""
 
 runs:
   using: composite
@@ -15,13 +18,26 @@ runs:
       shell: bash
       env:
         NVAT_VERSION: ${{ inputs.version }}
+        CARGO_MANIFEST: ${{ github.workspace }}/attestation-agent/attester/Cargo.toml
       run: |
         export RUSTFLAGS=""
+
+        if [ -z "${NVAT_VERSION}" ]; then
+          NVAT_VERSION="$(sed -n \
+            's/^nv-attestation-sdk[[:space:]]*=.*tag[[:space:]]*=[[:space:]]*"\([^"]*\)".*/\1/p' \
+            "${CARGO_MANIFEST}")"
+          [ -n "${NVAT_VERSION}" ] || {
+            echo "failed to read the nv-attestation-sdk tag from ${CARGO_MANIFEST}" >&2
+            exit 1
+          }
+        fi
+        echo "Using NVIDIA Attestation SDK ${NVAT_VERSION}"
+
         sudo apt-get update
         sudo apt-get install -y --no-install-recommends \
           build-essential clang cmake pkg-config libclang-dev \
           libcurl4-openssl-dev libssl-dev libxml2-dev \
-          libxmlsec1-dev libxmlsec1-openssl
+          libxmlsec1-dev zlib1g-dev
 
         tmpdir="$(mktemp -d)"
         git clone --depth=1 --branch "${NVAT_VERSION}" \
--- a/.github/workflows/coco-extension-image.yml
+++ b/.github/workflows/coco-extension-image.yml
@@ -40,8 +40,10 @@ env:
   # stops baking pause into the coco extension.
   PAUSE_IMAGE_REPO: docker://registry.k8s.io/pause
   PAUSE_IMAGE_VERSION: "3.10"
-  NVAT_VERSION: "2026.03.02"
   UMOCI_VERSION: v0.6.0
+  # The C++ SDK is linked by the Rust bindings, so the two have to be the same
+  # release; the crate pin below is the single source of truth for both.
+  NVAT_CARGO_MANIFEST: attestation-agent/attester/Cargo.toml
 
 jobs:
   build:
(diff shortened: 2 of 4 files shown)


.github/actions/install-nvat-sdk/action.yml
```diff
@@ -5,7 +5,7 @@ inputs:
   version:
     description: NVAT SDK version tag to checkout
     required: false
-    default: "2026.03.02"
+    default: "2026.06.09"
```
Xynnn007: @mythi once mentioned that do we have a way to set this automatically to align with the `Cargo.toml`'s?

.github/actions/install-nvat-sdk/action.yml
```diff
@@ -18,10 +18,13 @@ runs:
       run: |
         export RUSTFLAGS=""
         sudo apt-get update
+        # libxmlsec1-dev pulls in the openssl engine, whose package name is
+        # release specific (libxmlsec1-openssl on noble, libxmlsec1-openssl1
+        # later), so it is deliberately not named here.
```
Xynnn007: We do not need the notes here in code. just keep it in the commit message.

Xynnn007 (approved):

.github/actions/install-nvat-sdk/action.yml
fidencio: Took a stab at this one, addressed.

.github/actions/install-nvat-sdk/action.yml
fidencio: Addressed.

Xynnn007 (approved):
LGTM

.github/workflows/coco-extension-image.yml
```diff
@@ -8,6 +8,13 @@ name: Build and publish the CoCo guest extension image
 #   * an EROFS + dm-verity disk image (+ its dm-verity root hash), using the
 #     same layout as kata's rootfs-image-coco-extension asset.
 #
+# The payload is built inside a pinned Ubuntu container
+# (tools/coco-extension/Dockerfile.builder) rather than directly on the runner,
+# so the binaries match the guest rootfs ABI regardless of the runner image.
+# One variant is published per supported guest release, and every tag and
+# release asset is qualified with it (e.g. "<sha>-ubuntu26.04-amd64"): the
+# extension ships no libc, so a variant is only usable on a matching guest.
```
mythi, line 16: from libc perspective, an exact match is not needed: kata's 26.04 libc should provide backwards compatibility to run 24.04 binaries but it makes sense to have the mechanism to control the version match and decouple this from runner OS anyway.

mythi (approved):
LGTM. 

maybe the definition of all those arch flavors could be simplified later since we assume native building only but that's not relevant for this pr

.github/workflows/coco-extension-image.yml
fidencio: While I agree with this in the libc specific case, I am not so sure if the other libc linked deps would just behave in the very same way.

I'd rather take the safer path here, as I do think it can byte us in the future / testing.

.github/actions/install-nvat-sdk/action.yml
```diff
@@ -15,13 +18,26 @@ runs:
...
+          }
+        fi
+        echo "Using NVIDIA Attestation SDK ${NVAT_VERSION}"
+
         sudo apt-get update
         sudo apt-get install -y --no-install-recommends \
           build-essential clang cmake pkg-config libclang-dev \
           libcurl4-openssl-dev libssl-dev libxml2-dev \
-          libxmlsec1-dev libxmlsec1-openssl
+          libxmlsec1-dev zlib1g-dev
```
fitzthum, line 40: I guess the automatic updates will fail if nvat changes dependencies again, but that's ok.

.github/workflows/coco-extension-image.yml
```diff
@@ -90,66 +109,87 @@ jobs:
...
+        # variant; an empty version skips it entirely.
+        nvat_version=""
+        if [ "${ARCH}" = "x86_64" ]; then
+          nvat_version="$(sed -n \
+            's/^nv-attestation-sdk[[:space:]]*=.*tag[[:space:]]*=[[:space:]]*"\([^"]*\)".*/\1/p' \
+            "${NVAT_CARGO_MANIFEST}")"
+          [ -n "${nvat_version}" ] || {
+            echo "failed to read the nv-attestation-sdk tag from ${NVAT_CARGO_MANIFEST}" >&2
+            exit 1
+          }
```
fitzthum, line 141: We use this in a few places, but doesn't quite seem worth it to separate it out.

fitzthum (approved):
LGTM

PR #1646, opened 2026-08-10
api-server-rest: add unit tests for router

supritai:
Cover all branches reachable without a live ttrpc back-end:
response builder helpers (status + body), loopback IP guard,
`/info` GET/non-GET, unknown paths, and AA/CDH feature-not-enabled
404 responses across all defined sub-routes and methods.

The tests live in an inline `#[cfg(test)]` module at the bottom of
`router.rs` and use only `std`/`tokio`/`hyper` — no additional test
dependencies are introduced.

**Test coverage added:**

| Test | What it verifies |
|---|---|
| `response_status` | `build_response` sets the correct HTTP status code |
| `response_body` | `build_response` round-trips the body bytes |
| `non_loopback_rejected` | requests from non-loopback IPs get 403 |
| `info_get_ok` | `GET /aa/v1/info` returns 200 when AA is disabled (feature-not-enabled path) |
| `info_non_get_rejected` | non-GET verb on `/aa/v1/info` returns 405 |
| `unknown_path_404` | arbitrary path returns 404 |
| `aa_routes_404_when_disabled` | every AA sub-route + method returns 404 when `AA_FEATURE` is off |
| `cdh_routes_404_when_disabled` | every CDH sub-route + method returns 404 when `CDH_FEATURE` is off |

Assisted-by: IBM Bob

### DIFF
--- a/api-server-rest/src/router.rs
+++ b/api-server-rest/src/router.rs
@@ -292,3 +292,218 @@ impl Router {
         self.not_found()
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use hyper::{body::to_bytes, Method, Request, StatusCode};
+    use rstest::rstest;
+    use std::net::{IpAddr, Ipv4Addr, SocketAddr};
+
+    fn loopback() -> SocketAddr {
+        SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), 12345)
+    }
+
+    fn remote() -> SocketAddr {
+        SocketAddr::new(IpAddr::V4(Ipv4Addr::new(10, 0, 0, 1)), 12345)
+    }
+
+    fn router_no_clients() -> Router {
+        Router::new(None, None, "resource".to_string())
+    }
+
+    async fn body_str(resp: Response<Body>) -> String {
+        let bytes = to_bytes(resp.into_body()).await.unwrap();
+        String::from_utf8_lossy(&bytes).into_owned()
+    }
+
+    async fn get_info(router: &Router) -> Response<Body> {
+        let req = Request::builder()
+            .method(Method::GET)
+            .uri("/info")
+            .body(Body::empty())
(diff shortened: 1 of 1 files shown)


api-server-rest/src/router.rs
```diff
@@ -292,3 +292,227 @@ impl Router {
...
+    async fn loopback_address_passes_ip_guard() {
+        let mut router = router_no_clients();
+        let req = Request::builder()
+            .method(Method::GET)
+            .uri("/unknown")
+            .body(Body::empty())
+            .unwrap();
+        let resp = router.route(loopback(), req).await.unwrap();
+        assert_eq!(resp.status(), StatusCode::NOT_FOUND);
+    }
```
Xynnn007: Can we combine the two test with `rstest` to make the code more dry?

api-server-rest/src/router.rs
```diff
@@ -292,3 +292,227 @@ impl Router {
...
+
+    #[tokio::test]
+    async fn internal_error_returns_500_with_body() {
+        let router = router_no_clients();
+        let resp = router
+            .internal_error("something went wrong".to_string())
+            .unwrap();
+        assert_eq!(resp.status(), StatusCode::INTERNAL_SERVER_ERROR);
+        assert_eq!(body_str(resp).await, "something went wrong");
+    }
```
Xynnn007: Can we combine the test cases with `rstest`?

Xynnn007 (commented):
Overall LGTM, thanks! Let make the test cases more dry

api-server-rest/src/router.rs
supritai: sure, thanks for the suggestion.

api-server-rest/src/router.rs
supritai: sure

Xynnn007 (approved):

PR #1648, opened 2026-08-10
cdh: add crash-safe persistent LUKS volumes

chalabi2:
## Why

CDH needs a persistent block-volume path that survives guest restarts without letting a transient mount failure authorize reformatting. The host controls the backing device, including its LUKS2 metadata, so CDH must not pass an unauthenticated header to cryptsetup. This follows the threat described in the [Trail of Bits LUKS2 analysis](https://blog.trailofbits.com/2025/10/30/vulnerabilities-in-luks2-disk-encryption-for-confidential-vms/).

## Design

- Require a stable `volumeId` and an explicit existing CDH key reference.
- Format only after a complete scan confirms that the device is zero-filled.
- Store a detached 16 MiB LUKS2 header before the encrypted payload.
- Authenticate the header and two 4 KiB crash-state records with HMAC-SHA256 derived from the volume key.
- Copy the stored header into guest `/run`, verify it there, and only then invoke cryptsetup with that protected copy.
- Probe the decrypted filesystem explicitly and recover across interrupted initialization phases without reformatting an existing ext4 filesystem.

## Limits

The first mount reads the complete device. At 1 GB/s, a 1 TB volume takes about 16 minutes 40 seconds. Persistent mode does not yet support `dataIntegrity: true`; the HMAC protects the header and CDH state, not ciphertext integrity or rollback.

## Validation

- `cargo fmt --all -- --check`
- `cargo clippy -p confidential-data-hub --all-targets --no-default-features --features kbs,bin -- -D warnings`
- 11 focused persistent-mode unit tests
- 3 privileged Linux loop-device tests covering reopen, seven initialization phases, wrong keys, preserved data, and header tampering before mapper creation


zvonkok:
It would help if the comments explained not WHAT is being done, but WHY. The LLM-generated comments are lame. Beware the AI policy as mentioned in the other PR.

chalabi2:
> It would help if the comments explained not WHAT is being done, but WHY. The LLM-generated comments are lame. Beware the AI policy as mentioned in the other PR.

yessir, will go in and adhere to the policy wherever relevant!

fitzthum:
btw, you might want to check out this [blog post](https://blog.trailofbits.com/2025/10/30/vulnerabilities-in-luks2-disk-encryption-for-confidential-vms/). need to be very careful with how we handle the LUKS header.

chalabi2:
> btw, you might want to check out this [blog post](https://blog.trailofbits.com/2025/10/30/vulnerabilities-in-luks2-disk-encryption-for-confidential-vms/). need to be very careful with how we handle the LUKS header.

Not sure which path you want for the LUKS changes, so I had AI help me land on something that preserves my fix and addresses the blog post.

Xynnn007 (commented):
Looks like this PR lets CDH to store some metadata of mounted volumes for better lifetime tracking. I once had similar idea for this. I will get some time to take a look into the code and design.

chalabi2:
> Looks like this PR lets CDH to store some metadata of mounted volumes for better lifetime tracking. I once had similar idea for this. I will get some time to take a look into the code and design.

Thank you, and any guidance you can provide to help get this merged would be appreciated as its a requirement for providers on our network to support persistent storage in TEE.

manuelh-dev:
> > Looks like this PR lets CDH to store some metadata of mounted volumes for better lifetime tracking. I once had similar idea for this. I will get some time to take a look into the code and design.
> 
> Thank you, and any guidance you can provide to help get this merged would be appreciated as its a requirement for providers on our network to support persistent storage in TEE.

@chalabi2 - is this work in the Kata context? If so, do you have a Kata feature branch? Generally I would advise to first create a GitHub issue we can use for discussion/design of this, but this may just be my personal choice. An issue would especially help thinking of how, if used in Kata K8s context, this flow would work in production, do we use CSI drivers for this, ..., how does this scale. In any case, a bit more context would help here.

I've recently done some work on this, too, but got interrupted. I'm interested in collaborating here. I am sharing all my prior notes and feature branches here in a second message in a bit. I never got to open a GitHub issue here. My approach should cover the whole Kubernetes, Kata integration, Kata agent policy, attestation, ... scenario using r/o many and r/w single VolumeDevices.

If your scenario does not care about a Kata end-to-end use case, please still keep that use case in mind if it makes sense to you.

@fitzthum @Xynnn007 for vis (maybe it's time to sort requirements or design goals in a GitHub issue before adding new features to CDH, but, again, that's only my take).

manuelh-dev:
@chalabi2 - please let me know if this is useful for you. I had added a few TODOs here containing feedback and notes I had received but not yet adopted. also @Xynnn007 - if you had prior thoughts on this, maybe you can take a look here too, with a view on an end-to-end Kubernetes scenario with Kata containers and the agent security policy.

# Confidential Block Volumes for Kata Containers

## Goal

Enable an attested Kata workload to consume a persistent raw block volume whose contents remain confidential and integrity-protected from an untrusted host.

Kubernetes continues to attach the volume through its existing raw-block PVC and `volumeDevices` mechanism. Inside the guest, Confidential Data Hub verifies and decrypts the device before `kata-agent` exposes the activated block device to the container.

The design supports two storage profiles under the same activation model:

- immutable read-only volumes for pre-populated data such as models, datasets, or application assets
- single-pod read-write volumes for mutable confidential storage

The Pod annotation identifies the confidential volume and points to an attested manifest. The manifest selects the concrete protection profile.

## Storage formats

### Read-only profile

For read-only volumes, CI or an administrator prepares one raw disk image:

```text
┌──────────────────────────────────────┬─────────────────────┐
│ LUKS2 image region                   │ dm-verity hash tree │
│ header, metadata, encrypted data     │                     │
└──────────────────────────────────────┴─────────────────────┘
```

`dm-verity` covers the complete LUKS2 image region, including the LUKS2 header, metadata, encrypted filesystem, and encrypted data. The hash tree is appended immediately after that region.

The guest constructs:

```text
attached ciphertext device
        │
        ▼
read-only dm-verity mapping
        │
        ▼
read-only LUKS2 mapping
        │
        ▼
container block device
```

The host stores no plaintext key or trusted integrity root.

### Single-pod read-write profile

For read-write volumes, CI, an administrator, or storage tooling prepares one raw disk image/device:

```text
┌──────────────────────────────┬──────────────────────────────────────────────┐
│ LUKS2 header / metadata      │ encrypted data + dm-integrity metadata/tags  │
│ authenticated before use     │ mutable, integrity-protected payload area     │
└──────────────────────────────┴──────────────────────────────────────────────┘
```

The LUKS2 header is present on the attached disk, but it is not trusted directly. CDH copies the expected header region into guest tmpfs, authenticates that copied header against KBS-provided policy or verification material, and only then invokes `cryptsetup` with `--header` pointing at the authenticated tmpfs copy. `cryptsetup` then opens the mutable encrypted data area using the authenticated LUKS2 metadata, including its dm-integrity configuration.

The guest constructs:

```text
attached ciphertext device
        │
        ▼
authenticated LUKS2 header copied into guest tmpfs
        │
        ▼
read-write LUKS2 / dm-integrity mapping
        │
        ▼
container block device
```

This prevents `cryptsetup` from parsing host-controlled raw LUKS2 metadata directly.

## Workload interface

**TODO** REVISIT: we should consider using init-data for this.
A Pod identifies protected volumes using a neutral confidential-volume annotation whose value maps Kubernetes volume names to manifest references and guest device paths:

```yaml
metadata:
  annotations:
    io.katacontainers.volume.confidential: >-
      {"block-disk":{"manifestUri":"kbs:///kata-ci/storage-manifest/model-data-v1",
                     "devicePath":"/dev/hostdisk"}}
```

The Pod uses the normal Kubernetes block-volume interface:

```yaml
volumeDevices:
  - name: block-disk
    devicePath: /dev/hostdisk

volumes:
  - name: block-disk
    persistentVolumeClaim:
      claimName: model-data-pvc
      readOnly: true
```

For read-write volumes, the PVC and workload request are writable, but the intended ownership model is still single-pod read-write. This is not a read-write-many shared storage design.

A disabled-by-default runtime/operator gate enables support:

```toml
[runtime]
experimental = ["confidential_block_device"]
```

An annotated volume is rejected if the feature is disabled, malformed, or cannot be matched to exactly one block device. The requested access mode must be compatible with the manifest-selected protection profile.

## Trustee resources and manifest examples

Trustee stores immutable, versioned resources for each protected volume.

**TODO** REVISIT: the storage-manifest could be added to init-data instead
For a read-only volume:

```text
kbs:///kata-ci/storage-manifest/model-data-v1
kbs:///kata-ci/storage-key/model-data-v1
```

Example read-only manifest:

```json
{
  "version": 1,
  "volume": {
    "id": "model-data",
    "version": "v1"
  },
  "protection": {
    "type": "luks2-verity-ro",
    "keyUri": "kbs:///kata-ci/storage-key/model-data-v1",
    "luksUuid": "11111111-2222-3333-4444-555555555555",
    "verity": {
      "dataBlocks": 128000,
      "dataBlockSize": 4096,
      "hashBlockSize": 4096,
      "algorithm": "sha256",
      "rootHash": "...",
      "salt": "..."
    }
  }
}
```

For a read-write volume:

```text
kbs:///kata-ci/storage-manifest/rw-data-v1
kbs:///kata-ci/storage-key/rw-data-v1
```

Example read-write manifest:

```json
{
  "version": 1,
  "volume": {
    "id": "rw-data",
    "version": "v1"
  },
  "protection": {
    "type": "luks2-integrity-rw",
    "keyUri": "kbs:///kata-ci/storage-key/rw-data-v1",
    "luksUuid": "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee",
    "header": {
      "offsetBytes": 0,
      "lengthBytes": 16777216,
      "sha256": "..."
    }
  }
}
```

The manifest URI is encoded exactly in the measured `kata-agent` security policy. Trustee releases the manifest and key only to a guest presenting the expected attestation and Init-Data measurement. This prevents the host from substituting another manifest or volume version.

## CDH interface

Add a lifecycle-oriented `SecureVolumeService`:

```text
ActivateVolume(device ID, manifest URI)
    -> activation ID, plaintext device path

DeactivateVolume(activation ID)
```

CDH performs all security-sensitive operations:

1. Fetch and validate the manifest from Trustee.
2. Resolve the attached guest block device.
3. Select the activation profile from the manifest.
4. Retrieve the required key or verification material directly from Trustee.
5. Activate the protected mapping inside the guest.
6. Return the plaintext mapper path and retain cleanup state.
7. Remove mappings and temporary state during deactivation.

For read-only volumes, activation creates the verified `dm-verity` mapping, validates the expected LUKS metadata through that verified mapping, opens LUKS2 read-only, and returns the plaintext mapper.

For read-write volumes, activation copies the LUKS2 header into guest tmpfs, authenticates the copied header, opens LUKS2 read-write using the authenticated detached header, and returns the plaintext mapper.

Neither the shim nor `kata-agent` receives the key, trusted `dm-verity` parameters, or trusted header-authentication material.

The existing CDH `SecureMount` API remains unchanged. The new API is an additive block-device lifecycle API. Over time, this activation shape could be generalized into a common protected-storage lifecycle:

```text
ActivateStorage(request) -> activation handle + target
DeactivateStorage(handle)

target:
  kind: blockDevice
  path: /dev/mapper/...

target:
  kind: mount
  path: /run/...
```

Such a model could return either a block-device mapper path or a mounted filesystem path, while using the same activation handle for cleanup. `SecureMount` and `SecureVolume` could then remain as compatibility wrappers or profile-specific entry points.

## Kata changes

`runtime-rs` will:

- parse and validate the confidential-volume annotation
- associate it with the existing OCI block device
- check that the runtime gate is enabled
- validate that the requested access mode is compatible with the annotation and manifest intent
- hot-plug the physical device using the existing mechanism
- send `kata-agent` a typed secure-volume device request

`kata-agent` will:

- ask CDH to activate the volume
- replace the physical device in the container specification with CDH’s verified plaintext mapper
- grant device access according to the activated profile
- track the activation handle
- deactivate the volume during cleanup

A distinct secure-volume device type ensures that an older or unaware agent fails closed instead of exposing the physical ciphertext device normally.

## Security policy

`genpolicy` binds together:

- the exact annotation key and value
- Kubernetes volume name
- container device path
- manifest URI
- secure-volume device type
- requested access mode / PVC read-only state

At runtime, `kata-agent` policy rejects changed annotations, alternate manifests, unexpected devices, incompatible writable/read-only requests, or exposure of the original ciphertext device.

For read-only volumes, freshness and rollback protection for the expected data version comes from binding the manifest URI and exact annotation value into the measured policy and requiring attestation before Trustee releases the manifest and key.

For read-write volumes, the same policy binding protects the selected manifest and header-authentication policy. Additional replay handling for mutable disk state remains an area to investigate.

## Provisioning flow

For a read-only volume:

1. Create the raw disk image.
2. Format/populate the encrypted filesystem or data area.
3. Build the `dm-verity` hash tree over the full LUKS2 image region.
4. Store the LUKS unlock secret in Trustee/KBS.
5. Store the manifest containing verity parameters and key URI in Trustee/KBS.
6. Attach the image through a Kubernetes raw-block PV/PVC.
7. Reference the manifest URI through the confidential volume annotation.

For a read-write volume:

1. Create the block image or device.
2. Format it as LUKS2 with integrity protection, for example using `dm-integrity`.
3. Capture the expected LUKS2 header bytes, digest, MAC, or signature.
4. Store the unlock secret in Trustee/KBS.
5. Store the header authentication policy/material in Trustee/KBS.
6. Attach the device through a Kubernetes raw-block PV/PVC.
7. Reference the manifest URI through the confidential volume annotation.

Kata does not become a storage provisioner. The new mechanism defines how an already-attached block device is confidentially activated inside the guest.

## Validation

CI covers the end-to-end path for both profiles:

- read-only positive test: activate a protected image and verify known plaintext is readable and writes fail
- read-only negative test: corrupt ciphertext and verify activation or reads fail through `dm-verity`
- read-write positive test: activate a writable protected image, write data, sync it, and read it back
- read-write negative test: corrupt the LUKS2 header and verify CDH refuses activation before `cryptsetup` consumes unauthenticated metadata

## Repository delivery

The CDH and protocol implementation is developed in a guest-components feature branch and pushed to the `manuelh-dev` fork:
	- [manuelh-dev/guest-components at mahuber/confidential-ro-volume](https://github.com/manuelh-dev/guest-components/tree/mahuber/confidential-ro-volume)
	- [manuelh-dev/guest-components at mahuber/confidential-rw-volume](https://github.com/manuelh-dev/guest-components/tree/mahuber/confidential-rw-volume)

**TODO:** the branch is outdated - the versions.yaml references are nowadays made differently with these extension images, so dev flow needs to be adjusted
Kata’s `versions.yaml` pins that exact guest-components commit.

The corresponding `runtime-rs`, `kata-agent`, `genpolicy`, and CI work is developed in a Kata feature branch and built end-to-end from the Kata repository:
	- [manuelh-dev/kata-containers at mahuber/confidential-ro-volume](https://github.com/manuelh-dev/kata-containers/tree/mahuber/confidential-ro-volume)

## Why a new CDH API

The existing CDH `SecureMount` API already supports encrypted storage flows, including cases where CDH receives or retrieves encryption material and prepares a block device. However, `SecureMount` is mount-oriented: its natural contract is to turn a storage description into a mounted filesystem path.

Confidential block volumes preserve Kubernetes raw-block semantics. Kubernetes has already attached a specific raw block device, and `kata-agent` must expose a verified plaintext block mapper at the container’s requested `volumeDevices` path, such as `/dev/hostdisk`.

Reusing `SecureMount` would require overloading a mount API with “activate but do not mount” behavior, manifest-specific `dm-verity` or authenticated-header parameters, and a returned device-mapper path. The new `SecureVolumeService` keeps the responsibility split explicit: CDH owns confidential activation, key retrieval, integrity setup, LUKS setup, and cleanup state, while `kata-agent` owns container device placement and policy enforcement.

## Unsolved problems or aspects to investigate

- Replay / rollback protection for mutable read-write volumes, especially stale disk state, stale authenticated headers, or restored old KBS policy material.

- Recovery semantics for read-write filesystems, including use of ext4/XFS-style journaling and correct flush/FUA/barrier propagation through guest, virtio, QEMU, and host storage.

- Safe host storage configuration for read-write volumes, avoiding cache modes or backends that break the durability assumptions of the guest filesystem.

- Header update and maintenance flow for future key rotation, backup/restore, recovery, or LUKS2 header changes.

fitzthum:
yeah @chalabi2 turns out storage is pretty complicated. I'm curious about your approach here. Hopefully we can get something in for your use case.

confidential-data-hub/docs/use-cases/secure-mount-with-block-device.md
```diff
@@ -56,6 +57,77 @@ with `lazy_itable_init=0` unless the caller explicitly provides a
...
+}
+```
+
+The authenticated state slots do not provide anti-rollback protection. A
+storage provider can replay an older valid disk image. Preventing rollback
+requires an external monotonic authority.
+
+Persistent mode authenticates the LUKS2 header and CDH state, not the encrypted
+filesystem payload. Because persistent mode does not yet support
+`dataIntegrity: true`, callers that need tamper detection must add integrity at
```
manuelh-dev, line 124: would we not want to design a mechanism with tamper detection right away?

chalabi2:
> > > Looks like this PR lets CDH to store some metadata of mounted volumes for better lifetime tracking. I once had similar idea for this. I will get some time to take a look into the code and design.
> > 
> > 
> > Thank you, and any guidance you can provide to help get this merged would be appreciated as its a requirement for providers on our network to support persistent storage in TEE.
> 
> @chalabi2 - is this work in the Kata context? If so, do you have a Kata feature branch? Generally I would advise to first create a GitHub issue we can use for discussion/design of this, but this may just be my personal choice. An issue would especially help thinking of how, if used in Kata K8s context, this flow would work in production, do we use CSI drivers for this, ..., how does this scale. In any case, a bit more context would help here.
> 
> I've recently done some work on this, too, but got interrupted. I'm interested in collaborating here. I am sharing all my prior notes and feature branches here in a second message in a bit. I never got to open a GitHub issue here. My approach should cover the whole Kubernetes, Kata integration, Kata agent policy, attestation, ... scenario using r/o many and r/w single VolumeDevices.
> 
> If your scenario does not care about a Kata end-to-end use case, please still keep that use case in mind if it makes sense to you.
> 
> @fitzthum @Xynnn007 for vis (maybe it's time to sort requirements or design goals in a GitHub issue before adding new features to CDH, but, again, that's only my take).

@manuelh-dev yessir, this is in the Kata context. Our test path uses a raw block PVC passed into the guest, with CDH unlocking and mounting it for the workload.

I don't have a clean Kata feature branch for this yet. We have the integration code from our E2E testing, but it includes Akash-specific wiring, so I was treating this PR as the upstream feature work. I probably skipped some of the normal design process because I was trying to get persistent LUKS volumes working quickly.

The work you shared looks directly relevant. I'll go through it, and I'm happy to open an issue so we can compare the approaches and figure out what belongs in CDH versus Kata. I'd rather work with what you already started than create another competing path.

manuelh-dev:
> > > > Looks like this PR lets CDH to store some metadata of mounted volumes for better lifetime tracking. I once had similar idea for this. I will get some time to take a look into the code and design.
> > > 
> > > 
> > > Thank you, and any guidance you can provide to help get this merged would be appreciated as its a requirement for providers on our network to support persistent storage in TEE.
> > 
> > 
> > @chalabi2 - is this work in the Kata context? If so, do you have a Kata feature branch? Generally I would advise to first create a GitHub issue we can use for discussion/design of this, but this may just be my personal choice. An issue would especially help thinking of how, if used in Kata K8s context, this flow would work in production, do we use CSI drivers for this, ..., how does this scale. In any case, a bit more context would help here.
> > I've recently done some work on this, too, but got interrupted. I'm interested in collaborating here. I am sharing all my prior notes and feature branches here in a second message in a bit. I never got to open a GitHub issue here. My approach should cover the whole Kubernetes, Kata integration, Kata agent policy, attestation, ... scenario using r/o many and r/w single VolumeDevices.
> > If your scenario does not care about a Kata end-to-end use case, please still keep that use case in mind if it makes sense to you.
> > @fitzthum @Xynnn007 for vis (maybe it's time to sort requirements or design goals in a GitHub issue before adding new features to CDH, but, again, that's only my take).
> 
> @manuelh-dev yessir, this is in the Kata context. Our test path uses a raw block PVC passed into the guest, with CDH unlocking and mounting it for the workload.
> 
> I don't have a clean Kata feature branch for this yet. We have the integration code from our E2E testing, but it includes Akash-specific wiring, so I was treating this PR as the upstream feature work. I probably skipped some of the normal design process because I was trying to get persistent LUKS volumes working quickly.
> 
> The work you shared looks directly relevant. I'll go through it, and I'm happy to open an issue so we can compare the approaches and figure out what belongs in CDH versus Kata. I'd rather work with what you already started than create another competing path.

Excellent, @chalabi2 thank you, and great to hear this is in the Kata context. I don't care about ownership here, only about the feature. So, please feel free to just take / reuse any of this you deem useful.

Let's try to design this from the end user scenario top-down *and* end-to-end, if that makes sense:
- what is the flow - we'll want to use the features with CSI drivers - for instance, I'd be interested in a flow using nvmesh CSI (or looking into Kata direct volume CSI, but these are really just examples)
  - my prototype uses `volumeDevices`, is that what we're looking for, can we use `volumeMounts` or both, ...
- who is responsible for provisioning disks (if we allow for pre-provisioned disks as done in my prototype, *[edit] not locked into this*)
- how can we get to an attested solution (the end to end), integrity and rollback protection (to a reasonable extent), provisioning trustee with relevant information
- what API (changes/amendments) do we propose for CDH (to serve the kata-agent in our case but also to offer a good interface in general)


If we can get to a decent enough description of this, I am sure @Xynnn007 and @fitzthum will chip in with their guidance. Feel free to reach out on Slack if you like to work on a proposal together; anything works for me though.

chalabi2:
@manuelh-dev yes slack would be great would like to work closely with yall on this.

Xynnn007:
Thanks @manuelh-dev and @chalabi2 for the discussion. Persistent confidential storage is a critical need for CoCo. I'd like to join the technical design reviewment. Before that, a clean problem definition (I guess we already have), targets (what ability will user enjoy) and user guide will be good. Agree with @manuelh-dev 's idea to have a overall issue for design.


To me, there are some requirements/comments from my side
1. keep interfaces between kata-agent and CDH clean and de-coupled enough.
2. (as @manuelh-dev already wrote in the comment) pre-created volumes should be supported.
3. besides one-encrypted-rw, multi-r volumes, we should also support multi-r encrypted volumes.

chalabi2:
I found my way into the confidential-containers slack channel. name there is chalabi as well. Maybe we can move the convo for this over there.

zvonkok:
Is anyone in this thread considering replay protection for integrity-protected volumes?

noeljackson:
I opened a stacked follow-up in #1663:

- reject unknown persistent-volume options
- add journaled dm-integrity
- cover reopen, crash recovery, tamper, and corruption cases

PR #1651, opened 2026-08-11
Clippy/pre edition 2024

Xynnn007:
Please see concrete commit messages.


mythi (approved):
LGTM

> it was simply never hit because the clippy matrix in CI does not cover the `attester` crate with the `bin` feature enabled

I saw the same gap earlier. it might make sense to add separate `attester` checks to selected CI targets (e.g, weekly rust stable) in some later PR

PR #1652, opened 2026-08-11
aa: make lint cover all workspace packages under attestation-agent

Xynnn007:
Previously make lint only clippy'd the attestation-agent crate, so sibling crates (notably attester's evidence_getter via bin) were never linted. Clippy each package with -D warnings, and use --all-targets plus bin where binaries exist.

Note that sev parts are also included as they are in tree. I will make a separate PR to get rid of them altogether.

### DIFF
--- a/attestation-agent/Makefile
+++ b/attestation-agent/Makefile
@@ -117,13 +117,34 @@ TARGET := $(TARGET_DIR)/$(BIN_NAME)
 lint: $(LINT_OPTION)
 
 lint-default:
-	cd attestation-agent && cargo clippy --features kbs,coco_as,bin,grpc,ttrpc,all-attesters
+	cargo clippy -p attestation-agent --features kbs,coco_as,bin,grpc,ttrpc,all-attesters -- -D warnings
+	cargo clippy -p attester --all-targets --features all-attesters,bin -- -D warnings
+	cargo clippy -p coco_keyprovider --all-targets -- -D warnings
+	cargo clippy -p kbc --all-targets --features cc_kbc,all-attesters,rust-crypto -- -D warnings
+	cargo clippy -p kbs_protocol --all-targets --no-default-features \
+		--features background_check,passport,rust-crypto,all-attesters,bin,aa_ttrpc -- -D warnings
+	cargo clippy -p crypto --all-targets --features rust-crypto -- -D warnings
+	cargo clippy -p resource_uri --all-targets -- -D warnings
 
 lint-s390x:
-	cd attestation-agent && cargo clippy --no-default-features --features openssl,kbs,coco_as,bin,grpc,ttrpc
+	cargo clippy -p attestation-agent --no-default-features --features openssl,kbs,coco_as,bin,grpc,ttrpc -- -D warnings
+	cargo clippy -p attester --all-targets --no-default-features --features se-attester,bin -- -D warnings
+	cargo clippy -p coco_keyprovider --all-targets -- -D warnings
+	cargo clippy -p kbc --all-targets --no-default-features --features cc_kbc,se-attester,openssl -- -D warnings
+	cargo clippy -p kbs_protocol --all-targets --no-default-features \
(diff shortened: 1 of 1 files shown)


fitzthum (commented):
Is there a good way to just check all the packages? Maybe `cargo clippy --all`? This seems a little fragile as-is.

mythi:
> Clippy each package with -D warnings, and use --all-targets plus bin where binaries exist.

I wonder if `cargo clippy --workspace --all-targets  -- -D warnings` would be sufficient?

Xynnn007:
The `Makefile` now is under `attestation-agent` directory, thus as blackbox it's proper for it to only cover crates under the `attestation-agent/*`.

There are two ways to fix original issue

1. get this merged and improve clippy coverage **ASAP** and let's refactor the messy CI code in the repo
2. refactor the messy CI code in the repo, including `Makefile`s

I prefer 1, to resolve things asap and let's do "make it better" work then.

mythi:
my 2 cents is that the coverage gap is not critical enough that it cannot wait proper refactoring.

fitzthum (approved):
I guess I am fine with this in the short term. From the CI perspective, we don't really need to lint the components individually. It shouldn't take too much of a refactor to just lint the entire workspace instead. I guess we might want to keep this Make target around tho.

Xynnn007:
@mythi `cargo clippy --workspace --all-targets  -- -D warnings` does not always work because

For AA, different archs like `s390x`, `arm` and `x86` needs different feature for dependencies.

For packages like `image-rs`, `keywrap-grpc` and `keywrap-ttrpc` are inclusive., thus `--all-features` is unuseful.

`--all-targets` means for tests, integration tests, examples, and benchmarks but not for features. `--workspace` means every package. Above all, The feature combinations of features need to be enumerate.

Xynnn007:
wait for https://github.com/confidential-containers/guest-components/pull/1661#pullrequestreview-4935211059 and rebase

mythi:
> @mythi `cargo clippy --workspace --all-targets -- -D warnings` does not always work because

OK, I'm flexible. I've started to use this as a quick check for default features. 

> For AA, different archs like `s390x`, `arm` and `x86` needs different feature for dependencies.

Maybe we are creating complexity for ourselves due to a misconception that `cca-attester` needs to be linted on ARM. Only `nvidia-attester` has strictly arch specific coverage. For `cca-attester`, I'd argue that the `target_arch` switch is an unnecessary optimization also.

> For packages like `image-rs`, `keywrap-grpc` and `keywrap-ttrpc` are inclusive., thus `--all-features` is unuseful.

I did some work related to this in d64fc454bc21364f1ce4d16f5675d27b99f2d061. workspace level both get enabled because image-rs enables the other one by default and CDH enables the other one via image-rs custom features.

> `--all-targets` means for tests, integration tests, examples, and benchmarks but not for features. `--workspace` means every package. Above all, The feature combinations of features need to be enumerate.

Makes sense.

Xynnn007:
> Maybe we are creating complexity for ourselves due to a misconception that `cca-attester` needs to be linted on ARM. Only `nvidia-attester` has strictly arch specific coverage. For `cca-attester`, I'd argue that the `target_arch` switch is an unnecessary optimization also.

Besides nvidia nvat, intel dcap/tpm can only be built for x86. My idea is that the original reason behind this is "this platform has no lint error" and then "let's build on this platform".

1. For x86, `all-attesters` already covers all attesters (even "cca-attester" is included).
2. For `aarch64`, only `cca-attester` will work on that platform thus ok only cover `cca-attester`. 

what's your opinion to optimize on this?

> I did some work related to this in https://github.com/confidential-containers/guest-components/commit/d64fc454bc21364f1ce4d16f5675d27b99f2d061. workspace level both get enabled because image-rs enables the other one by default and CDH enables the other one via image-rs custom features.

Good. This part needs update for a long time.

anyway, let me get this in now. and I will make a new PR for CI tidying to make life easy

PR #1653, opened 2026-08-11
ocicrypt-rs drop unused env::set_var()

mythi:


Xynnn007 (commented):
I get the point that you want to get rid of `set_var` for `unsafe` reasons. 

This PR mainly changes the unit tests, and the `from_env` is a basic routine that ocicrypt(-rs) should support. 

It's ok to keep them as-is in unit test codes as they will not be compiled into binary. We can remove `set_var` parts in real logic. For this PR, you can add `from_file` test cases to UT which is missed in current codebase.

mythi:
> It's ok to keep them as-is in unit test codes as they will not be compiled into binary.

It is true that the PR is triggered by the 2024 edition change. If you read the commit message you notice 003e8ef removes dead code so why not to do the cleanup?

> For this PR, you can add from_file test cases to UT which is missed in current codebase.

I believe his is covered by e9f86f1.

Xynnn007:
Yes. https://github.com/confidential-containers/guest-components/pull/1653/changes/003e8efde06fb4babd027c49eb3cbfda821f4056 is ok.

about https://github.com/confidential-containers/guest-components/pull/1653/changes/e9f86f142159aa4189ebb608a8d2ebeeb4ad60aa, i am hesitating that `from_env` is a function so we usually do UT to cover the blackbox, instead of opening it though. As commit message says, it's only one line difference but breaks blackbox. A new separate `from_file` test would make sense to the code structure.

mythi:
Your comment on the other commit is fair so I just dropped it.

Xynnn007 (approved):

Xynnn007:
Thanks for adopting points!

fitzthum (approved):
LGTM 

looks like I missed the fun part of the discussions

PR #1656, opened 2026-08-12
Remove online-sev-kbc logic and documents

Xynnn007:
Related to #1276

This will be on top of #1602


.github/workflows/coco-extension-image.yml
```diff
@@ -99,8 +99,6 @@ jobs:
       ARCH: ${{ matrix.arch }}
       LIBC: ${{ matrix.libc }}
       ATTESTER: ${{ matrix.attester }}
-      # "kbs" enables cc_kbc, the KBS resource provider used on every arch.
```
Xynnn007, line 102: cc @mythi in the last commit, `RESOURCE_PROVIDER` has been removed because of no use. I bring coco-extension-image change into a separate commit. ptl

mythi: thanks, looks OK

confidential-data-hub/kms/src/plugins/kbs/mod.rs
```diff
@@ -19,14 +19,21 @@ use tokio::sync::Mutex;
...
 
-enum RealClient {
+#[async_trait]
+pub trait Kbc: Send + Sync {
+    async fn get_resource(&mut self, _rid: ResourceUri) -> Result<Vec<u8>>;
+}
+
+/// A fake KbcClient to carry the [`Getter`] semantics. The real `new()`
+/// and `get_resource()` will happen to the static variable [`KBS_CLIENT`].
+pub enum KbcClient {
```
mythi, line 29: would this allow us to rethink the env variable approach and how AA params are passed to the client?

mythi:
As you link 29f3af7 to #1276, I could update the issue rationale a bit to list things why we are doing this.

confidential-data-hub/kms/src/plugins/kbs/mod.rs
Xynnn007: Yes. I plan to do a refactor to inject params from function call than env later.

mythi: sounds great, thanks!

fitzthum (approved):
LGTM. Nice cleanup. There's some more stuff we can remove next.

mythi (approved):
Thanks!

Xynnn007:
Note that we should remove `RESOURCE_PROVIDER` when we bump gc version on kata-side this time.

mythi:
> Note that we should remove `RESOURCE_PROVIDER` when we bump gc version on kata-side this time.

We don't build gc on kata side anymore so this might be just a leftover in the build scripts.

PR #1660, opened 2026-08-14
cdh: reduce KBS client idle connection timeout

kimullaa:
KBS (Actix Web) closes idle keep-alive connections after 5 seconds by default.

This can occasionally cause ttrpc request errors due to the reuse of stale
pooled connections closed by KBS.

Configure a shorter client-side idle timeout to avoid reusing stale connections.

### DIFF
--- a/attestation-agent/kbs_protocol/src/builder.rs
+++ b/attestation-agent/kbs_protocol/src/builder.rs
@@ -89,6 +89,9 @@ impl<T> KbsClientBuilder<T> {
     pub fn build(self) -> Result<KbsClient<T>> {
         let mut http_client_builder = reqwest::Client::builder()
             .cookie_store(true)
+            // KBS uses a 5-second keep-alive timeout by default.
+            // Use a shorter idle timeout to avoid reusing stale pooled connections.
+            .pool_idle_timeout(std::time::Duration::from_secs(4))
             .user_agent(format!(
                 "attestation-agent-kbs-client/{}",
                 env!("CARGO_PKG_VERSION")


attestation-agent/kbs_protocol/src/builder.rs
```diff
@@ -89,6 +89,9 @@ impl<T> KbsClientBuilder<T> {
     pub fn build(self) -> Result<KbsClient<T>> {
         let mut http_client_builder = reqwest::Client::builder()
             .cookie_store(true)
+            // KBS uses a 5-second keep-alive timeout by default.
+            // Use a shorter idle timeout to avoid reusing stale pooled connections.
+            .pool_idle_timeout(std::time::Duration::from_secs(4))
```
Xynnn007, line 94: what's the difference between 4 and 5?

kimullaa: The client timeout just needs to be shorter than the KBS keep-alive timeout (5 seconds). 
I chose 4 seconds to leave some margin.

kimullaa:
The CI failures do not appear to be related to this change, so I created a separate PR to fix the CI issues on main.

https://github.com/confidential-containers/guest-components/pull/1661

Xynnn007:
@kimullaa #1661 is merged

attestation-agent/kbs_protocol/src/builder.rs
Xynnn007: ok. do you have any debug log or so to show some error happened because of this?

another point is why we try this, than to re-do the client request?

kimullaa: > ok. do you have any debug log or so to show some error happened because of this?

Sorry, it looks like this issue only occurs on an older version of the main branch. I'll close this MR.

> another point is why we try this, than to re-do the client request?

That's a good point. Retrying the request on connection errors would likely be a more robust solution.

---

Previously, I could reproduce the issue roughly once every 20 runs with:

```
$ seq 1 100 | xargs -I @@ bash -c "curl http://127.0.0.1:8006/cdh/resource/default/test/file; sleep 4.9"
```

Since this was a TCP-level failure, neither KBS nor CDH produced any meaningful logs indicating the root cause. The issue was only visible from packet captures.

However, after rebasing onto the latest main branch, I have not been able to reproduce the issue anymore. It looks like recent changes have altered the HTTP communication flow(HTTP connections are no longer being reused.).

prev:

```
$ tshark -r dump.pcap  http
  same sport 
   ... 127.0.0.1 → 127.0.0.1    HTTP 4950 GET /kbs/v0/resource/default/test/file HTTP/1.1
   ... 127.0.0.1 → 127.0.0.1    HTTP/JSON 588 HTTP/1.1 200 OK , JSON (application/json)

   ... 127.0.0.1 → 127.0.0.1    HTTP 4950 GET /kbs/v0/resource/default/test/file HTTP/1.1
   ... 127.0.0.1 → 127.0.0.1    HTTP/JSON 588 HTTP/1.1 200 OK , JSON (application/json)

   ... 127.0.0.1 → 127.0.0.1    HTTP 4950 GET /kbs/v0/resource/default/test/file HTTP/1.1
   ... 127.0.0.1 → 127.0.0.1    HTTP/JSON 588 HTTP/1.1 200 OK , JSON (application/json)
```

current(maybe after 1ab81f8a): 

```
$ tshark -r dump.pcap  http
    4   0.000261    127.0.0.1 → 127.0.0.1    HTTP/JSON 351 POST /kbs/v0/auth HTTP/1.1 , JSON (application/json)
    6   0.001658    127.0.0.1 → 127.0.0.1    HTTP/JSON 348 HTTP/1.1 200 OK , JSON (application/json)
    8   3.323894    127.0.0.1 → 127.0.0.1    HTTP/JSON 8403 POST /kbs/v0/attest HTTP/1.1 , JSON (application/json)
    9   3.330641    127.0.0.1 → 127.0.0.1    HTTP/JSON 5014 HTTP/1.1 200 OK , JSON (application/json)
   19   3.334008    127.0.0.1 → 127.0.0.1    HTTP 4950 GET /kbs/v0/resource/default/test/file HTTP/1.1
   21   3.336154    127.0.0.1 → 127.0.0.1    HTTP/JSON 588 HTTP/1.1 200 OK , JSON (application/json)
   29   4.542921    127.0.0.1 → 127.0.0.1    HTTP/JSON 351 POST /kbs/v0/auth HTTP/1.1 , JSON (application/json)
   31   4.544077    127.0.0.1 → 127.0.0.1    HTTP/JSON 348 HTTP/1.1 200 OK , JSON (application/json)
   33   7.742307    127.0.0.1 → 127.0.0.1    HTTP/JSON 8403 POST /kbs/v0/attest HTTP/1.1 , JSON (application/json)
   34   7.749127    127.0.0.1 → 127.0.0.1    HTTP/JSON 5014 HTTP/1.1 200 OK , JSON (application/json)
   44   7.752334    127.0.0.1 → 127.0.0.1    HTTP 4950 GET /kbs/v0/resource/default/test/file HTTP/1.1
   46   7.754381    127.0.0.1 → 127.0.0.1    HTTP/JSON 588 HTTP/1.1 200 OK , JSON (application/json)

```

kimullaa: @Xynnn007  This is not directly related to this MR, but would it make sense to reuse the result of the /attest request instead of performing attestation for every resource request?

Xynnn007: To reuse the attestation token (result of `/attest` request) is ok. There are some trade-off for the caller.
1. Not reusing means that always the latest tcb will be verified by KBS.
2. Using means quicker, but there will be an attacking window (expiration time for the token, defaults to 5 min) for malwared TCB updates

The token expiration time can be configured by trustee deployer to apply in concrete scenario, see [duration_min](https://github.com/confidential-containers/trustee/blob/main/kbs/docs/config.md#attestationtokenbroker).

kimullaa: Thanks for the additional information, that's very helpful.

Since I've confirmed that this issue cannot actually occur, I'll close this MR.
(I may open a separate MR for reuse in the future, but I'll first consider whether it makes sense and whether it would make the code unnecessarily complex.)

PR #1662, opened 2026-08-14
Use libnvat-dev package instead of building from source

fitzthum:
Building the nvat dependencies from source is fairly painful. Instead, we can use the libnvat-dev package. The only downside here is that the package version does not programmatically map to the crate version. The crate release notes do list the expected package version and the APIs are stable between major versions.


fitzthum:
Package has some issues building on 26.04, so let's keep building from source for the extension image for now. I will fix that when the next version of the package is released.

Xynnn007 (approved):
LGTM! just curiosity why different version number mechanism?

fitzthum:
> just curiosity why different version number mechanism?

You mean why does the package have one scheme and the crate/repo have another? I'm not totally sure. Looks like there is a time-based release of the entire repo (which includes the source of the package). That is fairly common at NVIDIA. Ofc it's a little tricky because that repo contains a few different components which would have different versions if using semver. So even if we had a nice semver version for the rust crate (and it were on crates.io), it probably wouldn't match the package version.

PR #1664, opened 2026-08-15
kms: add unit tests for plugin dispatch, offline-fs KBC and AWS client

rafsal-rahim:
Add serial_test and tempfile to [dev-dependencies] so that tests touching the filesystem and environment variables can run safely in parallel with the rest of the suite.

Three areas are covered:

plugins/mod.rs
  - new_getter / new_decryptor return an error for an unknown provider name and for an empty string provider name.

plugins/kbs/offline_fs.rs
  - init_with_file is a no-op when the path does not exist (matches the production fallback behaviour).
  - init_with_file returns an error for malformed JSON.
  - init_with_file populates the resource map from a valid file.
  - a second call to init_with_file with the same key silently overwrites the previous value without returning an error.

plugins/aws/client.rs
  - from_provider_settings returns an error when the credential file is absent.
  - from_provider_settings returns an error when the credential file contains malformed JSON.
  - from_provider_settings returns an error when the region field is missing from the provider settings.


Xynnn007 (approved):

chathuryaadapa (approved):

fitzthum (approved):
Ok. 

I'm not totally sure about the value of these sort of tests, but seems fine.

PR #1670, opened 2026-08-17
CI: tidy the CI matrix and job files

Xynnn007:
Current guest-components CI jobs are split in many different files. Some tests are overlapped. This PR's aim is to refactor  the CI and test code with the following idea

1. Let `attestation-agent`/`confidential-data-hub`/`image-rs`/`ocicrypt-rs` have their own Make commands to do lint check and test. The rust features will be determined by the arch and can be overwritten.
2. Combine duplicated CI jobs within same file to keep code dry. Rust suites will directly call the Makefile provided by 1. (except one case for `powerpc64le` cross-compilation test)
3. To avoid complex kbc feature matrix and also usuability, make `offline_fs_kbc` and `sample_kbc` into native code than under features.

Please see concrete commit for messages.


Xynnn007:
Updated. Rebased with latest action versions.
This PR is a continue of https://github.com/confidential-containers/guest-components/pull/1652

attestation-agent/kbc/src/lib.rs
```diff
@@ -15,16 +15,11 @@ use serde::{Deserialize, Serialize};
 
 pub use self::annotation_packet::AnnotationPacket;
 
-// Add your specific kbc declaration here.
-// For example: "pub mod sample_kbc;"
 #[allow(dead_code)]
 #[cfg(feature = "cc_kbc")]
 pub mod cc_kbc;
 
-#[cfg(feature = "offline_fs_kbc")]
```
fitzthum, line 24: Is there any risk to enabling this all the time? It reads from the guest fs.

attestation-agent/kbc/src/lib.rs
```diff
@@ -15,16 +15,11 @@ use serde::{Deserialize, Serialize};
...
 pub mod offline_fs_kbc;
-
-#[cfg(feature = "sample_kbc")]
```
fitzthum, line 27: And is there any risk to enabling this? It means that an attacker would know that a fixed key will be available. Wonder if that could be dangerous in any case.

Btw I think this and the fs kbs aren't both needed. We could probably get rid of one of them, but that's a separate question.

image-rs/src/image.rs
```diff
@@ -400,7 +399,6 @@ impl ImageClient {
             }
         }
 
-        #[cfg(feature = "signature")]
```
fitzthum, line 403: This seems good. Being able to disable signatures entirely seems like a bad thing. I wonder if there is still a potential footgun where we disable both cosign and simple signing. Do those need to be features?

image-rs/Cargo.toml
```diff
@@ -27,7 +27,7 @@ hex.workspace = true
 kbc = { path = "../attestation-agent/kbc", default-features = false, optional = true }
 loopdev = { git = "https://github.com/mdaffin/loopdev", rev = "c9f91e8f0326ce8a3364ac911e81eb32328a5f27" }
 nix = { workspace = true, features = ["mount", "fs"] }
-oci-client = { git = "https://github.com/oras-project/rust-oci-client", rev = "83092a67ebcaeed832eae090895a1c178d867344", default-features = false, optional = true }
+oci-client = { git = "https://github.com/oras-project/rust-oci-client", rev = "83092a67ebcaeed832eae090895a1c178d867344", default-features = false }
```
fitzthum: I think we may be able to use `v0.17.0` here.

fitzthum (commented):
Seems pretty good. A lot of different changes here. 

I have a couple questions about the features.

attestation-agent/kbc/src/lib.rs
Xynnn007: it just enable the module it supports. Which one is selected is decided by runtime config (aa param).
But your idea inspired me that do we still need other definitions like `KbcInstantiateFunc`? I will take a look and remove them if no imported by others anymore.

attestation-agent/kbc/src/lib.rs
Xynnn007: see above.

re whether this and fs kbc are needed. curretly the code path of image-rs/ocicrypt-rs still uses them for decryption key retrievement in `keywrap-native` feature. What we are using is `keywrap-ttrpc` to call CDH to get the key.

I think `keywrap-native` is still useful if a developer does not have CDH but want to save the keys locally.

image-rs/src/image.rs
Xynnn007: they are already features. see `signature-simple` and `signature-cosign`

attestation-agent/kbc/src/lib.rs
Xynnn007: removed.

mythi:
This PR seems to have 11 commits and only one of them is about CI. Let's split the work into smaller PRs.

ba35fd027af82b4f91451935ce46e92e1d7a0b10 removes changes that were only added one day earlier. What changed in 24 hours?

Xynnn007:
> This PR seems to have 11 commits and only one of them is about CI. Let's split the work into smaller PRs.

Ok. The CI change reported a lot of clippy errors, thus I fixed them in other commits. 



> [ba35fd0](https://github.com/confidential-containers/guest-components/commit/ba35fd027af82b4f91451935ce46e92e1d7a0b10) removes changes that were only added one day earlier. What changed in 24 hours?

They are doint different jobs. One is to bring up test coverage and the other is to dedup the code. I am ok to combine them into one and let the commit message cover the two things.

Xynnn007:
Let me squash the commits and put as separate PRs.

PR #1674, opened 2026-08-20
ci: publish CoCo extension images for pull requests

noeljackson:
## Why

Kata needs a published CoCo extension image to test a guest-components pull request before it merges. The existing pull-request workflow builds the images, but only uploads workflow artifacts; fork pull requests cannot publish directly because their `GITHUB_TOKEN` is read-only.

## How

Reuse `coco-extension-image.yml` for both normal builds and trusted pull-request artifact promotion:

- the existing `pull_request` path still builds without registry credentials and uploads its six matrix artifacts;
- a small `workflow_run` entry point calls the existing workflow with the completed run ID and a `pr-<head-sha>` version;
- the called workflow skips checkout and all PR-controlled build steps, validates and restores the artifacts, then uses the existing login, publish, attestation, release, and manifest logic;
- pull-request images are published as `pr-<head-sha>-ubuntu<release>`; and
- main, latest, and release tag behavior is unchanged.

A Kata branch can set `.externals.coco-guest-components.version` to, for example, `pr-<head-sha>-ubuntu24.04`.

The PR-scoped tags cannot replace a main or release tag. They remain in GHCR after the source PR closes; registry retention can be handled separately if their volume becomes material.

## Validation

- `actionlint` passes for the new workflow and reports only the existing custom `s390x` runner-label warning for the reused workflow.
- `zizmor --persona auditor .` reports no findings.
- The artifact contract was checked against all six outputs from successful CoCo PR workflow run 32327104540.
- A downloaded arm64 OCI archive was copied successfully with the same `skopeo` source transport used by promotion, and its disk artifacts were validated and restored under Kata's canonical filenames.

### DIFF
--- a/.github/workflows/coco-extension-image.yml
+++ b/.github/workflows/coco-extension-image.yml
@@ -15,13 +15,15 @@ name: Build and publish the CoCo guest extension image
 # release asset is qualified with it (e.g. "<sha>-ubuntu26.04-amd64"): the
 # extension ships no libc, so a variant is only usable on a matching guest.
 #
-# On pull requests everything is built and uploaded as workflow artifacts (no
-# registry push). On pushes to main the artifacts are published to ghcr.io
-# (tagged with the commit sha and "latest") and attested. On a published
-# release the same artifacts are published tagged with the release version, the
-# multi-arch manifests get the version tag, and the per-arch disk images and
-# their dm-verity root hashes are attached to the GitHub release so consumers
-# (e.g. kata-containers) can pin to a stable version.
+# On pull requests everything is built without registry credentials and
+# uploaded as workflow artifacts. A separate workflow_run publisher promotes a
+# successful build under a PR-scoped tag so consumers such as kata-containers
+# can test the proposed guest-components revision. On pushes to main the
+# artifacts are published to ghcr.io (tagged with the commit sha and "latest")
+# and attested. On a published release the same artifacts are published tagged
+# with the release version, the multi-arch manifests get the version tag, and
+# the per-arch disk images and their dm-verity root hashes are attached to the
+# GitHub release so consumers can pin to a stable version.
 on:
   pull_request:
   push:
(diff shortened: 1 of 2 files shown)


fitzthum:
I think the core idea here is that we want to make it easier for people to test out their extension images with Kata before merging the PR.

Currently, we build the extension images on PRs, but we don't publish them; partly, because workflows running on forks don't have the permissions to push artifacts.

So we could do this hybrid approach where we have another workflow that pushes the images. Another idea would be to create a `gc-committers` team that grants write access to people. Then they could open PRs again the repo directly and we could publish the images.

Xynnn007:
What I am worried about is that potentially a number of debug images will be pushed to ghcr thus occupying the space and make it messy.

A question would be: can kata repo specify the image source to your folk than the this repo? I mean, you can have a kata fork specifying your gc fork. This is my dev experience, though it's for peerpods than kata. What do you think of this way?

noeljackson:
> What I am worried about is that potentially a number of debug images will be pushed to ghcr thus occupying the space and make it messy.
> 
> 
> 
> A question would be: can kata repo specify the image source to your folk than the this repo? I mean, you can have a kata fork specifying your gc fork. This is my dev experience, though it's for peerpods than kata. What do you think of this way?

That could be a possibility let me see if it will work.

mythi:
> > What I am worried about is that potentially a number of debug images will be pushed to ghcr thus occupying the space and make it messy.
> > A question would be: can kata repo specify the image source to your folk than the this repo? I mean, you can have a kata fork specifying your gc fork. This is my dev experience, though it's for peerpods than kata. What do you think of this way?
> 
> That could be a possibility let me see if it will work.

It should be possible to pass any path https://github.com/kata-containers/kata-containers/blob/7739b6262fccf27fcc7afd4fa39de5b4fb0b53a0/versions.yaml#L294. It's then up to kata maintainers to accept the PR workflows to run for arbitrary extension images.

PR #1676, opened 2026-08-21
Add configfs-tsm Support for AMD SEV-SNP Attester

amd-cmaxwell:
Purpose:
---
The Linux kernel introduced configfs-tsm (Trusted Security Module via configfs) as a standardized, vendor-neutral interface for TEE attestation, available at /sys/kernel/config/tsm/report/. This replaces the need for each TEE platform to expose its own bespoke kernel interface for attestation.

Intel TDX has already adopted configfs-tsm as its primary attestation path in the guest-components repository. AMD SEV-SNP currently uses only the legacy path via the sev Rust crate, which communicates with the kernel's /dev/sev-guest driver. The goal of this task is to bring SNP to parity with TDX by adding configfs-tsm support.

Summary:
---
- Add support for Linux kernel `configfs-tsm` (`sysfs`) as the preferred path for generating SEV-SNP attestation reports, with 
- Update `detect_platform()` to recognize SEV-SNP via configfs-tsm in addition to the legacy `/sys/devices/platform/sev-guest` path
- Update `get_snp_host_data()` to use configfs-tsm when available, falling back to the sev crate's Firmware API otherwise
- Add the `tsm-report` feature as a dependency of snp-attester in `Cargo.toml`
- Add debug/error logging across the SNP attester module

Potential Breaking changes
---
- Change visibility of constant `attester::snp::VMPL` to `pub`
- Add new variant to `GetHostDataError`:
  - `GetHostDataError::GetTsmReportError(#[from] TsmReportError)`
- Change enum variant `TsmReportData::Sev(u8, Vec<u8>)` to be `TsmReportData::Sev(Vec<u8>)`

Dependent PRs:
---
https://github.com/confidential-containers/guest-components/pull/1696


attestation-agent/attester/src/snp/mod.rs
```diff
@@ -42,15 +42,20 @@ pub struct SnpAttester {
 
 impl Default for SnpAttester {
     fn default() -> Self {
+        debug!("Creating default SnpAttester...");
```
Xynnn007: We might not need this one; also, for rust if `new()` does not return a `Result`, we usually have only `default()` to work as `new()`. It means we can leave only `default()` w/o `new()`

attestation-agent/attester/src/snp/hostdata.rs
```diff
@@ -4,22 +4,48 @@
...
+        let mut tsm_report_path = TsmReportPath::new(TsmReportProvider::Sev)?;
+
+        // Generate the attestation report from the TSM Report path.
+        report_bytes = tsm_report_path
+            .attestation_report(TsmReportData::Sev(VMPL, report_data.to_vec()))?;
+    } else {
+        tracing::debug!("No file found at TSM_REPORT_PATH: {}", TSM_REPORT_PATH);
+        tracing::debug!("sysfs for SEV-SNP is not supported, which requires kernel version >= 6.16.");
+        let mut firmware = Firmware::open()?;
+        report_bytes = firmware.get_report(None, Some(report_data), Some(0))?;
```
Xynnn007: We can reuse the `VMPL` here for `Some(0)`. Thus `VMPL` could be a mod level const.

attestation-agent/attester/src/snp/mod.rs
```diff
@@ -26,24 +30,105 @@ struct SnpEvidence {
...
+    /// and the registers.
+    supports_tsm_measurements: bool,
+}
+
+impl Default for SnpAttester {
+    fn default() -> Self {
+        debug!("Creating default SnpAttester...");
+        Self::new()
+    }
+}
```
Xynnn007: Let's move all the `new()` stuff inside `default`, and only leave `default()` function

attestation-agent/attester/src/snp/mod.rs
```diff
@@ -26,24 +30,105 @@ struct SnpEvidence {
...
+        Self {
+            supports_tsm_measurements,
+        }
+    }
+}
 
 #[async_trait::async_trait]
 impl Attester for SnpAttester {
     async fn get_evidence(&self, mut report_data: Vec<u8>) -> Result<TeeEvidence> {
+        debug!("Getting evidence using {:?} ...", self);
```
Xynnn007: Note that we already have debug info later for which (tsm/sev) mechanism we will use. Thus no need for this

attestation-agent/attester/src/snp/mod.rs
```diff
@@ -26,24 +30,105 @@ struct SnpEvidence {
...
         if report_data.len() > 64 {
             bail!("SNP Attester: Report data must be no more than 64 bytes");
         }
 
         report_data.resize(64, 0);
 
-        let mut firmware = Firmware::open()?;
-        let data = report_data.as_slice().try_into()?;
+        // The VMPL value is set to 0, which means the report is generated at the highest privilege level.
+        const VMPL: u8 = 0;
```
Xynnn007: see comment above

attestation-agent/attester/src/snp/mod.rs
```diff
@@ -26,24 +30,105 @@ struct SnpEvidence {
...
+
+        // The attestation report generated by the SEV-SNP firmware, which includes the report data, measurement registers, and other information.
+        let report: Vec<u8>;
```
Xynnn007: we do not need this. For rust we can direct use `let report = xxx` for both declaration and assign

attestation-agent/attester/src/snp/mod.rs
```diff
@@ -26,24 +30,105 @@ struct SnpEvidence {
...
+
+        // The certificate chain generated by the SEV-SNP firmware, which includes the platform certificate, the CEK certificate, and the VCEK certificate.
+        let certs: Option<Vec<CertTableEntry>>;
```
Xynnn007: also this one

attestation-agent/attester/src/snp/mod.rs
```diff
@@ -26,24 +30,105 @@ struct SnpEvidence {
...
+                            error!("Failed to parse certificate table data: {:?}", e);
+                            e
+                        })?;                    
+                    debug!("Certificates before sorting: {:?}", certificates);
+                    certificates.sort();
+                    debug!("Certificates after sorting: {:?}", certificates);
+                    Some(certificates) 
+                }
+            };
+            debug!("Certificate chain set to: {:?}", certs);
```
Xynnn007: I guess you can try this

```suggestion
let certs = cert_table_data
            .map(|data| -> Result<Vec<CertTableEntry>> {
                let mut certificates: Vec<CertTableEntry> =
                    CertTableEntry::vec_bytes_to_cert_table(&mut data).map_err(|e| {
                        error!("Failed to parse certificate table data: {:?}", e);
                        e
                    })?;
                certificates.sort();
                Ok(certificates)
            })
            .transpose()?;
```

attestation-agent/attester/src/tsm_report/mod.rs
```diff
@@ -103,27 +107,147 @@ impl TsmReportPath {
             return Err(TsmReportError::InblobLen);
         }
 
-        std::fs::write(report_path.join("inblob"), report_data)
-            .map_err(|e| TsmReportError::Access("inblob", e))?;
+        debug!("Attempting to write Report Data to inblob: {:?}", &report_data);
```
Xynnn007: We might not need these. If surely needed, `trace!` would be preferred.

attestation-agent/attester/src/tsm_report/mod.rs
```diff
@@ -103,27 +107,147 @@ impl TsmReportPath {
...
+        self.write_inblob(&report_data)?;
+        debug!("Wrote Report Data to inblob: {:?}", &report_data);
 
+        debug!("Attempting to read outblob from report path: {:?}", report_path);
```
Xynnn007: ditto

attestation-agent/attester/src/tsm_report/mod.rs
```diff
@@ -103,27 +107,147 @@ impl TsmReportPath {
...
         let q = std::fs::read(report_path.join("outblob"))
-            .map_err(|e| TsmReportError::Access("outblob", e))?;
+            .map_err(|e| {
+                tracing::error!("Failed to read outblob from report path: {:?}, error: {:?}", report_path, e);
+                TsmReportError::Access("outblob", e)
+            })?;
+        debug!("Read outblob from report path: {:?}: {:?}", report_path, q);
```
Xynnn007: ditto

attestation-agent/attester/src/tsm_report/mod.rs
```diff
@@ -154,10 +133,108 @@ impl TsmReportPath {
...
+            .parse::<u32>()
+            .map_err(TsmReportError::Parse);
+        
+        generation
+    }
+
+    /// check_tsm_report_generation checks if the expected generation of this TsmReportPath instance matches the generation of the Configfs-TSM report maintained by the kernel.
+    /// Returns an error if the expected generation does not match the current generation in the kernel.
+    /// 
+    /// This is meant to detect concurrent modifications to the TsmReportPath instance by other processes.
```
Xynnn007: Does this mean a race and concurrency problem? Could you take an example to explain this?

Xynnn007 (commented):
Thanks @amd-cmaxwell for this. Could you help to
1. squash the commits into one or two
2. remove the debug macros - another way is to downgrade them to `tracing`. I prefer to remove them at all as error messages would bring enough message.

attestation-agent/attester/src/snp/mod.rs
```diff
@@ -26,24 +30,105 @@ struct SnpEvidence {
...
+        debug!("Creating default SnpAttester...");
+        Self::new()
+    }
+}
+
+impl SnpAttester {
+    pub fn new() -> Self {
+        debug!("Creating new SnpAttester...");
+
+        let supports_tsm_measurements = Path::new(TSM_REPORT_PATH).exists();
```
mythi: Tdx used something like: https://github.com/confidential-containers/guest-components/blob/b4c6900f72f9ab7e491aca84a0c869e820d394bf/attestation-agent/attester/src/tdx/mod.rs#L132-L141

Let's use a similar structure and not expose `TSM_REPORT_PATH`.

attestation-agent/attester/src/snp/hostdata.rs
```diff
@@ -4,22 +4,48 @@
...
+
+        // Get and verify the TSM Report path instance for SEV-SNP.
+        let mut tsm_report_path = TsmReportPath::new(TsmReportProvider::Sev)?;
+
+        // Generate the attestation report from the TSM Report path.
+        report_bytes = tsm_report_path
+            .attestation_report(TsmReportData::Sev(VMPL, report_data.to_vec()))?;
+    } else {
+        tracing::debug!("No file found at TSM_REPORT_PATH: {}", TSM_REPORT_PATH);
+        tracing::debug!("sysfs for SEV-SNP is not supported, which requires kernel version >= 6.16.");
```
mythi: nit: tsm-reports was added in 6.7 so it's been there for a looong time (even in 24.04 lts). maybe you read Tdx code which suggested 6.16 but that's for a different feature and not related to TSM reports.

attestation-agent/attester/src/tsm_report/mod.rs
```diff
@@ -79,22 +86,19 @@ impl TsmReportPath {
...
+        let report_path = self.path.clone();
 
         let report_data = match provider_data {
             TsmReportData::Cca(inblob) => inblob,
             TsmReportData::Tdx(inblob) => inblob,
             TsmReportData::Sev(privlevel, inblob) => {
-                // TODO: untested
-                std::fs::write(report_path.join("privlevel"), vec![privlevel])
-                    .map_err(|e| TsmReportError::Access("privlevel", e))?;
+                self.set_privlevel(privlevel)?;
```
mythi: we get `VMPL0` by default so would it be possible we skip `privlevel` write completely for the initial transition.  With that, only this match arm needs updating. A later PR could add the support for other vmpl levels (if needed)

attestation-agent/attester/src/tsm_report/mod.rs
mythi: Setting `privlevel` increments `generation` too it seems so the value becomes 2 for this Sev flow.

attestation-agent/attester/src/snp/mod.rs
amd-cmaxwell: @Xynnn007 Can you explain the reason behind this change? This appears to call `CertTableEntry::vec_bytes_to_cert_table()` on every byte in the arrary.

Xynnn007: Oh. The original idea is to make the code shorter. But the `cert_table_data` is `Vec<_>` than `Option`. Let me resolve this comment.

attestation-agent/attester/src/tsm_report/mod.rs
```diff
@@ -79,73 +88,233 @@ impl TsmReportPath {
...
         &self,
         provider_data: TsmReportData,
     ) -> Result<Vec<u8>, TsmReportError> {
-        let report_path = self.path.as_path();
+        let report_path = self.path.clone();
 
         let report_data = match provider_data {
             TsmReportData::Cca(inblob) => inblob,
             TsmReportData::Tdx(inblob) => inblob,
-            TsmReportData::Sev(privlevel, inblob) => {
```
mythi, line 94: My comment was to keep this as but simply handle privlevel != 0 as an error and remove  `std::fs::write(report_path.join("privlevel")`. Why this PR still needs to modify `generation` count if you only write `inblob` like others.

mythi (changes requested):
this isn't ready for review I think. it needs all the debug prints removed and code formatted properly.

attestation-agent/attester/src/tsm_report/mod.rs
amd-cmaxwell: >this isn't ready for review I think.

Yes, I have not re-requested review yet as I am still addressing the previous comments.

mythi: OK, sorry, I saw a push so I thought there's another rev. would it be good to move this back to draft then?

amd-cmaxwell: > would it be good to move this back to draft then?

that probably won't be necessary

>but simply handle privlevel != 0 as an error

Sure, but why allow for a privlevel to be specified at all if we aren't going to write the value to the report?

Can't we change `TsmReportData::Sev(u8, Vec<u8>)` to be `TsmReportData::Sev(Vec<u8>)` since that tuple structure already matches the other variants of `TsmReportData` enum?


> [we get `VMPL0` by default so would it be possible we skip `privlevel` write completely for the initial transition. With that, only this match arm needs updating. A later PR could add the support for other vmpl levels (if needed)](https://github.com/confidential-containers/guest-components/pull/1676#discussion_r3841245240)

I will be separating the logic to set the report VMPL & track the report generation to a new PR.

mythi: > Can't we change `TsmReportData::Sev(u8, Vec<u8>)` to be `TsmReportData::Sev(Vec<u8>)`

If the goal is to support `privlevel` setting in the future, I'd keep it as it is. If you think it's never going to be useful the change works for me

mythi:
Let us know when it's ready for another review round.

amd-cmaxwell:
It is ready for another round of review

fitzthum:
btw @amd-cmaxwell does the confifgs support the derived key API? I seem to recall that it does not which might conflict with https://github.com/confidential-containers/guest-components/issues/1678 (although issue might not be going anywhere).

attestation-agent/attester/src/tsm_report/mod.rs
```diff
@@ -92,9 +95,21 @@ impl TsmReportPath {
             TsmReportData::Cca(inblob) => inblob,
             TsmReportData::Tdx(inblob) => inblob,
             TsmReportData::Sev(privlevel, inblob) => {
```
mythi: we should avoid the dependency to `snp-attester` in this module.
```suggestion
            TsmReportData::Sev(privlevel, inblob) => {
                const DEFAULT_VMPL: u8 = 0;
```

attestation-agent/attester/src/snp/mod.rs
```diff
@@ -38,12 +46,57 @@ impl Attester for SnpAttester {
...
+
+                    // Use the TSM Report path to generate the attestation report and get the certificate chain.
+
+                    // Get and verify the TSM Report path instance for SEV-SNP.
+                    let tsm_report_path: TsmReportPath = TsmReportPath::new(TsmReportProvider::Sev)
+                        .context("SEV-SNP Attester: failed to create TSM Report path")?;
+
+                    // Generate the attestation report from the TSM Report path.
+                    // The attestation report generated by the SEV-SNP firmware, which includes the report data, measurement registers, and other information.
+                    let report: Vec<u8> = tsm_report_path
```
mythi: `hostdata.rs` had it like this and it should work here too:
```suggestion
                    let report: Vec<u8> = tsm
```

attestation-agent/attester/src/tsm_report/mod.rs
```diff
@@ -106,8 +121,13 @@ impl TsmReportPath {
         std::fs::write(report_path.join("inblob"), report_data)
             .map_err(|e| TsmReportError::Access("inblob", e))?;
 
-        let q = std::fs::read(report_path.join("outblob"))
-            .map_err(|e| TsmReportError::Access("outblob", e))?;
+        let q = std::fs::read(report_path.join("outblob")).map_err(|e| {
+            error!(
+                "Failed to read outblob from report path: {:?}, error: {:?}",
+                report_path, e
+            );
```
mythi: can we trace these errors on the caller side?

amd-cmaxwell:
> btw @amd-cmaxwell does the confifgs support the derived key API? I seem to recall that it does not which might conflict with #1678 (although issue might not be going anywhere).

@fitzthum I am not sure, I will ask around, 
this is the document i was following for tsm support https://www.kernel.org/doc/Documentation/ABI/testing/configfs-tsm

attestation-agent/attester/src/tsm_report/mod.rs
hgowda-amd: @mythi @amd-cmaxwell I agree with keeping this PR focused on the core configfs-tsm SNP path.

Since, we are not writing privlevel for VMPL0 I would recommend to remove the privlevel-related logic from this PR rather than keeping guards or fields we don’t use yet. Privlevel handling and generation tracking can come in a follow-up PR. That keeps this smaller and easier to review.

attestation-agent/attester/src/tsm_report/mod.rs
amd-cmaxwell: These errors bubble up through 3-4 function calls across multiply files before being logged by 
`error!("AA (ttrpc): get evidence failed:\n {e:?}");` in `attestation-agent::ttrpc_dep::server::AA::get_evidence()`. 
This proved notable difficult to trace on the caller side while debugging so I logged them close to the source for convenience.

Additionally the repr of `TsmReportError::Access` 
```
#[error("Failed to access TSM Report attribute: {0} ({1})") ]
Access(&'static str, #[source] std::io::Error),
```
does not expose the `report_path`, which is the first thing I would want to know while debugging a file read/write issue.

I will remove these logging statements if that's what's necessary, but perhaps error tracing should be revisited in the future.

attestation-agent/attester/src/tsm_report/mod.rs
amd-cmaxwell: Changed `TsmReportData::Sev(u8, Vec<u8>)` to be `TsmReportData::Sev(Vec<u8>)` for the time being

attestation-agent/attester/src/snp/mod.rs
mythi: One clarification: I meant that you could take |_tsm| -> tsm as the tsm report instance

attestation-agent/attester/src/tsm_report/mod.rs
mythi: > does not expose the report_path, which is the first thing I would want to know while debugging a file read/write issue

The tsm report configfs attribute is included in the error. There is little you can do with the full path because the temp dir is removed once the TsmReport instance goes out of scope but if there is a case where it's useful we can add it in a separate commit/pr

amd-cmaxwell:
@fitzthum you are corrrect Configfs TSM does not support the derived key API

amd-cmaxwell:
@mythi @hgowda-amd Can we mark this ready for test?

hgowda-amd:
> @mythi @hgowda-amd Can we mark this ready for test?

@mythi or Fabiano should be able to add "ok-to-test" flag.

fitzthum:
Don't really need that label in this repo. Just need a maintainer to approve the tests since it's your first contribution to the repo.

amd-cmaxwell:
All tests have passed and feedback applied. Can we please green-light this for merge? @Xynnn007 @mythi

attestation-agent/attester/src/snp/mod.rs
```diff
@@ -38,12 +46,57 @@ impl Attester for SnpAttester {
 
         report_data.resize(64, 0);
 
-        let mut firmware = Firmware::open()?;
-        let data = report_data.as_slice().try_into()?;
+        let (report, certs): (Vec<u8>, Option<Vec<CertTableEntry>>) =
+            TsmReportPath::new(TsmReportProvider::Sev).map_or_else(
+                |e| {
+                    warn!("TSM Report not supported. Attestation report data will be obtained via the sev Rust crate instead. Reason: {}", e);
```
Xynnn007: There is also a term named `report data`. To avoid confusion let's just call `report` here.

Also, for users they might not care about the crate but the tech, so i recommend to use `ioctl` here

```suggestion
                    warn!("TSM Report not supported. Attestation report will be obtained via ioctl. Reason: {}", e);
```

attestation-agent/attester/src/snp/mod.rs
```diff
@@ -38,12 +46,57 @@ impl Attester for SnpAttester {
...
+                            certificates.sort();
+                            Some(certificates)
+                        }
+                    };
 
-        let (report, certs) = firmware
-            .get_ext_report(None, Some(data), Some(0))
-            .context("Failed to get attestation report")?;
+                    Ok((report, certs))
+                }
```
Xynnn007, line 98: I saw some overlap logic between here and the initdata part. We could reuse some part. But ok to refactor in later PR.

Xynnn007 (approved):
Only some nits. LGTM

amd-cmaxwell:
@mythi can you please approve or remove your blocking change request?

attestation-agent/attester/src/snp/mod.rs
```diff
@@ -38,12 +46,57 @@ impl Attester for SnpAttester {
...
+                },
+                |_tsm|{
+                    debug!("TSM Report for SEV-SNP is supported. The TSM Report path will be used.");
+
+                    // Use the TSM Report path to generate the attestation report and get the certificate chain.
+
+                    // Get and verify the TSM Report path instance for SEV-SNP.
+                    let tsm: TsmReportPath = TsmReportPath::new(TsmReportProvider::Sev)
+                        .context("SEV-SNP Attester: failed to create TSM Report path")?;
+
```
mythi, line 70: no need to instantiate two paths:
```suggestion
                |tsm|{
                    debug!("TSM Report for SEV-SNP is supported. The TSM Report path will be used.");

                    // Use the TSM Report path to generate the attestation report and get the certificate chain.
```

attestation-agent/attester/src/tsm_report/mod.rs
mythi: OK

attestation-agent/attester/src/snp/mod.rs
amd-cmaxwell: This redundancy is fixed in the subsequent PR #1696.

mythi: my review comment was to fix it in this PR

mythi (approved):
LGTM

PR #1677, opened 2026-08-22
image-rs: add opt-in anonymous fallback for rejected credentials

joaosa:
## Summary

When image-rs is configured with registry credentials, a rejected credential (e.g. an expired token) will fail the image pull for public images. In a measured guest, where the credential is baked into the rootfs, one expired token will break every pull on the node.

This PR adds the `anonymous_fallback_on_unauthorized` config switch. When enabled, a pull that fails with `UnauthorizedError` under non-anonymous auth is retried once with a fresh anonymous client. This flag is disabled by default, so default behavior is unchanged, i.e. a rejected credential will fail the image pull.

### DIFF
--- a/image-rs/src/image.rs
+++ b/image-rs/src/image.rs
@@ -5,6 +5,7 @@
 use anyhow::{bail, Context};
 use oci_client::{
     client::{Certificate, CertificateEncoding, ClientConfig, ClientProtocol},
+    errors::OciDistributionError,
     manifest::{OciDescriptor, OciImageManifest},
     secrets::RegistryAuth,
     ParseError, Reference,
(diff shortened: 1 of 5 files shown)


Xynnn007 (commented):
Hi @joaosa The code looks fine. I only have an overall design comment. From the architecture design view, whether to raise an error when the given auth is wrong has different answers based on different needs. So could you help to add a switch in image-rs's config? probably named `anonymous_fallback_on_unauthorized` and give it a default value if not provided.

mkulke:
I think we aim for consistent behaviour with container utilities like podman or docker, afaiu those would fail similarly and not attempt anonymous pulls? 

(please consider the project's AI policy when creating submissions: https://github.com/confidential-containers/confidential-containers/blob/main/AI.md#dont-use-ai-for-prose)

joaosa:
Thank you for the feedback @Xynnn007 @mkulke . My bad missing the project's AI policy 🙏🏼 . 
Rewrote the PR desc and added the suggested `anonymous_fallback_on_unauthorized` config switch.

Xynnn007 (approved):
LGTM. 

FWIW, making it configurable and defaulting to `false` would not change anything, but provide flexibility. wdyt? @mkulke @fitzthum

fitzthum (approved):
LGTM

image-rs/src/image.rs
```diff
@@ -372,17 +341,77 @@ impl ImageClient {
...
+                    RegistryAuth::Anonymous
+                }
+            },
+        };
+
+        let mut client = self.new_pull_client(&task, &auth)?;
 
-        let (image_manifest, image_digest, image_config, manifest_list_digest) =
-            client.pull_manifest().await?;
+        let (image_manifest, image_digest, image_config, manifest_list_digest) = match client
```
mkulke, line 395: would a if/let construct make this more readable?

```rust
let mut result = client.pull_manifest().await;

if let Err(PullLayerError::PullManifestError { source: OciDistributionError::UnauthorizedError { .. } }) = &result 
    && self.config.anonymous_fallback_on_unauthorized
    && !matches!(auth, RegistryAuth::Anonymous)
{
      warn!("registry rejected the configured credential for {image_url}; retrying anonymously");

      client = self.new_pull_client(&task, &RegistryAuth::Anonymous)?;
      result = client.pull_manifest().await;
}

let (image_manifest, image_digest, image_config, manifest_list_digest) = result?;
```

mkulke (commented):
small nit

image-rs/src/image.rs
joaosa: Done in 4a356153. Edition 2021 has no let chains, so the error is checked with `matches!` on the result instead of `if let ... &&`, same shape otherwise.

joaosa: that does look cleaner. rework this bit

mkulke (approved):

PR #1684, opened 2026-08-24
AA: align all-attesters across workspace

Xynnn007:
## Summary

Split from #1670. The `all-attesters` feature set was incomplete and wired inconsistently across AA crates, so attesters we claim to support were not always compiled under `make ATTESTER=all-attesters`. Aligns the feature graph so build and lint exercise the same set.

Helped with [Cursor](https://cursor.com)


mkulke (approved):

attestation-agent/attester/Cargo.toml
```diff
@@ -73,8 +73,10 @@ required-features = ["bin"]
 
 [features]
 default = ["all-attesters"]
+
 all-attesters = [
     "tdx-attester",
+    "sgx-attester",
```
mythi, line 79: we had c5d58c46c164fad3954fdebf5835c9c46b1500cd but I don't remember why I kept AA/KBS protocol features enabled.

Xynnn007: There are two design ways
1. `all-attesters` means enabling all attesters, regardless it's for VM or process. Note that we do not build binaries with this feature, usually Makefile helps us to select specific attester.
2. change `all-attesters` to `cvm-attesters`. Thus semantically match the behavior if `sgx-sttester` is not included.

Xynnn007: Anyway, this is a little change. If no strong blocker, i'd like to get it merged asap.

mythi (approved):

PR #1685, opened 2026-08-24
KBC: drop unused registry APIs and wire offline_fs into native

Xynnn007:
## Summary

Split from #1670. After AA moved secret handling to CDH, KBC still carried a plugin-registry API that nothing calls anymore. Removes that dead surface and routes native keywrap through offline_fs_kbc, which tests and local workflows already depend on.

Helped with [Cursor](https://cursor.com)

### DIFF
--- a/ocicrypt-rs/src/keywrap/keyprovider/native.rs
+++ b/ocicrypt-rs/src/keywrap/keyprovider/native.rs
@@ -6,11 +6,15 @@
 use std::sync::{Arc, LazyLock};
 
 use anyhow::*;
-use kbc::{cc_kbc::Kbc as CcKbc, sample_kbc::SampleKbc, AnnotationPacket, KbcInterface};
+use kbc::{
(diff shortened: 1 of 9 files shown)


ocicrypt-rs/src/keywrap/keyprovider/native.rs
```diff
@@ -21,6 +25,7 @@ async fn initialize_channel(kbs_addr: &str, kbc: &str) -> Result<()> {
     let channel = match kbc {
         "cc_kbc" => Kbc::Cc(CcKbc::new(kbs_addr.to_owned())?),
         "sample_kbc" => Kbc::Sample(SampleKbc::new(kbs_addr.to_owned())),
+        "offline_fs_kbc" => Kbc::OfflineFs(OfflineFsKbc::new()),
```
fitzthum, line 28: btw we do need to be careful that the default ocicrypt config never includes entries forthe sample or offline_fs providers.

fitzthum (approved):
LGTM 

Good simplifications.

PR #1687, opened 2026-08-24
ocicrypt-rs: make RPIT captures explicit

mythi:
Make the return-position impl Trait (RPIT) capture sets explicit for the synchronous and asynchronous layer readers before migrating to edition 2024.

The opaque return types capture only the input reader type R and, for encrypt_layer, its existing 'a lifetime. Configuration, annotations, key-option data, and digest inputs are used only to construct owned cipher state and are not borrowed by the returned readers. Precise use<...> bounds preserve this contract under Rust 2024's broader RPIT capture rules.

These changes can be generated with:

    cargo fix --edition --allow-dirty -p ocicrypt-rs --lib \
        --no-default-features \
        --features async-io,block-cipher-openssl

Commit message was proof-read by Copilot but the changes are from a manual cargo fix

### DIFF
--- a/ocicrypt-rs/src/encryption.rs
+++ b/ocicrypt-rs/src/encryption.rs
@@ -211,7 +211,7 @@ pub fn encrypt_layer<'a, R: 'a + Read>(
     annotations: Option<&BTreeMap<String, String>>,
     digest: &str,
 ) -> Result<(
-    Option<impl Read + EncryptionFinalizer + 'a>,
+    Option<impl Read + EncryptionFinalizer + 'a + use<'a, R>>,
     EncLayerFinalizer,
 )> {
     let mut encrypted = false;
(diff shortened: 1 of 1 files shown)


Xynnn007 (approved):

mkulke (approved):
uh this is tricky, is my understanding correct?

`Option<impl Read + EncryptionFinalizer + 'a + use<'a, R>>`

is required to make the code compile/lint on both Rust 2021 and 2024, but when we migrate to Rust 2024 we can drop `'a` and `+ 'a + use<'a, R>` from the signature completely b/c in 2024 everything will be captured?

mythi:
this is tricky indeed and to be honest I spent way too long trying to understand what is going on. I did not check for any build failures, this is what `cargo fix --edition` suggested.

let me try: with `use<'a, R>` and 2024 we tell the borrow checker that all inputs (2024 default) don't have to be captured

mythi:
two approvals and CI is green so let me get this merged

PR #1699, opened 2026-08-28
Improve image-rs logging for signed images

fitzthum:
Two little tweaks to expose more information when an image signature fails. Previously you would get something like `Error: Image policy rejected: Denied by policy: rejected by sigstoreSigned rule` regardless of whether the policy failed or a resource required to fulfill the policy was unavailable. This should help distinguish those cases. 

There shouldn't be any risk of revealing secret stuff in the log here.

### DIFF
--- a/image-rs/src/signature/mod.rs
+++ b/image-rs/src/signature/mod.rs
@@ -33,7 +33,7 @@ pub enum SignatureError {
     #[error("Illegal image digest: {0}")]
     IllegalImageDigest(String),
 
-    #[error("Denied by policy: {source}")]
+    #[error("Denied by policy: {source:#}")]
     DeniedByPolicy {
         #[source]
         source: anyhow::Error,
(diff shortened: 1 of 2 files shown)


image-rs/src/signature/mod.rs
```diff
@@ -33,7 +33,7 @@ pub enum SignatureError {
     #[error("Illegal image digest: {0}")]
     IllegalImageDigest(String),
 
-    #[error("Denied by policy: {source}")]
+    #[error("Denied by policy: {source:#}")]
```
Xynnn007, line 36: I guess you mean `{source:#?}`?

fitzthum: Well I think `?` will add all the debug info on top of the air messages. That wasn't what I was thinking, but can add it if you would prefer.

Xynnn007: Oh it's fine to keep it as-is. I once thought `source:#` is illegal but I found this https://doc.rust-lang.org/std/fmt/#sign0 .

Xynnn007 (approved):

PR #1683, opened 2026-08-24
Makefile: add workspace fmt/lint/test make targets

Xynnn007:
## Summary

Split from #1670. Local lint/test today uses different feature knobs than build (especially in attestation-agent), so `make ATTESTER=… test` does not always check what we ship.

Helped with [Cursor](https://cursor.com)

PR #491, opened 2024-02-29
image-rs: Add cargo build for multiarch

stevenhorsman:
- Add building image-rs on s390x and powerpc64le
Note: At the moment this will probably fail with:
```
error: failed to run custom build command for `ring v0.16.20`

Caused by:
  process didn't exit successfully: `/root/guest-components/target/debug/build/ring-f931d4c26d8deee2/build-script-build` (exit status: 101)
  --- stderr
  thread 'main' panicked at /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ring-0.16.20/build.rs:358:10:
  called `Option::unwrap()` on a `None` value
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
whilst we wait for #422 to get fixed, so we shouldn't merge it until after that is fixed

PR #1706, opened 2026-09-01
ci: consolidate Rust CI and binary release workflows

Xynnn007:
Split from #1670. Follow-up to #1683–#1686. Rust CI had grown into many overlapping workflows with duplicated setup and release jobs. Consolidates fmt/clippy/test and binary publishing behind path-filtered workflows that call the workspace make targets from #1683.

Assisted-by Cursor


Xynnn007:
cc @confidential-containers/guest-components-maintainers 

This would be a useful change.

.github/workflows/rust.yml
```diff
@@ -0,0 +1,443 @@
...
+
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+          fetch-depth: 0
+
+      - uses: dorny/paths-filter@ceb8a2b8f2d89434be7ff52d3de7ec3738c5cc9d # v4.0.3
+        id: filter
+        with:
+          filters: |
```
mythi, line 50: do you have any numbers how much these conditionals save time? I'm thinking if we could simplify the flow even further and check everything unconditionally (with proper caches the overhead may not be that much)

.github/workflows/rust.yml
```diff
@@ -0,0 +1,443 @@
...
+
+      - name: Build
+        working-directory: ./image-rs
+        run: make build
+
+      - name: Cross-compile for powerpc64le
+        if: matrix.instance == 'ubuntu-24.04'
+        run: |
+          sudo -E PATH=$PATH -s RUSTFLAGS=" -C linker=powerpc64le-linux-gnu-gcc" \
+            cargo build --target powerpc64le-unknown-linux-gnu -p image-rs --features default
```
mythi, line 342: What is the use-case for this? I forget. Looks like it's also `image-rs` only...

.github/workflows/rust.yml
```diff
@@ -0,0 +1,443 @@
...
+          eval make ttrpc=true ${MAKE_ARGS} && make install PREFIX=${HOME}/.local
+
+      - name: Run rust lint check
+        working-directory: ./attestation-agent
+        run: sudo -E PATH=$PATH -s make lint
+
+      - name: Run rust lint check with openssl
+        if: matrix.instance == 'ubuntu-24.04'
+        working-directory: ./attestation-agent
+        run: sudo -E PATH=$PATH -s make OPENSSL=1 lint
```
mythi: does `lint` need sudo?
```suggestion
        run: make OPENSSL=1 lint
```

mythi (commented):
the other commit message is slightly outdated (`sev` related mentions)

.github/workflows/rust.yml
Xynnn007: It's used to test ppc64 building. https://github.com/confidential-containers/guest-components/blob/main/.github/workflows/image_rs_build.yml#L107

.github/workflows/rust.yml
Xynnn007: The conditions here is [not something new](https://github.com/confidential-containers/guest-components/blob/main/.github/workflows/aa_basic.yml#L7-L11).

Back to your question, let's divide into two parts: the PR latency (how long the CI will run), and the total consumption (the time sum for all CI jobs).

From recent full-matrix runs, the PR latency dependens the most time-consumption job：
1. [cold run](https://github.com/confidential-containers/guest-components/actions/runs/34797699390) it lasted about 33 minutes
2. [warm run with hitting cache](https://github.com/confidential-containers/guest-components/actions/runs/33585685754) about 16 minutes

for the total consumption, the conditionals here help most on single-crate and non-Rust PRs. With my AI helper compiling recent merges, it showed that roughly half are Cargo.lock/workspace and still run everything, so average savings are maybe ~40–50%. 

Unconditional + cache would mainly increase the total consumption of runners, but nearly no influence on PR latency.

.github/workflows/rust.yml
Xynnn007: it comes from https://github.com/confidential-containers/guest-components/blob/main/.github/workflows/aa_basic.yml#L114. I can remove them and see if it still works.

.github/workflows/rust.yml
mythi: Yes, no need to change anything here but it looks odd we get CI coverage for image-rs only.

Xynnn007: Historically, it's brought together with s390x https://github.com/confidential-containers/guest-components/pull/491/changes for image-rs only. Then s390x adds more matrix with self-hosted runner.

It might be better to have a full routine for ppc64le separately within the matrix, than live together with x86-64?
If you agree, I can make it in a separate PR.

.github/workflows/rust.yml
mythi: > The conditions here is [not something new](https://github.com/confidential-containers/guest-components/blob/main/.github/workflows/aa_basic.yml#L7-L11).

What's new is that we have to pull a 3rd party workflow. Maybe it's not that bad but the alternative would be to keep using `paths` for the workflow trigger to filter out everything not relevant to Rust and just run everything. 

> Unconditional + cache would mainly increase the total consumption of runners, but nearly no influence on PR latency.

same runners could be reused between workspace members?

.github/workflows/rust.yml
mythi: makes sense

Xynnn007: https://github.com/confidential-containers/guest-components/issues/1734

.github/workflows/rust.yml
Xynnn007: > What's new is that we have to pull a 3rd party workflow. Maybe it's not that bad but the alternative would be to keep using `paths` for the workflow trigger to filter out everything not relevant to Rust and just run everything.

Yes. This can also be a case. To sum up, different ways
1. Use `dorny/paths-filter` to do fine grained Rust path filters, avoid useless Rust suite running. The pro is computation resource control. The con is 3rd dep and some more code lines.
2. Use `paths` to coarse-grained filter rust and non-rust runs. The pro is easy and no 3rd dep. The con is extra consumption for total computation resource.

In most cases, the PR latency would be the same. But for corner cases, like only `ocicrypt-rs` is changed, way 1 would have less time than way 2. The reason is that now the most time-cost job is for image-rs, and in such case way 1 would skip image-rs runs.

I am ok with both. wdyt?

> same runners could be reused between workspace members?

Do you mean, e.g. to run 
```bash
make lint -C attestation-agent
make test -C attestation-agent
make lint -C confidential-data-hub
make test -C confidential-data-hub
...
```
On a same runner?

Now we use rust cache for jobs, which will share both crates and `target`. In the new way would save some of the duplication building process, but the TODO list for each job would be longer. Up to now no tests are taken to compare with the previous version, but I am afraid that it will take longer.

Still, they are about the optimization of the whole CI pipeline. Maybe we can do this in a separate thread. This PR mainly aims to refactor the code and to follow the original logic.

mythi: > On a same runner?

Yes, that's what I had in mind.

> Still, they are about the optimization of the whole CI pipeline. Maybe we can do this in a separate thread. This PR mainly aims to refactor the code and to follow the original logic.

This sounds OK to me. now that everything gets moved to a single `rust.yml` updates are simpler.

mythi (approved):
LGTM

PR #1710, opened 2026-09-01
Improve secret-cli packaging 

fitzthum:
We have a number of little issues with how we release the secret-cli.

First, make sure that the artifact just contains the binary. Previously,
we were releasing a tarred directory containing the binary. There is no
need for that.

Next, let's make it a multi-arch image and make the tags a little
simpler (just use `latest` or the release version).

Next, let's not add these artifacts to the release. We don't add any of
our other binaries to the release. No need for this one, which is one of
the least prominent.

Finally, let's rename the package (on ghcr) from secret_cli to
secret-cli. This is more consistent with our other packages.

It turns out the name of the binary is actually just `secret` (not `secret-cli`).
Update the docs and the source in a few places to be more consistent.
We can discuss renaming the tool, but for now let's keep it as is.

### DIFF
--- a/.github/workflows/secret_cli_release.yml
+++ b/.github/workflows/secret_cli_release.yml
@@ -18,14 +18,10 @@ permissions:
   contents: read
 
 jobs:
-  build-and-upload:
-    name: Build and upload secret CLI
+  build-and-publish:
+    name: Build and publish secret CLI
     permissions:
-      # Write scopes are only exercised by the release-only upload/publish
-      # steps below (gated on env.IS_PUBLISH). On pull_request the job just
-      # builds the binary to catch breakage early; fork PRs get a read-only
-      # token regardless of what is requested here.
-      contents: write # required to upload release assets via gh release upload
+      # Write only used for releases, when the binary is pushed to ghcr
       packages: write # required to publish OCI artifact to ghcr.io
     strategy:
       fail-fast: false
@@ -40,8 +36,8 @@ jobs:
     runs-on: ${{ matrix.platform.runner }}
     env:
       RUST_TARGET: ${{ matrix.platform.arch }}-unknown-linux-${{ matrix.platform.libc }}
-      # KMS providers bundled into the released binary. Building all of them
-      # into one artifact keeps the released tool provider-agnostic.
+      OCI_ARCH: ${{ matrix.platform.arch == 'x86_64' && 'amd64' || 'arm64' }}
+      # Build all KMS providers
       CLI_FEATURES: cli,aliyun,aws,kbs
       REGISTRY: ghcr.io
       IMAGE_NAME: ${{ github.repository }}
@@ -65,7 +61,7 @@ jobs:
       if: env.IS_PUBLISH == 'true'
       uses: oras-project/setup-oras@1d808f7d7f6995cc68b7bf507bfe5c5446e1dc9d # v2.0.1
       with:
-        version: 1.2.0
+        version: 1.3.3
 
     - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       with:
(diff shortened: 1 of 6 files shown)


.github/workflows/secret_cli_release.yml
```diff
@@ -102,32 +98,47 @@ jobs:
...
-          "${image}:${TAG}-${RUST_TARGET},latest-${RUST_TARGET}" \
-          "${ASSET}"
+        image="${REGISTRY}/${IMAGE_NAME}/secret-cli"
+        for tag in "${TAG}" latest; do
+          docker manifest create "${image}:${tag}" \
+            --amend "${image}:${TAG}-x86_64-unknown-linux-musl" \
+            --amend "${image}:${TAG}-aarch64-unknown-linux-musl"
+          docker manifest annotate --arch amd64 --os linux "${image}:${tag}" "${image}:${TAG}-x86_64-unknown-linux-musl"
+          docker manifest annotate --arch arm64 --os linux "${image}:${tag}" "${image}:${TAG}-aarch64-unknown-linux-musl"
+          docker manifest push "${image}:${tag}"
```
Xynnn007: We have `oras manifest index create` [command](https://oras.land/docs/commands/oras_manifest_index_create) for this, which might be more native

.github/workflows/secret_cli_release.yml
```diff
@@ -18,14 +18,10 @@ permissions:
...
-    name: Build and upload secret CLI
+  build-and-publish:
+    name: Build and publish secret CLI
     permissions:
-      # Write scopes are only exercised by the release-only upload/publish
-      # steps below (gated on env.IS_PUBLISH). On pull_request the job just
-      # builds the binary to catch breakage early; fork PRs get a read-only
-      # token regardless of what is requested here.
-      contents: write # required to upload release assets via gh release upload
+      # Write only used for releases, when the binary is push to ghcr
```
Xynnn007:
```suggestion
      # Write only used for releases, when the binary is pushed to ghcr
```

Xynnn007 (commented):
Otherwise looks to me, only one thing about release binary. I think the cli tools or so is ok to be left in release binaries, as they can be directly downloaded and used. wdyt?

.github/workflows/secret_cli_release.yml
fitzthum: good idea

fitzthum:
>  I think the cli tools or so is ok to be left in release binaries, as they can be directly downloaded and used. wdyt?

I just find it a bit inconsistent that the only tool we include in any of our releases is the secret-cli. I can add it back if you want but I think it's a bit confusing for users to have some stuff attached to the release (but not any of the things they actually want).

Xynnn007:
> I just find it a bit inconsistent that the only tool we include in any of our releases is the secret-cli. I can add it back if you want but I think it's a bit confusing for users to have some stuff attached to the release (but not any of the things they actually want).

Yes. We now have guest-components that's not released via release binaries. CoCo keyprovider is container image because skopeo and other suites are also needed thus complex.

I wonder if we can publish user-facing tools (CLIs), like this one, and maybe aa-cli, cdh-cli binary in future. Looks like CLIs are usually published in other projects. GC binaries do not need to.

fitzthum:
> I wonder if we can publish user-facing tools (CLIs), like this one, and maybe aa-cli, cdh-cli binary in future. Looks like CLIs are usually published in other projects. GC binaries do not need to.

Yeah, in theory it might make sense to have the CLI tools as part of the release artifacts. I just think we should do that as a separate topic; decide if we want the CLIs here and then add all of them if we do. We added the secret-cli here without really discussing it much.

.github/workflows/secret_cli_release.yml
```diff
@@ -65,7 +61,7 @@ jobs:
       if: env.IS_PUBLISH == 'true'
       uses: oras-project/setup-oras@1d808f7d7f6995cc68b7bf507bfe5c5446e1dc9d # v2.0.1
       with:
```
mythi, line 63: Could we just use the default version so that we don't have to bump it in two places here?

mythi (commented):
LGTM but one question

Xynnn007 (approved):
Let's package binaries into release in a separate PR

Xynnn007:
Also there will be overlapping with https://github.com/confidential-containers/guest-components/pull/1706, i can do rebase if this gets in first

.github/workflows/secret_cli_release.yml
fitzthum: You mean leave off the hash? The CI image doesn't have oras installed and we have two different jobs here, so we install it twice, unless I am missing something.

mythi: I was thinking we could drop:
```
with:
    version: a.b.c
```

if the tool version is not relevant (we use the default from the pinned action).

fitzthum: oh. got it. good idea.

Issue #1513, opened 2026-06-12
RFC: ccm_kbc - CDH-Side KBC Plugin for Vendor Verifier + KMS Deployments


prabhanjan-gururaj:
# RFC: ccm_kbc — A CDH-Side KBC Plugin for Vendor Verifier + KMS Deployments

---

## Summary

This RFC proposes adding a fourth KBC implementation, `ccm_kbc`, as a sub-module under `confidential-data-hub/kms/src/plugins/kbs/`, alongside the existing `cc_kbc`, `online_sev_kbc`, and `offline_fs_kbc`. It implements CDH's local `Kbc` trait and is selected at runtime via `AA_KBC_PARAMS=ccm_kbc::default`. The plugin links the upstream `attester` crate as an in-process library for TEE evidence, sends the resulting quote + ephemeral CSR to a remote verifier (Fortanix CCM) for a short-lived X.509 leaf certificate, and uses that certificate as a client identity for mTLS to a remote key-management service (Fortanix DSM) for resource release.

The contribution is small in surface area — one match arm in `kms/src/plugins/kbs/mod.rs`, one new Cargo feature, feature-gated new code and additive: no existing component (AA, image-rs, ocicrypt-rs, the `Kbc` trait, any sibling KBC) is modified. The broader argument is that **broadening the set of deployment shapes CoCo supports — adding vendor-managed verifier + KMS integrations alongside the existing self-operated Trustee path — accelerates enterprise adoption by letting each organization match CoCo's components to its own operational and compliance posture**. This plugin establishes the pattern that other vendors can follow.

---

### Two deployment shapes, both legitimate

The CoCo project ships a attestation and resource-release stack centered on Trustee (verifier + KBS). Some organizations may prefer to deploy Trustee themselves and operate it.

A different class of organization will want to consume the verifier and KMS as services they don't operate, integrated with infrastructure they already run. For these organizations and a large share of the regulated-enterprise market that CoCo aims to reach at scale — three recurring patterns shape the deployment decision:

**1. KMS integration with existing certified infrastructure.** Most regulated industries already operate a FIPS KMS that's been through extensive internal review, external audit, and regulatory acceptance. Confidential-container keys ideally live inside that established compliance perimeter rather than alongside it. 

[...]

PR #1722, opened 2026-09-07
Fortanix ccm_kbc and ccm_as integration

kiran-kannur:
**What**

Implements the `ccm_as` / `ccm_kbc` design proposed in #1513. This allows a confidential guest to release a key from Fortanix DSM, with the release gated on successful attestation to Fortanix CCM.

This adds two paired, opt-in plugins:

### `attestation-agent`

Adds a new `ccm_as` token type alongside the existing `kbs` and `coco_as` token types.

The Attestation Agent (AA) performs local attestation to CCM over vsock, with SEV-SNP and TDX supported today. Once attestation succeeds, CCM returns a short-lived workload X.509 certificate and private key, which the AA exposes as the token.

The credential is cached for the lifetime of the AA process. The AA only performs attestation again when the cached certificate expires.

### `confidential-data-hub`

Adds a new `ccm_kbc` KBS client, alongside the existing `cc_kbc`, `online_sev_kbc`, and `offline_fs_kbc` implementations.

For resource URIs in the form:

```text
kbs:///dsm/key/<uuid>
```

the KBC:

1. Fetches the `ccm_as` credential from the AA over ttrpc.
2. Uses that credential as an mTLS client identity when connecting to DSM.
3. Requests the specified key from DSM.

Both plugins are opt-in through the `ccm_as` and `ccm_kbc` Cargo features. They are disabled by default, so existing builds don't incur any additional compile-time or runtime cost, and existing KBCs and token types continue to behave exactly as before.

### Why `ccm_kbc` doesn't use the attester directly

The attestation and evidence collection are handled entirely by `ccm_as` on the AA side.

`ccm_kbc` only needs to:

* Fetch the credential from AA over ttrpc.
* Present that credential to DSM.

This uses the same ttrpc mechanism already used for CDH ↔ AA token fetching and follows the existing two-phase model used by `cc_kbc`: the AA handles attestation and produces a credential, while the KBC presents that credential to the actual KMS.

## Out of scope

The `coco_keyprovider` changes are intentionally not part of this PR.

This includes:

* DSM-wrapped-KEK image encryption.
* The corresponding CDH-side `decrypt_with_kek` / `Decrypter` support.

Those changes will be handled in a follow-up PR.

For this PR, `ccm_kbc` only implements the `kbs:///dsm/key/<uuid>` resource-release path through `get_resource()`. It does not add image-decryption support.

Ref : [#1513](https://github.com/confidential-containers/guest-components/issues/1513)

Assisted-by: Claude Sonnet 5

### DIFF
--- a/attestation-agent/Makefile
+++ b/attestation-agent/Makefile
@@ -60,14 +60,14 @@ else
 endif
 
 ifeq ($(ATTESTER), none)
-    AA_FEATURES := kbs,coco_as,bin,$(RPC),$(CRYPTO)
-    AA_CHECK_FEATURES := kbs,coco_as,bin,ttrpc,grpc,$(CRYPTO)
+    AA_FEATURES := kbs,coco_as,ccm_as,bin,$(RPC),$(CRYPTO)
+    AA_CHECK_FEATURES := kbs,coco_as,ccm_as,bin,ttrpc,grpc,$(CRYPTO)
     ATTESTER_PKG_FEATURES := bin
     KBC_FEATURES := cc_kbc,$(CRYPTO)
     KBS_PROTOCOL_FEATURES := background_check,passport,$(CRYPTO),bin,aa_ttrpc,pqc-experimental
 else
-    AA_FEATURES := kbs,coco_as,bin,$(RPC),$(CRYPTO),$(ATTESTER)
-    AA_CHECK_FEATURES := kbs,coco_as,bin,ttrpc,grpc,$(CRYPTO),$(ATTESTER)
+    AA_FEATURES := kbs,coco_as,ccm_as,bin,$(RPC),$(CRYPTO),$(ATTESTER)
+    AA_CHECK_FEATURES := kbs,coco_as,ccm_as,bin,ttrpc,grpc,$(CRYPTO),$(ATTESTER)
     ATTESTER_PKG_FEATURES := $(ATTESTER),bin
     KBC_FEATURES := cc_kbc,$(ATTESTER),$(CRYPTO)
     KBS_PROTOCOL_FEATURES := background_check,passport,$(CRYPTO),$(ATTESTER),bin,aa_ttrpc,pqc-experimental
@@ -115,7 +115,7 @@ ifneq ($(RUSTFLAGS_ARGS),)
 endif
 
 build:
-	cd attestation-agent && $(RUST_FLAGS) cargo build $(release) --no-default-features --features "$(AA_FEATURES)" $(binary) $(LIBC_FLAG)
+	cd attestation-agent && env -u DESTDIR $(RUST_FLAGS) cargo build $(release) --no-default-features --features "$(AA_FEATURES)" $(binary) $(LIBC_FLAG)
 	mv $(TARGET_DIR)/$(binary_name) $(TARGET)
 
 TARGET := $(TARGET_DIR)/$(BIN_NAME)
--- a/attestation-agent/attestation-agent/Cargo.toml
+++ b/attestation-agent/attestation-agent/Cargo.toml
@@ -47,6 +47,16 @@ toml.workspace = true
 tonic = { workspace = true, optional = true }
 ttrpc = { workspace = true, features = ["async"], optional = true }
 
+# SEV-SNP/TDX/NVIDIA attestation and CCM workload-certificate issuance, used
+# by the ccm_as token type.
+attestation = { git = "https://github.com/fortanix/attestation.git", rev = "fade7bf57c93fefaeaca77fe453b87a7909c9cd4", default-features = false, features = [
+    "sev-guest",
+    "tdx-guest",
+    "nvidia-evidence",
+], optional = true }
+client = { git = "https://github.com/fortanix/attestation.git", rev = "fade7bf57c93fefaeaca77fe453b87a7909c9cd4", optional = true }
+x509-cert = { version = "0.2.5", features = ["pem"], optional = true }
+
 [dev-dependencies]
 rstest.workspace = true
 serial_test.workspace = true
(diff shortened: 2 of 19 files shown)


attestation-agent/attestation-agent/Cargo.toml
```diff
@@ -47,6 +47,16 @@ toml.workspace = true
...
 ttrpc = { workspace = true, features = ["async"], optional = true }
 
+# SEV-SNP/TDX/NVIDIA attestation and CCM workload-certificate issuance, used
+# by the ccm_as token type.
+attestation = { git = "https://github.com/fortanix/attestation.git", rev = "fade7bf57c93fefaeaca77fe453b87a7909c9cd4", default-features = false, features = [
+    "sev-guest",
+    "tdx-guest",
+    "nvidia-evidence",
+], optional = true }
+client = { git = "https://github.com/fortanix/attestation.git", rev = "fade7bf57c93fefaeaca77fe453b87a7909c9cd4", optional = true }
```
beraldoleal, line 57: I couldn't find this commit on master, only on coco-integration. Do you plan to merge that branch into master at some point?

Just a bit worried about pinning the build to a commit that only lives on a working branch.

confidential-data-hub/kms/src/plugins/kbs/ccm_kbc/aa_token_client.rs
```diff
@@ -0,0 +1,43 @@
...
+
+const TIMEOUT_NANOS: i64 = 50 * 1000 * 1000 * 1000;
+
+#[derive(Deserialize)]
+pub(super) struct CcmAsCredential {
+    pub cert_pem: String,
+    pub key_pem: String,
+}
+
+pub(super) async fn get_ccm_as_credential(aa_socket: &str) -> Result<CcmAsCredential> {
```
beraldoleal, line 26: nit:  duplicates AATokenProvider::get_token in kbs_protocol/src/token_provider/aa/mod.rs, just a different TokenType. Maybe factor this into a shared helper instead?

attestation-agent/attestation-agent/src/config/mod.rs
```diff
@@ -111,6 +114,10 @@ pub struct TokenConfigs {
     /// This config item is used when `kbs` feature is enabled.
     #[cfg(feature = "kbs")]
     pub kbs: Option<kbs::KbsConfig>,
+
+    /// This config item is used when `ccm_as` feature is enabled.
+    #[cfg(feature = "ccm_as")]
+    pub ccm_as: Option<ccm_as::CcmAsConfig>,
```
beraldoleal, line 120: Adding this field without #[serde(default)] breaks the 8 existing TokenConfigs { ... } test literals in this file's mod tests. Fails with E0063: missing field.

Unless I'm missing something, `make check` and `make test` are broken by this PR as-is.

attestation-agent/Makefile
```diff
@@ -60,14 +60,14 @@ else
...
+    AA_FEATURES := kbs,coco_as,ccm_as,bin,$(RPC),$(CRYPTO)
+    AA_CHECK_FEATURES := kbs,coco_as,ccm_as,bin,ttrpc,grpc,$(CRYPTO)
     ATTESTER_PKG_FEATURES := bin
     KBC_FEATURES := cc_kbc,$(CRYPTO)
     KBS_PROTOCOL_FEATURES := background_check,passport,$(CRYPTO),bin,aa_ttrpc,pqc-experimental
 else
-    AA_FEATURES := kbs,coco_as,bin,$(RPC),$(CRYPTO),$(ATTESTER)
-    AA_CHECK_FEATURES := kbs,coco_as,bin,ttrpc,grpc,$(CRYPTO),$(ATTESTER)
+    AA_FEATURES := kbs,coco_as,ccm_as,bin,$(RPC),$(CRYPTO),$(ATTESTER)
+    AA_CHECK_FEATURES := kbs,coco_as,ccm_as,bin,ttrpc,grpc,$(CRYPTO),$(ATTESTER)
```
beraldoleal, line 70: Not blaming this PR, since its just following the convention... but just wondering... as more vendor plugins get added this way the AA binary just keeps growing no matter what you actually use.

Maybe worth making this opt-in at some point?

cc @Xynnn007 , @mythi

attestation-agent/attestation-agent/src/config/mod.rs
kiran-kannur: You're right, I missed this. I'll add the fix by adding the field to each of the 8 literals

attestation-agent/attestation-agent/Cargo.toml
kiran-kannur: Currently, we have a few internal dependencies that still run on the `master` branch, and we're gradually migrating them to the new changes. Until that migration is complete, the changes for this integration will remain on the `coco-integration` branch. Once the migration is complete, we plan to merge the integration changes into `master`.

attestation-agent/Makefile
kiran-kannur: Makes sense, any thoughts here @Xynnn007 , @mythi  ?

fitzthum:
Please see community [AI Policy](https://github.com/confidential-containers/confidential-containers/blob/main/AI.md). It will help us to review your PR if you more deliberately craft your PR description and commit messages, especially since there are still a number of unsettled design questions per https://github.com/confidential-containers/guest-components/issues/1513

beraldoleal:
@kiran-kannur, please, rebase this branch to solve the conflicts?

Could you also fold the follow-up fix commits into the commits they correct, and avoid merge commits in the PR history?

kiran-kannur:
@beraldoleal Done. Do you still have a concern with : https://github.com/confidential-containers/guest-components/pull/1722#discussion_r3961432738 ?

PR #1723, opened 2026-09-08
fix(coco_keyprovider): fail when the KBS rejects a KEK registration

tjw-ecora:
## Problem

`register_kek` discards the HTTP response. The `?` on `send()` covers **transport** errors only, so a KBS answering 401, 403 or 5xx returns `Ok(())` — and the caller then logs `"register KEK succeeded."`

```rust
let _ = client
    .post(resource_url)
    .header("Content-Type", "application/octet-stream")
    .bearer_auth(token)
    .body(key)
    .send()
    .await?;

Ok(())
```

## Why it matters

The failure is silent and surfaces a long way from its cause. ocicrypt still wraps and the image is still pushed, so the encryption pipeline reports success — with **no key registered in the KBS**. Nothing is noticed until much later, at runtime, when the guest attests *successfully* and then key release finds nothing to release. It therefore reads as an attestation or guest problem rather than a key-registration one.

The trigger is ordinary rather than exotic: the registration JWT is minted with a two-hour expiry a few lines above (`Claims::create(Duration::from_hours(2))`), so a long-lived keyprovider outlives its own credential and every subsequent registration 401s, silently. A wrong or stale `--auth-private-key` does the same.

## Change

Check the status and `bail!` with the status, the kid and the response body, so an operator learns which registration failed and why.

This appears to be an isolated omission rather than a pattern — every other HTTP call site in the repository already matches on `res.status()`: `attestation-agent/src/token/coco_as.rs`, and `kbs_protocol`'s `token_client.rs` and `rcar_client.rs` (all three sites).

## Tests

Three cases against a one-shot loopback HTTP listener — no network access and no new runtime dependency:

- `401` → error, carrying status, kid and body
- `5xx` → error
- `2xx` → success

Confirmed to be a real regression guard rather than a vacuous test: with the status check removed, the two failure cases fail; with it, all pass. `accepts_success` passes either way, as it should.

`tokio` gains `macros`/`net` as a **dev-dependency only**, so the shipped binary is unchanged.

## Verification

`cargo test -p coco_keyprovider` — 10 passed on this branch. `cargo clippy -p coco_keyprovider --all-targets` and `cargo fmt --check` clean.

Also verified against `v0.19.0`, where the defect is byte-identical, since that is the release we consume.


tjw-ecora:
Resubmitted as #1736 — this was closed in error (the commit was missing its DCO sign-off, and I closed the PR rather than respinning it). The resubmission is rebased on current `main`, signed off, and carries the same fix and tests.

PR #1724, opened 2026-09-09
fix(coco_keyprovider): enable a TLS backend so an https KBS is reachable

tjw-ecora:
## Problem

`coco_keyprovider` is built with no TLS backend, so it cannot register a KEK against a KBS that serves TLS.

The workspace pins reqwest with `default-features = false`:

```toml
# Cargo.toml
reqwest = { version = "0.13", default-features = false }
```

and `coco_keyprovider` inherits that verbatim:

```toml
# attestation-agent/coco_keyprovider/Cargo.toml
reqwest.workspace = true
```

Nothing ever turns a TLS feature back on for this crate, so `https` is not a supported scheme in the built binary. Passing `--kbs https://...` fails before a request leaves the process:

```
register KEK failed
  0: error sending request for url (https://<kbs>/kbs/v0/resource/<repo>/<type>/<tag>)
  1: client error (Connect)
  2: invalid URL, scheme is not http
```

A KBS terminating TLS is the default posture — `insecure_http` is opt-in — so the sanctioned auto-registration flow (`--auth-private-key` + `--kbs`) cannot complete against one today. A port-forward does not help: it relocates a TLS endpoint that the client still cannot dial, and a CA bundle is useless to a client with no TLS stack compiled in.

The failure also surfaces mid-encrypt, on the `WrapKey` path, rather than at startup, so it reads as a KBS problem rather than a missing client feature.

## Fix

Enable `reqwest`'s `rustls` feature for this crate, rather than widening the workspace pin, so the TLS posture of the other crates that inherit it is unchanged.

This follows what the workspace already does elsewhere:

- `attestation-agent/kbs_protocol/Cargo.toml` maps `rust-crypto = ["reqwest/rustls", ...]` and `openssl = ["reqwest/native-tls-vendored", ...]`
- `confidential-data-hub/kms/Cargo.toml` enables `reqwest/rustls` directly

I have taken the `kms` shape (enable it directly) because `coco_keyprovider` has no crypto-backend feature split to hang it off, and it always needs TLS. If you would rather it mirror `kbs_protocol` and offer a `rustls` / `openssl` pair, say so and I will respin.

## Private and internal CAs

reqwest 0.13's `rustls` feature pulls in `rustls-platform-verifier`, which on Linux calls `rustls_native_certs::load_native_certs()`. That honours `SSL_CERT_FILE` and `SSL_CERT_DIR`, so a KBS fronted by a private or internal CA is reachable by pointing those at the CA — with no roots bundled into the binary.

## No Cargo.lock change

The rustls stack is already in the committed lock, because other workspace members enable it. reqwest's own lock entry already lists `rustls`, `rustls-platform-verifier`, `hyper-rustls` and `tokio-rustls`.

```
$ cargo metadata --format-version 1 --locked >/dev/null && git status --porcelain
 M attestation-agent/coco_keyprovider/Cargo.toml
```

Cargo.lock is untouched.

## Testing

```
$ cargo test -p coco_keyprovider --locked
   Compiling rustls-platform-verifier v0.6.2
   Compiling hyper-rustls v0.27.7
   Compiling reqwest v0.13.2
running 7 tests
test result: ok. 7 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

Verified in the built binary, not just the source: a release build of `coco_keyprovider` with this change links the rustls-native-certs path (`SSL_CERT_FILE` / `SSL_CERT_DIR` are present in the binary and absent from a build without the feature), and an `https://` KBS URL now produces a TLS-layer result instead of `invalid URL, scheme is not http`.


attestation-agent/coco_keyprovider/Cargo.toml
```diff
@@ -22,7 +22,14 @@ protos = { path = "../../protos", default-features = false, features = [
...
 ] }
 rand.workspace = true
-reqwest.workspace = true
+# The workspace pins reqwest with `default-features = false`, which leaves it with no TLS
+# backend. Registering a KEK against a KBS served over TLS then fails before a request is
+# sent, with "invalid URL, scheme is not http". Opt in to rustls for this crate, the way
+# kbs_protocol and confidential-data-hub/kms already do, rather than widening the
+# workspace pin. reqwest 0.13's `rustls` feature verifies through rustls-platform-verifier,
+# which loads the platform trust store -- and honours SSL_CERT_FILE -- so a KBS fronted by
+# a private CA needs no extra roots bundled into the binary.
```
Xynnn007, line 31: let's only include these information into the commit message.

Xynnn007 (approved):
Thanks, only one nit

tjw-ecora:
Thanks for the review. Agreed — and that explanation is already in the commit message body, so the comment in `Cargo.toml` is purely a duplicate of it.

I have deliberately not force-pushed, so as not to disturb your approval. Say the word and I will push the branch with the comment block removed and rebased on current `main` — the diff then becomes the single `reqwest = { workspace = true, features = ["rustls"] }` line. If you would rather keep the review as it stands, I am equally happy for you to merge as-is.

fitzthum:
Looks ok, but please see [AI Policy](https://github.com/confidential-containers/confidential-containers/blob/main/AI.md). For a one line change we just need a simple, human explanation or example. 

Force push as needed, it will not dismiss the review.

PR #1727, opened 2026-09-12
coco_keyprovider: encrypt images against a non-exportable DSM KEK, and unwrap them inside DSM

shashn-fortanix:
## What This PR Implements

This is the work [#1722](https://github.com/confidential-containers/guest-components/pull/1722) lists as out of scope: image encryption using a non-exportable Fortanix DSM key encryption key, and the matching in-KMS unwrap on the guest side.

**This branch is based on [#1722](https://github.com/confidential-containers/guest-components/pull/1722).** Its commit `5c962bc5` will appear in the diff below until that PR merges; only the three commits on top are new here. The `Cargo.lock` conflict against `main` is known and will be resolved by a rebase once #1722 lands.

### CoCo Keyprovider

`coco_keyprovider` gains a DSM path, enabled with `--dsm-endpoint` and `--dsm-api-key-file`. It creates an AES-256 key encryption key in DSM without `EXPORT` in its `key_ops` and has DSM wrap each layer's data encryption key server-side, so the KEK's raw bytes never leave DSM and only the wrapped key is written to the registry. `--kek-mode` selects one KEK for the whole image or one per layer.

`AnnotationPacket` gains two optional fields, `provider` and `annotations`, recording which plugin should unwrap the key at run time and the parameters it needs. Both are optional, so annotations produced by the existing path are unchanged.

`hack/dsm-encrypt.sh` drives the flow with skopeo and ships in the keyprovider image alongside the existing `encrypt-image.sh`.

### Confidential Data Hub

The `Kbc` trait gains `decrypt_with_kek`, with a default implementation, so every existing KBC compiles unchanged and only KBCs whose KMS can unwrap server-side opt in. `CcmKbc` is the only type that overrides it, asking DSM to decrypt the wrapped key with the KEK in place rather than releasing it. `DecryptorProvider::Dsm` routes annotations naming the `dsm` provider to that path; annotations using the default `kbs` provider are untouched.

The `ccm_kbc` DSM client is built on the `sdkms` crate, the same client `coco_keyprovider` uses.

## Backward Compatibility

Everything sits behind the existing `ccm_kbc` feature and the new annotation fields are optional, so the default build is unchanged. Images encrypted by the current keyprovider continue to decrypt through `get_resource`, and `get_resource` remains available for `ccm_kbc` so it still supports the standard [secret-resource feature](https://confidentialcontainers.org/docs/features/get-resource/).

---

RFC: [#1513](https://github.com/confidential-containers/guest-components/issues/1513)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -35,7 +35,7 @@ version = "0.8.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
 dependencies = [
- "cfg-if",
+ "cfg-if 1.0.4",
  "cipher 0.4.4",
  "cpufeatures 0.2.17",
 ]
@@ -355,15 +355,38 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
 
 [[package]]
+name = "attestation"
+version = "0.1.0"
+source = "git+https://github.com/fortanix/attestation.git?rev=fade7bf57c93fefaeaca77fe453b87a7909c9cd4#fade7bf57c93fefaeaca77fe453b87a7909c9cd4"
+dependencies = [
+ "b64-ct",
+ "der 0.7.10",
+ "iocuddle 0.1.1",
+ "log 0.4.29",
+ "mbedtls",
+ "nvml-wrapper",
+ "once_cell",
+ "pkix",
+ "serde",
+ "sgx-isa",
+ "sgx_pkix",
+ "tdx-ql",
+ "thiserror 1.0.69",
+ "yasna 0.3.2",
+]
+
+[[package]]
 name = "attestation-agent"
 version = "0.1.0"
 dependencies = [
  "anyhow",
  "async-trait",
+ "attestation",
  "attester",
  "base64 0.23.1",
  "byteorder",
  "clap",
+ "client",
  "config",
  "const_format",
  "crypto",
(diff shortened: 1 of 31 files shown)


fitzthum:
Needs commit messages and a human PR description. 

Since the very beginning we have avoided a tight coupling between images and the KBS. This is why we don't put the KBS URI in the image annotations, for instance. Having to recreate your images to use CCM seems like a big red flag. Are these changes just a helper to upload your stuff to the DSM or do images need different annotations to be used with Foratnix?

shashn-fortanix:
I am not sure I understand the specifics of your comment. I will try answering them, let me know if I misunderstand any of your questions.

> Since the very beginning we have avoided a tight coupling between images and the KBS.

If you are talking about the "provider" field, we used it as it is already a field in AnnotationPacketV2. Without a field like that, the CDH will default to the default KBS provider and the flow breaks. 

> Having to recreate your images to use CCM seems like a big red flag. 

Isn't this expected if we want to use a masterkey in DSM to encrypt the image? We will have to re-encrypt the image to use it with ccm_kbc.

> Are these changes just a helper to upload your stuff to the DSM or do images need different annotations to be used
with Foratnix?

Both. These changes cover key creation inside DSM, and for the image to be used with ccm_kbc, the annotations should include iv and tag params as crypto context.

fitzthum:
> If you are talking about the "provider" field, we used it as it is already a field in AnnotationPacketV2. Without a field like that, the CDH will default to the default KBS provider and the flow breaks.

If you're talking about the ocicrypt keyprovider abstraction, that is just a pointer to the CDH/AA. That step should be the same in either case.

> Isn't this expected if we want to use a masterkey in DSM to encrypt the image? We will have to re-encrypt the image to use it with ccm_kbc.

No. The image just contains a kid that we can map to a resource uri. You should be able to serve the same resource URIs with the DSM, so all you would need to do is load the appropriate key into the DSM with the expected resource URI and an existing image would work.

Btw, I think I mentioned in the past that you guys might not be using the resource URI correctly. Resource URIs should be interchangeable between backends and should not contain any fotanix-specific fields. If that is not the case you might run into issues here.

PR #1728, opened 2026-09-12
coco_keyprovider: read the algorithm parameter from the right key

shashn-fortanix:
## Problem

`parse_input_params` reads the `algorithm` parameter from the wrong map key:

```rust
let keypath = map.get("keypath").map(|p| p.to_string());
let algorithm = map
    .get("keypath")          // <- should be "algorithm"
    .map(|alg| (*alg).try_into().unwrap_or_default())
    .unwrap_or_default();
```

Both lines read `keypath`, so the `algorithm` parameter documented in the README is never looked up. Passing `algorithm=A256CTR` has no effect and encryption always falls back to the default `A256GCM`.

## Fix

One word: read `algorithm` from `map.get("algorithm")`.


Xynnn007 (approved):

mythi (approved):
LGTM

Looks like the function could be hardened with proper error handling of invalid/unsupported values and improved unit test coverage

fitzthum (approved):

PR #1729, opened 2026-09-13
fix(ttrpc-aa): fix attestation_sock argument handling

### LINKED ISSUE #1580: attestation-agent (ttrpc): --attestation_sock handling is wrong
https://github.com/confidential-containers/guest-components/blob/b2fad038ddb67a54749f484a8a24d9174b75f2ca/attestation-agent/attestation-agent/src/bin/ttrpc-aa.rs#L134-L140

I was doing manual runs as suggested by the cli docs

 ```
   /// This Unix socket address which the Attestation ttRPC service
    /// will listen to, for example:
    ///
    /// `--attestation_sock unix:///tmp/attestation`
```

but noticed the arg handling has problems. Copilot found some more.

1. `DEFAULT_UNIX_SOCKET_DIR` is created always but the parent of  `--attestation_sock` is not
2. a typo like `unix://relative/path/to` is not checked (AI suggested to check for absolute path) 
3. (any) provided path is unconditionally removed (AI suggested to check that the provided `--attestation_sock`  path has a socket file

### PR DESCRIPTION
**Pull request by siddu-os**
## Problem

Three bugs exist in the `--attestation_sock` startup path and `clean_previous_sock_file()` in `attestation-agent/attestation-agent/src/bin/ttrpc-aa.rs`.

Closes https://github.com/confidential-containers/guest-components/issues/1580

---

### Bug 1 — Parent directory only created for the default path

`DEFAULT_UNIX_SOCKET_DIR` was unconditionally created at startup, but if
a user passed a custom path such as `--attestation_sock unix:///custom/path/aa.sock`,
the parent `/custom/path/` was never created, causing a bind failure at runtime.

**Fix:** Extract the parent directory from the user-supplied (or default) socket path
after stripping the `unix://` prefix and call `create_dir_all` on it. The hardcoded
`DEFAULT_UNIX_SOCKET_DIR` constant is no longer needed and has been removed.

---

### Bug 2 — No validation that the path is absolute

After `strip_prefix("unix://")`, a relative path such as `unix://relative/path/to.sock`
was accepted silently and would misbehave at bind time with no useful error message.

**Fix:** After stripping the prefix, check that the resulting path starts with `/`.
Return an error immediately with a descriptive message if it does not.

---

### Bug 3 — `clean_previous_sock_file` removes files unconditionally

The function called `remove_file(path)` whenever the path existed, regardless of
whether it was actually a Unix socket. A user pointing `--attestation_sock` at a
regular file or directory would cause silent data loss.

**Fix:** Check `FileTypeExt::is_socket()` before removing. If the path exists but is
not a Unix socket, return an error rather than deleting it.

---

## Changes

- `attestation-agent/attestation-agent/src/bin/ttrpc-aa.rs` — only file changed

## Testing

- `cargo check --bin ttrpc-aa --features bin,ttrpc,rust-crypto --no-default-features` passes with zero errors and zero warnings.

Signed-off-by: siddaram-sonnagi <siddaram.sonnagi@ibm.com>

### CODE REVIEW

**Review by Xynnn007 — APPROVED** [review]
Thanks @siddu-os 

I have an open question whether we need to detect the socket file is not used by other process before removing it. wdyt?

**Comment by Xynnn007** [comment]
Also, you will need to `cargo fmt` the code.

File: attestation-agent/attestation-agent/src/bin/ttrpc-aa.rs [hunk 1]
```diff
@@ -131,12 +127,22 @@ rpc: ttrpc
...
     debug!(config = ?config, "Using config");
 
-    if !Path::new(DEFAULT_UNIX_SOCKET_DIR).exists() {
-        std::fs::create_dir_all(DEFAULT_UNIX_SOCKET_DIR).expect("Create unix socket dir failed");
+    let sock_path = cli
+        .attestation_sock
+        .strip_prefix(UNIX_SOCKET_PREFIX)
+        .context("socket address scheme is not expected")?
+        .to_owned();
+    if !sock_path.starts_with('/') {
```
mythi: perhaps check this via `sock_dir.is_absolute()` instead

siddu-os:
> Thanks @siddu-os
> 
> I have an open question whether we need to detect the socket file is not used by other process before removing it. wdyt?

Thanks for the review and the good point! In the AA startup context, a stale socket left over from a previous crash/restart is the expected case - a live concurrent AA instance on the same socket would be unusual and would typically be caught at bind time anyway (the kernel rejects a second bind on an in-use socket).

Adding a connect() probe before removal is possible but introduces a small TOCTOU window between the check and the remove_file. I'd lean towards keeping the current behaviour (fail fast if the path is not a socket, remove if it is) and let the bind itself surface any "already in use" error with a clear OS error message.

Happy to add the connect-probe if you feel it's needed for correctness - just let me know! 
Thank you.

siddu-os:
> Also, you will need to `cargo fmt` the code.

Done, cargo fmt applied. Thanks.

attestation-agent/attestation-agent/src/bin/ttrpc-aa.rs
siddu-os: Done, updated to use Path::is_absolute(). 
Thanks for the review.

Xynnn007 (approved):
LGTM. I guess the existing working  socket file with a same path is no-usual. Let's keep it as-is.

attestation-agent/attestation-agent/src/bin/ttrpc-aa.rs
```diff
@@ -199,13 +208,29 @@ rpc: ttrpc
...
-    let path = unix_socket
-        .strip_prefix(UNIX_SOCKET_PREFIX)
-        .ok_or_else(|| anyhow!("socket address scheme is not expected"))?;
+/// Remove a leftover socket file from a previous run.
+///
+/// Only removes the file when it already exists **and** is a Unix socket.
+/// Returns an error if the path exists but is not a socket (e.g. a regular
+/// file or a directory), so callers are not surprised by silent data loss.
+fn clean_previous_sock_file(path: &str) -> Result<()> {
+    let p = Path::new(path);
```
mythi, line 217: would it make sense to pass `&Path` and `sock_path_buf`.

attestation-agent/attestation-agent/src/bin/ttrpc-aa.rs
```diff
@@ -131,12 +129,23 @@ rpc: ttrpc
...
+    let sock_dir = sock_path_buf
+        .parent()
+        .context("attestation_sock path has no parent directory")?;
+    if !sock_dir.exists() {
+        std::fs::create_dir_all(sock_dir).context("create unix socket parent dir")?;
     }
 
-    clean_previous_sock_file(&cli.attestation_sock)
-        .context("clean previous attestation socket file")?;
+    clean_previous_sock_file(&sock_path).context("clean previous attestation socket file")?;
```
mythi, line 148: could also be under `else` for `!sock_dir.exists()`

mythi (commented):
two minor suggestions.

attestation-agent/attestation-agent/src/bin/ttrpc-aa.rs
siddu-os: Fixed. Thank you.

attestation-agent/attestation-agent/src/bin/ttrpc-aa.rs
siddu-os: Done. Thanks.

mythi (approved):
LGTM, thanks for taking care of this

siddu-os:
> LGTM, thanks for taking care of this

Thank you for the review.

PR #1731, opened 2026-09-14
kbs-protocol: Include machine type in RCAR request 

annu0905:
Read the machine type from /proc/sysinfo and include it in the extra_params field of the RCAR Request payload. The AS uses this to decide whether to request a firmware hash from the Ultravisor during challenge generation.
On non-s390x systems where /proc/sysinfo is absent, the field is simply omitted and the existing attestation flow is unchanged.

Signed-off-by: Annu Sharma [annu-sharma@ibm.com]


attestation-agent/kbs_protocol/src/client/rcar_client.rs
```diff
@@ -65,6 +78,65 @@ async fn get_request_extra_params(attestation_policy_selector: Option<&str>) ->
 
     if let Some(attestation_policy_selector) = attestation_policy_selector {
         extra_params[ATTESTATION_POLICY_SELECTOR_JSON_KEY] = json!(attestation_policy_selector);
+
+/// Detect IBM Z machine type from /proc/sysinfo
+/// Returns machine generation string ("z16", "z17") or None if not detected.
+#[cfg(target_arch = "s390x")]
+fn detect_se_machine_type() -> Option<String> {
```
mkulke, line 81: shouldn't this code be part of the attester crate? (in the ibm se module)

annu0905: You're right that the hardware detection logic belongs in the attester. The reason it currently lives in rcar_client.rs is that rcar_client.rs is the only consumer, it needs the machine type to populate extra_params on the KBS Request before the challenge-response begins. Moving it to attester/src/se/ as a pub function is clean since kbs_protocol already depends on attester. Happy to move it after the follow up if required.

Xynnn007: It might be hard to directly integrite into attester crate as we do not have a direct code path to generate extra information. Let me put a PR first to have this.

Xynnn007: This PR would help https://github.com/confidential-containers/guest-components/pull/1735

PR #1735, opened 2026-09-16
KBS Protocol | Add TEE Metadata/parameters to Request message

Xynnn007:
A building block to #1731 in a decoupled style. Please check the concrete commit message.

Also, this requires kbs-types / kbs protocol document to be updated. See https://github.com/confidential-containers/kbs-types/pull/92/changes

Also, this can be one of the TODO for KBS protocol v 1.0.0. See https://github.com/confidential-containers/trustee/issues/1613#issuecomment-5629114492


fitzthum:
Interesting. I think there are a couple different threads here.

1) Novel requirement for s390x challenge stuff per #1731 
2) Changes to tow we identify devices per https://github.com/confidential-containers/trustee/pull/1590
3) Little gap in KBS protocol where the initial message only contains the primary TEE

For 3, it's pretty reasonable to extend the request to include all the devices, although I'm not sure if this is worth a breaking change.

For 2, I was thinking that any additional device identifiers would be supplied by the verifier, not the attester. I guess we could do this in the attester, but it seems like we still have some discussion outstanding on this topic, so maybe we should keep it separate. 

Unfortunately, I don't think either one of those things necessarily solves the first issue. Apparently we need to identify a particular generation. Maybe the one-off patch like in #1731 is the best thing for this.

Xynnn007:
Overall, there are different ways to design to bring s390x related info in the message.

1. The original way in #1731. It can work, but as @mkulke pointed out it brings TEE specific logic into `kbs-protocol` than `attester`. It breaks decoupling but easy to code.
2. To make it more generic. This way we bring a mechanism for attesters to generate metadata that will send along with the Request message to the AS and it would help AS to call `generate_supplemental_challenge` API. Now only `s390x` leverages this API. In this way the protocol changes but the semantics keeps decoupling and have a clear boarder. The protocol level change would also touch your point three, and I'd like to admit that currently it surely only works for s390x, but it's for future.

Note that now the PR implements way 2 and the TEE metadata field will be delivered to AS to generate original challenge:
- https://github.com/confidential-containers/trustee/pull/1616/changes#diff-f0041a445f53d71eb7d44469efa9bc01830e24503eb024ca19eb9e2fd664a741R173-R179
- https://github.com/confidential-containers/trustee/pull/1616/changes#diff-41b55ad1748c6bf85c3d9db9000f0703eb76c4f3afd4fd236bbb70ebdaf33494R322

For s390x it's enough to implement `get_tee_metadata` on the attester side and the Request will carry information.
For verifier side, https://github.com/confidential-containers/trustee/pull/1616/changes#diff-41b55ad1748c6bf85c3d9db9000f0703eb76c4f3afd4fd236bbb70ebdaf33494R322 is the counterpart.

> For 2, I was thinking that any additional device identifiers would be supplied by the verifier, not the attester. I guess we could do this in the attester, but it seems like we still have some discussion outstanding on this topic, so maybe we should keep it separate.

Yes. The two threads are somehow different. TEE metadata is used to help the AS to generate challenge, and no need to carry hardware type information. Note that TEE metadata's integrity is not protected and thus can be modified. Thus verifier logic and the attestation flow design should care about this.

The hardware type PR (I changed the aim to bring the attestation tech https://github.com/confidential-containers/trustee/pull/1590#issuecomment-5630671857) is the verifier side to generate. They are derived from the evidence by the verifier side, and authenticity can be assured. So it can be put into the attestation token claims, while the TEE metadata can not.

fitzthum:
I guess one option would be to just provide an optional hook for each attester to provide additional fields for the request. That might be a bit simpler than doing something more structured, although it would only be supported for primary attesters I guess. wdyt?

Xynnn007:
> I guess one option would be to just provide an optional hook for each attester to provide additional fields for the request. That might be a bit simpler than doing something more structured, although it would only be supported for primary attesters I guess. wdyt?

It's workable also. So we can got different design here

1. To deliver such metadata on the [extra-params](https://github.com/confidential-containers/kbs-types/blob/main/src/lib.rs#L83). Note that currently, we are delivering kbs-protocol semantical things like [hash algorithms and policy selectors](https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/kbs_protocol/src/client/rcar_client.rs#L39-L43) in that field also. So the attester and verifier side should have a contract about what the field of interest should be. The pro is easy to handle w/o breaking change upon kbs protocol. The con is that semantics coupled.
2. To have some more structured request message, s.t. having TEE specific field to take metadata like the current implementation. The pro is that kbs protocol semantics and attester semantics are layered, also it's future facing to include TEE typology. The con is more code and a breaking change.

I think both way has its trade-off. And personally prefer way 2 (as always, I like decoupled design), but I can live with way 1, or other better ideas with strong reason.
cc also @mythi @mkulke if you have time and interested : )

mkulke:
to unblock the #1731 ibm se use case I think using extra params to carry attester specific-payload over is fine, if we can do this without a breaking change to the kbs protocol. We can extend the attester trait with `fn extra_params() -> Option<SerdeJson::Value>` and then do it like Tobin suggested I guess: attach in into some `extra_params.tee` hierarchy.

We can consider generalizing this use case in a larger kbs protocol update, but I'd agree that this PR conflates a few things and it's not trivial to review. maybe it would better to have an RFC for such a meaningful kbs_protocol change  first (or some more details in the PR description, like an example with the shape of the old + new json payload so it's a bit more tangible).

Xynnn007:
Updated based on the comments. w/o changing the kbs protocol, but added to `request.extra-params`

